PR #46, opened 2022-10-03
Update docs to build with pre built containers

noamblitz:


underdarknl (approved):
lets version pin this for the next release

PR #99, opened 2022-12-27
Enable auto-reload of code for all services

dekkers:
Only bytes and rocky did autoreload when code changes, this change will run octopoes, katalogus and keiko uvicorn with reload option. Also adds missing mount of code in the octopoes container.


ammar92 (approved):
Cool, this adds some consistency 👍 

Two side notes to keep in mind:
- When running this natively/ without docker: using the `--reload` option in a PyCharm debugging session might not always work depending on the IDE and package version
- The `--reload` option seems to slow down closing the application/ service

reincode050 (approved):

underdarknl (approved):

PR #125, opened 2023-01-31
Add Debian install guide

dekkers:
This is the Debian install guide from the wiki updated with the recent xtdb multinode changes.


ring-ring-ring (approved):
nice!

underdarknl (approved):

PR #133, opened 2023-02-06
Add postgresql database for scheduler

jpbruinsslot:


docker-compose.release-example.yml
```diff
@@ -185,4 +200,5 @@ volumes:
   bytes-db-data:
   bytes-data:
   katalogus-db-data:
+  scheduler-db:
```
dekkers, line 203: This should be `scheduler-db-data`

dekkers (changes requested):

dekkers (approved):

PR #148, opened 2023-02-08
small clarification

ring-ring-ring:
## Changes
small edits to the install guide to clarify the process and options

### DIFF
--- a/docs/source/technical_design/localinstall.rst
+++ b/docs/source/technical_design/localinstall.rst
@@ -1,6 +1,6 @@
-==========================
-Developer or local install
-==========================
+====================================
+make kat: developer or local install
+====================================
 
 **Important:** The version of OpenKAT you are about to install is a **development environment**, which is used by the developers to build OpenKAT. This is not a version that would be used in a production environment, and it requires some knowledge about Linux, compiling software, and Docker.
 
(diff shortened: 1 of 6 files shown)


docs/source/technical_design/localinstall.rst
```diff
@@ -84,6 +84,11 @@ In some cases this may not work because Docker does not yet know your user name.
 
 Then OpenKAT is built, including all the parts such as Octopoes and Rocky.
 
+Front end
+*********
+
+Find the frontend of your OpenKAT install at port 8000 or 8443 of your localhost and follow the 'on boarding flow' to test your setup and start using your development setup of OpenKAT.
```
underdarknl, line 90: we should clarify the protocol here.
http for 8000
https for 8443
Using the wrong protocol to connect will result the browser and server not being happy,  which will not give a useful error message.

underdarknl: We should also clarify that using http only works when connecting to localhost due to the security flags on the session and xsrf cookies. Localhost specifically is whitelisted to allow secure cookies over an unsecure connection, whereas connecting to any other IP over http be result in these cookies being disregarded. resulting in XSRF warnings when logging in.

ring-ring-ring: done, thnx!

dekkers (approved):

PR #178, opened 2023-02-15
Bytes: Add better API feedback.

Donnype:
The main feature is handling IntegrityErrors when committing a transaction of BoefjeMeta and NormalizerMeta objects. This gives better feedback in the boefje/normalizer logs as we see duplicate task handling happening.

Original PR: https://github.com/minvws/nl-kat-bytes/pull/56

### DIFF
--- a/bytes/bytes/api/v2/router.py
+++ b/bytes/bytes/api/v2/router.py
@@ -1,16 +1,17 @@
 import logging
-from typing import Dict, Optional, List
+from typing import Optional, List
 
 from fastapi import APIRouter, Depends, HTTPException, Request, Query
 from fastapi.responses import Response
 from pydantic import BaseModel
+from starlette.responses import JSONResponse
 
 from bytes.auth import authenticate_token
 from bytes.events.events import RawFileReceived, NormalizerMetaReceived
 from bytes.events.manager import EventManager
 from bytes.models import BoefjeMeta, MimeType, NormalizerMeta, RawData, RawDataMeta
 from bytes.rabbitmq import create_event_manager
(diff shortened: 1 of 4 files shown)


bytes/bytes/api/v2/router.py
```diff
@@ -71,17 +77,22 @@ def create_normalizer_meta(
...
+        )
 
     event = NormalizerMetaReceived(
         organization=normalizer_meta.boefje_meta.organization,
         normalizer_meta=normalizer_meta,
     )
     event_manager.publish(event)
 
-    return {"status": "success"}
+    return JSONResponse({"status": "success"}, status_code=201)
```
ammar92, line 95: Good catch 👍 Should indeed be a 201

bytes/bytes/database/sql_meta_repository.py
```diff
@@ -195,6 +200,11 @@ def __init__(self, cls: Type[SQL_BASE], **kwargs):  # type: ignore
         super().__init__(f"The object of type {cls} was not found for query parameters {kwargs}")
 
 
+class MetaIntegrityError(Exception):
+    def __init__(self, message: str):
+        super().__init__(message)
```
ammar92, line 205: Technically, this initializer is unnecessary since it will forward all arguments by default to the parent's `__init__`. Unless you really want to force the user to specify an error message. Imho I would leave it out since the default (str) representation of an exception is the passed arguments, which is often always a message and used by others as such

ammar92 (dismissed):
One suggestion that could be considered, but consider it approved otherwise 👍

ammar92 (approved):

underdarknl (approved):

underdarknl (approved):

PR #287, opened 2023-02-21
Remove unused fields

jpbruinsslot:
Removal of unused fields in scheduler models


mula/scheduler/models/boefje.py
```diff
@@ -23,7 +17,8 @@ class BoefjeMeta(BaseModel):
     id: str
     boefje: Boefje
     input_ooi: Optional[str]
-    arguments: Dict[str, Any]
```
ammar92, line 26: This is actually a nice hint, unless the key type is changed?

mula/scheduler/models/boefje.py
```diff
@@ -23,7 +17,8 @@ class BoefjeMeta(BaseModel):
...
+    arguments: Dict = {}
```
ammar92, line 20: Mutable type, use a `Field` with `default_factory` argument

ammar92 (commented):
Just one remark (and one question)

mula/scheduler/models/boefje.py
jpbruinsslot: I changed it to resemble the same structure in boefjes. However I'm ok with reverting it, we should consider changing it to both to the correct form.

mula/scheduler/models/boefje.py
```diff
@@ -23,7 +17,8 @@ class BoefjeMeta(BaseModel):
     id: str
     boefje: Boefje
     input_ooi: Optional[str]
-    arguments: Dict[str, Any]
+    arguments: Dict[str, Any] = Field(default=dict)
```
ammar92, line 20:
```suggestion
    arguments: Dict[str, Any] = Field(default_factory=dict)
```

ammar92 (approved):

Donnype (approved):

underdarknl (approved):

PR #293, opened 2023-02-21
Add locking functionality to the priority queue

jpbruinsslot:
Solve race conditions of multiple workers popping off the same task from the queue


dekkers:
The `for_update` isn't necessary when we always hold the lock when calling `pop`.

I also think we will keep running into issues like this if we keep on using threading, because threading is just very hard to get right. I also can't think of a reason why we need to use threading, so I think it would be a good idea if we switch to asyncio in the future and get rid of most concurrency issues doing this. FastAPI is build around asyncio, so it shouldn't be hard to do, see https://fastapi.tiangolo.com/async/

With asyncio we can the just use `for_update` and don't need to do any locking in mula itself. In my opinion we should just make formal that we don't support any other database than PostgreSQL because we really want to be able to use all the power of PostgreSQL and not be limited to only things that are also supported by other databases.

jpbruinsslot:
> The `for_update` isn't necessary when we always hold the lock when calling `pop`.
> 
> I also think we will keep running into issues like this if we keep on using threading, because threading is just very hard to get right. I also can't think of a reason why we need to use threading, so I think it would be a good idea if we switch to asyncio in the future and get rid of most concurrency issues doing this. FastAPI is build around asyncio, so it shouldn't be hard to do, see https://fastapi.tiangolo.com/async/
> 
> With asyncio we can the just use `for_update` and don't need to do any locking in mula itself. In my opinion we should just make formal that we don't support any other database than PostgreSQL because we really want to be able to use all the power of PostgreSQL and not be limited to only things that are also supported by other databases.


> The `for_update` isn't necessary when we always hold the lock when calling `pop`.
> 
> I also think we will keep running into issues like this if we keep on using threading, because threading is just very hard to get right. I also can't think of a reason why we need to use threading, so I think it would be a good idea if we switch to asyncio in the future and get rid of most concurrency issues doing this. FastAPI is build around asyncio, so it shouldn't be hard to do, see https://fastapi.tiangolo.com/async/
> 
> With asyncio we can the just use `for_update` and don't need to do any locking in mula itself. In my opinion we should just make formal that we don't support any other database than PostgreSQL because we really want to be able to use all the power of PostgreSQL and not be limited to only things that are also supported by other databases.

yes was about to remove it, this is an artifact of debugging. I'll have a look for the implications of using asyncio. And I agree on the formalization of postgresql

ammar92 (approved):
Looks good! Although I'm not sure yet if using `queue.Queue` in the tests is the way to go since that queue implementation itself is already thread-safe

Donnype (approved):

underdarknl (approved):

PR #304, opened 2023-02-23
Create volume out of the bytes data dir and change the default

Donnype:
## Changes
This builds rocky as root to have write permissions in the locale directory.

_This section may be skipped or omitted. Uncomment and answer the below questions if relevant._

## Checklist for author(s):
- [ ] This PR comes from a `feature` or `hotfix` branch, in line with our git branching strategy;
- [ ] This PR is "bite-sized" and only focuses on a single issue, problem, or feature;
- [ ] I am not reinventing the wheel: there is no high-quality library that already has this feature;
- [ ] I have performed a self-review of my own code;
- [ ] I have commented my code, particularly in hard-to-understand areas;

If a non-trivial PR:
- [ ] This PR is part of a milestone and has appropriate labels;
- [ ] This PR is properly linked to the project board (either directly or via an issue);
- [ ] I have added screenshots or some other proof that my code does what it is supposed to do;

```

### What works:

### What doesn't work:

### Bug or feature?:
```

```
```


bytes/Dockerfile
```diff
@@ -15,6 +15,7 @@ WORKDIR /app/bytes
 ENV PATH=/home/bytes/.local/bin:${PATH}
 
 RUN mkdir /data && chown bytes: /data
+RUN mkdir ./bytes-data && chown bytes: ./bytes-data
```
ammar92, line 18: You can actually set the file mode directly using `mkdir`s `--mode` option

ammar92 (commented):
One small remark, but consider it to be approved

bytes/Dockerfile
dekkers: I don't understand why an extra data directory is created instead of just using /data? I think we should add a bytes-data volume and mount in on /data in the same way as what we do in the docker-compose.release-example.yml. Then we also don't need to do special things to make sure the directory also exists in the mounted volume and we don't pollute the source repository with data. 

@ammar92 The owner of the directory is being set however, not the mode.

bytes/bytes/raw/file_raw_repository.py
```diff
@@ -21,7 +21,6 @@ def create_raw_repository(settings: Settings) -> RawRepository:
 
 class FileRawRepository(RawRepository):
     UUID_INDEX = 3  # To reduce the number of subdirectories based on the uuid, see self._index()
-    _ENCODING = "utf-8"
```
dekkers, line 24: What is the purpose of this change?

rocky/Makefile
```diff
@@ -9,7 +9,7 @@ export COMPOSE_DOCKER_CLI_BUILD=1
 build: build-rocky build-rocky-frontend
 
 build-rocky:
-	docker-compose run --rm rocky make build-rocky-native
+	docker-compose run -u root --rm rocky make build-rocky-native  # root for write permissions compilemessages
```
dekkers, line 12: Compilemessages is already run in the Dockerfile. Is there any reason to run it again here or can we just remove it from the build-rocky-native step?

dekkers (changes requested):

bytes/bytes/raw/file_raw_repository.py
Donnype: Discovered during looking into these issues that this is not used anywhere

rocky/Makefile
Donnype: The issue was the mount of the source code that removes the compiled messages when they are not present on the host (when debugging an install). Now that I am looking into the container file permissions, I realize that this might actually had to do with the host files being mounted may have had stricter permissions. I will have a look.

bytes/Dockerfile
Donnype: @dekkers This is partly a debugging artifact for the particular issue we were solving, but it got me thinking: what if users did not use the `docker-compose-release-example.yml` and still have the default value of `./bytes-data`. Then the build assumes you will run your containers with specific environment variables, which seemed a bit strange unless we explicitly set `ENV BYTES_DATA_DIR=/data` by default.

But given that this data residing in the source repository is mainly a legacy setup, I agree it is a lot nicer to add a bytes-data volume. Then we could add a command that copies all data from the previous default `./bytes-data` to this volume in the case people want to migrate the development docker-compose file?

`docker-compose run --rm --no-deps --entrypoint /bin/sh bytes -c "cp -r ./bytes-data/* /data"`

bytes/bytes/raw/file_raw_repository.py
underdarknl: Raw files are also not guaranteed to be text, right?

bytes/Dockerfile
dekkers: I am not sure whether we really need to provide migrations for this. It's a development setup, so there should only be test data in there?

I agree that we should also change the default value to `/data`.

Donnype:
> Styling with Markdown is supportedCommentCancel
> Resolve conversation


This still shows some unexpected behavior, still looking into this

bytes/Dockerfile
Donnype: True, although it wouldn't hurt to add a line or two about it in the release notes of course

Donnype:
@dekkers It turned out the issue was cause by Docker being run in rootless mode, causing files mounted from the host not to show the same UID and GID as the host but always having root permissions. I changed it back, but we might have to add either some text to the documentation or provide workaround for this?

dekkers (approved):

PR #317, opened 2023-02-27
Add Cookie OOIs

### LINKED ISSUE #213: Add cookie objects

noamblitz:
## Changes
Add a cookie models and parsers

Closes #213 

## Proof
_Please paste some screenshots or other proof of your (working) change here. If you feel that this is not required (e.g. this PR is trivial), note that here._

To test you can either organically find a resource that has a Set-Cookie header or add a RawCookie manually

## Checklist for author(s):
- [ ] This PR comes from a `feature` or `hotfix` branch, in line with our git branching strategy;
- [ ] This PR is "bite-sized" and only focuses on a single issue, problem, or feature;
- [ ] I am not reinventing the wheel: there is no high-quality library that already has this feature;
- [ ] I have performed a self-review of my own code;
- [ ] I have commented my code, particularly in hard-to-understand areas;

If a non-trivial PR:
- [ ] This PR is part of a milestone and has appropriate labels;
- [ ] This PR is properly linked to the project board (either directly or via an issue);
- [ ] I have added screenshots or some other proof that my code does what it is supposed to do;

```

### What works:

### What doesn't work:

### Bug or feature?:
```

```
```


noamblitz:
Needs some tests

octopoes/octopoes/models/ooi/web.py
```diff
@@ -284,3 +284,51 @@ def format_reference_human_readable(cls, reference: Reference) -> str:
...
+    _natural_key_attrs = ["name", "value", "domain", "path"]
+
+    @classmethod
+    def format_reference_human_readable(cls, reference: Reference) -> str:
+        t = reference.tokenized
+
+        return f"{t.name} @ {t.domain.name}{t.path}"
+
+
+class RawCookie(OOI):
```
dekkers, line 322: Also I don't really understand the purpose of RawCookie. We already have the raw cookie in the HTTP header, what is the added value of duplicating that into a RawCookie instead of just parsing the Cookie from the HTTP header?

dekkers (changes requested):
As you already commented, this still needs tests.

octopoes/octopoes/models/ooi/web.py
noamblitz: It is mostly to decouple cookies from header so that users can add cookies from other sources, such as manually without having to add website -> resource -> header etc

dekkers: What is the use case for this? Because I can't really think of any.

dekkers: But that's a whole different concept than cookies that KAT observes on websites. I don't think it is a good idea to mix these two things and have such loggedin user cookies end up on the web interface and in reports etc. and users being able to use them to login to things they aren't supposed to be able to login to. It also don't think it is a good idea to duplicate all the cookie data just for this.

We should create a different request object for this and probably treat it like a secret. I also wonder how good this will work in practice given that sessions usually have a limited lifetime and might also be bound to a specific IP address.

underdarknl: There's a functional difference between having a Cookie in a HTTP header, and a cookie from any other context.
An observed http header is just that, an http header. From there we extract a cookie in its raw form. A parsed cookie on the other hand has had operations done on it. The reason we do these operations is the same with how we parse Url's into complex objects. This logic should be part of a business rule, and not be part of the model because the model cannot hold turing complete code like Bits can.
Separating these two types also allows us to create an UnparseableCookie finding related to the rawcookie object if parsing fails. It also allows other sources of cookies to be added, (eg, user input, network traffic analysers, browser cookiejars).

Last, the concern about keeping live credentials in the graph makes sense, but is the only reasonable way to create session based logins from KAT. Doing this would require someone to configure the values somewhere as OOI's wich can then be mashed into session cookies by BIT logic (or form posts). These session details will always be present in KAT somewhere for this to function, and having them in the graph means we can invalidate them over time (as the cookie expires), and also stores them in a forensic way.

User Cookies as seen in the browser from some hapless user will not be ingested into Kat, unless there's a boefje specifically looking for them, in which case it would be logical to show them in the reports too. If any cleaning/anonymizing is to be done, it should be done in the boefje doing that gathering, or perhaps they should be stored as encrypted ooi's.

rocky/tools/ooi_form.py
```diff
@@ -61,7 +61,7 @@ def generate_form_fields(
             elif issubclass(field.type_, Enum):
                 fields[name] = generate_select_ooi_type(field)
             elif issubclass(field.type_, str):
-                fields[name] = forms.CharField(max_length=256, **default_attrs)
+                fields[name] = forms.CharField(max_length=2056, **default_attrs)
```
underdarknl, line 64: Practical user agent implementations have limits on the number and
   size of cookies that they can store.  General-use user agents SHOULD
   provide each of the following minimum capabilities:
   
   At least 4096 bytes per cookie (as measured by the sum of the
      length of the cookie's name, value, and attributes).

noamblitz: Okay, should I change it to 4096? This is for all str fields btw

underdarknl: I think so, yes. A better handling of max-lengths needs to be done by checking the input against the model's fields and their max length.

octopoes/octopoes/models/ooi/web.py
underdarknl: A cookie is not a Httpheader. 
"set-cookie: $cookie" is a Httpheader.
"$cookie" is a cookie.
"Cookie: $cookie" is a Httpheader. 

Those are different things, and offer different use cases.

Cookies can be found in headers, and most likely are, but  might come from other places like the user. 
Cookies traveling towards the server have a `Cookie` header-name instead of a `set-cookie` header-name.

By not having the cookie parse logic in the http header parser we keep that parser cleaner, and have a better separation of concerns where The cookie parsing BIT only needs to parse rawcookies instead of all httpheaders. I think we should do the same for CSP headers, header comes in, we create an unparsed raw CSP object, and parse that into the CSP stanza's. This also makes sure that we have a logical place to store the parsing code outside of the Model, as the model wont be able to hold parsing code once the model becomes a declarative system.

Regarding response / requests:
Ideally we move to a response / request OOI model at some point, where  we also generate new possible requests from url's + cookies. However this would flood the graph with possibly short lived OOI's.
Since cookies travel both ways, with a request to the server, and with a response from the server, both would be reasonable references. However, since cookies are hostname + path bound, (which might include subdomains), I feel like a reference to the hostname is enough, including the direction would not make sense unless we also bind it to a request or response OOI and then use that direction to validate if the cookies would have survived this trip and be accepted by the browser based on hostname / path mentioned in the cookie itself, eg, find mismatches. 
I'd think this is for now beyond the scope of this cookie model. Whereas the proposed model would already give us lots of useful BIT logic to detect various (missing) flags, cookie-origins, suspicious third party vendors and the like.

underdarknl: There is one issue remaining though.
Multiple cookies in a single set-cookie header. We are seeing conflicting results based on concatenation of the cookie string, or the combination of multiple similar named headers.

For Cookie headers: we see concatenation of cookies with a semicolon:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies

For http2 we see the same happening with the set-cookie values:
https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2.5

For http header concatenation we see comma separation:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)].

This seems to be reflected in other parsers:
https://github.com/nfriedly/set-cookie-parser/blob/master/test/set-cookie-parser.js

And also what python requests emits.

However the python simplecookie module does not understand the comma separation when the last argument of the cookie is just a keyword. Eg, Secure, instead of secure=True. (same is probably true for the Httponly keyword).

Separation on semicolon would make parsing hard to do with future keywords being added, as the only way to properly split multiple cookies is to recognize (or more specifically not recognize) the name=value and being a new cookie during parsing.

underdarknl: `The section 4.1.1 of RFC 6265 defines the Secure attribute and specifies that it is a "cookie attribute" that can be present with no value or with the value true. If the Secure attribute is present without a value, it is treated as if the value is true.`

And

`A <cookie-value> can optionally be wrapped in double quotes and include any US-ASCII character excluding a control character, Whitespace, double quotes, comma, semicolon, and backslash.`

Cookie values 'should' also be encoded (base64)
Cookie Names are also not ascii or utf8 per-se. (but might be rendered as such when shown to a user)

noamblitz: CookieName1=002230; path=/; Httponly; Secure, _CookieName2=lfsjndldjn; Path=/

Fails but

CookieName1=002230; path=/; Httponly; Secure; SameSite=Lax,  _CookieName2=lfsjndldjn; Path=/

Doesnt fail

dekkers: If we only include the response hostname in the rawcookie objects, we will be unable to report where a problematic cookie originated. For example if we have two paths, /foo and /bar, and /foo sends a good cookie with path=/, but /bar sends a cookie also with path=/ and but missing the secure flag and we have a bit that reports on that. Without the information from which url the cookie originated we can only report "on hostname.com there is a bad cookie somewhere" instead of "https://hostname.com/bar sends a bad cookie". I think the latter is a lot more useful to include in reports.

underdarknl:
To conclude the above comments. We will move this functionality to sprint 1.9 as the current Python implementation is buggy and on top of that does not follow all valid rfc's.

Simplecookie follows RFC2109 (but fails to do so correctly) -> https://github.com/python/cpython/issues/81458

However RFC2109 is obsolete and many frameworks and browsers follows RFC2616, and other rfc's also are valid concatenation strategies for headers.

https://pypi.org/project/cookies/ might be an option to look into, and we could decide that http-headers containing cookies following the obsolete RFC2109 might even warrant an informational finding.

sigio:
RFC2109 is zo te zien obsoleted door 2965
RFC2616 is obsoleted door [RFC 7230](https://2rfc.net/7230), [RFC 7231](https://2rfc.net/7231), [RFC 7232](https://2rfc.net/7232), [RFC 7233](https://2rfc.net/7233), [RFC 7234](https://2rfc.net/7234), and [RFC 7235](https://2rfc.net/7235)

PR #318, opened 2023-02-27
Feature/member edit view

### LINKED ISSUE #272: Add edit view to organization members
Currently we can't set clearance level on users to another level than 0 or 4. We want to be more flexible, but since that setting doesn't fit in the table it is time to have a dedicated edit view for organization members.

This edit view should contain a title containing the username like 'Edit: <username> and a form with, for now, only the following form fields:

- [ ] Clearance level (see screenshot below)
- [ ] Verified (is a on/off toggle, checkbox will work for this)

image

Rieven:
## Changes

- Update view for a particular member (Short view with edits for only member name and trusted clearance level)
- Especially added the radio select for 'trusted clearance level' using new paw design

Created different users for unit tests (superuser, admin, redteamer, client):
1. Updated users and objects of member on the tests
2. Tests for access rights for different members at the member edit view. (ex. clients/redteamers cannot access member edit page, etc..)

Closes #272

## Proof
Added an edit column with an edit button:
image

Update view: (admins and superusers cannot change their own status, blocking themself out)
image

But can edit redteamer or clients
image

Warning trusted clearance level below acknowledged level + automatically setting it at the same level
image

Warning acknowledged level greater than trusted clearance level. User must acknowledge this level to use it.
image

Superuser cannot change trusted clearance level for itself, not neccesary:
image

Superuser can change admin 
image

## Checklist for author(s):

If a non-trivial PR:

```

### What works:

### What doesn't work:

### Bug or feature?:
```

```
```


TwistMeister:
A few remarks, after some functional testing:

- [ ] Organisation admins should be able to edit oragnisation members as wel, with the exception of edition a superuser. In addition: I think we wan't to disable editing the clearance level on a super user. It seems logical to give them clearance
- [ ] level 4 in all cases.
- [ ] We also wan't to lower the "acknowledged clearance level" to the same value, when lowering the "trusted clearance level". When increasing the "trusted clearance level" the "acknowledged level" should stay the same, until the organisation member himself agrees to take the trusted clearance level.

ammar92:
No technical remarks on the code. But please see @TwistMeister's results.

underdarknl:
Can we also show the 'accepted clearance level'? This might help admins understand why a user cannot yet add indemnifications.
Maybe even add a warning if the trusted level is not the same as the accepted level to hint the admin about this.
Obviously, the accepted clearance level should not be editable by the admin.

Rieven:
> Can we also show the 'accepted clearance level'? This might help admins understand why a user cannot yet add indemnifications. Maybe even add a warning if the trusted level is not the same as the accepted level to hint the admin about this. Obviously, the accepted clearance level should not be editable by the admin.

Added:
image

Rieven:
> A few remarks, after some functional testing:
> 
> * [ ]  Organisation admins should be able to edit oragnisation members as wel, with the exception of editing a superuser.
> * [ ]  In addition: I think we wan't to disable editing the clearance level on a super user. It seems logical to give them clearance level 4 in all cases.
> * [x]  We also wan't to lower the "acknowledged clearance level" to the same value, when lowering the "trusted clearance level". When increasing the "trusted clearance level" the "acknowledged level" should stay the same, until the organisation member himself agrees to take the trusted clearance level.

Logged in as admin cannot change superuser (also navigating using the edit link gives a Forbidden 403)
image

rocky/account/forms/account_setup.py
```diff
@@ -168,10 +172,42 @@ class Meta:
...
+        widget=forms.RadioSelect(),
+        help_text=_("Select a clearance level you trust this member with."),
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if self.instance.user.is_superuser:
+            self.fields["trusted_clearance_level"].disabled = True
+        if "admin" in self.instance.user.groups.all() or self.instance.user.is_superuser:
+            # There could be a case where you block yourself out of the system
```
TwistMeister, line 194: So it's possible an admin blocks himself? Maybe its better to disable the checkbox for the current user? For sanities sake?

rocky/rocky/templates/partials/cat_paws.html
```diff
@@ -0,0 +1,6 @@
+<ul class="level-indicator l{% if level < 0 %}0{% else %}{{ level }}{% endif %} {{ custom_class }}">
+  <li></li>
+  <li></li>
+  <li></li>
+  <li></li>
+</ul>
```
TwistMeister, line 6: We already had "scan_level_indicator.html" which does this, but more dynamically (instead of having the `li`'s static, it loops over the amount of `scan_levels`. I think you can use that partial and remove this one.

rocky/rocky/templates/organizations/select_clearance_level_radio_input.html
```diff
@@ -0,0 +1,27 @@
...
+  {% for choice in field %}
+    <div class="horizontal-view">
+      <input type="radio"
+             name="{{ field.name }}"
+             value="{{ choice.data.value }}"
+             id="id_{{ field.name }}_{{ forloop.counter0 }}"
+             {% if choice.data.attrs.disabled %} disabled{% endif %}
+             {% if choice.data.attrs.checked %} checked{% endif %}/>
+      <label>
+        <ul class="level-indicator l{{ forloop.counter0 }}">
```
TwistMeister, line 16: I think you could use the `scan_level_indicator` partial for rendering the amount of paws after each radio button.

TwistMeister (changes requested):
Few remarks, haven't reviewed the tests since I don't feel I'm equipped to judge about them.

rocky/account/forms/account_setup.py
Rieven: > 

Which checkbox? I have disabled the dropdown for admins to change their status.

rocky/rocky/templates/partials/cat_paws.html
Rieven: The scan level indicator is used differently :
image

rocky/account/forms/account_setup.py
TwistMeister: Yes, but an organization admin shouldn't be able to block themselfs. If that's not the case then this is fine.

rocky/rocky/templates/partials/cat_paws.html
TwistMeister: They are both ul's with li's that represent cat pawns? Don't really see what's different...

Rieven: `scan_level_indicator` is looping without radio button in the loop

rocky/tests/test_members.py
```diff
@@ -0,0 +1,162 @@
...
+    This will test if a admin from one organization can edit
+    a admin from another organization at the member edit view.
+    """
+
+    request = setup_request(rf.get("organization_member_edit"), admin_member.user)
+    # from OrganizationView
+    with pytest.raises(Http404):
+        OrganizationMemberEditView.as_view()(
+            request, organization_code=admin_member_b.organization.code, pk=admin_member_b.id
+        )
```
dekkers, line 119: What is missing is the case where the user is an admin for organization A but a client for organization B and tries to edit B.

What is also missing is test for the actual post that changes things.

rocky/tests/test_members.py
```diff
@@ -0,0 +1,162 @@
...
+    resulted_request = setup_request(rf.get(response.url), admin_member.user)
+    resulted_response = OrganizationDetailView.as_view()(
+        resulted_request, organization_code=admin_member.organization.code
+    )
+    assert resulted_response.status_code == 200
+
+    # member list has been updated
+    assertContains(resulted_response, "Member name test")
+    assertContains(resulted_response, "Active")
+    assertContains(resulted_response, "Yes (L4)")
```
dekkers, line 145: Instead of calling the OrganizationDetailView and trying to parse the response, just check if it is saved in the database Something like

```
redteam_member.refresh_from_db()
assert redteam_member.trusted_clearance_level == 4
```

rocky/tests/test_members.py
```diff
@@ -0,0 +1,162 @@
...
+        ),
+        admin_member.user,
+    )
+    response = OrganizationMemberEditView.as_view()(
+        request, organization_code=redteam_member.organization.code, pk=redteam_member.id
+    )
+    assert response.status_code == 200
+    assertContains(response, "Member name test")
+    assertContains(response, "blocked")
+    assertContains(response, "l0")
```
dekkers, line 162: You also need to this here to check if the status has changed.

dekkers (changes requested):

rocky/rocky/views/organization_member_edit.py
```diff
@@ -1,53 +1,78 @@
...
+class OrganizationMemberEditView(PermissionRequiredMixin, UserPassesTestMixin, OrganizationView, UpdateView):
+    form_class = OrganizationMemberEditForm
     model = OrganizationMember
     template_name = "organizations/organization_member_edit.html"
     object: OrganizationMember
-    permission_required = "tools.change_organization"
+    permission_required = "tools.change_organizationmember"
+
+    def test_func(self):
+        return not self.get_object().user.is_superuser or self.request.user.is_superuser
```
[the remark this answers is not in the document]
Rieven: This happens at the permission_required attribute which only admins have

rocky/tests/conftest.py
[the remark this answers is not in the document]
Rieven: And why did you you change all test cases from being tested as normal user to being tested as superuser? That can hide problems that happen for normal users but don't happen for superusers because of permissions etc.

All tests used superuser but was called `my_user`

rocky/rocky/templates/partials/cat_paws.html
TwistMeister: Why not do it something like this?

`<input type="radio"
             name="{{ field.name }}"
             value="{{ choice.data.value }}"
             id="id_{{ field.name }}_{{ forloop.counter0 }}"
             {% if choice.data.attrs.disabled %} disabled{% endif %}
             {% if choice.data.attrs.checked %} checked{% endif %}/>
      <label>
          {% include 'partials/scan_level_indicator.html' with value=forloop.counter0 %}
      </label>
      `
      
That way you get pawns rendered as an UL with LI's with unfilled pawns where needed. That's exactly what this component was meant to do.

TwistMeister:
It's possible to assign a clearance level to a client from the Members view. However, clients cannot accept any clearances.

rocky/rocky/templates/organizations/organization_member_edit.html
```diff
@@ -8,35 +8,36 @@
...
-            <div>
-              {% translate 'Goal' %}
-              <div>{{ object.goal }}</div>
-            </div>
+            <legend>
+              {% translate 'Edit' %} {{ object.user.full_name }}
+            </legend>
             {% for field in form %}
-              {% include 'partials/form/field_input_wrapper.html' %}
+              {% if field.name == "trusted_clearance_level" %}
```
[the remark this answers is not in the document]
Rieven: > These 2 fields have an integer field but must be converted to cat paws with scan levels

These 2 fields have an integer field but must be converted to cat paws with scan levels

rocky/account/forms/account_setup.py
Rieven: Logic moved to Views, needed the request

rocky/rocky/views/organization_member_edit.py
```diff
@@ -1,53 +1,78 @@
...
+                    % (tcl, acl, tcl),
+                )
+            if int(tcl) > int(acl):
+                messages.add_message(
+                    self.request,
+                    messages.WARNING,
+                    _(
+                        "You have trusted this member with a higher trusted level than member acknowledged. "
+                        "Member must first accept this level to use it."
+                    ),
```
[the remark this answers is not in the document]
Rieven: was a request from @underdarknl 
https://github.com/minvws/nl-kat-coordination/pull/318#issuecomment-1447771338

rocky/tests/test_members.py
Rieven: See commit: https://github.com/minvws/nl-kat-coordination/pull/318/commits/d2b2b4d21514943d91be4767237549db64ea86e6

rocky/rocky/templates/organizations/organization_member_list.html
```diff
@@ -19,69 +19,89 @@ <h2>{% translate 'Members' %}</h2>
...
-              <td>
-                {% for group in member.user.groups.all %}
-                  {% for permission in group.permissions.all %}
-                    {% if 'can_set_clearance_level' in permission.codename %}
-                      {% if not member.user.is_superuser %}
-                        {% if member.trusted_clearance_level > 0 %}
-                          {% include 'partials/single_action_form.html' with btn_text="Revoke" action='withdraw_clearance' key='member_id' value=member.id btn_class="ghost" %}
-                        {% else %}
-                          {% include 'partials/single_action_form.html' with btn_text="Grant" action='give_clearance' key='member_id' value=member.id %}
+            {% if not member.user.is_superuser or request.user.is_superuser %}
```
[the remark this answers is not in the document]
Rieven: We don't want superuser to end up in the frontend listed nor editing.

rocky/rocky/templates/organizations/organization_member_list.html
```diff
@@ -19,69 +19,89 @@ <h2>{% translate 'Members' %}</h2>
             <th>{% translate 'Status' %}</th>
             <th>{% translate 'Role' %}</th>
             <th>{% translate 'Added on' %}</th>
-            <th>{% translate 'Has trusted permission to set OOI clearance levels' %}</th>
-            <th>{% translate 'Acknowledged permission to set OOI clearance levels' %}</th>
-            <th>{% translate 'Grant/revoke permission to set OOI clearance levels' %}</th>
+            <th>{% translate 'Has trusted permisson to set OOI clearance levels' %}</th>
+            <th>{% translate 'Acknowledged permisson to set OOI clearance levels' %}</th>
+            <th>{% translate 'Grant/revoke permisson to set OOI clearance levels' %}</th>
```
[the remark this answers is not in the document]
Rieven: Moved grant revoke before edit button:
image

rocky/rocky/views/organization_member_edit.py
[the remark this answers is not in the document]
dekkers: I just asked @underdarknl and he agrees that both messages should be INFO instead of WARNING.

rocky/rocky/views/organization_member_edit.py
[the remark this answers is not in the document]
dekkers: The permission can only say that the user is an admin of an organization, not whether the user is an admin of this organisation. If the user is an admin of organization A, but only a client of organization B, the user should not be able to edit members of organisation B. The current code doesn't enforce that.

rocky/tests/conftest.py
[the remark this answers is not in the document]
dekkers: This was not how it was originally implemented, but I see that it somehow got changed from user to superuser in https://github.com/minvws/nl-kat-rocky/pull/115/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R27. This was not correct, tests should use a normal user, so all places that used to ues the my_user fixture should be changed to have user, clientuser or client_member fixture instead of superuser.

rocky/rocky/templates/organizations/organization_member_list.html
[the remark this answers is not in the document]
dekkers: I can see that, but that isn't a reason to revert the fix of the spelling error...

rocky/rocky/templates/organizations/organization_member_edit.html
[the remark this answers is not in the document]
dekkers: You can do that in a widget, see https://docs.djangoproject.com/en/4.1/ref/forms/widgets/. But this is also broader than just these changes, so I created #506 for this to clean this up.

rocky/rocky/templates/organizations/organization_member_list.html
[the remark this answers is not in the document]
Rieven: owwww , I didnt see the typo

rocky/tests/test_members.py
dekkers: This means the code that calls OrganizationDetailView can be removed from the test. This makes the test more efficient. Testing OrganizationDetailView itself is already done in other tests.

rocky/tests/test_members.py
dekkers: What you test there is user that is an admin of organization A and not a member of organization B. That is different from a user that is an admin of organization A and a non-admin member or organiation B.

rocky/rocky/urls.py
```diff
@@ -84,7 +84,7 @@
         name="organization_add",
     ),
     path(
-        "<organization_code>/<path:pk>/edit/",
+        "<organization_code>/<int:pk>/edit/",
```
dekkers, line 87: The pk is not necessary here actually, we already have the code, so this can just be `<organization_code>/edit/`

dekkers (changes requested):

rocky/rocky/urls.py
Rieven: Edited in https://github.com/minvws/nl-kat-coordination/pull/318/commits/1602747e044eec605a21b070131e1a33448541b6

rocky/tests/test_members.py
Rieven: You are indeed right. Here admin member of orgA is requesting the MemberEditView of another admin member from orgB using the pk=admin_member_b.id  which will throw this 404, either way if the user is a client or a redteamer it will have the same effect. This comes form the OrganizationView that a user/member cannot even view things from another org. We can remove this test or do you have any idea how to mock this?

dekkers: The issue is when the user is an admin member of orgA and **also a non-admin member of orgB**. This will not throw a 404, because the user is a member of orgB. Permissions are global, so this user will also have the "tools.change_organizationmember" permission. So this user will be able to edit organization members while not being an admin member of orgB.

TwistMeister:
As I said before: It's possible to assign a clearance level to a client from the Members view. However, clients cannot accept any clearances.

Maybe I wasn't clear on this, but it should not be possible to assign a clearance level to clients. It should only be possible to set a clearance level on red teamers and admins (super admins already have it by default).

rocky/rocky/templates/organizations/organization_member_list.html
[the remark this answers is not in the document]
TwistMeister: I may have missed/forgotten the reason, but why don't we want superusers to be listed?

Darwinkel:

### What works:
* New UI is nice and works overall
* No unhandeled errors

### What doesn't work:
* Superadmin member name is "Unset"
image

* "Acknowledged clearance level" does not match actual acknowledged clearance level (should be four paws instead of zero paws)
image


### Bug or feature?:
* The "Grant" button automatically assigns L4 clearance to a user, but this is mentioned nowhere. Does it make sense to automatically assume L4?

image
* From a user perspective, "Unset" and L0 have the same effect. What exactly is the difference between the two from a user perspective?

Rieven:
@Darwinkel 

> * Superadmin member name is "Unset"
Yes, this will be fixed in another branch, the superuser name is indeed Unset but it is nowhere to be set, it will be in @TwistMeister branch

> * "Acknowledged clearance level" does not match actual acknowledged clearance level (should be four paws instead of zero paws)

This is fixed now

> * The "Grant" button automatically assigns L4 clearance to a user, but this is mentioned nowhere. Does it make sense to automatically assume L4? 

This was not added in this feature, but the idea was to grant full access till this moment. Will be refactored in @TwistMeister extended branch

> * From a user perspective, "Unset" and L0 have the same effect. What exactly is the difference between the two from a user perspective?
The idea was that -1 is really unset and untouched by the user. When an admin sets it for example to L0 then we really know that it was intentionally set to L0. If it was L0 at the beginning and the admin sets it to L0, so same you couldn't know the difference.

rocky/rocky/views/organization_member_edit.py
```diff
@@ -26,9 +26,11 @@ def test_func(self):
 
     def get_form(self):
         form = super().get_form()
-        if (self.object.user == self.request.user) or self.request.user.is_superuser:
+        if self.object.user.is_superuser or self.object.user.groups.filter(name=GROUP_ADMIN).exists():
             # There could be a case where you block yourself out of the system
             form.fields["status"].disabled = True
+        if self.object.user.groups.filter(name=GROUP_CLIENT).exists():
```
dekkers, line 32: This does two database queries to check if the user is in a specific group. We should just fetch the groups with self.request.user.groups.all() and then check if the specific group is in the list.

dekkers (approved):

underdarknl (approved):

PR #402, opened 2023-03-01
Cleanup rocky requirements

dekkers:
Remove django-markdownify as it is no longer used

Replace pytz with datetime.timezone. Also fixes a bug that initial form data didn't use a lambda and had the time of process startup as initial datetime instead of the current time.


ammar92 (approved):
Thanks for the cleanup and nice find about the wrong initial value for the widget.

By the way, it doesn't necessarily have to be a lambda, could be any callable (e.g. a factory method or partial objects)

underdarknl (approved):

PR #414, opened 2023-03-02
Update docs about clearance levels

TwistMeister:
Fixes explanation on the various levels we use in KAT


ammar92 (approved):

underdarknl:
ideally, i'd also have this information stored inside a table or config somewhere in KAT, translatable if possible.
I forsee a future where users of OpenKAT might want to change these levels themselves, or add levels themselves.

TwistMeister:
@underdarknl Yes, that's a great idea, also for when we'll render these descriptions anywhere else; one source of truth. But it seems out of scope for this PR, since I only wanted to rectify this small error in the current docs. I will create an issue for your suggestion.

docs/source/introduction/howdoesitwork.rst
```diff
@@ -42,9 +42,15 @@ Indemnification
 OpenKAT works with a system of indemnities for scanning, linked to intrusion levels. An organization gives an indemnification for a certain intensity of scan. For each intrusion level, the level is specified, in order to prevent unexpected problems for a production system.
 
 Intrusion levels or indemnities:
-:L1: 'do not touch'
+
+:L0: 'do not touch'
```
underdarknl, line 46: I feel like we might need to explain this some more. Don't touch also applies to level 1, but level 0 never gathers information from this OOI. Whereas Level 1 does, but never directly by touching the system itself.

TwistMeister: Good point, will add this.

dekkers (approved):

PR #445, opened 2023-03-06
Specify install requirements for running Docker in rootless mode.

Donnype:
## Changes
This would add the option to get a working Kat install while running Docker in rootless mode. This works because docker-compose uses the `COMPOSE_FILE` environment variable to determine which docker-compose.yml to use. When you set 

```shell
export KAT_COMPOSE_FILE=$PWD/docker-compose.yml:$PWD/docker-compose.mountless.yml
```

This is the same as running docker-compose as follows from $PWD:

```shell
docker-compose -f docker-compose.yml -f docker-compose.mountless.yml
```

[More info](https://docs.docker.com/compose/environment-variables/envvars/).

In the `docker-compose.mountless.yml` we can override volume mounts that crash because containers need write privileges in the source code directory.

Closes https://github.com/minvws/nl-kat-coordination/issues/415


Donnype:
I asked someone running rootless Docker to test this setup. To be continued.

Donnype:
Closing for now as this will be an incomplete solution.

Issue #428, opened 2023-03-06
Failing to create organization for external services in post_create creates inconsistency in Rocky


Donnype:
When starting up services and some failing to get up, creating an organization in Rocky at several times has caused inconsistency in my setup since a failed post_create does not delete the organization. This means we also cannot retry creating an organization. I would suggest to make this (and perhaps other?) more atomic, meaning we only finish the creation of an organization when all services succeed to create one and else delete organizations everywhere. Another option would be to support some interface to retry this, but since the existence of organizations is crucial to functionality I would not opt for this now perhaps. Retrying the creation calls in `post_create` is already a good first step by the way.

dekkers:
I doubt that retrying again will help, because most of the time this fails because there is some misconfiguration that needs to be fixed. What probably would help a lot is if we have good health checks everywhere and don't allow creation of an organization when not all services are healthy.

Donnype:
Agreed that basic retries would not capture most current cases. Only checking health of services would not capture bugs/programming errors in those services though, still leaving Rocky open to inconsistencies. The trade-off would be consistency vs. availability I guess, although it is arguable how `available` Rocky really is when you can create organizations without having the organization in the Katalogus or an XTDB node for that organization.

PR #455, opened 2023-03-07
Check Octopoes and Katalogus health before creating orgs, "rollback" on exceptions

Donnype:
Closes https://github.com/minvws/nl-kat-coordination/issues/428.

This makes adding organizations through Rocky a more atomic operation to reduce inconsistencies.


Donnype:
I also added a fix for the bug in `rocky/middleware/onboarding.py`: redteamers might not have an organization member after an organization has been deleted.

dekkers (changes requested):
This looks good to me, but we should also catch RockyError in  OnboardingOrganizationSetupView and return a proper error message to the user. We currently only do this in the admin (see rocky/rocky/admin.py). The saving is done in form_valid, so we should catch the exception when calling super and also add the success notification after saving.

Donnype:
@dekkers See update:

image

rocky/tools/models.py
```diff
@@ -73,42 +75,76 @@ def get_absolute_url(self):
...
+            except Exception as e:
+                raise RockyError(
+                    f"Could not delete organization in the Katalogus after failing to create the "
+                    f"organization in the Katalogus: {e}"
+                ) from e
+
             raise RockyError(f"Octopoes returned error creating organization: {e}") from e
 
+    @staticmethod
+    def _get_healthy_katalogus_and_octopoes(organization_code) -> Tuple[KATalogusClientV1, OctopoesAPIConnector]:
```
ammar92, line 124: It makes more sense to split this into two functions

rocky/rocky/middleware/onboarding.py
```diff
@@ -24,12 +24,15 @@ def middleware(request):
                 or request.path.startswith("/api/")
             ):
                 if not member_onboarded:
-                    if is_red_team(request.user):
+                    member = OrganizationMember.objects.filter(user=request.user)
+
+                    # There might be redteamers without an organization after an organization is deleted.
+                    if member.exists() and is_red_team(request.user):  #
```
ammar92, line 30:
```suggestion
                    if member.exists() and is_red_team(request.user):
```

ammar92 (commented):
Looks good, some remarks/ questions

rocky/rocky/middleware/onboarding.py
underdarknl: This is not just limited to red-teamers. Admins and clients are mostly the same.

Donnype: @Rieven Was there a specific reason to filter on redteamers here?

Rieven: > @Rieven Was there a specific reason to filter on redteamers here?

@Donnype Yes, when member is not onboarded at startup it is redirected to the redteamers onboarding flow/views.

Donnype: Answer: yes clients do not necessarily need to be onboarded(?)

Darwinkel:
- [x] I confirmed that the PR's advertised `feature` or `hotfix` works as intended;

### What works:
* No functional regressions under normal circumstances
* Proper error message when Octopoes is down
* Proper error message when Katalogus is down

Nice work! I think this may actually be extremely helpful in debugging what exactly goes wrong in the CI's Robot tests!

Darwinkel (approved):

PR #456, opened 2023-03-07
Seed on startup and add missing method 

Donnype:
## Changes
When running more granular builds we run into the issue of missing the LOCAL repository sometimes for the docker setup. This should fix that. I also implemented the delete method for the SQL implementation that apparently was missing, and added a check now allowing the LOCAL repository to be deleted.


boefjes/boefjes/katalogus/routers/repositories.py
```diff
@@ -42,5 +42,7 @@ def remove_repository(
     repository_id: str,
     storage: RepositoryStorage = Depends(get_repository_store),
 ):
+    if repository_id == RESERVED_LOCAL_ID:
```
ammar92, line 45: Although this is the correct response, I'm still a bit hesitant if this is the right layer to block such calls.

You could still delete the local repository using the `SQLRepositoryStorage` or any other storage repository client (I'm not sure what happens in that case). This doesn't necessarily mean this is a bad idea because by convention we can still assume that these clients will only be used by this specific API backend which handles the application logic (e.g. prohibiting deletion of the local repository)

Donnype: Also popped in my mind indeed, but I would say that you would want the option to delete this one through the repository instead of the repository knowing of certain business requirements. And at this stage there is no service/business layer between the API and repository..

dekkers (approved):

Issue #425, opened 2023-03-06
Certificates and wildcard domains


noamblitz:
Subject Alternative Names must be created for each domain when a certificate has a wildcard in it

noamblitz:
Few thoughts:

There should be a coupling (other than Subject Alternative Names) between certificates and hostnames. This means changing the current model to support that, which is a risk because of migrations.

So the subjects and the SANS should be of type str with a coupling OOI between those and Hostnames. Problem is that bits can not query XTDB with plain text, which means that for "normal" subjects and sans can be added easily, but wildcard domains not.

A solution to get rid of the false positives is to add an OOI "WildcardSubjectAlternativeName" and check in the bit that makes the finding whether the hostname is either in the normal sans or part of a wildcard. This should solve all false potiives

underdarknl:
Its more complicated, but also more simple (from a Bits perspective):
There should be a coupling from a cert and a hostname+ip+port combo.

A given service using a given protocol on a given port with a given hostname uses a certain certificate. These Might bind to a wildcard cert, these might bind to a wrong cert, these might not include a hostname (eg, a server using a named cert when accessed directly on the IP without specificying a hostname).

noamblitz:
Hmm this would add a lot of challenges regarding migrations. 

Also, same challenge regarding regarding wildcards remain

underdarknl:
Currently we are looking in the graph to see if we know of any cert that might satisfy the required certificate need for an ecnrypted connection. We should instead be looking at what cert was in fact used by a given service when requesting it to connect using the associated hostname.

noamblitz:
After a discussion with @underdarknl, we came to the following conclusions:

1. We will introduce a SubjectAlternativeNameWildcard OOI with a str attribute for the wildcard and a netloc attribute of type Hostname. When a website serves a certificate, the bit will check the subject, SubjectAlternativeNames and SubjectAlternativeNameWildcards. This will fix all false positives.
2. Website OOIs will have the Hostname as an optional attribute. When hostname is None, this is because the IPAddress in the IPService functions as netloc.
3. If we want to add other IPServices that serve certificates like FTP servers and mailservers, those will need to be added as OOIS that point to a cert.

underdarknl:
Agreed:
no 2 requires a rework of various bits who now use the available hostname to construct urls. This also allows us to add websites with just the ip-address as their location.

noamblitz:
@dekkers

PR #468, opened 2023-03-09
check for wildcards

noamblitz:
## Changes
Check for wildcards in subject and subject alternative names of certs before creating finding

Fixes https://github.com/minvws/nl-kat-coordination/issues/425


noamblitz:
Found a bug (which also existed before this PR):

Because certificate is an optional attr of website, the bit is not (always) correctly updated when a new website points to an already existing certificate. This means sometimes, new websites do not get a finding although they deserve one.

ammar92 (approved):

underdarknl (approved):

Issue #232, opened 2023-02-07
As a KAT-user I want to export reports through an endpoint


TwistMeister:
As any KAT user I want to be able to export a KAT report so that I'm able to distribute it (automatically, periodically) outside of KAT.

Donnype:
So the main blocker would be that users do not have access to the keiko API directly? Or are there no PDF reports in Keiko that users can easily distribute?

Donnype:
As a suggestion, perhaps we can enhance the current `/api/v1` specification as follows:

```
[GET, POST] /organization/
[GET, PUT, PATCH, DELETE] /organization/{organization_id}/
[GET] /report?organization_id={org_id}&clearance_level={level}&clearance_type={type}?valid_time={time}?status={status}
[POST] /report/
```

Where the JSON of a POST request could be something along the lines of
```json
{
  "id": "str",
  "valid_time": "str", 
  "organization_id": "int",
  "status": "Literal['pending', 'finished', 'failed']"

  "ooi_id": "Optional[str]",
  "clearance_level": "Optional[int]",
  "clearance_type": "Optional[str]",
}
```

And we can retrieve this JSON metadata or delete the report through 
```
[GET, DELETE]
/report/{report_id}/metadata
```

And access the pdf file once `"status" == "finished"` at the location field, which would equal
```
[GET]
/report/{report_id}
```

Some considerations:
-  We could choose to create a Report database model, try to add some persistent info about reports into keiko or accept that we cannot filter old reports through Rocky for now (as there is no info in keiko mapping report ids to metadata of their contents).
- We could try to wait for the reports during the request, but this can be risky. A status field may resolve this as in this example
- The id itself could match either the keiko report id,  or the primary key of a database model in Rocky if we create one.
-  The location field could be left out of course, and we could also make the metadata json the default and return the actual file at `/report/{report_id}/data`.
-  We might have to make the organization_id optional in the future if we do cross-organization reporting, or we can start off making this a many-to-many by making a list of primary keys out of this field: `"organizations": [1, 2, 4, ...]`. If we are **never** going to do this, it could be worth considering namespacing the reports endpoints on the `organization_id`.
- Clearance filters are supported for all object list views (see `BaseOOIListView::get_queryset()`. I added these here for completeness with respect to the new button, but I do not know what the convention is for Findings and if it makes sense to have these filters here (@Lisser ?).

[...]

PR #469, opened 2023-03-09
Feature/organization reports through api

Donnype:
## Changes
This adds a button and endpoint to generate Finding reports on the organization level. It updates Keiko accordingly to handle more generic Finding reports.

https://github.com/minvws/nl-kat-coordination/issues/232

When you login with an client you can now download a pdf report at (e.g. for localhost and organization `abc`):
http://localhost:8000/nl/abc/findings/report/pdf

Other languages are not supported so changing nl for en still returns a Dutch report.

## Proof
Here is the new button  on the Findings page:

new_button

The new report is compatible with the previous ooi reports, but can now also produce organization level titles:

rep

### DIFF
--- a/rocky/rocky/views/ooi_report.py
+++ b/rocky/rocky/views/ooi_report.py
@@ -201,7 +201,8 @@ class OOIReportPDFView(SingleOOITreeMixin, ConnectorFormMixin, View):
         # reuse existing dict structure
         report_data = build_findings_list_from_store(self.tree.store)
         report_data["valid_time"] = str(self.get_observed_at())
-        report_data["ooi"] = get_ooi_dict(self.ooi)
+        report_data["report_source_type"] = self.ooi.object_type
+        report_data["report_source_value"] = self.ooi.human_readable
 
         # request pdf from keiko
         try:
@@ -210,26 +211,7 @@ class OOIReportPDFView(SingleOOITreeMixin, ConnectorFormMixin, View):
             messages.error(self.request, _("Error generating report: {}").format(e))
             return redirect(get_ooi_url("ooi_report", self.ooi.primary_key, self.organization.code))
 
-        # generate file name
-        report_name = "bevindingenrapport"
-        org_code = self.organization.code
-        ooi_id = self.ooi.primary_key
-        valid_time = self.get_observed_at().isoformat()
-        current_time = datetime.now(timezone.utc).isoformat()
-        language = "nl"
-        report_file_name = "_".join(
-            [
-                report_name,
-                language,
-                org_code,
-                ooi_id,
-                valid_time,
-                current_time,
-            ]
-        )
-        # allow alphanumeric characters, dashes and underscores, replace rest with underscores
-        report_file_name = re.sub("[^0-9a-zA-Z-]", "_", report_file_name)
-        report_file_name = f"{report_file_name}.pdf"
+        report_file_name = self._generate_report_file_name()
 
         # open pdf as attachment
         try:
@@ -239,6 +221,23 @@ class OOIReportPDFView(SingleOOITreeMixin, ConnectorFormMixin, View):
                 self.request, _("Error generating report: Timeout reached. See Keiko logs for more information.")
             )
             return redirect(get_ooi_url("ooi_report", self.ooi.primary_key, self.organization.code))
+
+    def _generate_report_file_name(self):
+        report_file_name = "_".join(
+            [
+                "bevindingenrapport",
+                "nl",
+                self.organization.code,
+                self.ooi.primary_key,
+                self.get_observed_at().isoformat(),
+                datetime.now(timezone.utc).isoformat(),
+            ]
+        )
+        # allow alphanumeric characters, dashes and underscores, replace rest with underscores
+        report_file_name = re.sub("[^0-9a-zA-Z-]", "_", report_file_name)
+        report_file_name = f"{report_file_name}.pdf"
+
+        return report_file_name
 
 
 """
--- a/rocky/rocky/views/ooi_report.py
+++ b/rocky/rocky/views/ooi_report.py
@@ -6,6 +6,7 @@ from django.contrib import messages
 from django.core.exceptions import BadRequest
 from django.http import FileResponse
 from django.shortcuts import redirect
+from django.urls import reverse
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 from django.views import View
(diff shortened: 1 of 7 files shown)


Donnype:
@dekkers What would be the best route to allow access to this view through DRF as well?

rocky/rocky/views/ooi_report.py
```diff
@@ -198,47 +243,54 @@ def get(self, request, *args, **kwargs):
...
+            self.organization.name,
+            {finding.primary_key: finding for finding in findings},  # Create store from finding list
+            self._generate_report_file_name(),
+            reverse("finding_list", kwargs={"organization_code": self.organization.code}),
+        )
+
+    def _generate_report_file_name(self):
+        file_name = "_".join(["bevindingenrapport_nl", self.organization.code, datetime.now(timezone.utc).isoformat()])
+        # allow alphanumeric characters, dashes and underscores, replace rest with underscores
+        file_name = re.sub("[^0-9a-zA-Z-]", "_", file_name)
```
ammar92, line 290:
```suggestion
        file_name = re.sub("[^\w-]", "_", file_name)
```

rocky/rocky/views/ooi_report.py
```diff
@@ -198,47 +243,54 @@ def get(self, request, *args, **kwargs):
...
+
+        return self.get_report(
+            "Organization",
+            self.organization.name,
+            {finding.primary_key: finding for finding in findings},  # Create store from finding list
+            self._generate_report_file_name(),
+            reverse("finding_list", kwargs={"organization_code": self.organization.code}),
+        )
+
+    def _generate_report_file_name(self):
```
ammar92, line 287: I was wondering if this replacement is necessary since neither of the joined elements contain these characters (except for the colon used in the datetime's format, but that doesn't necessarily justify a regex replacement since you can format your own datetime string)

rocky/rocky/views/ooi_report.py
```diff
@@ -186,9 +184,56 @@ def get_context_data(self, **kwargs):
...
+        for finding_type, finding_group in findings_grouped.items():
+            list_of_findings = []
+            for finding in finding_group["list"]:
+                # Either take the human_readable or the primary key of the OOI from the Finding in the store
+                if finding["ooi"]:
+                    ooi_field = str(finding["ooi"]["human_readable"])
+                else:
+                    ooi_field = str(store[finding["id"]].ooi)
+
+                updated_x = {**finding, "ooi": ooi_field}
```
ammar92, line 224: I think `updated` or `updated_finding` is a more appropriate name

rocky/rocky/views/ooi_report.py
Donnype: Not sure to be honest, but perhaps primary keys of ooi's could give some issues?

Donnype: Ah that's the other one. But organization codes can contain hyphens perhaps?

Donnype: Made some updates for this one :+1:

Donnype:
I have been able to fix the API route issue by creating a ViewSet from the view as well. This is a temporary fix though.

rocky/rocky/views/ooi_report.py
```diff
@@ -198,47 +249,64 @@ def get(self, request, *args, **kwargs):
...
+                "bevindingenrapport_nl",
+                self.organization.code,
+                datetime.now(timezone.utc).strftime(self.FILE_NAME_FRIENDLY_DATE_FORMAT),
+            ]
+        )
+
+        return f"{file_name}.pdf"
+
+
+class APIFindingReportPDFView(ViewSet, FindingReportPDFView):
```
dekkers, line 305: This is really ugly as hell. A viewset is something different than a standard Django class based view. You shouldn't create classes that inherent from both. It just doesn't make any sense wrt object oriented programming.

It will probably give a lot of issues, but it is hard to really understand all the implications of this which itself is a good reason we should not do this. One thing I can think of is error handling, if something goes wrong the view will add a message and redirect to the finding view. I don't think we should ever do something like that in an API.

dekkers (changes requested):

rocky/rocky/views/ooi_report.py
Donnype: I agree, this was a quick attempt to get somewhat of an endpoint but setting reporting and the API up properly is definitely a better decision at this stage.

dekkers (changes requested):

rocky/tools/management/commands/generate_report.py
```diff
@@ -0,0 +1,101 @@
...
+class Command(BaseCommand):
+    help = (
+        "Generates a Finding report for all findings in an organization. "
+        "Requires either the id (-i) or the unique code (-c) of an organization."
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument("--id", "-i", type=int, help="The primary key of the organization.")
+        parser.add_argument("--code", "-c", type=str, help="The unique organization code.")
+        parser.add_argument(
```
ammar92, line 34: Usually, in most CLI apps this option is consistently called `--output/-o`. Also, I think it's better to mention (in the help) that it writes to `stdout` by default, the user itself can decide what to do with it.

rocky/rocky/views/finding_list.py
```diff
@@ -1,41 +1,59 @@
...
 def sort_by_severity_desc(findings) -> List[Dict[str, Any]]:
     return sorted(findings, key=lambda x: x["risk_level_score"], reverse=True)
 
 
+def generate_findings_metadata(
+    findings: OOIList, severity_filter: Optional[List[RiskLevelSeverity]] = None
+) -> List[Dict[str, Any]]:
+    findings_meta = []
+
+    for finding in findings[: findings.count]:
```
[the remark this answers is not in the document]
Donnype: This effectively has the same effect as doing
```python
for finding in findings[: 99999999]:
    ...
```

Do we really want to implicitly have this hard limit set or should we perhaps let users of this class decide how to handle this performance issue, e.g. by doing the above?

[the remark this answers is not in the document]
Donnype: Let's handle further details when picking up [this](https://github.com/minvws/nl-kat-coordination/issues/555) ticket, (as you suggested in our chat)

Darwinkel:
- [x] I confirmed that the PR's advertised `feature` or `hotfix` works as intended;

I checked the original issue, and I think this PR satisifies zcert's initial requirements - the ability to generate reports programatically. 

### What works:
* "Download PDF" button
* Command line interface with various filters (launched from the Rocky container)

### What doesn't work:
* In the Dutch report, it says "Organization" while it should say "Organisatie"
image

dekkers (approved):

dekkers:
I created #602 the "Organization" instead of "Organisatie"

PR #473, opened 2023-03-12
Add codespell pre-commit hook and fix everything it finds

dekkers:


underdarknl:
This also touches some migrations. will that give us any headaches?

dekkers:
> This also touches some migrations. will that give us any headaches?

No. It only changes the text description of the choices which isn't saved in the database schema. If you change them in both the migration and the model definition it will work fine.

underdarknl (approved):

underdarknl (dismissed):

underdarknl (approved):

Issue #429, opened 2023-03-06
XTDB Out-of-memory Java heap space


zcrt:
**Describe the bug**
When running KAT 1.6.2 for a few days on multiple organisations, at some point the `findings` and `objects` pages become unreachable, resulting in `Server Error 500` after a timeout of multiple minutes.

**To Reproduce**
1. Run KAT 1.6.2 for a few days (with multiple organisations?)
2. Click on 'objects'
3. Enjoy `Server Error 500`

**Expected behavior**
KAT should quickly load the requested pages.

**OpenKAT version**
1.6.2

**Additional context**
In our case, it looked like the XTDB container crashed due to using more than 1GB of RAM, before its garbage could be properly collected.

Some logs:
```java
WARN  o.e.j.util.thread.QueuedThreadPool - 
java.lang.OutOfMemoryError: Java heap space
07:57:50.189 [qtp857410557-25-acceptor-0@ea59010-ServerConnector@71b38b11{HTTP/1.1, (http/1.1, h2c)}{0.0.0.0:3000}] WARN  o.e.jetty.server.AbstractConnector - 
java.lang.OutOfMemoryError: Java heap space
07:57:50.190 [qtp857410557-4736] WARN  org.eclipse.jetty.server.HttpChannel - /_xtdb/z-cert/query
java.lang.IllegalStateException: COMPLETED
	at org.eclipse.jetty.server.HttpChannelState.sendError(HttpChannelState.java:915)
	at org.eclipse.jetty.server.Response.sendError(Response.java:471)
	at juxt.clojars_mirrors.ring_jetty_adapter.v0v14v2.ring.adapter.jetty9$proxy_handler$fn__16448.invoke(jetty9.clj:59)
	at juxt.clojars_mirrors.ring_jetty_adapter.v0v14v2.ring.adapter.jetty9.proxy$org.eclipse.jetty.server.handler.AbstractHandler$ff19274a.handle(Unknown Source)
	at org.eclipse.jetty.server.handler.HandlerList.handle(HandlerList.java:59)
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:127)
	at org.eclipse.jetty.server.Server.handle(Server.java:516)
	at org.eclipse.jetty.server.HttpChannel.lambda$handle$1(HttpChannel.java:388)
	at org.eclipse.jetty.server.HttpChannel.dispatch(HttpChannel.java:633)
[...]

PR #488, opened 2023-03-14
implement ooi list performance optimization

Lisser:
## Changes
This PR improves the OOI List performance. This is achieved by removing the outer-join from the query, which basically provided default scan-profiles for OOI's without a scan profile. 

**Context:**  
Scan-profiles are calculated asynchronously by Octopoes. Ideally, a default empty scan profile is created in the `on_create_ooi` event hook. This usually happens pretty fast after creating an OOI. 

However, the previously described event processing might fail in case of system crash, DB downtime, etc. Fault tolerance is achieved here by a periodic `recalculate_scan_profiles` job.

In the worst case scenario, a scan-profile object is not present for an OOI between OOI creation and the next iteration of `recalculate_scan_profiles` (default interval is set to 60 seconds). 

To allow filtering of the OOI List based on scan profiles, a join is made to the scan-profile. This results in the simplest case in an inner join, so that only OOIs are retrieved that actually have a scan profile. During the scenario in which an OOI does not have a scan profile temporarily, the OOI will not appear in the list. The previous solution added default scan-profile at query time to prevent this. However, it appears that that resulted in extreme performance loss with larger datasets.

Hence the removal of this part of the query. This PR does reintroduce the worst case scenario of new OOI's not appearing, however it attempts to minimize it by adding the scan-profile in the `on_create_ooi` event hook.

_I ran into some obscure concurrency issues in the robot tests, so I introduced a slight timeout before starting the polling for the celery task completions._

https://github.com/minvws/nl-kat-coordination/issues/429

## Proof

Benchmark: total of 785 objects

Old: 7542ms
Screenshot 2023-03-14 at 16 49 31

New: 288ms
Screenshot 2023-03-14 at 16 48 52

-N/A

## Checklist for author(s):

If a non-trivial PR:

```

### What works:

### What doesn't work:

### Bug or feature?:
```

```
```

### DIFF
--- a/octopoes/tests/robot/CeleryMonitor.py
+++ b/octopoes/tests/robot/CeleryMonitor.py
@@ -1,4 +1,5 @@
 import threading
+from collections import defaultdict
 
 from celery import Celery
 from robot.api import logger
@@ -66,12 +67,19 @@ class Monitor:
             self._started.set()
             self._receiver.capture(limit=None, timeout=None, wakeup=True)
 
+    @property
+    def sum_by_type(self) -> dict:
+        sums = defaultdict(lambda: 0)
+        for task_type in self.tasks.values():
+            sums[task_type["type"]] += 1
+        return sums
+
 
 class CeleryMonitor:
     ROBOT_LIBRARY_SCOPE = "TEST"
 
     def __init__(self):
-        self._monitor = None
+        self._monitor: Monitor = None
 
     def start_monitoring(self, queue_uri: str):
         self._monitor = Monitor(queue_uri)
@@ -84,6 +92,7 @@ class CeleryMonitor:
     def count_tasks(self, event_type: str) -> int:
         event_type = "task-" + event_type.lower()
 
+        logger.info(self._monitor.sum_by_type)
         count = len(list(filter(lambda v: v["type"] == event_type, self._monitor.tasks.values())))
         logger.info(f"Counting tasks of type {event_type}: {count}")
 
--- a/octopoes/tests/robot/CeleryMonitor.py
+++ b/octopoes/tests/robot/CeleryMonitor.py
@@ -1,5 +1,6 @@
 import threading
-from collections import defaultdict
+from collections import Counter
+from operator import itemgetter
 
 from celery import Celery
 from robot.api import logger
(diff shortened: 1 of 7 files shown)


octopoes/tests/robot/CeleryMonitor.py
```diff
@@ -66,12 +67,19 @@ def _monitor_events(self, celery_app: Celery) -> None:
             self._started.set()
             self._receiver.capture(limit=None, timeout=None, wakeup=True)
 
+    @property
```
ammar92, line 70: Just a few remarks:
- I think based on the implementation this is not a typical property, but just a method (calc_x_.., get_sum_by_type, ...)
- `defaultdict(int)` would also suffice, since `int()` gives int 0 as well

octopoes/tests/robot/CeleryMonitor.py
```diff
@@ -84,6 +92,7 @@ def stop_monitoring(self):
     def count_tasks(self, event_type: str) -> int:
         event_type = "task-" + event_type.lower()
 
+        logger.info(self._monitor.sum_by_type)
```
ammar92, line 95: Perhaps add some text to make clear what the output is? Also, the `msg` parameter is usually a string

ammar92 (commented):
Cool stuff! Just a few remarks

I was also wondering since the query is much different, do we really not lose any functionality or data by removing these OR and AND conditions in the query?

octopoes/tests/robot/CeleryMonitor.py
ammar92: To make it even fancier:
```
from collections import Counter
from operator import itemgetter
sums = Counter(map(itemgetter("type"), self.tasks.values())
```

Lisser: 😻

Lisser: Nice one, thanks. Implemented :)

octopoes/tests/robot/CeleryMonitor.py
Lisser: Implemented :) Robot does not allow standard logger formatting (`'%s'` notation), so I used a n f-string instead

ammar92 (approved):
Looks good to me 👍

Darwinkel:

### What works:
* There do not appear to be any functional regressions
* Haven't run any objective benchmarks - but with 500+ objects the UI still feels fast

Darwinkel (approved):

dekkers (approved):

PR #536, opened 2023-03-21
Fix/testing different users

Rieven:
## Changes
While making tests for different users I came across other tests where superuser is used instead of pinpointing what type of member is requesting a view, it is important because different members has different roles and different permissions. For example: a redteamer should be able to upload a CSV with objects to scan that has scan permissions. This PR changes the general use of superuser everywhere.

thanks @Donnype for helping out!


Donnype (dismissed):
Looks like a good refactor given the new fixtures :+1:. It wouldn't hurt to duplicate tests that trigger some crucial superuser/redteamer only functionality with a `client_member` and verify that regular clients do not have the right permissions. But I understand if this is out of scope.

dekkers (dismissed):

dekkers (approved):

PR #548, opened 2023-03-22
Upgrade to manon v15.1.3 (Layout)

HeleenSG:
Updated to manon v15.1.3. Added main layout based on spacing and grid as designed.

## Changes
- Updated to manon v15.1.3 (Which contains the new manon open layout.) 
- Upgrading new kat layout. Based on the figma designs. https://www.figma.com/file/WdyhI30BFLptnts8zaiYYu/KAT-%7C-Design-System-%7C-Fundamentals?node-id=207%3A10843&t=93W2RhhIJvZAva7S-1
- Tile styling
   - Cover image border radius
   - Font size
   - Spacing
 - Added spacing set
 - Added body text set
 
## Todo 

- [x] Main
- [x] Section
- [x] Header
- [x] Main nav
- [x] Footer

```
```

Solves: 
https://github.com/minvws/nl-kat-coordination/issues/888
https://github.com/minvws/nl-kat-coordination/issues/894
https://github.com/minvws/nl-kat-coordination/issues/876
https://github.com/minvws/nl-kat-coordination/issues/882
https://github.com/minvws/nl-kat-coordination/issues/885
https://github.com/minvws/nl-kat-coordination/issues/883


rocky/assets/src/bundles/app/css/themes/soft/soft.scss
```diff
@@ -24,7 +24,7 @@
 @import "manon/button-ghost"; /* Ghost button */
 @import "manon/button-icon"; /* Icon button */
 @import "manon/filter";
-@import "manon/layout";
+@import "manon/layout-set";
```
dekkers, line 27: If I understand this correctly manon/layout.scss is not used anymore and the file can be deleted?

HeleenSG: Yes that is true. But for the moment I kept the file as a reference until the transition between the two is complete. I will change the status on this pr from draft to ready for review once it's ready to be checked.

HeleenSG:
Thanks for testing @Darwinkel 

- **Double "Select filters" button.** This isn't a regression as it is visible on main. I agree we should fix this but as it is outside of the scope of this pr I would like suggest doing so on a separate pr. 
- **It's not apparent that URL/Type is not fillable. Could it be greyed-out or something?** This input field is missing a "disabled" state. But as this is also an issue in main I would suggest solving this in a separate pr as well. 
- **Perhaps slightly increase the paw contrast to a darker shade of purple for all levels?** Agreed, and fixed.
- **Maybe align the Voorvallen more to the center?** Agreed, that element needed some attention. Is fixed now. 
- **Checked checkbox could use some increased contrast. I'm guessing you meant radio buttons?** I fixed that one. If you did mean checkboxes we should look at the design. For now I have added the yellow on both. I also fixed the button alignment but that will be visible after the next manon release as it needed to be fixed there. So I suggest releasing manon and upgrading the version one more time on this pr before merging. 
- **Spacing difference between the first and second row in the Status column (in this case, Active and Blocked).** Agreed and fixed. 
- **Significant whitespace between OpenKAT and Privacyverklaring in the footer.** This is the designed spacing or am I missing something that is broken?

Darwinkel:
> Thanks for testing @Darwinkel
> 
>     * **Double "Select filters" button.** This isn't a regression as it is visible on main. I agree we should fix this but as it is outside of the scope of this pr I would like suggest doing so on a separate pr.
> 
>     * **It's not apparent that URL/Type is not fillable. Could it be greyed-out or something?** This input field is missing a "disabled" state. But as this is also an issue in main I would suggest solving this in a separate pr as well.
> 
>     * **Perhaps slightly increase the paw contrast to a darker shade of purple for all levels?** Agreed, and fixed.
> 
>     * **Maybe align the Voorvallen more to the center?** Agreed, that element needed some attention. Is fixed now.
> 
>     * **Checked checkbox could use some increased contrast. I'm guessing you meant radio buttons?** I fixed that one. If you did mean checkboxes we should look at the design. For now I have added the yellow on both. I also fixed the button alignment but that will be visible after the next manon release as it needed to be fixed there. So I suggest releasing manon and upgrading the version one more time on this pr before merging.
> 
>     * **Spacing difference between the first and second row in the Status column (in this case, Active and Blocked).** Agreed and fixed.
> 
>     * **Significant whitespace between OpenKAT and Privacyverklaring in the footer.** This is the designed spacing or am I missing something that is broken?
> 
>     * **I can't find the button to create a new organization anymore** This is incoming with @TwistMeister's work.
> 
>     * **Changing the language doesn't work on all pages.** @TwistMeister I haven't looked at this one just yet.

Thanks for all the fixes! I created some new issues for the remaining ones:
- https://github.com/minvws/nl-kat-coordination/issues/1145
- https://github.com/minvws/nl-kat-coordination/issues/1146

>I'm guessing you meant radio buttons?

Ah yes indeed, apologies :)

>Significant whitespace between OpenKAT and Privacyverklaring in the footer

Oops, looks like this was my bad. I thought they were one menu entry.

dekkers (approved):

underdarknl (approved):

PR #549, opened 2023-03-22
pass on underscore domains

noamblitz:


underdarknl (approved):
We need to fix this properly in the model for 1.8

Issue #486, opened 2023-03-14
Mula, keep track of the changeset of OOI_types when new plugins are enabled and query for those objects


underdarknl:
When Mula refreshes its KAT-alogus info, it should keep track of the changeset of OOI_Types, (and possibly related minimal scan levels). This elta can be used to directly ask Octopoes for objects that would from then on be eligible for scheduling.
This would give our users a better instant satisfaction when enabling new boefjes

jpbruinsslot:
Related: https://github.com/minvws/nl-kat-coordination/issues/204

jpbruinsslot:
To get ooi's and boefjes scheduled, when a new boefje gets enabled I would recommend leveraging the work on this pr into a feature branch https://github.com/minvws/nl-kat-coordination/blob/a1d041d072839a478fc5c0cc948c67481f4b8df7/mula/scheduler/schedulers/boefje.py#L193

This requires a change in octopoes https://github.com/minvws/nl-kat-coordination/issues/227 to relay a list of ooi's from an organization that an enabled boefje can run on. `get_objects_by_object_type()` @Lisser  thoughts?

I'm a bit hesitant to add another accounting/cache to keep track of in the scheduler. And I think this would solve the problem of "directly ask Octopoes for objects that would from then on be eligible for scheduling."

PR #574, opened 2023-03-27
Feature/mula new boefjes

jpbruinsslot:
Related issue: https://github.com/minvws/nl-kat-coordination/issues/486


mula/scheduler/connectors/services/katalogus.py
```diff
@@ -26,13 +41,21 @@ def _flush_organisations_plugin_cache(self) -> None:
...
 
         for org in orgs:
+            if org.id not in self.organisations_plugin_cache:
+                self.organisations_plugin_cache[org.id] = {}
+                self.organisations_new_boefjes_cache[org.id] = {}
+
+            plugins = self.get_plugins_by_organisation(org.id)
             self.organisations_plugin_cache[org.id] = {
-                plugin.id: plugin for plugin in self.get_plugins_by_organisation(org.id)
+                plugin.id: plugin for plugin in plugins if plugin.enabled is True
```
ammar92, line 50:
```suggestion
                plugin.id: plugin for plugin in plugins if plugin.enabled
```

mula/scheduler/connectors/services/katalogus.py
```diff
@@ -42,6 +65,9 @@ def _flush_organisations_boefje_type_cache(self) -> None:
                 if plugin.type != "boefje":
                     continue
 
+                if plugin.enabled is False:
```
ammar92, line 68:
```suggestion
                if not plugin.enabled:
```

mula/scheduler/schedulers/boefje.py
```diff
@@ -242,166 +247,328 @@ def push_tasks_for_scan_profile_mutations(self) -> None:
...
+            self.logger.warning(
+                "Could not connect to rabbitmq queue: %s [org_id=%s, scheduler_id=%s]",
+                f"{self.organisation.id}__scan_profile_increments",
+                self.organisation.id,
+                self.scheduler_id,
+            )
+            if self.stop_event.is_set():
+                raise e
+
+        if new_boefjes is None or len(new_boefjes) == 0:
```
ammar92, line 275:
```suggestion
        if new_boefjes is None or not new_boefjes:
```

mula/scheduler/connectors/services/katalogus.py
```diff
@@ -115,3 +155,26 @@ def get_normalizers_by_org_id_and_type(self, organisation_id: str, normalizer_ty
...
+        # Get the enabled boefjes for the organisation from katalogus
+        plugins = self.get_plugins_by_organisation(organisation_id)
+        enabled_boefjes = {
+            plugin.id: plugin for plugin in plugins if plugin.enabled is True and plugin.type == "boefje"
+        }
+
+        # Check if there are new boefjes
+        new_boefjes = []
+        for boefje_id, boefje in enabled_boefjes.items():
+            if boefje_id in self.organisations_new_boefjes_cache.get(organisation_id, {}):
```
underdarknl, line 169: Would this not see each boefje as a new boefje if the cache has been cleaned? This would result in a large query to Octopoes for all active boefje's their cumulative OOI-Type set instead of just the actual change.
Im sot sure a Cache is the right way to do this. Maybe a set compare to the last run is better?

underdarknl (commented):
Is a cache really the way forward here?

mula/tests/integration/test_boefje_scheduler.py
```diff
@@ -258,6 +258,214 @@ def test_push_tasks_for_scan_profile_mutations_item_on_queue(
         self.assertEqual(task_db.id.hex, task_pq.id)
         self.assertEqual(task_db.status, models.TaskStatus.QUEUED)
 
+    @mock.patch("scheduler.schedulers.BoefjeScheduler.is_task_running")
+    @mock.patch("scheduler.schedulers.BoefjeScheduler.is_task_allowed_to_run")
+    @mock.patch("scheduler.schedulers.BoefjeScheduler.has_grace_period_passed")
```
ammar92, line 263: These lines are repeated for every test. If I'm not mistaken, you can also decorate the entire `TestCase` class.

mula/tests/integration/test_connectors.py
ammar92, line 1: What does this do?

ammar92 (commented):
Looks good in general, nice tests. Just some styling remarks. Please see the last two paragraphs of this section https://pep8.org/#programming-recommendations

mula/scheduler/connectors/services/katalogus.py
jpbruinsslot: It sets the `enabled_boefjes` as the current collection of boefjes (checked at that moment). It should at a new call of `get_new_boefjes_by_org_id` check the difference since the last call to the method.

mula/tests/integration/test_connectors.py
jpbruinsslot: removed it

mula/tests/integration/test_boefje_scheduler.py
jpbruinsslot: Picked up in https://github.com/minvws/nl-kat-coordination/issues/591

mula/tests/integration/test_connectors.py
```diff
@@ -0,0 +1,6 @@
+import unittest
+
+
+class KatalogusTestCase(unittest.TestCase):
+    def test_truthy(self):
+        self.assertTrue(True)
```
dekkers, line 6: This doesn't look very useful?

Darwinkel:
No functional regressions, the scheduler runs correctly as usual. But I cannot say anything feels different from a user perspective either.

underdarknl (approved):

PR #600, opened 2023-03-29
Add indices for Bytes

Donnype:
## Changes
Add an index on the `boefje_meta` table: (organization, boefje_id). This due to the high load for a large amount of boefje_meta entries in Bytes, which was called several times per second through the API on staging.


ammar92 (dismissed):
Good one; should probably have been indexed the first time. Most foreign keys are usually indices

Donnype:
Checking whether an index on the raw_file table makes sense at this point!

dekkers (dismissed):

underdarknl (approved):

Issue #582, opened 2023-03-27
Add RBAC support


TwistMeister:
Add support for Role-Based Access Control (RBAC) where we can support both organization specific roles as well as roles transcending the organization. For this it is is needed that we have a `groups` field on both `user` and `organization_member` models. This will also make it possible for users to have different roles for different organizations. It will also make it possible to define custom roles and/or change the default roles.

The Django `Group` model will function as role. To make it possible for users to add/delete/change the roles (groups), all places where we currently check if the user is in a group must be refactored to check whether the user has the permission. The user has a permission if either one of the groups on the `User` objects has the permission or one of the groups on the `OrganisationMember` object for the specific organization has the permission.

### Implementation

- [x] Copy the "groups" field from the `user` model to the `organization_member` model;
- [x] Copy the current values from the `user.groups` to the `organization_member.groups`;
- [x] `user.groups` can be cleared (not deleted! Just empty this field)
- [x] Create our own version of the `PermissionRequiredMixin` where override `has_permission` that will also check the `OrganisationMember` permissions.
- [x] Create our own template tag similar to the Django built-in `perms` tag to check for permission inside templates. This template tag will also need the organization as argument.
- [x] All places that currently check for the groups need to be refactored to check for permission. The mixin and template tag should be used where possible.
- [x] Enable setting both `user.groups` and `organization_member.groups` when adding a user to an organization. `user.groups` can only be set by a superuser (global admin). `organization_member.groups` can be set by users which have the `add_organizationmember` permission.

TwistMeister:
https://github.com/minvws/nl-kat-coordination/pull/606

PR #606, opened 2023-03-29
Groups and Permissions for Organization member

Rieven:
## Changes

- Add groups to the organization member model
- Migrate/update groups from the user model to the organization member model
- Clear all groups from the user model
- Permissions are set on groups, so check permissions using the organization member groups instead of user groups.
- Change PermissionrequiredMixin to our own custom RockyPermissionMixin to check permissions not only based on users but also for organization members.
- Change all views needing permissions to use the RockyPermissionMixin
- Change template based permissions to use our own custom template tag to check also for organization member permissions.
- Permissions set on user level overrule organization member permissions

https://github.com/minvws/nl-kat-coordination/issues/582

## Proof
Groups in member
image

Group in user, same as member:
image

_This section may be skipped or omitted. Uncomment and answer the below questions if relevant._

## Checklist for author(s):

If a non-trivial PR:

```

### What works:

### What doesn't work:

### Bug or feature?:
```

```
```


rocky/tools/models.py
```diff
@@ -161,6 +161,7 @@ class STATUSES(models.TextChoices):
 
     user = models.ForeignKey("account.KATUser", on_delete=models.PROTECT, related_name="members")
     organization = models.ForeignKey(Organization, on_delete=models.CASCADE, related_name="members")
+    groups = models.ManyToManyField(Group, related_name="members", blank=True)
```
TwistMeister, line 164: I feel like it would be logical that this defaults to 'client', since that's the group with the least permissions, and I think groups should never be empty. But I'm curious what @dekkers, @underdarknl or @praseodym think of this.

praseodym: Why don't we remove `blank=True` and always require a group to be set? I don't think we'd need a default, we should be able to ensure a group is always set.

TwistMeister: Yes, that's maybe even better. That enforces the user to make a decision. I like that suggestion.

Rieven: Here is decided that we can have a none group member.

rocky/tools/models.py
```diff
@@ -170,6 +182,55 @@ class STATUSES(models.TextChoices):
...
+        codenames = []
+        if permissions is None:
+            raise ImproperlyConfigured("Permission(s) not defined")
+        if isinstance(permissions, str):
+            perms = (permissions,)
+        else:
+            perms = permissions
+        for perm in perms:
+            try:
+                _, codename = perm.split(".", 1)
```
ammar92, line 207: Wouldn't counting the number of dots be simpler?
```if perm.count(".") != 1: raise ValueError```

File: rocky/tools/models.py [hunk 3]
```diff
@@ -170,6 +182,55 @@ class STATUSES(models.TextChoices):
...
+
+    @property
+    def is_redteam(self):
+        return self.groups.filter(name=GROUP_REDTEAM).exists()
+
+    @property
+    def is_client(self):
+        return self.groups.filter(name=GROUP_CLIENT).exists()
+
+    def get_permissions_codenames(self, permissions):
```
ammar92, line 197: Where are the tests for this? Based on just the code I can see that `"app_label."` also works, which contradicts with the error message

rocky/tools/models.py
```diff
@@ -170,6 +182,55 @@ class STATUSES(models.TextChoices):
...
+    def is_admin(self):
+        return self.groups.filter(name=GROUP_ADMIN).exists()
+
+    @property
+    def is_redteam(self):
+        return self.groups.filter(name=GROUP_REDTEAM).exists()
+
+    @property
+    def is_client(self):
+        return self.groups.filter(name=GROUP_CLIENT).exists()
```
ammar92, line 195: I think these properties should be cached. Also, are they still necessary since with this approach we don't have to check for fixed groups anymore but permissions?

Rieven: Indeed caching here is a good idea. We use these props in just a few cases where we want to know if a user is in a particular group for redirecting purposes, mostly used in onboarding.

Darwinkel:
## Checklist for QA:

### What works:
* No apparent regressions
* No issues with existing user management

rocky/tools/models.py
dekkers: Those checks should be changed, because we should not hardcode these groups and that means those groups don't have to exist.

Rieven: Here decided that we are going to fix this in another PR. We have to think about the design of how we are going to use these groups in the future.

rocky/tools/models.py
```diff
@@ -157,6 +171,7 @@ class STATUSES(models.TextChoices):
 
     user = models.ForeignKey("account.KATUser", on_delete=models.PROTECT, related_name="members")
     organization = models.ForeignKey(Organization, on_delete=models.CASCADE, related_name="members")
+    groups = models.ManyToManyField(Group, related_name="members", blank=True)
```
dekkers, line 174: I think "members" as related name for groups is confusing as it won't really give the members of the group, but the relation between organization, user and group. I think the default organizationmember_set would be fine if we would use it in the code.

rocky/tools/models.py
```diff
@@ -170,6 +185,55 @@ class STATUSES(models.TextChoices):
...
+
+    def has_member_perms(self, permissions: Union[str, tuple] = None) -> bool:
+        if self.user.is_superuser and self.user.is_active and self.status is not OrganizationMember.STATUSES.BLOCKED:
+            return True
+        codenames = self.get_permissions_codenames(permissions)
+        has_permissions = []
+
+        for codename in codenames:
+            for group in self.groups.all():
+                group_permissions_codenames = group.permissions.values_list("codename", flat=True)
```
dekkers, line 227: This is wrong. Both the app and codename should match. The permission "app1.can_do_something" is not the same as a permission "app2.can_do_something". It also does too many queries.

We can get all the permissions with a single query:

```
perms = ["%s.%s" % (ct, name) for ct, name in Permission.objects.filter(group__organizationmember=member).values_list("content_type__app_label", "codename")]
```

rocky/tools/migrations/0034_update_perms_move_and_clear_groups.py
```diff
@@ -0,0 +1,55 @@
...
+
+User = get_user_model()
+
+
+# https://stackoverflow.com/a/40092780/1336275
+def migrate_permissions(apps, schema_editor):
+    for app_config in apps.get_app_configs():
+        app_config.models_module = True
+        create_permissions(app_config, apps=apps, verbosity=0)
+        app_config.models_module = None
```
dekkers, line 17: There is no need for this in this migration as far as I can see. Either the admin and redteam group exists, but then we have already run migrate before and the permissions have been created, or the groups don't exist but then we also don't need the permissions.

rocky/tools/migrations/0034_update_perms_move_and_clear_groups.py
```diff
@@ -0,0 +1,55 @@
...
+    except Group.DoesNotExist:
+        pass
+
+
+def migrate_user_groups_to_organizationmember(apps, schema_editor):
+    members = OrganizationMember.objects.all()
+    for member in members:
+        user_groups = member.user.groups.all()
+        for group in user_groups:
+            member.groups.add(group)
```
dekkers, line 38: This can be done in one query using:

`member.groups.add(*user_groups)`

dekkers (changes requested):

rocky/tools/models.py
```diff
@@ -170,6 +185,55 @@ class STATUSES(models.TextChoices):
...
+                    if ContentType.objects.filter(app_label=app_label, model="organization").exists():
+                        codenames.append(codename)
+            else:
+                raise ValueError("Permissions should be defined as: 'app_label.codename'.")
+        return codenames
+
+    def has_member_perms(self, permissions: Union[str, tuple] = None) -> bool:
+        if self.user.is_superuser and self.user.is_active and self.status is not OrganizationMember.STATUSES.BLOCKED:
+            return True
+        codenames = self.get_permissions_codenames(permissions)
```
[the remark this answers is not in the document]
Rieven: This is how it is implemented now. The permission parameter comes in as string or a tuple of string perms, the `get_permissions_codenames` will split the app_label and perm and see if it exists. According to your other comments this part will be refactored to a more simpler approach.

rocky/account/templates/account_detail.html
```diff
@@ -63,7 +64,8 @@ <h1>{% translate "Account details" %}</h1>
                 </dl>
             </div>
         </section>
-        {% if perms.tools.can_scan_organization %}
+        {% has_organization_perms "tools.can_scan_organization" request.user organization as can_scan_organization %}
+        {% if can_scan_organization %}
```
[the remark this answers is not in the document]
Rieven: I have made a MemberPermissionMixin where we can pass it to the context as an instance like:  `perms = MemberPermissionMixin()`

then we can do `perms.has_member_perms("tools.can_scan_organization", member)`

dekkers:
I implemented organization member permissions check in the templates in the same way as django itself does using the `perms` context variable. In the OrganizationView we set the `perms` context variable which will overrule the context processor. This means that we can just use `perms` everywhere and for views that inherit from OrganizationView this means that the permissions from the OrganizationMember groups are added to the perms.

I also moved the organization member permissions checks to the OrganizationMember model as this is where they belong. This makes it also easy to cache the permissions similar as how it done on the User model to minimize database queries.

I also renamed `RockyPermissionRequiredMixin` to `OrganizationPermissionRequiredMixin` as the mixin is specific to the organization views and not for all of rocky.

Rieven:
> I implemented organization member permissions check in the templates in the same way as django itself does using the `perms` context variable. In the OrganizationView we set the `perms` context variable which will overrule the context processor. This means that we can just use `perms` everywhere and for views that inherit from OrganizationView this means that the permissions from the OrganizationMember groups are added to the perms.
> 
> I also moved the organization member permissions checks to the OrganizationMember model as this is where they belong. This makes it also easy to cache the permissions similar as how it done on the User model to minimize database queries.
> 
> I also renamed `RockyPermissionRequiredMixin` to `OrganizationPermissionRequiredMixin` as the mixin is specific to the organization views and not for all of rocky.

Very well implemented!

praseodym:
In the members view I can no longer see which role a member has:
image

Rieven:
> In the members view I can no longer see which role a member has: image

fixed in commit: https://github.com/minvws/nl-kat-coordination/pull/606/commits/fb293e068a6887fbdbb5ab4c4f668a9240701877

praseodym:
lgtm!

praseodym (approved):

Darwinkel:
## Checklist for QA:


### What doesn't work:
* After creating a new redteamer, logging in, and doing the onboarding:
image
* As a redteamer without any clearance, I can start L1+ scans. This issue remains after an admin has modified the assigned clearance level. In other words: authorization is broken. 
* To make matters worse, I can, as a client, initiate scans.

### Bug or feature?:
* Not sure if redteamers should be onboarded if being created by an admin. Currently, logging in as a redteamer triggers a flow which encourages a new user to add a new object.

Rieven:
As a redteamer without any clearance, I can start L1+ scans. This issue remains after an admin has modified the assigned clearance level. In other words: authorization is broken.

@Darwinkel 
All and all good catch! This problem is not directly related to this feature as it is also broken in main. The problem here is that if admin creates a new redteam member without setting clearance levels, then a redteam will encounter this problem anyways, because there is no acknowledgement for a scan level L2.  No levels are set for this new member. For the onboarding purpose we need to create an extra screen where the redteamer must acknowledge the L2 to scan for DNS. 

This is something not yet implemented and should have been thought off in the first place.

Darwinkel:
## Checklist for QA:


### What works:
- Clients can no longer initiate scans. I talked with @TwistMeister and there is something to be said for a client being able to run (some) scans, as this is also what the scheduler does in the background. It may be relevant for them to generate a more recent report with the newest information. The same essentially goes for low-clearance redteamers: shouldn't they be able to perform manual scans as long as this is something that the scheduler would trigger by itself?

### What doesn't work:
* https://github.com/minvws/nl-kat-coordination/issues/915
* https://github.com/minvws/nl-kat-coordination/pull/783
* https://github.com/minvws/nl-kat-coordination/issues/908

Rieven:
Shouldn't they be able to perform manual scans as long as this is something that the scheduler would trigger by itself?

No, they shouldn't indeed....Clearance checks should be checked for manual scans.

dekkers (approved):

PR #621, opened 2023-03-30
Fixes the redirect to the onboarding flow

### LINKED ISSUE #620: Onboarding flow not working
**Describe the bug**

- After OpenKAT install and login in with superuser, it will not redirect to the onboarding flow.
- There is a crisis room link at to unauthorized screen which is not needed in onboarding:
image
- 

**To Reproduce**
Do a fresh install `make kat` and after login user should redirect to the admin/superuser onboarding 
- unauthorized link was found by accident/coincidence while testing user groups.

TwistMeister:
https://user-images.githubusercontent.com/16188579/228907718-276aef1c-88d0-4d93-b5be-11923fcb7f11.mov

Resolves #620

### DIFF
--- a/rocky/rocky/middleware/onboarding.py
+++ b/rocky/rocky/middleware/onboarding.py
@@ -9,7 +9,8 @@ def OnboardingMiddleware(get_response):
     def middleware(request):
         response = get_response(request)
         if request.user.is_authenticated:
-            member_onboarded = OrganizationMember.objects.filter(user=request.user, onboarded=True)
+            member_onboarded = list(filter(lambda o: o.onboarded, request.user.organization_members))
+
             # do not redirect itself, otherwise it will endup in endless loop
             # with too many redirects
             # exclude admin urls
--- a/rocky/rocky/middleware/onboarding.py
+++ b/rocky/rocky/middleware/onboarding.py
@@ -9,7 +9,7 @@ def OnboardingMiddleware(get_response):
     def middleware(request):
         response = get_response(request)
         if request.user.is_authenticated:
-            member_onboarded = filter(lambda o: o.onboarded, request.user.organization_members)
+            member_onboarded = OrganizationMember.objects.filter(user=request.user, onboarded=True)
             # do not redirect itself, otherwise it will endup in endless loop
             # with too many redirects
             # exclude admin urls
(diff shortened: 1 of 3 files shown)


rocky/rocky/middleware/onboarding.py
```diff
@@ -9,7 +9,7 @@ def OnboardingMiddleware(get_response):
     def middleware(request):
         response = get_response(request)
         if request.user.is_authenticated:
-            member_onboarded = filter(lambda o: o.onboarded, request.user.organization_members)
+            member_onboarded = OrganizationMember.objects.filter(user=request.user, onboarded=True)
```
praseodym, line 12: The old code was there so the cached_property would be used, avoiding a SQL query here. But I'm not sure why it didn't work..

rocky/rocky/middleware/onboarding.py
```diff
@@ -9,7 +9,8 @@ def OnboardingMiddleware(get_response):
     def middleware(request):
         response = get_response(request)
         if request.user.is_authenticated:
-            member_onboarded = filter(lambda o: o.onboarded, request.user.organization_members)
+            member_onboarded = list(filter(lambda o: o.onboarded, request.user.organization_members))
```
underdarknl, line 12: I think we should move this to be a property on the organisationMember, which should be a property of the user available trough request.
So available as:
request.user.member_onboarded

Which is loaded when needed, and cached for the duration of the request.

underdarknl (dismissed):

rocky/rocky/middleware/onboarding.py
praseodym: `request.user.organization_members` is already a cached property of `KATUser`; what advantage would moving this line to a property have?

underdarknl: I like keeping data local to where other related data lives. In this case, this member_onboarded flag is an abstraction of the data on the organization_members list and should be available from that data. It makes the API cleaner, and makes sure there's no need to reinvent this code if we want to know if a user is onboarded anywhere else.
come to think of it, it should just be a property on the user object, not of the organization_members property.

dekkers: I agree this would be cleaner API. It would also be better to use `any(o.onboarded for o in organization_members)` which will result in a boolean instead of making a list.

praseodym: If we want to know if an `OrganizationMember` is onboarded, we can simply look at its `onboarded` property.
`member_onboarded` is a list of onboarded `OrganizationMember`s, which we only use to look if the current user is onboarded on any organization (we only onboard the user on the first organization?).

We could make this a boolean flag and move it to the `KATUser` if we agree that's better?

dekkers (approved):

PR #629, opened 2023-03-31
Don't set KAT_VERSION by default in .env-dist

dekkers:


underdarknl (commented):
Do we default to Main / unstable when no version is set?

dekkers:
> Do we default to Main / unstable when no version is set?

Given that it is an example I think it is okay to not have a default. I also think defaulting to main / unstable for an example to be used in production is not a good idea.

underdarknl (approved):

PR #645, opened 2023-04-03
Feature/no bits without clearence

noamblitz:
## Changes
Only run bits on oois with matching clearence level. Problem is that oois dont get a clearence upon creation, so inference now also runs when:
- scan level or source ooi gets changed
- oois get a scan level inheritance
- scan levels are reset
- scan levels are unset


octopoes/octopoes/core/service.py
```diff
@@ -129,10 +129,18 @@ def _run_inference(self, origin: Origin, valid_time: datetime):
...
         try:
-            resulting_oois = BitRunner(bit_definition).run(source, list(parameters.values()))
+            try:
+                level = self.scan_profile_repository.get(source.reference, valid_time).level
+            except ObjectNotFoundException:
+                level = 0
+            if level >= 1:
+                resulting_oois = BitRunner(bit_definition).run(source, list(parameters.values()))
+            else:
+                resulting_oois = []
```
ammar92, line 139: This would be better:
```python3
try:
    level = ...
    if level > 0:
        resulting_oois = ...
except ...:
    resulting_oois = []
```

noamblitz: I refactored this part a bit, better?

ammar92: The condition check wasn't the issue, but the code blocks

Darwinkel:

### What works:
* There do not appear to be regressions
* No errors when setting scan levels and changing OOI's
* Findings are correctly re-created after changing scan levels and deleting them

octopoes/octopoes/core/service.py
```diff
@@ -128,11 +128,20 @@ def _run_inference(self, origin: Origin, valid_time: datetime):
...
+        except ObjectNotFoundException:
+            level = 0
+
+        if level >= bit_definition.min_scan_level:
+            try:
+                resulting_oois = BitRunner(bit_definition).run(source, list(parameters.values()))
+            except Exception as e:
+                logger.exception("Error running inference", exc_info=e)
+                return
+
```
Lisser, line 144: @ammar92 

The logic here: when a level is not found (could be created with a slight delay, asynchronously), it should default to 0. Then proceed with the normal flow of checking the scan level and running the Bit

Lisser:
I changed to logic to match the OOI scan level with the Bit's min_scan_level. Because some Bits, like url_classification, should always run.

ammar92 (approved):

underdarknl (approved):

PR #662, opened 2023-04-05
Implement openmetrics endpoint in mula

jpbruinsslot:
Example output:

```
$ curl http://localhost:8004/metrics

# HELP app_settings_info Scheduler configuration settings
# TYPE app_settings_info gauge
app_settings_info{katalogus_cache_ttl="30",monitor_organisations_interval="60",pq_maxsize="1000",pq_populate_grace_period="86400",pq_populate_interval="60",pq_populate_max_random_objects="50"} 1.0
# HELP scheduler_qsize Size of the scheduler queue
# TYPE scheduler_qsize gauge
scheduler_qsize{scheduler_id="normalizer-dev"} 0.0
scheduler_qsize{scheduler_id="boefje-dev"} 0.0

```


mula/scheduler/app.py
```diff
@@ -252,8 +252,41 @@ def run(self) -> None:
...
+                scheduler_id=s.scheduler_id,
+            ).set(
+                s.queue.qsize(),
+            )
+
+        # Collect metrics from the katalogus
+        k = self.ctx.services.katalogus
+        self.ctx.metrics_katalogus.info(
+            {
+                "plugin_cache_expiry_time": k.organisations_plugin_cache.start_time.isoformat(),
```
praseodym, line 285: These should be separate metrics, not labels on a single metric. Using labels with high cardinality values creates separate time series that can't be analysed but do result in high storage costs.

A good example of a metric and label combination would be `http_requests_total{method="$method"}` where `$method` could be `GET`, `POST` or `DELETE` and the value of the metric is the total number of http requests for each method.

https://prometheus.io/docs/practices/naming/#labels has more information about metrics and labels usage.

mula/scheduler/context/context.py
```diff
@@ -85,6 +84,39 @@ def __init__(self) -> None:
...
+        # Metrics collector registry
+        self.metrics_registry: CollectorRegistry = CollectorRegistry()
+
+        Info(
+            name="app_settings",
+            documentation="Scheduler configuration settings",
+            registry=self.metrics_registry,
+        ).info(
+            {
+                "monitor_organisations_interval": str(self.config.monitor_organisations_interval),
```
praseodym, line 103: These should also be separate metrics instead of labels.

praseodym (changes requested):

mula/scheduler/context/context.py
```diff
@@ -30,9 +31,6 @@ def __init__(self) -> None:
         """Initializer of the AppContext class."""
         self.config: settings.Settings = settings.Settings()
 
-        if not self.config.database_dsn.startswith("postgresql"):
```
praseodym, line 33: I think this accidentally reverse a change.

jpbruinsslot: Initially, it was removed accidentally, however, I've moved it a bit lower on line 87

mula/scheduler/app.py
jpbruinsslot: Cool, thanks! I'll have a look at it.

mula/scheduler/context/context.py
jpbruinsslot: Is that also true for values of configuration settings? I guess this should probably be known by the person who configured the scheduler., and should not be present at all in a metrics endpoint. But thought of these settings as: `"version": "0.0.1"` etc. hence I opted for `Info`.

I referenced the example of

```python
from prometheus_client import Info
i = Info('my_build_version', 'Description of info')
i.info({'version': '1.2.3', 'buildhost': 'foo@bar'})
```

praseodym: I see! Since these are config values they shouldn't change often so that should be fine. Apparently the [Prometheus docs also include `foobar_build_info` as an example](https://prometheus.io/docs/practices/naming/).

mula/scheduler/app.py
jpbruinsslot: @praseodym I've made some changes and updated the example output at the top. Would this be the correct way to do it?

praseodym: Yes that looks a lot better! There's no metric for `scheduler_qsize` now though?

jpbruinsslot: Ok cool, thanks! Ah yeah I see. Think it is because I haven't added an organisation yet. But I'll make sure that's the case.

jpbruinsslot: Yeah it had to do with an organisation not yet being created, when it has been created the following output will be generated by the endpoint:

```
HTTP/1.1 200 OK
content-length: 1088
content-type: text/plain; charset=utf-8
date: Thu, 06 Apr 2023 07:05:42 GMT
server: uvicorn

# HELP app_settings_info Scheduler configuration settings
# TYPE app_settings_info gauge
app_settings_info{monitor_organisations_interval="60",pq_maxsize="1000",pq_populate_grace_period="86400",pq_populate_interval="60",pq_populate_max_random_objects="50"} 1.0
# HELP scheduler_qsize Size of the scheduler queue
# TYPE scheduler_qsize gauge
scheduler_qsize{scheduler_id="normalizer-dev"} 0.0
scheduler_qsize{scheduler_id="boefje-dev"} 0.0
# HELP app_katalogus_info Katalogus service information
# TYPE app_katalogus_info gauge
app_katalogus_info 1.0
# HELP katalogus_cache_expiry Katalogus cache expiry
# TYPE katalogus_cache_expiry gauge
katalogus_cache_expiry{cache_type="plugin"} 1.680764699e+09
katalogus_cache_expiry{cache_type="boefje"} 1.68076476e+09
katalogus_cache_expiry{cache_type="normalizer"} 1.68076476e+09
# HELP katalogus_cache_start Katalogus cache start
# TYPE katalogus_cache_start gauge
katalogus_cache_start{cache_type="plugin"} 1.680764669e+09
katalogus_cache_start{cache_type="boefje"} 1.680764669e+09
katalogus_cache_start{cache_type="normalizer"} 1.680764669e+09

```

praseodym (approved):

Donnype:
Unrelated but happened to notice: does this PR somehow contain a lot of linting changes with a shorter line length?

jpbruinsslot:
> Unrelated but happened to notice: does this PR somehow contain a lot of linting changes with a shorter line length?

Yeah, had problems with formatting and keeping the pre-commit happy

mula/scheduler/app.py
```diff
@@ -252,8 +252,65 @@ def run(self) -> None:
...
+            interval=1,
+        )
+
         # Main thread
         while not self.stop_event.is_set():
             time.sleep(0.01)
 
         self.shutdown()
+
+    def collect_metrics(self) -> None:
```
Donnype, line 268: For my understanding: the only advantage of running this in a thread vs. on every request is some performance benefits when we get multiple requests per second? A Gauge is "forgetful" right?

jpbruinsslot: I went back on forward on this. I decided eventually to implement it this way because otherwise (if I remember correctly) application-wide metrics meant that I had to pass the app as an argument to the server in order to get those.

mula/scheduler/context/context.py
```diff
@@ -85,6 +84,53 @@ def __init__(self) -> None:
...
+            documentation="Size of the scheduler queue",
+            registry=self.metrics_registry,
+            labelnames=["scheduler_id"],
+        )
+
+        self.metrics_katalogus = Info(
+            name="app_katalogus",
+            documentation="Katalogus service information",
+            registry=self.metrics_registry,
+        )
```
dekkers, line 122: This one is never set... what is the purpose of this?

mula/scheduler/context/context.py
```diff
@@ -85,6 +84,53 @@ def __init__(self) -> None:
...
+            registry=self.metrics_registry,
+            labelnames=["cache_type"],
+        )
+
+        self.metrics_katalogus_cache_start = Gauge(
+            name="katalogus_cache_start",
+            documentation="Katalogus cache start",
+            registry=self.metrics_registry,
+            labelnames=["cache_type"],
+        )
```
dekkers, line 136: I have a hard time understanding how these two metrics can be used. The cache start/expiry is basically just a random time based on when a request was made after the last expiration if I understand it correctly. How would you create a graph based on this that can help diagnose problems or how would you set alerts on a metric like this?

mula/setup.py
```diff
@@ -1,3 +1,8 @@
 from setuptools import setup, find_packages
 
-setup(name="mula", author="MinVWS", packages=find_packages(exclude="tests"), include_package_data=True)
+setup(
+    name="mula",
+    author="MinVWS",
+    packages=find_packages(exclude="tests"),
+    include_package_data=True,
+)
```
dekkers, line 8: There is a lot of unnecessary reformatting noise in this PR that makes code reviewing what actually has changed hard.

dekkers (changes requested):

mula/scheduler/context/context.py
jpbruinsslot: I agree with you that it doesn't give any useful information at the moment, and if the request is: knowing what plugins are known to the scheduler at a specific moment for debugging purposes. We might better be off with an endpoint referencing/reporting it. Additionally, I'm not too fond of using caches for this ( because it introduces a new set of problems), I'd rather have the katalogus have more filter options then there won't be a need to know what is present in the scheduler at a given moment. Although this would create more api chattiness from the scheduler to the katalogus.

For now I'll remove the katalogus metric from the openmetrics implementation, and we can pick up the 'freshness' in a different issue.

Darwinkel (approved):
## Checklist for QA:

### What works:
* All organisations are represented with a normalizer and boefje entry
* Queue sizes are correctly updated for both boefje and normalizer
* App settings are correctly updated after `.env` changes

Darwinkel:
Ready for merge after fixing the merge conflict.

mula/scheduler/context/context.py
jpbruinsslot: Made separate issue of this https://github.com/minvws/nl-kat-coordination/issues/856

dekkers (approved):

PR #670, opened 2023-04-05
Add MutedFinding OOI and filter them in report

### LINKED ISSUE #624: Create new MutedFinding OOI type
Currently, we have no way to Mute Findings.
We should allow users to Mute findings.

Given our future ambitions with Procedures on findings, creating a MuteOOI that relates to a specific finding (and possibly some other to be created AnswerOOI), would allow is to 'hide' findings from reports or other views.

Proposed is a new OOI type:

- Who (currently untracked) optional
- Which finding
- Which proof (eg, the answer) optional

In the future we can use the valid-till time on the OOI to mute findings for a give period But this is out of scope for now.

* MuteOOI's can be created from the rocky UX by binding them to the specific finding.
* MuteOOI's can be deleted from the rocky UX
* MuteOOI's can in the future be created by a normaliser acting on the Uploaded answers from a questionOOI.

### LINKED ISSUE #636: Filter findings in report if MuteFinding OOI exists
See #624 for background

Nice to haves (can also be implemented later):

- Appendix in the report with muted findings

dekkers:
This adds the MutedFinding OOI model and will filter findings for which a MuteFinding OOI exists from reports.

It also changes the sorting to sort on severity and name of the finding type instead of just severity so that finding types with the same severity will be sorted by name.

Fixes #624, fixes #636


ammar92 (dismissed):
Looks good to me. Although I think that `MutedFinding` is a better name since it says something about its state, whilst *mute finding* might be mistaken for an action

underdarknl (dismissed):

ammar92 (approved):

underdarknl (approved):

PR #569, opened 2023-03-24
strip appending dot from records to avoid duplicates

tobiasBDO:
## Changes
Add `rstrip(".")` to ensure no double OOI's are raised during default onboarding behavior. The toplevels such as `.nl` and `.net` are still being raised duplicate as `Hostname` also appended with the dot. I don't know if that is something that should be fixed as well, or just excluded. Love to hear what you guys think!

image

### before
image

### after
image

## Extra instructions for others
_This section may be skipped or omitted. Uncomment and answer the below questions if relevant._

## Checklist for author(s):
- [x] All the commits in this PR are properly PGP-signed and verified;
- [x] This PR comes from a `feature` or `hotfix` branch, in line with our git branching strategy;
- [x] This PR is "bite-sized" and only focuses on a single issue, problem, or feature;
- [x] I am not reinventing the wheel: there is no high-quality library that already has this feature;
- [ ] I have changed the example `.env` files if I added, removed, or changed any config options, and I have informed others that they need to modify their `.env` files if required;
- [x] I have performed a self-review of my own code;
- [ ] I have commented my code, particularly in hard-to-understand areas;
- [ ] I have made corresponding changes to the documentation, if necessary;
- [ ] I have written unit, integration, and end-to-end tests for the change that I made;

If a non-trivial PR:
- [ ] This PR is part of a milestone and has appropriate labels;
- [ ] This PR is properly linked to the project board (either directly or via an issue);
- [ ] I have added screenshots or some other proof that my code does what it is supposed to do;

```
## Checklist for functional reviewer(s):
- [ ] If a non-trivial PR: This PR is properly linked to an issue on the project board;
- [ ] I have checked out this branch, and successfully ran `make kat`;
- [ ] I have ran `make test-rf` and all end-to-end Robot Framework tests pass;
- [ ] I confirmed that the PR's advertised `feature` or `hotfix` works as intended;
- [ ] I confirmed that there are no unintended functional regressions in this branch;

### What works:
* _bullet point + screenshot (if useful) per tested functionality_

### What doesn't work:
* _bullet point + screenshot (if useful) per tested functionality_

### Bug or feature?:
* _bullet point + screenshot (if useful) if it is unclear whether something is a bug or an intended feature._
```

```
## Checklist for code reviewer(s):
- [ ] The code passes the CI tests and linters;
- [ ] The code does not bypass authentication or security mechanisms;
- [ ] The code does not introduce any dependency on a library that has not been properly vetted;
[...]

PR #675, opened 2023-04-05
only create fqdn hostnames

noamblitz:
Only create hostnames that end with a dot


Darwinkel:
Should this be merged in conjunction with https://github.com/minvws/nl-kat-coordination/pull/569, or instead of it?

noamblitz:
> Should this be merged in conjunction with #569, or instead of it?

Instead of it

Issue #588, opened 2023-03-28
Scheduler crashes when postgres database becomes unavailable


sigio:
When the postgres database used by scheduler becomes temporarily unavailable, for example due to a restart or a temporary network issue, it goes into a state where it no longer functions, and requests for /(lang)/(code)/tasks result in error-pages even after the database has been available again for a long time.

This can be resolved by restarting the scheduler, but this should not be needed, scheduler should retry connections to the database (with exponential backoff) in case of failed connections.

jpbruinsslot:
Able to reproduce with:

1. Go through onboarding, add task
2.

```
$ docker-compose kill postgres; sleep 15; docker-compose up -d postgres
```

Eventually, it will crash during a transaction and will shutdown the scheduler.

---

In the case of the scheduler continues to run, while in the meantime the database connection has been lost. And that the tasks endpoint continues to return error-pages. I wasn't able to reproduces it. I've added [`pool_recycle`](https://docs.sqlalchemy.org/en/13/core/pooling.html#setting-pool-recycle) setting to the connect, this might alleviate this problem.

---

As far as the first issue I can implement a retrying query class in sqlalchemy, so that it will retry failed sqlalchemy queries when a transaction failed.

PR #697, opened 2023-04-11
Retry on db disconnects in mula

jpbruinsslot:


mula/scheduler/repositories/sqlalchemy/datastore.py
```diff
@@ -55,3 +57,22 @@ def __init__(self, dsn: str) -> None:
...
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            for i in range(max_retries):
+                try:
+                    return func(*args, **kwargs)
+                except Exception as e:
+                    if i == max_retries - 1:
+                        raise e
+
+                    print(f"Retrying {func.__name__} in {retry_delay} seconds ({i+1}/{max_retries}): {str(e)}")
```
ammar92, line 73: Perhaps a log statement instead of print would be better?

ammar92: `str(e)` isn't necessary, the f-string will always use the stringified version of objects

mula/scheduler/repositories/sqlalchemy/datastore.py
```diff
@@ -55,3 +57,22 @@ def __init__(self, dsn: str) -> None:
         self.session = orm.sessionmaker(
             bind=self.engine,
         )
+
+
+def retry(max_retries=3, retry_delay=5):
```
ammar92, line 62:
```suggestion
def retry(max_retries: int = 3, retry_delay: float = 5.):
```

mula/scheduler/repositories/sqlalchemy/datastore.py
```diff
@@ -55,3 +57,22 @@ def __init__(self, dsn: str) -> None:
...
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            for i in range(max_retries):
+                try:
+                    return func(*args, **kwargs)
+                except Exception as e:
```
ammar92, line 69: I was wondering if it might be better to retry on a specific error (such as `sqlalchemy.exc.DBAPIError`)  rather than anything, otherwise, it could repeat calls caused by programming errors or user input.

The type(s) of exception(s) could also be an argument for the decorator, this would make the decorator even fancier.

ammar92 (commented):
Nice stuff. I was expecting something that was specifically for DB disconnects, but this could definitely work. Have you also considered using the `pool_recycle` option when creating the connection?

I have a few remarks that might help improve it even more. And I think a small test would also be very useful.

jpbruinsslot:
Thanks for the review!

> Nice stuff. I was expecting something that was specifically for DB disconnects, but this could definitely work. Have you also considered using the `pool_recycle` option when creating the connection?
> 
> I have a few remarks that might help improve it even more. And I think a small test would also be very useful.

Yeah, I've indeed added pool_recycle as well. (https://github.com/minvws/nl-kat-coordination/pull/697/files#diff-62bb74de4f65c5805017b3069b71d5e56e5262116de9a33d53033b6d80010b26R50) However, regarding the initial issue, it wasn't able to alleviate completely this, because of the following:

> The implication that this has for the notion of “retrying” a statement is that in the default case, when a connection is lost, the entire transaction is lost. There is no useful way that the database can “reconnect and retry” and continue where it left off, since data is already lost. For this reason, SQLAlchemy does not have a transparent “reconnection” feature that works mid-transaction, for the case when the database connection has disconnected while being used. The canonical approach to dealing with mid-operation disconnects is to retry the entire operation from the start of the transaction, often by using a custom Python decorator that will “retry” a particular function several times until it succeeds, or to otherwise architect the application in such a way that it is resilient against transactions that are dropped that then cause operations to fail. (https://docs.sqlalchemy.org/en/20/faq/connections.html#how-do-i-retry-a-statement-execution-automatically)

Since the scheduler will then throw an exception on the transaction.

mula/scheduler/repositories/sqlalchemy/datastore.py
jpbruinsslot: Yeah you're right I'll update it

Darwinkel:
## Checklist for QA:

### What works:
* When postgres dies and is restarted sometime later, KAT (including the scheduler) resumes normal functionality.

### Bug or feature?:
* As all KAT modules currently depend on a single instance of Postgres, the entire application becomes unusable and throws Django errors until Postgres is reachable again. Not related to this PR per se, but I'll create a new ticket to think and implement proper error handling for this.

Darwinkel (approved):

dekkers (changes requested):
I don't really like the catching of very broad exceptions and just retrying them. It will for example catch things like ProgrammingError and IntegrityError which doesn't make sense to retry.

I also don't like that we do the same thing in every function instead of handling errors depending on what the function does.  For example `pop` and `peek` can probably just return None, while for `update` or `remove` we can retry (but we still need to handle that this might not succeed).

Did we actually find the exact problem why the scheduler didn't work anymore? It's not really clear to me, but if it is caused by that we don't handle exceptions correctly somewhere then this won't solve the problem because we still throw the same exceptions when the retries also fail.

jpbruinsslot:
> I don't really like the catching of very broad exceptions and just retrying them. It will for example catch things like ProgrammingError and IntegrityError which doesn't make sense to retry.

Isn't that covered by https://github.com/minvws/nl-kat-coordination/pull/697/files#diff-62bb74de4f65c5805017b3069b71d5e56e5262116de9a33d53033b6d80010b26R73-R76 or are you referring to something else?

> I also don't like that we do the same thing in every function instead of handling errors depending on what the function does. For example `pop` and `peek` can probably just return None, while for `update` or `remove` we can retry (but we still need to handle that this might not succeed).

When the system is in error, e.g. connection to a database is lost. Returning `None` on `pop` and `peek` would confuse what is actually happening, because `None` is something significant in the context of `pop` and `peek`. I'd rather have the system than raise an exception when the retrying fails. 

> Did we actually find the exact problem why the scheduler didn't work anymore? It's not really clear to me, but if it is caused by that we don't handle exceptions correctly somewhere then this won't solve the problem because we still throw the same exceptions when the retries also fail.

In the original issue associated with this PR (https://github.com/minvws/nl-kat-coordination/issues/588) it was requested that the scheduler would try to reconnect when the connection to postgres would be temporarily lost. In the issue you can also read my comment on trying to reproduce the error.

dekkers:
> > I don't really like the catching of very broad exceptions and just retrying them. It will for example catch things like ProgrammingError and IntegrityError which doesn't make sense to retry.
> 
> Isn't that covered by https://github.com/minvws/nl-kat-coordination/pull/697/files#diff-62bb74de4f65c5805017b3069b71d5e56e5262116de9a33d53033b6d80010b26R73-R76 or are you referring to something else?

`DBAPIError` is  very broad base class. Catching all exceptions derived from this class will mean we retry a lot more circumstances than when PostgreSQL in not available, almost all errors that can happen. See https://peps.python.org/pep-0249/#exceptions. Looking at the descriptions there, we should only retry on `OperationalError` and `InternalError`, and not on `DataError`, `IntegrityError`, `ProgrammingError` and `NotSupportedError`.

dekkers (approved):

dekkers (approved):

dekkers (approved):

underdarknl (approved):

Issue #554, opened 2023-03-23
Scheduler throws api exception on expired dict when katalogus just got up


Donnype:
Found this one when having the KATalogus down for a while and bringing it back up, perhaps this can be handled?

![Image](https://user-images.githubusercontent.com/46660228/227162052-e74bb2ce-863e-4c2c-becc-98566fa13f03.png)

PR #700, opened 2023-04-11
Add reset() to expiring dict

jpbruinsslot:
Fixes https://github.com/minvws/nl-kat-coordination/issues/554


ammar92 (approved):

Donnype (approved):
I was able to reproduce the issue with a lifetime of 3 for the Katalogus and starting the Katalogus 10 seconds after starting the scheduler. This branch seems to fix that issue :+1:

underdarknl (approved):

PR #706, opened 2023-04-11
fix translation block bug

noamblitz:


dekkers:
The same problem is also in rocky/rocky/templates/oois/ooi_detail_findings_list.html. But I think this should be double quotes and we shouldn't change the text in the user interface because of djlint. Can't we escape the quotation mark or ignore the djlint rule for these cases?

ammar92 (commented):
I was wondering if these were intended as double quotes, why not just escape the double quote characters instead of changing symbols?

ammar92:
> I was wondering if these were intended as double quotes, why not just escape the double quote characters instead of changing symbols?

Hadn't read the comment, but @dekkers already mentioned this

ammar92 (approved):

dekkers (approved):

PR #713, opened 2023-04-12
Only have hostnames without a dot

dekkers:
The canonical way to write hostnames is without a dot and we should use that in OpenKAT. The alternative of having all hostnames end with a dot means we will either have to remove the dot everywhere or have strange URLs such as  "https://www.example.com./". 

This also removed the fqdn field on Hostname as that isn't necessary anymore.


ammar92 (approved):
Does make sense to me, although I don't know what the implications are (yet)

Lisser (approved):

dekkers:
@Lisser and I have been testing this code. If we force Hostname to not have a dot in the Hostname `__init__` this gives problems after upgrading when there are still existing Hostname objects that have a dot in the database. It will render both Hostname objects without dot in the object list causing doubles. Links from other object types will still render with a dot in the UI, but if you click on it will get an exception.

The solution is to no longer force Hostname to not have a dot in `__init__` but make sure that all normalizers and bits create Hostname objects without a dot. This will means everything works fine after upgrading and the Hostname objects with a dot will disappear after the boefjes/normalizers/bits have run.

underdarknl (approved):

Issue #543, opened 2023-03-22
[Bug] Common open ports are not reported as findings


zcrt:
**Describe the bug**
Common ports such as 443 do not end up in KATs findings. However, not every service with port 443 open was also intended to have that port reachable by KAT. In a report, as an end-user I want to know all open ports that KAT has found.

**To Reproduce**
1. Scan a website
2. Generate a report
3. There is no finding for a common open port

**Expected behavior**
Report all open ports seen by KAT.

**OpenKAT version**
c799f181cdd4569304925f625affcf9f58092607 (latest)

**Additional context**
https://github.com/minvws/nl-kat-coordination/blob/82d64a588f641ab3551a7c338393235dfd0e4774/octopoes/bits/port_classification/port_classification.py#L35

underdarknl:
There are two ways for this to be done.
The first would be to move the various lists of ports from inside the BIT, to inside the graph, where they can be added/changed by users. The Bit would then look for this ConfigOOI and match ports listed there against the ports seen on an IP. This would also allow more fine-grained control of these settings, as you could bind a ConfigOOI to a Network, but also directly to an IPAddress to make specific changes based on your requirements. Following that would allow you to have more or less findings based on your config.

The second would be to include all ports in the report you are looking at. Since all known ports are present in the Graph, we could just include them in the query that produces the report. This for the moment is a bit too complex, but is something that would become much easier when octopoesV3 becomes available.

zcrt:
I would opt for the second option as I always want to know all ports that are open on all IPs I scan with KAT. Why is it complex at the moment? It is already possible to filter on IPPort on the object page, so adding similar logic to all reports should be enough

PR #720, opened 2023-04-12
:sparkles: enhance port findings

zcrt:
## Changes
Improves port findings, includes a recommendation finding for common TCP port, so that they will also always end up in a report. Also makes sure TCP port findings are only raised for TCP ports.

https://github.com/minvws/nl-kat-coordination/issues/543

## Proof
Trivial

## Checklist for author(s):

```

### What works:

### What doesn't work:

### Bug or feature?:
```

```
```


octopoes/bits/port_classification/port_classification.py
```diff
@@ -1,42 +1,43 @@
...
-            finding_type=kat.reference,
-            ooi=input_ooi.reference,
-            description=f"Port {port} is not a common port and should possibly not be open.",
-        )
+    for finding in yield_port(
+        ooi_reference=input_ooi.reference,
+        finding_id=finding_id,
+        finding_description=f"Port {port} is usually {port_type} port and found to be open.",
+    ):
+        yield finding
```
ammar92, line 43: You can replace the loop with `yield from yield_port(...)`

ammar92:
Looks good to me 👍 @noamblitz do you see any problems?

rocky/OOI_database_seed.json
```diff
@@ -476,9 +476,9 @@
     "fields": {
       "last_updated": "2021-08-09T14:50:09.835Z",
       "data": {
-        "description": "A system administration port is open.",
-        "recommendation": "Secure this port",
-        "risk": "Middle"
+        "description": "A port commonly used for system administration is open.",
+        "recommendation": "Manually validate whether this open port should actually be open.",
+        "risk": "Low"
```
dekkers, line 481: Open sysadmin ports risk should not be changed to low. This is a big security risk, but there are also a few usecases for having ssh open, so I think that keeping it middle instead of high is okay.

rocky/OOI_database_seed.json
```diff
@@ -489,8 +489,8 @@
     "fields": {
       "last_updated": "2021-08-09T14:50:09.835Z",
       "data": {
-        "description": "A database port is open.",
-        "recommendation": "Secure this port",
+        "description": "A port commonly used by databases is open.",
+        "recommendation": "Manually validate whether this open port should actually be open.",
```
dekkers, line 493: What is the use case for having a database open to the internet? I can't think of any and it is a big security risk so I think the recommendation should be that the port be closed.

rocky/OOI_database_seed.json
```diff
@@ -502,13 +502,26 @@
...
   },
+  {
+    "model": "tools.ooiinformation",
+    "pk": "KATFindingType|KAT-OPEN-COMMON-TCP-PORT",
+    "fields": {
+      "last_updated": "2021-08-09T14:50:09.835Z",
+      "data": {
+        "description": "A port commonly used for TCP traffic was found to be open.",
+        "recommendation": "Manually validate whether this open port should be open.",
+        "risk": "Recommendation"
```
dekkers, line 520: I don't think we should make findings for this. Having a finding for every webserver that has port 80 and 443 open isn't useful. As @underdarknl said in the linked ticket, we want to have policy objects for this. Until we have those we shouldn't create lots of useless findings in my opinion.

dekkers (changes requested):

rocky/OOI_database_seed.json
zcrt: Not every application hosting a webserver is actually expected by the system administrator to host a webserver. In addition, not every application hosting a webserver should also be reachable by a KAT installation (maybe you only intennd your webserver to be reachable in your internal network). In both these cases it is a relevant finding if port 80 and/or 443 are reachable by KAT.
In addition, adding a finding for ports makes sure any port find by KAT wil also end up in a report (export). A clean table with IPs and found open ports would be a nicer but more complicated solution in this case though.

underdarknl: Would a filtered tree view (ip's + their ports) exported as PDF / CSV not make more sense to include in the report bundle?
I agree some common ports might be unexpected on some machines, but that hinges on knowledge we don't have in the graph at the moment. Creating findings from those ports regardless just to include them in a finding report creates more objects than required, and diminishes the value of the concept Finding.

rocky/OOI_database_seed.json
underdarknl: True, we should probably at some point differentiate between ports, and actual fingerprinted services on them.

underdarknl:
To keep things configurable the functionality proposed in this PR has been merged as a separate bit:
https://github.com/minvws/nl-kat-coordination/pull/734

PR #736, opened 2023-04-14
Add PR template code blocks for easier reviewing

Darwinkel:
### Changes
Trivial: turns the code review and QA checklists into code blocks for easy copy-pasting into comments, as it was before.


praseodym:
Maybe we can move them out of the PR template and link to them instead, so that the PR template is shorter and less confusing?

Darwinkel:
> Maybe we can move them out of the PR template and link to them instead, so that the PR template is shorter and less confusing?

More like this? https://github.com/minvws/nl-kat-coordination/blob/053796b523e8c883418e881ebc28e52707f747df/.github/pull_request_template.md

underdarknl (approved):

PR #740, opened 2023-04-17
Restructure mocks on tests

jpbruinsslot:


mula/tests/integration/test_normalizer_scheduler.py
```diff
@@ -55,17 +55,13 @@ def setUp(self):
...
-        mock_is_task_allowed_to_run,
-        mock_is_task_running,
-    ):
+
+@mock.patch("scheduler.schedulers.NormalizerScheduler.is_task_running")  # index: 3
+@mock.patch("scheduler.schedulers.NormalizerScheduler.is_task_allowed_to_run")  # index: 2
+@mock.patch("scheduler.schedulers.NormalizerScheduler.get_normalizers_for_mime_type")  # index: 1
+@mock.patch("scheduler.context.AppContext.services.raw_data.get_latest_raw_data")  # index: 0
+class NormalizerSchedulerTestCase(NormalizerSchedulerBaseTestCase):
+    def test_push_tasks_for_received_raw_file(self, *mocks):
```
Donnype, line 64: Hm, this doesn't really make it more clear what happens?

Donnype (commented):
My perhaps unpopular opinion if you are getting tired of the mocks: try to avoid or reduce them with dependency injection and stubs.

mula/tests/integration/test_normalizer_scheduler.py
jpbruinsslot: you mean the comment or the code?

Donnype: More using indices instead of having explicit arguments

jpbruinsslot: Yeah, the trade-off here was that it would mean I would have all the mocks decorating the methods and then a slew of args. This was at the moment the solution that would make it more DRY, albeit I'm not that happy with the `*mocks` and referencing it by their index. If there are other (idiomatic) strategies that would allow to keep it more "clean" (reducing the use of mock.patch decorators and a lot of method arguments) I'm all ears.

jpbruinsslot: > My perhaps unpopular opinion if you are getting tired of the mocks: try to avoid or reduce them with dependency injection and stubs.

Could do just that xD

ammar92: This is what `pytest` does better at this, you can have (optional and named) fixtures and mocks injected wherever you need them

ammar92: Although it's a bit repetitive, personally I would write the names out and use them as positional arguments rather than `*args`. The next developer/ maintainer can then easily copy the signature and read the mock names rather than dealing with indices.

dekkers: > This is what `pytest` does better at this, you can have (optional and named) fixtures and mocks injected wherever you need them

Yes and you can also define the side effect of the mock in the pytest fixture so you also avoid duplicating that in multiple tests.

dekkers (approved):

PR #765, opened 2023-04-19
Move the ID of the button to the parent element

TwistMeister:
Due to a small UX flaw I moved the ID from the button to the parent/root element of the plugin instance. With the ID on the button in the tiles view the tile went out of the scroll view, with the button aligning with the top of the view. Instead of the whole plugin aligning at the top of the view.

**Before this fix**

https://user-images.githubusercontent.com/16188579/233085911-8f7e6bf3-83bc-4fc5-b665-79b77dea5ed0.mov

**After this fix**

https://user-images.githubusercontent.com/16188579/233085580-1cd84a84-2d40-4441-abd1-ef60e5361a9e.mov

### DIFF
--- a/rocky/katalogus/templates/katalogus.html
+++ b/rocky/katalogus/templates/katalogus.html
@@ -50,7 +50,7 @@
                             </thead>
                             <tbody>
                                 {% for plugin in object_list %}
-                                    <tr>
+                                    <tr id="enableform_{{ plugin.id|slugify }}">
                                         <td scope="row">{{ plugin.name }}</td>
                                         <td>{{ plugin.description }}</td>
                                         <td>
--- a/rocky/katalogus/templates/katalogus.html
+++ b/rocky/katalogus/templates/katalogus.html
@@ -50,7 +50,7 @@
                             </thead>
                             <tbody>
                                 {% for plugin in object_list %}
-                                    <tr id="enableform_{{ plugin.id|slugify }}">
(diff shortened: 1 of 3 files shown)


rocky/katalogus/templates/katalogus.html
```diff
@@ -50,7 +50,7 @@ <h1>{% translate "KAT-alogus" %}</h1>
                             </thead>
                             <tbody>
                                 {% for plugin in object_list %}
-                                    <tr>
+                                    <tr id="enableform_{{ plugin.id|slugify }}">
```
dekkers, line 53: Should we still name the id enableform if it is put on the row instead of the form?

Darwinkel:
## Checklist for QA:

### What works:
* Fix works as intended.

Darwinkel (dismissed):

rocky/katalogus/templates/katalogus.html
ammar92: Indeed, it doesn't make sense to use that prefix anymore. E.g. `plugin_` should suffice, but it also makes sense to not use any prefix

underdarknl: agreed, we should call it enableplugin instead.

TwistMeister: Yes, it was a bit lazy to just move the ID and not rethink its prefix. Just fixed it and altered it to "plugin_"

dekkers (approved):

PR #778, opened 2023-04-20
Add dev target to Dockerfiles

praseodym:
### Changes

During development, code is mounted in the Docker container through volumes. This means the code doesn't have to be added to the container image, and not doing so prevents images from being rebuilt during development unless the Python dependencies change.

Docker makes it easy to skip the code copying step by adding a target to the `Dockerfile` that represents the container image just before the code is copied in. This target is then used in the Docker Compose configuration only during development.

### Issue link
n/a

### Proof
n/a

---
### Code Checklist

### Communication

---

---
## Checklist for QA:
- [ ] I have checked out this branch, and successfully ran a fresh `make kat`.

### What works:

### What doesn't work:

### Bug or feature?:


dekkers (dismissed):

dekkers:
I tested the github container images, ready for merge.

boefjes/Dockerfile
```diff
@@ -30,6 +30,8 @@ RUN --mount=type=cache,target=/root/.cache \
     && rm /tmp/boefjes-requirements.txt \
     && if [ "$ENVIRONMENT" = "dev" ]; then pip install -r requirements-dev.txt; fi
 
+FROM dev
+
 COPY octopoes/ /tmp/octopoes
 RUN cd /tmp/octopoes && python setup.py bdist_wheel
 RUN pip install /tmp/octopoes/dist/octopoes*.whl
```
Donnype, line 37: This means that the development containers do not have octopoes installed as a package in their environment anymore. (It still works since we mount octopoes at the $PYTHONPATH.) This might expose us potential issues whe e.g. octopoes introduces new dependencies in `install_requires` or other setup changes.

Donnype: It is indeed a leftover, but if we choose this path perhaps we should remove the setup.py etc. as well? It's mainly the inconsistency that might introduce a small risk over a small performance gain in this case.

dekkers: I think we shouldn't remove setup.py because we currently still install octopoes in the container images and Debian packaging and changing this should probably be done as part of a bigger reorganization. But we can remove the install_required from setup.py, we don't really use it at the moment and that would get rid of the inconsistency.

underdarknl (dismissed):

boefjes/Dockerfile
praseodym: Created #807 to do this and clean up `setup.py` across modules.

praseodym:
Found a bug: Keiko reports don't work because it tries to write them to the volume that is changed to read-only. I'll fix this by adding a volume to the Keiko container.

praseodym:
> Found a bug: Keiko reports don't work because it tries to write them to the volume that is changed to read-only. I'll fix this by adding a volume to the Keiko container.

There was a volume, but the default config was not using this location to write reports. Fixed in this PR.

docker-compose.yml
```diff
@@ -126,12 +130,13 @@ services:
...
       context: .
       dockerfile: rocky/Dockerfile
+      target: dev
       args:
         ENVIRONMENT: dev
     command: python3 manage.py runserver 0.0.0.0:80
     volumes:
-      - ./rocky:/app/rocky
-      - ./octopoes/octopoes:/app/rocky/octopoes
+      - ./rocky:/app/rocky:ro
```
dekkers, line 138: How are you supposed to run makemigrations if the volume is read only? Makemigrations won't be able to write the migration file. Same for the modules that use alembic.

ammar92: Migration files are usually made on the development machine and not inside the container, right?

Donnype: The Makefile targets often use the containers (this also holds for Katalogus/Bytes)

praseodym: Good point, `makemigrations` is a valid use case for having a writable volume. Do we revert the read-only change for all modules using Django or alembic?

Donnype: Cool!

Darwinkel:
## Checklist for QA:

### What works:
* No apparent regressions, everything works as usual, including generating PDF reports.

Darwinkel (approved):

dekkers (approved):

PR #788, opened 2023-04-24
DB schema update scheduler

jpbruinsslot:


mula/scheduler/alembic/env.py
```diff
@@ -40,6 +40,8 @@ def run_migrations_offline():
         target_metadata=target_metadata,
         literal_binds=True,
         dialect_opts={"paramstyle": "named"},
+        compare_server_default=True,
+        compare_type=True,
```
Donnype, line 44: Did these help? I recall having the issues detecting string lengths as well but that does seem to work now for the Katalogus for instance.

dekkers (approved):

Donnype (dismissed):

mula/scheduler/alembic/env.py
jpbruinsslot: No, also wasn't able to automatically detect the changes with alembic. Had to create the migration by hand

dekkers (approved):

Issue #510, opened 2023-03-16
`admin` organization code causes route priority issues


Darwinkel:
**Describe the bug**
If the organization code name is "admin", users are redirected to the Django admin interface. 

**To Reproduce**
1. Do a clean `make kat`
2. Create an organization with the code `admin`
3. Notice that you are redirected to the Django admin interface.

**Expected behavior**
We should probably not allow reserved organization codes, such as admin and static, to prevent route priority bugs.

Include these comments:
https://github.com/minvws/nl-kat-coordination/issues/790

underdarknl:
I think the only way around this is to have a deny-list of url's that we cannot use because they are already taken.

underdarknl:
I think the only way around this is to have a deny-list of url's that we cannot use because they are already taken.

Darwinkel:
Verification without Django-admin has been implemented recently :tada:

PR #800, opened 2023-04-24
Add deny list and validator for organization code

Rieven:
### Changes
We must exclude some organization codes that cannot be used when creating a new organization. Some organization codes like `admin` conflicts with the Django admin. We might extend the deny list when more confliction occurs or things we really want to deny.

### Issue link
Closes https://github.com/minvws/nl-kat-coordination/issues/510

### Proof
image

From admin:
image

From the api:
image

---
### Code Checklist

### Communication

---

---
## Checklist for QA:
- [ ] I have checked out this branch, and successfully ran a fresh `make kat`.

### What works:

### What doesn't work:

### Bug or feature?:


rocky/tools/models.py
```diff
@@ -54,6 +62,7 @@ class Organization(models.Model):
             "A slug containing only lower-case unicode letters, numbers, hyphens or underscores "
             "that will be used in URLs and paths"
         ),
+        validators=[organization_code_validator],
```
dekkers, line 65: Validators aren't run by `save` and I am not sure these validators are run by django-rest-framework. I think we should also add a test case to check that these codes are also rejected by the API.

rocky/tests/test_organization.py
```diff
@@ -315,3 +315,22 @@ def test_admin_edits_organization(rf, admin_member, mocker):
     assertContains(resulted_response, "tags-color-1-light plain")  # default color
     assertContains(resulted_response, "tag1")
     assertContains(resulted_response, "tag2")
+
+
+def test_organization_code_validator(rf, superuser_member, mocker, mock_models_octopoes):
+    mocker.patch("katalogus.client.KATalogusClientV1")
+    if DENY_ORGANIZATION_CODES:
+        for organization_code in DENY_ORGANIZATION_CODES:
```
dekkers, line 323: The if is always true, so the if is useless here. I think it also enough to just test one denied code, testing all the code only makes the test suite longer for no gain.

dekkers (changes requested):

rocky/tools/models.py
```diff
@@ -44,6 +46,12 @@ def css_class(self):
         return f"tags-{self.color} {self.border_type}"
 
 
+def organization_code_validator(value: str) -> None:
+    if value in DENY_ORGANIZATION_CODES:
+        raise ValidationError("Choose another organization code")
```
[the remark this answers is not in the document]
Rieven: I kept it simple here to not give the user extra info what urls is reserved by OpenKAT, for security reasons to not let them try out unauthorised urls.

rocky/tools/models.py
Rieven: Once the organization code is created it cannot be changed, not from the admin panel + also not from api. Saving is here not used. I can use clean method here instead is also a possibility to also include save().

rocky/tools/models.py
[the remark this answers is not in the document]
Darwinkel: > I kept it simple here to not give the user extra info what urls is reserved by OpenKAT, for security reasons to not let them try out unauthorised urls.

I'm a bit torn on this: exposing as little security information as possible is a good idea, but on the other hand, KAT is open source and a dedicated attacker could easily figure out the reserved keywords. I'm inclined to agree with @dekkers here - as a user it's nice to know _why_ they are not allowed to do something.

rocky/tools/models.py
Rieven: Moved it to `clean` method

rocky/rocky/settings.py
```diff
@@ -333,6 +333,34 @@
...
+    "tasks",
+    "bytes",
+    "kat-alogus",
+    "boefjes",
+    "mula",
+    "keiko",
+    "octopoes",
+    "rocky",
+    "fmea",
+]
```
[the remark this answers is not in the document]
underdarknl: Being able to add extra denied names is a nice feature though, as we might run into situation where people install other apps, or just don' want some words to be used in their install.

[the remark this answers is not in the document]
Rieven: What is also possible is to have 2 lists, one for internal usage and an extra extended one for reserved codes. Let me know what can be decided here @underdarknl @dekkers

underdarknl:
this closes https://github.com/minvws/nl-kat-coordination/issues/510

rocky/rocky/settings.py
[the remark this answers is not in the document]
dekkers: I don't see how we can run into a situation where people install other apps and even if they do that they have to modify the code anyway. Adding organisations is also not something that can be done by everyone. I don't think we should turn this simple PR that denies the names that conflict with urls into something more complex feature that supports users being able to configure denying certain words in organisation names. I think we have more important things to work on...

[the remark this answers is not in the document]
underdarknl: Agreed, let keep it a non-user-configurable setting.

dekkers:
Looks good, moving to QA.

rocky/tools/models.py
```diff
@@ -104,8 +132,20 @@ def delete(self, *args, **kwargs):
 
         super().delete(*args, **kwargs)
 
+    def clean(self):
+        if self.code in DENY_ORGANIZATION_CODES:
+            raise ValidationError(
```
Donnype, line 137: This validation error should be handled in organization_add.py. The ServiceException is handled there as well and perhaps adding the ValidationError to the except is enough, but now a user without DEBUG=True will have no idea why they cannot create their organization.

Donnype (changes requested):

rocky/tools/models.py
Rieven: @Donnype I put this directly in the model because we also have an /api connection to create organizations, also in the django admin. Then I would need to create separate validations for each view.

Rieven: The form used in `organization_add` is different as well as those used for the api and admin.

Darwinkel:
## Checklist for QA:

### What works:
* Deny list when creating an organization works correctly during the onboarding as well as through the organization management page

rocky/tools/models.py
Donnype: Throwing the error here is fine! But organization_add.py will now give us an error page as your screenshot shows. When debug is turned off, this will be a generic 500 error without the user knowing that he used a denied name. So I would add an 'except ValidationError' in that view and tell the user either the denylist or at least that they added a denied name either through messaging or an error page.

(How the API handles it is a different scenario indeed, which perhaps should be checked as well but might be less important because of the limited usage (in terms of users) of the API and the fact that the current API users probably have access to logs as well.)

Rieven: The 500 error is at the API url not from `organization_add` view.

Donnype: My bad

Donnype (approved):

dekkers (approved):

PR #813, opened 2023-04-26
Workflow cleanup rdo

sigio:
Rework all rdo-packaging workflows into a single unified workflow
- Try to optimize workflow (no double tasks)
- Update to latest workflows
- Only upload releases if workflow completes builds


sigio:
Tegen verkeerde branch... retry tegen main ;)

PR #828, opened 2023-04-28
Create boefjes.md

ring-ring-ring:
### Changes
Moving all content from the readmes to dev docs


dekkers:
I think the READMEs in each individual directory is not really useful and we can just delete them?

dekkers:
> > 
> 
> There is some content left in these readmes. The idea of moving them here is to make everything available in one place and start updating as needed.

I don't mean the content that is moved to the documentation, but the README.rst that are added in each directory telling you to go to docs.openkat.nl, isn't having that in the toplevel directory enough?

ring-ring-ring:
> > > 
> > 
> > 
> > There is some content left in these readmes. The idea of moving them here is to make everything available in one place and start updating as needed.
> 
> I don't mean the content that is moved to the documentation, but the README.rst that are added in each directory telling you to go to docs.openkat.nl, isn't having that in the toplevel directory enough?

Yes it might be enough, we can leave them out. Thanks for clarifying, I misunderstood.

ammar92 (approved):

docs/source/developer_documentation/rocky.md
```diff
@@ -1,3 +1,140 @@
...
+```
+ROCKY_DB_HOST=
+ROCKY_DB_PORT=
+ROCKY_DB=
+ROCKY_DB_USER=
+ROCKY_DB_PASSWORD=
+```
+
+
+## KATalogus View Structure
```
Donnype, line 75: Since I moved these diagrams in another PR as well, there is a duplicate for the Katalogus views diagram now (see below)!

Donnype (changes requested):

Donnype (approved):

dekkers (approved):

PR #832, opened 2023-05-01
Optimize scheduler

jpbruinsslot:
* rabbitmq basic consume instead of basic get for the retrieval of messages of queue
* population of boefje queue (mutations, new/enabled boefjes, random objects) run in separate threads
* I/O tasks for information gathering from API's 'off-loaded' with threadpoolexecutor
* don't ask bytes if a task is running, when internal scheduler task specifies it is still running
* quick win: flush cache when orgs get added (start of with a filled plugin cache, don't wait for a call to the katalogus api to flush cache)
* refactoring some code, removing some dead code
---

TODO:
* [x] test katalogus requests with increased cache ttl
* [x] test if removal cache reset on update mutation is helpful


mula/scheduler/connectors/listeners/raw_data.py
```diff
@@ -1,18 +1,25 @@
...
-            return None
+    def listen(self) -> None:
+        self.basic_consume(self.queue, False)
 
-        return RawDataReceivedEvent(**response)
+    def dispatch(self, body: bytes) -> None:
+        # Convert body into a RawDataReceivedEvent
+        body_str = body.decode("utf-8")
+        body_dict = json.loads(body_str)
+        model = RawDataReceivedEvent(**body_dict)
```
ammar92, line 22:
```suggestion
        # Convert body into a RawDataReceivedEvent
        model = RawDataReceivedEvent.parse_raw(body)
```

mula/scheduler/connectors/listeners/scan_profile.py
```diff
@@ -1,31 +1,26 @@
...
+        self.basic_consume(self.queue, True)
 
-    @exception_handler
-    def get_scan_profile_mutations(self, queue: str, n: int) -> Optional[List[ScanProfileMutationModel]]:
-        oois: List[ScanProfileMutationModel] = []
+    def dispatch(self, body: bytes) -> None:
+        # Convert body into a ScanProfileMutationModel
+        body_str = body.decode("utf-8")
+        body_dict = json.loads(body_str)
+        model = ScanProfileMutationModel(**body_dict)
```
ammar92, line 23:
```suggestion
        model = ScanProfileMutationModel.parse_raw(body)
```

mula/scheduler/schedulers/normalizer.py
```diff
@@ -40,172 +40,165 @@ def __init__(
...
                     self.scheduler_id,
                 )
                 return
 
+        # TODO: duplicates possible here?
+        normalizers = []
+        for mime_type in latest_raw_data.raw_data.mime_types:
+            normalizers_by_mime_type = self.get_normalizers_for_mime_type(mime_type.get("value"))
+            if not normalizers_by_mime_type:
+                continue
```
ammar92, line 91: Assuming that the function above can return an empty list, we don't necessarily need to check this emptiness when extending the other list

```suggestion
```

mula/scheduler/connectors/listeners/listeners.py
```diff
@@ -81,7 +82,10 @@ def callback(
         properties: pika.spec.BasicProperties,
         body: bytes,
     ) -> None:
-        self.logger.debug(" [x] Received %r", body)
+        """Callback function that is called when a message is received on the
+        queue.
+        """
+        self.logger.debug("Received message %r on queue %s", body, method.routing_key)
```
ammar92, line 88: (subjective) you might want to flip the order of arguments since it's probably easier to read the queue name before the (usually bigger) message body

ammar92 (commented):
Cool stuff! A few remarks that could improve this even more

Darwinkel:
## Checklist for QA:

### What works:
* No apparent regressions
* Scheduling of new tasks and normalizers "feels" snappy, with two organizations, with most boefjes enabled

ammar92 (approved):

Donnype (approved):
These look like good changes!

I think it also wouldn't hurt to create some scripts in the process of improving performance that would allow us to do better stress testing in the future. Perhaps mimicking production load is quite hard because of the unpredictability, but an optional test target that fills up the database an fires events/calls to the scheduler would allow us to measure the impact of changes such as these :grin: !

dekkers:
I am moving to QA, because the code looks good and my comments are just suggestions for further improvements.

dekkers (approved):

Issue #836, opened 2023-05-01
Create declarative xtdb query from Path


Donnype:
## Requirements

From [this](https://github.com/minvws/nl-kat-coordination/issues/288) issue and discussion we concluded that Octopoes needs the ability to resolve a relationship Path to a set of results, which is a typical xtdb query use-case. An example of the structure of such a query would be

```edn
{:query
    {:find [(pull finding [*])]
        :where [
            [e :xt/id "IPPort|internet|xxxx::xxxxx|tcp|23"]
            [e :IPPort/address f]
            ( or 
                [f :IPAddressV4/network n]
                [f :IPAddressV6/network n]
            )
            [finding :Finding/ooi n]
            [finding :Finding/finding_type "KATFindingType|KAT-500"]
        ]
    }
}
```

This needs to be generated from the scenario where we _want_ to find all Findings for the network an IPPort is attached to, while _having_ only an IPPort reference (`"IPPort|internet|xxxx::xxxxx|tcp|23""`) and the relationship path (`"address.network"`).

Things to keep in mind here are:
- Octopoes has abstract types that require an `or` (as above)
- Octopoes has an abstraction for relationships already: the `Path` object
- The where clauses are _commutative_, meaning that the resultset is independent of the order of the clauses

---

## Functionality

For the query above, it would be pretty flexible if we could generate it as follows:

```python
query = Query(Finding)\
    .where(IPPort, id="IPPort|internet|xxxx::xxxxx|tcp|23", address=IPAddress)\
    .where(IPAddress, network=Network)\
    .where(Finding, ooi=Network)\
   .where(Finding, finding_type="KATFindingType|KAT-500")
```

Then we could leverage the Path object to generate it.

PR #761, opened 2023-04-18
initial setup of config oois

noamblitz:
### Changes
Adds the possibility to add config oois to kat. With the correct bit definition, this config ooi is fetched and the config wil be send to be used in the bit.

### Issue link
Mentioned in https://github.com/minvws/nl-kat-coordination/issues/288#issuecomment-1511452781

---
### Code Checklist
- [ ] This PR only contains functionality relevant to the issue; tickets have been created for newly discovered issues.
- [ ] I have written unit tests for the changes or fixes I made.
- [ ] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [ ] I have performed a self-review of my code and refactored it to the best of my abilities.

### Communication
- [ ] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [ ] I have made corresponding changes to the documentation, if necessary.

---
## Checklist for code reviewers:
- [ ] The code does not violate Model-View-Template and our other architectural principles.
- [ ] The code prioritizes readability over performance where appropriate.
- [ ] The code does not bypass authentication or security mechanisms.
- [ ] The code does not introduce any dependency on a library that has not been properly vetted.
- [ ] The code contains docstrings, comments, and documentation where needed.

---
## Checklist for QA:
- [ ] I have checked out this branch, and successfully ran a fresh `make kat`.
- [ ] I confirmed that there are no unintended functional regressions in this branch:
    - [ ] I have managed to pass the onboarding flow
    - [ ] Objects and Findings are created properly
    - [ ] Tasks are created and completed properly
- [ ] I confirmed that the PR's advertised `feature` or `hotfix` works as intended.

### What works:
* _bullet point + screenshot (if useful) per tested functionality_

### What doesn't work:
* _bullet point + screenshot (if useful) per tested functionality_

### Bug or feature?:
* _bullet point + screenshot (if useful) if it is unclear whether something is a bug or an intended feature._

Issue #288, opened 2023-02-21
Question OOIs with json-schema


Donnype:
See #130 
Draft PR: #1394

Some considerations:

- We want questions to be formulated by json-schemas and answers as json
- In the frontend we can create forms from json-schema's and json from form submissions
- ~We might need to generate id's for questions~ Unique Questions per OOI are probably fine?
- Schema's are defined in Bits
- Per answer, the right Config needs to be created. Determining which Config can be done through:
  - **The mime types, by adding an id per schema (best practice for json-schema's) and triggering the right normalizer**. However, this would also be communicating the bit_id for Configs, so we are somewhat abusing notation here.
  - ~Adding a identifier field in the answer~
  - ~Adding an argument (but I'm not a fan of the mostly unused arguments field, see https://github.com/minvws/nl-kat-coordination/issues/763)~
  - ~Try to infer it through the input_ooi~ This will not work if we start attaching multiple configs to a Network.

Still, Rocky (saving the raw file) needs to know what the "type" of Question is, so somehow the Question object (either as a field or in its schema) needs to carry an identifier.

Questions:
- Do we need to add a new page with an overview? Quickly seeing unanswered question would be nice I'd say. 

Config flow and cardinality:

```mermaid
classDiagram

direction RL
    class Question
    class Config
    class JsonAnswer
    class JsonAnswer2["JsonAnswer"]

    Question : ooi[OOI]
    Question : schema[str]

    Config : ooi[OOI]
    Config : config[str]
    Config : bit_id[str]

    OOI <|-- Question
    Question <|-- JsonAnswer
    Question <|-- JsonAnswer2
    JsonAnswer <|-- Config
    JsonAnswer2 <|-- Config
```

Resulting ERD:

```mermaid
classDiagram

direction RL
    class Question
    class Config
    class OOI

    Question : ooi[OOI]
    Question : schema[str]

    Config : ooi[OOI]
    Config : config[str]
    Config : bit_id[str]

    OOI <|-- Question
    OOI <|-- Config
```

#### The Flow

```mermaid
sequenceDiagram
    participant User
    participant Rocky
    participant Normalizer
    participant Octopoes
    participant Bits
    participant Bytes

    Normalizer->>Octopoes: Add Network
    Bits->>Octopoes: Add Question["What ports may be open for this Network?"]
    Rocky->>Octopoes: Get Question
    Rocky->>User: Prompt Question to user
    User->>Rocky: Give answer (form)
[...]

PR #851, opened 2023-05-01
Feature/octopoes path queries

Donnype:
### Changes
Adds functionality with which we can build the query we need for https://github.com/minvws/nl-kat-coordination/issues/288 and in this PR: https://github.com/minvws/nl-kat-coordination/pull/761

Integration with the other PR can be done after adding this feature, or we could merge them.

### Issue link
fixes https://github.com/minvws/nl-kat-coordination/issues/836

### Proof
See the tests. I also tested the generated queries on my environment for the use-case and they worked.

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


octopoes/tests/test_query.py
```diff
@@ -0,0 +1,199 @@
...
+    [ Finding :Finding/ooi Network ]
+    [ Finding :object_type "Finding" ]
+    [ IPPort :IPPort/address IPAddress ]
+    [ IPPort :IPPort/primary_key "IPPort|internet|xxx:xxx:x|tcp|23" ]
+    [ IPPort :object_type "IPPort" ]]}}
+"""
+    )
+
+
+def test_create_query_from_relation_path():
```
dekkers, line 140: Test function should also be renamed to `from_path`.

Lisser (approved):
Neat!

It's difficult to consider all the intricacies that come along with this kind of 'meta'-programming. So I can't be 100% certain that this abstraction 'hits the nail on its head', or that we might find limitations in this abstraction later on. 

One example that comes to mind: we are using the type name (e.g. `Hostname`) as the 'variable-name' in the translated Datalog. So, I wonder what happens if I'd like to make Path query that traverses the same type multiple types. E.g. `Hostname -> DNSTXTRecord -> Hostname -> DNSARecord -> IPAddress`

Does the double Hostname appear as the same symbol in the translated Datalog?

But overall, if this can replace most/all of the plain Datalog queries, then I'd consider this a win already.

octopoes/octopoes/xtdb/query.py
```diff
@@ -0,0 +1,195 @@
...
+            if segment.direction is Direction.OUTGOING:
+                query = query.where(segment.source_type, **{segment.property_name: segment.target_type})
+                continue
+
+            query = query.where(segment.target_type, **{segment.property_name: segment.source_type})
+
+        return query
+
+    def query(self, ooi_type: Type[OOI]) -> "Query":
+        """Change the target object type of the Query after initialization, e.g. when using from_relation_path()"""
```
dekkers, line 74: This doesn't seem to be used anymore, at least not in `from_path`, and also doesn't have test coverage. Is this still needed?

Lisser:
I will try to come up with a test for the edge-case I could think of named above

octopoes/octopoes/xtdb/query.py
```diff
@@ -0,0 +1,195 @@
...
+    def from_path(cls, path: Path) -> "Query":
+        """
+        Create a query from a Path.
+
+        The last segment in the path is assumed to be the queries OOI Type. You can change this by calling the
+        query() method after initialization for the required target OOI Type.
+        """
+
+        ooi_type = path.segments[-1].target_type
+        query = Query(ooi_type)
```
dekkers, line 62: This should be `cls(ooi_type)`.

octopoes/octopoes/xtdb/query.py
```diff
@@ -0,0 +1,195 @@
...
+    _offset: Optional[int] = None
+
+    def where(self, ooi_type: Type[OOI], **kwargs) -> "Query":
+        for field_name, value in kwargs.items():
+            self._where_field_is(ooi_type, field_name, value)
+
+        return self
+
+    def format(self) -> str:
+        return "\n" + self._compile(separator="\n    ") + "\n"
```
dekkers, line 50: Why start the string with a newline?

octopoes/octopoes/xtdb/query.py
```diff
@@ -0,0 +1,195 @@
...
+        query() method after initialization for the required target OOI Type.
+        """
+
+        ooi_type = path.segments[-1].target_type
+        query = Query(ooi_type)
+
+        for segment in path.segments:
+            if segment.direction is Direction.OUTGOING:
+                query = query.where(segment.source_type, **{segment.property_name: segment.target_type})
+                continue
```
dekkers, line 67: I think an `else:` is better here than a `continue` here, because `continue` signals that we skip the rest of the loop for some reason, but that is not really the case.

Donnype:
@Lisser This definitely doesn't cover all of it, but this scenario in particular is a shortcoming indeed: we cannot traverse the same type in multiple places without the query claiming it has to be the _same_ entity. But as this solves the current use-case I think this is something we can iterate on and think of how we want to use the Query object to convey that specific behaviour.

Lisser:
> @Lisser This definitely doesn't cover all of it, but this scenario in particular is a shortcoming indeed: we cannot traverse the same type in multiple places without the query claiming it has to be the same entity. But as this solves the current use-case I think this is something we can iterate on and think of how we want to use the Query object to convey that specific behaviour.

Exactly, good to be aware of it, but doesn't need to fixed before merging this. Maybe good to do create an issue for this though, so we don't forget.

octopoes/octopoes/xtdb/query.py
Donnype: Was playing around with some formatting in the tests, this looked pretty easy to read. But the newlines at the start and end perhaps don't add mucht?

octopoes/octopoes/xtdb/query.py
Donnype: Ah perhaps not for now indeed

Donnype:
@Lisser https://github.com/minvws/nl-kat-coordination/issues/911

Lisser (approved):

dekkers (approved):

PR #901, opened 2023-05-02
Clean up Makefile

praseodym:
### Changes

**Clean up Makefile**

Avoid data loss when running `make kat`, do not recreate containers
unnecessarily, use Makefile prerequisites where possible and add
comments to all targets.

**Remove most Docker Compose dependencies for Rocky**

Only PostgreSQL is needed to start Rocky. Avoiding other services
greatly speeds up the build process and avoids unnecessary container
recreations when running `make kat`.

**Enable Parcel cache**

On my machine this saves about 5s for a new build.

**Wait for database in Rocky build**

Prevents 'Connection refused' errors in migration if the database has
not started before Django migrations are attempted.

### Issue link
n/a

### Proof

```
$ time make reset
[..]
Executed in   21.20 secs
```
---

### Code Checklist

### Communication

---

---
## Checklist for QA:

### DIFF
--- a/.github/workflows/build-rdo-package.yml
+++ b/.github/workflows/build-rdo-package.yml
@@ -5,7 +5,7 @@ concurrency:
   cancel-in-progress: true
 
 on:
-  create:
+  push:
     tags:
       - v*
 
--- a/.github/workflows/keiko_test.yml
+++ b/.github/workflows/keiko_test.yml
@@ -28,7 +28,17 @@ jobs:
           cache: 'pip' # caching pip dependencies
 
       - name: Install Latex
-        run: sudo apt-get install -y --no-install-recommends texlive-latex-base texlive-latex-extra texlive-latex-recommended texlive-lang-european texlive-fonts-recommended
+        run: |
+          sudo apt-get install -y --no-install-recommends \
+            chktex \
+            fonts-recommended \
+            latexmk \
+            texlive-fonts-recommended \
+            texlive-lang-european \
+            texlive-latex-base \
+            texlive-latex-extra \
+            texlive-latex-recommended \
+            texlive-xetex
 
       - name: Install requirements-dev.txt
         run: pip install -r requirements-dev.txt
--- a/keiko/Dockerfile
+++ b/keiko/Dockerfile
@@ -14,8 +14,15 @@ ENV PATH=/home/keiko/.local/bin:${PATH}
 # LateX dependencies
 RUN apt update -y \
     && apt install -y --no-install-recommends \
-        texlive-latex-base texlive-latex-extra texlive-latex-recommended \
-        texlive-lang-european texlive-fonts-recommended chktex \
+        chktex \
+        fonts-recommended \
+        latexmk \
+        texlive-fonts-recommended \
+        texlive-lang-european \
+        texlive-latex-base \
+        texlive-latex-extra \
+        texlive-latex-recommended \
+        texlive-xetex \
     && rm -rf /var/lib/apt/lists/*
 
 ENV KEIKO_REPORTS_FOLDER=/reports
--- a/keiko/debian/control
+++ b/keiko/debian/control
@@ -7,6 +7,6 @@ Section: python
 Priority: optional
 Architecture: any
 Pre-Depends: ${misc:Pre-Depends}
-Depends: ${python}, texlive-latex-base, texlive-latex-extra, texlive-latex-recommended, texlive-lang-european, texlive-fonts-recommended, chktex, ${misc:Depends}, ${shlibs:Depends}
+Depends: ${python}, chktex, fonts-recommended, latexmk, texlive-fonts-recommended, texlive-lang-european, texlive-latex-base, texlive-latex-extra, texlive-latex-recommended, texlive-xetex, ${misc:Depends}, ${shlibs:Depends}
 Description: Keiko
  The PDF report generation tool for KAT.
--- a/keiko/keiko/keiko.py
+++ b/keiko/keiko/keiko.py
@@ -5,11 +5,11 @@ import os
 import shutil
 import subprocess
 import tempfile
-from logging import getLogger
+from logging import DEBUG, ERROR, getLogger
 from pathlib import Path
-from typing import Dict, Set, Tuple
+from typing import Any, Dict, Set, Tuple
 
-from jinja2 import Environment, FileSystemLoader, select_autoescape
+from jinja2 import Environment, FileSystemLoader
 from opentelemetry import trace
 from opentelemetry.trace import Status, StatusCode
 
@@ -21,6 +21,39 @@ logger = getLogger(__name__)
 tracer = trace.get_tracer(__name__)
 
 DATA_SHAPE_CLASS_NAME = "DataShape"
+
+LATEX_SPECIAL_CHARS = str.maketrans(
+    {
+        "&": r"\&",
+        "%": r"\%",
+        "$": r"\$",
+        "#": r"\#",
+        "_": r"\_",
+        "{": r"\{",
+        "}": r"\}",
+        "~": r"\textasciitilde{}",
+        "^": r"\^{}",
+        "\\": r"\textbackslash{}",
+        "\n": "\\newline%\n",
+        "-": r"{-}",
+        "\xA0": "~",  # Non-breaking space
+        "[": r"{[}",
+        "]": r"{]}",
+    }
+)
+
+
+def latex_escape(input: Any) -> str:
+    """Escape characters that are special in LaTeX.
+
+    References:
+    - https://github.com/JelteF/PyLaTeX/blob/ecc1e6e339a5a7be958c328403517cd547873d7e/pylatex/utils.py#L68-L100
+    - http://tex.stackexchange.com/a/34586/43228
+    - http://stackoverflow.com/a/16264094/2570866
+    """
+    if not isinstance(input, str):
+        input = str(input)
+    return input.translate(LATEX_SPECIAL_CHARS)
 
 
 def baretext(input_: str) -> str:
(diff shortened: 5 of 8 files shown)


Makefile
```diff
@@ -5,90 +5,77 @@ SHELL := bash
...
 
 # Export Docker buildkit options
 export DOCKER_BUILDKIT=1
 export COMPOSE_DOCKER_CLI_BUILD=1
 
-kat: env-if-empty clean # This should give you a clean install
-	make build
-	make up
+# Build and bring up all containers (default target)
+kat: env-if-empty build up
```
Donnype, line 15: Make does not guarantee that dependencies run sequentially as I found in [this](https://stackoverflow.com/questions/9159960/order-of-processing-components-in-makefile/9160175#9160175) discussion before. In theory this could affect `update` and `reset` as well.

Donnype: Also, the semantics of `make kat` change here since we omit a `clean` now. As this is probably the target that is most used by devs (and perhaps other users), we need to make sure we communicate such a change properly.

Having said that, I do think `make kat` simply meaning "guarantee that I have a running system of the current code" and `make reset` meaning "get me an empty version of the current code" make sense.

Makefile
```diff
@@ -5,90 +5,77 @@ SHELL := bash
...
+# Build will prepare all services: migrate them, seed them, etc.
+build:
+ifeq ($(UNAME),Darwin)
+	docker-compose build --pull --build-arg USER_UID="$$(id -u)"
 else
-	docker-compose build --build-arg USER_UID="$$(id -u)" --build-arg USER_GID="$$(id -g)"
+	docker-compose build --pull --build-arg USER_UID="$$(id -u)" --build-arg USER_GID="$$(id -g)"
 endif
-
-build: build-all
```
Donnype, line 86: Combining `build` and `build-all` does mean that without https://github.com/minvws/nl-kat-coordination/pull/860 the user gets prompted to create a superuser on upgrades as well (if I recall correctly that that was the reason to split them before).

rocky/Makefile
```diff
@@ -9,9 +9,11 @@ export COMPOSE_DOCKER_CLI_BUILD=1
 build: build-rocky build-rocky-frontend
 
 build-rocky:
-	docker-compose run --rm rocky make build-rocky-native
+# Set DATABASE_MIGRATION=false to prevent entrypoint from running migration
+	docker-compose run --rm -e DATABASE_MIGRATION=false rocky make build-rocky-native
```
Donnype, line 13: 👍

Makefile
```diff
@@ -5,90 +5,77 @@ SHELL := bash
...
-	make up
-endif
-
-reset:
-	-docker-compose down --remove-orphans --volumes --timeout 0
-	make up
-	make -C boefjes build
-	make -C rocky almost-flush
+# Bring down containers, remove all volumes, and bring them up again
+reset: clean kat
```
Donnype, line 21: I suppose there's no need for the almost-flush functionality anymore?

Donnype (commented):
Good stuff! We've come a long way with the Makefile since the repo merge. Do take a look at the caveat of potential parallel target dependencies!

Makefile
praseodym: I removed it because it can be invoked using the Rocky Makefile directly. If needed we could make a target that invokes it in the main Makefile?

Makefile
praseodym: Indeed, I tested these changes on top of #860.

Makefile
praseodym: The default target causing data loss has bit me before, so this was an intended change. Good point about communicating this to the team after merge.

praseodym: > Make does not guarantee that dependencies run sequentially as I found in [this](https://stackoverflow.com/questions/9159960/order-of-processing-components-in-makefile/9160175#9160175) discussion before. In theory this could affect `update` and `reset` as well.

Didn't think of that. Maybe we should consider explicitly disabling parallelism for some targets, as described in https://www.gnu.org/software/make/manual/html_node/Parallel-Disable.html?

Makefile
Donnype: Yes perhaps, although I am not sure it was really being used (which we might find out by removing it altogether of course 😀)

Makefile
```diff
@@ -5,90 +5,77 @@ SHELL := bash
...
-build-all:  # Build should prepare all other services: migrate them, seed them, etc.
-ifeq ($(UNAME), Darwin)
-	docker-compose build --build-arg USER_UID="$$(id -u)"
+# Build will prepare all services: migrate them, seed them, etc.
+build:
+ifeq ($(UNAME),Darwin)
+	docker-compose build --pull --build-arg USER_UID="$$(id -u)"
 else
-	docker-compose build --build-arg USER_UID="$$(id -u)" --build-arg USER_GID="$$(id -g)"
+	docker-compose build --pull --build-arg USER_UID="$$(id -u)" --build-arg USER_GID="$$(id -g)"
```
Donnype, line 70: Now that we're looking into the performance, what do you think of building the images in parallel with `--parallel`, perhaps optionally? The output does become rather ugly.

Makefile
Donnype: Seems like a good solution if we can fix this with a one liner!

Makefile
praseodym: On my machine, Docker Compose (version v2.17.3) already builds the images in parallel. Can't really find when this became the default, but we could add `--parallel` explicitly for older Docker Compose versions.

Donnype: Interesting, I'll upgrade especially given this message from the docker documentation: 

image

praseodym: I've added `--parallel` explicitly, doesn't hurt.

Makefile
praseodym: Let's keep it out for now and add it back if anyone requests it.

Makefile
praseodym: Added `.NOTPARALLEL:`.

Donnype (approved):

Darwinkel:
## Checklist for QA:

### What works:
* `make kat` is faster
* `make kat` is data persistent (including objects, findings, users, organizations, katalogus)
* `make reset` results in a clean installation (including objects, findings, users, organizations, katalogus)
* `make upgrade` checks out the right version (with and without params)
* `.env` is properly created including random superuser credentials

Nice work!

### What doesn't work:
- Removing the `.env` and then running `make kat` (without wiping data) results in database errors for Rocky and the Katalogus:

```
django.db.utils.OperationalError: FATAL:  password authentication failed for user "rocky_app"

sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) FATAL:  password authentication failed for user katalogus_app"
```

Should we provide some good error handling for this scenario?

praseodym:
> - Removing the `.env` and then running `make kat` (without wiping data) results in database errors for Rocky and the Katalogus:
> 
> 
> 
> ```
> 
> django.db.utils.OperationalError: FATAL:  password authentication failed for user "rocky_app"
> 
> 
> 
> sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) FATAL:  password authentication failed for user katalogus_app"
> 
> ```
> 
> 
> 
> Should we provide some good error handling for this scenario?

The cause is that the .env file has different database credentials than the credentials that are active in the database. I'm not sure if we can detect this scenario to provide good error handling. Unless someone has a better idea, I think our best option is to mention this error in the developer docs with the suggestion to run `make reset`.

Darwinkel:
> The cause is that the .env file has different database credentials than the credentials that are active in the database. I'm not sure if we can detect this scenario to provide good error handling. Unless someone has a better idea, I think our best option is to mention this error in the developer docs with the suggestion to run `make reset`.

Yeah, I figured. Updating the docs is probably sufficient. We should keep an eye out in the issues and community chat for people encountering this scenario.

Darwinkel (dismissed):

docs/source/technical_design/localinstall.rst
```diff
@@ -107,6 +107,22 @@ Go to the directory containing openkat:
 	$ cd nl-kat-coordination
 	$ make update
 
+Create a new superuser for the new version. You can delete the old superuser after the update. This is not pretty, but has the advantage that your databases remain intact. Check that you are on the most recent version everywhere, especially Rocky sometimes hangs because of yarn.lock.
```
dekkers, line 110: Why do you need to create a new superuser each time you update?

docs/source/technical_design/localinstall.rst
```diff
@@ -107,6 +107,22 @@ Go to the directory containing openkat:
...
+If you to start over with a clean slate, you can do so with the following commands:
+
+.. code-block:: sh
+
+	$ cd nl-kat-coordination
+	$ make reset
+
+This removes all Docker containers and volumes, and then brings up the containers again.
+
+Optionally, first remove the ``.env`` file (``rm .env``) before running ``make reset`` to also reset all configuration in environment variables. This should also reolve issues such as database authentication errors (``password authentication failed``).
```
dekkers, line 124: Typo `reolve` -> `resolve`

docs/source/technical_design/localinstall.rst
praseodym: This is a merge problem, this paragraph was removed last week in https://github.com/minvws/nl-kat-coordination/commit/5a2315e29ba20488b7a2bae3049444615cf52539#diff-174ecb96956e599b0bfe65c2f1d191c2212fdcb5a8bff742178fcd2e45d1b189 but somehow got back in with this PR.

docs/source/technical_design/localinstall.rst
praseodym: Good catch, fixed

docs/source/technical_design/localinstall.rst
praseodym: Removed this paragraph like intended.

Darwinkel (approved):
Still works fine, ready for merge.

dekkers (approved):

PR #912, opened 2023-05-04
Add indices for hash fields

jpbruinsslot:
Added indices on hash fields of the `Task` and `PrioritizedItem` and migrated the `p_item` and `data` to `JSONB`.

Had to manually update the alembic migration, since alembic doesn't recognize a functional index (text)  on a json field (https://www.mail-archive.com/sqlalchemy@googlegroups.com/msg38546.html)


jpbruinsslot:
Have to fix the test, but can be reviewed

mula/Makefile
```diff
@@ -109,9 +109,9 @@ endif
 
 itest: ## Run the integration tests.
 ifneq ($(file),)
-	$(ci-docker-compose) run --rm mula python -m pytest tests/integration/${file} $(if $(function),-k $(function))
+	$(ci-docker-compose) run --rm mula sh -c "sleep 10; python -m pytest tests/integration/${file} $(if $(function),-k $(function))"
```
ammar92, line 112: I think 10 seconds is quite a lot. Another way to do this is to use health checks (see https://github.com/minvws/nl-kat-coordination/pull/920). I believe this should also work for Postgres containers (see `pg_isready`)

dekkers (changes requested):

mula/tests/integration/test_api.py
```diff
@@ -426,53 +430,58 @@ def test_get_tasks_value(self):
 
     def test_get_tasks_min_and_max_created_at(self):
         # Get tasks based on datetime, both min_created_at and max_created_at, should return 2 items
-        min_created_at = self.first_item_api.get("created_at")
-        max_created_at = self.second_item_api.get("created_at")
+        min_created_at = urllib.parse.quote(self.first_item_api.get("created_at"))
```
ammar92, line 433: I was still wondering if this wrap is really needed since the client used here is usually really good at quoting and encoding. What I for sure know is that if you're using the `params` keyword the query parameters will get encoded properly. I just tested this and it should also work if you pass the url only:
```python3
>>> import requests
>>> res = requests.get('https://google.com/?this="iasd"a"compleX#@query&param=2313X#~~~')
>>> res.url
'https://www.google.com/?this=%22iasd%22a%22compleX#@query&param=2313X%23~~~'
```

ammar92 (commented):
Looks good in general, just a few remarks/ questions.

I was also wondering why you chose `jsonb` over the previous `json` type

mula/Makefile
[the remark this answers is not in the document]
jpbruinsslot: I commented above with an implementation that was done on bytes that I'd want to look at as a solution. And in order to keep it in line with the rest of the project. See https://github.com/minvws/nl-kat-coordination/pull/920/files for implementation of healthcheck

jpbruinsslot: Yeah commented above as well to use that as well

jpbruinsslot:
> Looks good in general, just a few remarks/ questions.
> 
> I was also wondering why you chose `jsonb` over the previous `json` type

If I'm correct a JSON field its contents are stored as text and must be parsed for querying, JSONB should be more efficient for querying and indexing.

mula/Makefile
jpbruinsslot: Fix implemented

mula/tests/integration/test_api.py
jpbruinsslot: Fix implemented

dekkers:
> > Looks good in general, just a few remarks/ questions.
> > I was also wondering why you chose `jsonb` over the previous `json` type
> 
> If I'm correct a JSON field its contents are stored as text and must be parsed for querying, JSONB should be more efficient for querying and indexing.

`json` type is useful for just storing json when you don't want to query the conten. if you want to query specific fields like we do `jsonb` is better.

dekkers:
Looks good to me, moving to QA.

Darwinkel:
## Checklist for QA:

I haven't verified that the scheduler is, in fact, faster with larger datasets.

### What doesn't work:
* Opening plugin details in the Katalogus gives an error. Not sure it's directly related to this PR though.
image

Darwinkel (approved):

dekkers (approved):

Issue #653, opened 2023-04-03
Crisis Room and Findings pages very slow when many findings are present


praseodym:
The Crisis Room page loads very slowly (>50s) when many findings are present. This seems to be caused in part because the `get_queryset` function of the `Top10FindingListView` class in `rocky/rocky/views/finding_list.py` queries all findings only to then truncate it to the top 10. The query should be adjusted so that only the top 10 findings are retrieved from Octopoes/xtdb.

praseodym:
This seems related to https://github.com/minvws/nl-kat-coordination/issues/555. As discussed improvements in this area will be part of Octopoes v3, which should allow us query only the data we need from xtdb.

Lisser:
Related to:

https://github.com/minvws/nl-kat-coordination/issues/811
https://github.com/minvws/nl-kat-coordination/issues/461
https://github.com/minvws/nl-kat-coordination/issues/470
https://github.com/minvws/nl-kat-coordination/issues/152

Lisser:
Implementation plan:
- add fields to FindingType models
- add Boefjes/Normalizers to enrich data
- replace logic in Rocky and Octopoes API to do the heavy lifting by XTDB

PR #921, opened 2023-05-05
feature(octopoes): fields to finding type model

Lisser:
### Changes
Adds fields to FindingType model

### Issue link
Implements part 1 of:  
https://github.com/minvws/nl-kat-coordination/issues/653

### Proof
N/A

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


octopoes/octopoes/models/ooi/findings.py
```diff
@@ -1,13 +1,30 @@
...
-class FindingType(OOI):
+class RiskLevelSeverity(Enum):
+    CRITICAL = "critical"
+    HIGH = "high"
+    MEDIUM = "medium"
+    LOW = "low"
+    RECOMMENDATION = "recommendation"
+
+
+class FindingType(OOI, abc.ABC):
```
dekkers, line 19: Why inherit from `abc.ABC`? This on itself doesn't really add anything. `abc.ABC` has very specific functionality that we don't seem to be using, so I don't really see the point of inheriting from `abc.ABC`.

Lisser: @dekkers 

`FindingType` is an abstract class, so it should never be instantiated directly. So, IMO better to enforce it?

dekkers: But that is not what `abc.ABC` does. See for example:

```
In [1]: import abc

In [2]: class Test(abc.ABC):
   ...:     pass
   ...: 

In [3]: a = Test()

In [4]: print(a)
<__main__.Test object at 0x7f5423fba4d0>
```

ammar92: Good points. Technically it doesn't add anything, semantically it has meaning. I don't really have a strong opinion on this, but I'd suggest leaving it out as it's not supposed to be used like this.

octopoes/octopoes/models/ooi/findings.py
```diff
@@ -1,13 +1,30 @@
...
     id: str
 
+    description: Optional[str]
+    source: Optional[AnyUrl]
+
+    risk_score: Optional[float]
+    risk_severity: Optional[RiskLevelSeverity]
```
Donnype, line 26: Do these make more sense on a FindingType versus on a Finding? I can imagine two Findings having a different risk_severity depending on which object they point to. (E.g. A `RetireJSFindingType` on your main website has higher risk than on third party websites found by KAT.) In that case risk attributes on the FindingTypes could also be interpreted as defaults.

But then again I have not gone into the risk modelling that much yet so far.

Donnype (commented):
Are these

octopoes/octopoes/models/ooi/findings.py
dekkers: Findings like the KAT findings currently have a static score that is used for all the findings that are created and we need to have a place to store those. They are currently stored in the rocky database, but we are moving those to xtdb and thus need the fields also there.

Donnype: Yes seems legitimate, but I was just questioning the fact that all Findings of the same type have the same risk score and severity as there is no way to differentiate the risk between two findings of the same type now.

Lisser: @Donnype 

I think your point is valid, a Finding of a certain FindingType might be more or less severe depending on its relation to the Organisation. However, this is a separate use-case and needs more thought on how, when and where to implement this functionality. 

For now, as @dekkers says, this story concerns 'just' moving the FindingType metadata to XTDB/Octopoes to improve performance of several views and the reports.

octopoes/octopoes/models/ooi/findings.py
Lisser: Fixed

ammar92 (approved):

underdarknl (approved):

PR #922, opened 2023-05-08
Fix active state in language switcher

TwistMeister:


HeleenSG (approved):

Donnype:
@TwistMeister I guess we can do a style fix and merge this one?

Issue #464, opened 2023-03-09
Update settings support in Katalogus/Rocky to json


Donnype:
Since we now support JSON settings in the Katalogus validated with json-schema, we should update Rocky to make use of this as this functionality is not used at this stage. Some considerations now that I have taken a look at the current code are the following:

Rocky has some functionality to add and update single settings. Since settings are (optionally) encrypted as a blob, and uses json validation, I think it would make more sense interacting with the list of settings as a whole from Rocky as well. This would make validating the json settings as a whole a lot easier. (This is also how e.g. AWS secretsmanager handles their secrets: one page to add/edit all of them with one save button.) This would make both Rocky as well as the Katalogus API simpler and remove some redundancy for the endpoints that edit a single key (where the backend now gets the whole value, updates one key, and saves the whole value again..).

Donnype:
Related: we should make the settings namespace more consistent, since I believe the path should contain an extra `plugins` before the `plugin_id` to have the same setup as the endpoint for the plugins.

Issue #108, opened 2023-01-10
Boefje settings integer handling


zcrt:
**Describe the bug**
An "integer" setting in a boefje will only work if it is set after the boefje is enabled. If it is set before enabling the boefje the following error is generated:

```python
HTTPError at /kat-alogus/plugins/boefje/<boefje-id>/False/

404 Client Error: Not Found for url: http://katalogus:8000/v1/organisations/_dev/repositories/LOCAL/plugins/<boefje-id>

Request Method: 	POST
Request URL: 	http://localhost:8000/kat-alogus/plugins/boefje/<boefje-id>/False/
Django Version: 	3.2.16
Exception Type: 	HTTPError
Exception Value: 	

404 Client Error: Not Found for url: http://katalogus:8000/v1/organisations/_dev/repositories/LOCAL/plugins/<boefje-id>

Exception Location: 	/usr/local/lib/python3.8/site-packages/requests/models.py, line 1021, in raise_for_status
Python Executable: 	/usr/local/bin/python3
Python Version: 	3.8.16
Python Path: 	

['/app/rocky',
 '/usr/local/lib/python38.zip',
 '/usr/local/lib/python3.8',
 '/usr/local/lib/python3.8/lib-dynload',
 '/usr/local/lib/python3.8/site-packages']
```
The Katalogus will throw the error: `boefjes.katalogus.storage.interfaces.SettingsNotConformingToSchema`:
```python
Failed validating 'type' in schema['properties']['TOP_PORTS']:
    {'description': 'Scan TOP_PORTS most common ports. Defaults to 250.',
     'title': 'TOP_PORTS',
     'type': 'integer'}

On instance['TOP_PORTS']:
    '33'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/exceptions.py", line 64, in __call__
    await self.app(scope, receive, sender)
  File "/usr/local/lib/python3.8/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/usr/local/lib/python3.8/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/routing.py", line 680, in __call__
    await route.handle(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/routing.py", line 275, in handle
[...]

PR #939, opened 2023-05-09
Better JSON schema support for plugin settings in Rocky

Donnype:
### Changes
Simplify the settings setup to support JSON schema validation from the plugins. This means no add/edit/delete per key, only an upsert (idempotent `PUT`/`create_or_update`) and a delete. Since we validate settings as a whole against the schema it makes more sense to treat it as one entity as well.

### Issue link
Fixes https://github.com/minvws/nl-kat-coordination/issues/464.
Fixes https://github.com/minvws/nl-kat-coordination/issues/108.

### Proof

Some jsonschema errors in the form:

image

image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


underdarknl:
Nice! functionally this looks like a great step in the right direction.

rocky/katalogus/forms/plugin_settings.py
```diff
@@ -12,65 +15,41 @@ class PluginSchemaForm(forms.Form):
...
+        cleaned_data = super().clean()
 
-    def __init__(self, plugin_schema, setting_name, setting_value=None, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.plugin_schema = plugin_schema
-        self.setting_name = setting_name
-        self.setting_value = setting_value
-        self.populate_field()
+        # The form assigns "" and in some scenario's null to all unfilled (optional) fields
+        cleaned_data = {key: value for key, value in cleaned_data.items() if value is not None and value != ""}  # noqa
```
ammar92, line 47: What error does this violate and why should we suppress this with a `noqa`?

ammar92 (dismissed):
Looks good I don't have much to say except other than that single linter suppression

rocky/katalogus/forms/plugin_settings.py
Donnype: What is?

dekkers: It would be better to use `# noqa: PLC1901`. That makes it clear what rule is silenced and doesn't silence any other rule.

Darwinkel:
## Checklist for QA:

### What works:
* Integer validation works on the nmap boefje
image
image

### What doesn't work:
* When a string is entered in the integer field, a success message is given but the field becomes Unset
image
image

rocky/katalogus/forms/plugin_settings.py
ammar92: > Because we should accept the value being 0 for instance, so something like `if value` or `if bool(value)` will not work. Checking both the type _and_ truthiness is overkill

I was under the impression that cleaned data is always a dictionary of string-(optional)string pairs, that's why I was surprised that value could be anything other than a `str`.

Donnype:
@Darwinkel Firefox seems to send empty strings when you pass text in a number field. (Chromium does not allow text, although you can bypass that.) So I guess this is expected behavior? Also because one of the (optional) fields is actually being set?

boefjes/boefjes/job_handler.py
```diff
@@ -83,7 +83,7 @@ def get_environment_settings(boefje_meta: BoefjeMeta, environment_keys: List[str
             f"{settings.katalogus_api}/v1/organisations/{boefje_meta.organization}/{boefje_meta.boefje.id}/settings"
         ).json()
 
-        return {k: v for k, v in environment.items() if k in environment_keys}
+        return {k: str(v) for k, v in environment.items() if k in environment_keys}
     except RequestException:
         logger.exception("Error getting environment settings")
```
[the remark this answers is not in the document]
Donnype: It returns an empty dict below. But I think it wouldn't hurt to have another look at the environment/settings setup, since we can also check wether the boefje defines a schema and skip this call altogether if it doesn't. Then we could crash when the API call fails.

Also in the current setup where these handlers run in the same context as the Katalogus, we do not need to make API calls but can simply query the settings from the repository itself. This would also reduce the chance of crashing since we bypass an extra API call.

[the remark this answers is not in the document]
Donnype: I did the check and reraising, not bypassing the API (yet).

Darwinkel:
> @Darwinkel Firefox seems to send empty strings when you pass text in a number field. (Chromium does not allow text, although you can bypass that.) So I guess this is expected behavior? Also because one of the (optional) fields is actually being set?

Fair point. Would it be much work to add a new error message for when text is somehow passed to a number field (or a similar situation)?

dekkers (approved):

PR #943, opened 2023-05-09
Dependabot prs

ammar92:


dekkers:
> I saw that this PR updates [reportlab](https://www.reportlab.com), but are we even using that?

Sort of. It is a dependancy of xhtml2pdf which is used in the fmea code. The fmea code itself is currently not used, but is still in the repo and in INSTALLED_APPS.

Issue #916, opened 2023-05-04
Clients and red teamers should be able to run certain manual scans


Darwinkel:
(As of PR https://github.com/minvws/nl-kat-coordination/pull/606)

Clients can no longer initiate scans. I talked with @TwistMeister and there is something to be said for a client being able to run (some) scans, as this is also what the scheduler does in the background. It may be relevant for them to generate a more recent report with the newest information. The same essentially goes for low-clearance redteamers: shouldn't they be able to perform manual scans as long as this is something that the scheduler would trigger by itself?

_Originally posted by @Darwinkel in https://github.com/minvws/nl-kat-coordination/issues/606#issuecomment-1534682075_

underdarknl:
Correct, Clients should be able to start scans that the scheduler would have started.

The main difference between clients and red-teamers shoud be: 
A red teamer should be able to add indemnifications on objects. (where clients cannot)
A red teamer can enable / configure boefjes, where a client cannot.

Any action that OpenKAT would do automatically, should be in reach for regular users if they want/need it to be scheduled sooner.
Some grey area is present here, as a user adding hundreds of heavy jobs could be damaging to remote systems, but this should be something we guard against in the scheduler at some point (regardless of those jobs being human or auto-generated)

TwistMeister:
PR: https://github.com/minvws/nl-kat-coordination/pull/945

PR #945, opened 2023-05-09
Clients can start scans on objects without changing clearance

Rieven:
### Changes
Client can perform scans but cannot set or update clearance on OOIs.

### Issue link
Fixes https://github.com/minvws/nl-kat-coordination/issues/916

### Dependency
https://github.com/minvws/nl-kat-coordination/pull/950

### Proof

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


rocky/katalogus/views/plugin_detail.py
```diff
@@ -149,20 +163,16 @@ def get_form_filtered_consumable_oois(self):
         oois = self.get_form_consumable_oois()
         return [ooi for ooi in oois if ooi.scan_profile.level >= self.plugin.scan_level.value]
 
-    def get_oois_with_clearance_level(self, selected_oois):
-        """Return a list of selected oois that meets clearance level."""
-        oois_with_clearance_level = []
+    def get_sorted_oois(self, selected_oois: List[str]) -> Dict[str, Any]:
```
dekkers, line 166: Naming this `get_sorted_oois` is confusing because this doesn't return a list of sorted OOIs, but a dict with two lists specifying which have clearance level and which haven't.

rocky/tools/migrations/0037_update_client_permission.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.2.18 on 2023-05-09 11:34
+
+from django.db import migrations
+
+
+def add_group_permissions(apps, schema_editor):
+    Group = apps.get_model("auth", "Group")
+    Permission = apps.get_model("auth", "Permission")
+    try:
+        admin = Group.objects.get(name="clients")
```
dekkers, line 10: The object is the clients group, not the admin group

Rieven:
> In general this PR can't be merged until we have created separate permissions for things that we currently use `can_scan_organisation` for but a client shouldn't be able to do, because else clients will have more permission than they should.

I have added the dependency PR in this PR description

praseodym:
Blocked on #950

dekkers:
We also need to add a migration that adds the permissions to the existing groups.

praseodym:
I can't reproduce the issues, after a `make reset` the onboarding works fine. Do you see any build errors? It also looks like your containers are outdated (Python 3.8 instead of 3.11).

Darwinkel:
Looks like `make` has some strange issues which we're currently investigating.

ammar92 (dismissed):
Looks good to me. Also tested this and appears to work.

Darwinkel:
## Checklist for QA:

### What works:
* A redteamer can enable and disable boefjes (n.b. even without clearance)
* A redteamer cannot change OOI clearances to a higher level than that the redteamer has clearance
* A redteamer can manually schedule boefjes on existing OOI's
* A redteamer can create objects and findings
* A client cannot enable/disable boefjes
* A client can manually schedule boefjes on existing OOI's
* A client cannot change clearance levels

### Bug or feature?:
* A redteamer can reduce an OOI clearance level to a level that the redteamer has clearance to (e.g. from L3 to L0).
* A client can add objects (L0) and findings
* A client cannot access the boefje details in the katalogus

TwistMeister:
- A redteamer can reduce an OOI clearance level to a level that the redteamer has clearance to (e.g. from L3 to L0).
Desirable? Not sure... Maybe we would prefer a "can't touch" policy for OOI's with a higher clearance level than a RedTeamer has permission for, but I can't say this for sure, yet. **Is this a bug or risk and therefor blocking?** No, it's save like this and it follows the specs that we agreed upon. It's an edge case that we didn't really account for, explicitly

- A client can add objects (L0) and findings.
I don't think this is desirable. Will check with @underdarknl.

- A client cannot access the boefje details in the katalogus.
Not blocking for this PR, but I don't know if this is intended.

Darwinkel:
> * A redteamer can reduce an OOI clearance level to a level that the redteamer has clearance to (e.g. from L3 to L0).
>       Desirable? Not sure... Maybe we would prefer a "can't touch" policy for OOI's with a higher clearance level than a RedTeamer has permission for, but I can't say this for sure, yet. **Is this a bug or risk and therefor blocking?** No, it's save like this and it follows the specs that we agreed upon. It's an edge case that we didn't really account for, explicitly
> 
>     * A client can add objects (L0) and findings.
>       I don't think this is desirable. Will check with @underdarknl.
> 
>     * A client cannot access the boefje details in the katalogus.
>       Not blocking for this PR, but I don't know if this is intended.

Ah, another thing I noticed: a redteamer can also edit objects of higher clearance. This may also not be desirable.

underdarknl:
@TwistMeister / @Darwinkel 
Agreed, a user should not be able to alter objects that have a higher clearance level than his own acknowledged clearance level. As this would be a route trough which he can (inadvertently) perform scans on objects that we would otherwise not be able to.

Users adding an object at any level should be a separate permission, which we could grant to users.
Since A client cannot  (or should not be able to set a clearance level (including level 0) this should not be a problem in most cases. An empty clearance level is not the same as a 0 clearance level.
If the graph decides to inherit a clearance level to this object, i'd say this is works as intended if the object gets scanned.

The katalogus page might also show settings at the moment, maybe we should differentiate on that and keep those available for only those users who have the rights to also change them?

TwistMeister:
@underdarknl about showing boefje settings to certain users. In Figma @zcrt raised this question as well and preferred to not show them at all on the detail page when not explicitly asked for, for security sake (unauthorised people watching the screen). I quote @zcrt from Figma:
> "One could even argue that it should not be visible to an editor. In fact, many security-aware applications never show an API-key more than once after generation. As an analogy, you would also not show your plain-text password on the main settings screen of your computer even though you are an administrator"

underdarknl:
> 

I agree some settings should not be readily shown. However, this would only be possible if we designate those fields as sensitive (only show when the user hover/activates the input field), or secrets, only allow resets, never show previous data.
There is no json-schema field-type for this, but we can add an extra indicator for these fields and render accordingly.

Most other settings however are not secret, and should just be visible to the users who can set them (and possibly also to clients), eg the number of Top ports nmap scans is no secret and might be useful for anyone involved to be visible.

Darwinkel:
I've created a follow-up ticket here: https://github.com/minvws/nl-kat-coordination/issues/1064

How about merging this PR and then creating a new one for the remaining refinements?

rocky/tools/migrations/0039_update_permissions.py
```diff
@@ -0,0 +1,28 @@
...
+    try:
+        redteam = Group.objects.get(name=GROUP_REDTEAM)
+        redteam.permissions.add(Permission.objects.get(codename="can_mute_findings"))
+        redteam.permissions.add(Permission.objects.get(codename="can_view_katalogus_settings"))
+        redteam.permissions.add(Permission.objects.get(codename="can_set_katalogus_settings"))
+
+        clients = Group.objects.get(name=GROUP_CLIENT)
+        clients.permissions.add(Permission.objects.get(codename="can_scan_organization"))
+    except Group.DoesNotExist:
+        pass
```
dekkers, line 18: Can the `can_recalculate_bits` permission be added to the admin group in this migration? We forgot to do this earlier and if we add it here we don't need to add an extra migration for this.

Rieven: Done

dekkers (approved):

PR #947, opened 2023-05-09
Fix radio button alignment

TwistMeister:
**Note: this merges to working branch "chore/manon-update"**

### DIFF
--- a/rocky/rocky/templates/partials/form/field_input_radio.html
+++ b/rocky/rocky/templates/partials/form/field_input_radio.html
@@ -3,7 +3,7 @@
     {{ field.label }}
   </legend>
   {% for choice in field %}
-      <div class="horizontal-view">
+      <div class="radio">
           <input type="{% if choice.data.value < 0 %}hidden{% else %}radio{% endif %}"
                   name="{{ field.name }}"
                   value="{{ choice.data.value }}"
--- a/rocky/rocky/templates/partials/form/field_input_radio.html
+++ b/rocky/rocky/templates/partials/form/field_input_radio.html
@@ -1,24 +1,25 @@
-<div>
-    {{ field.label_tag }}
+<fieldset>
+  <legend>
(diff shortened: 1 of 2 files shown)


rocky/rocky/templates/partials/form/field_input_radio.html
```diff
@@ -1,24 +1,25 @@
-<div>
-    {{ field.label_tag }}
+<fieldset>
+  <legend>
+    {{ field.label }}
+  </legend>
+  {% for choice in field %}
+      <div class="horizontal-view">
```
HeleenSG, line 6: This is within a form with the horizontal-view class right? If tat's the case, we shouldn't need this horizontal-view for each object. The styling should work based on the fact that the content is grouped within a div.

If it's added for checkbox or radio-button alignment I would suggest adding the "checkbox" and "radio" classes instead. For examples see: 
-https://minvws.github.io/nl-rdo-manon/documentation/components/checkbox.html  
-https://minvws.github.io/nl-rdo-manon/documentation/components/fieldset-radio.html

rocky/rocky/templates/partials/form/field_input_radio.html
```diff
@@ -1,24 +1,25 @@
-<div>
-    {{ field.label_tag }}
+<fieldset>
+  <legend>
+    {{ field.label }}
```
HeleenSG, line 3: If I remember correctly, it's just the name field.label right? As it actually contains the content of the legend but can't be called legend because of Django expectations/behaviour. If that is the case it's not ideal but ok. If we can have field.legend that would be the correct name.

TwistMeister: This is correct, unfortunately tunatly Django doesn't provide a native "legend" attribute to use for this AFAIK.

HeleenSG (approved):

Issue #808, opened 2023-04-25
Docs: remove references to `make kat-stable`


praseodym:
The docs mention `make kat-stable` but this Make target seems to have been removed.

praseodym:
> I guess that the bigger question is whether we want to provide the shortcut for developers. https://docs.openkat.nl/technical_design/containers.html provides alternative instructions for a production environment.

Those docs are also out of date, e.g. the mention Celery which we are no longer using.

underdarknl:
> > I guess that the bigger question is whether we want to provide the shortcut for developers. https://docs.openkat.nl/technical_design/containers.html provides alternative instructions for a production environment.
> 
> Those docs are also out of date, e.g. the mention Celery which we are no longer using.
> 
> But of course the question still remains. Installing using the Debian packages is very unwieldy, although the community-provided installer shell script makes it a lot easier.

We should aim for a meta-package which does most of the default stuff for a deb based install.
Hardening / configuring could be handles by scripts that we then offer to the user, either as config options trough dpk-reconfigure, or as helper scripts.

PR #948, opened 2023-05-10
Update containers.md

ring-ring-ring:
removed mention of Celery as requested by @praseodym  https://github.com/minvws/nl-kat-coordination/issues/808#issuecomment-1527394169

### DIFF
--- a/docs/source/technical_design/containers.md
+++ b/docs/source/technical_design/containers.md
@@ -102,8 +102,7 @@ We have two container images that are used to run multiple containers. What the 
 | boefjes         | normalizer  | Normalizers runtime                                                               |
 | boefjes         | katalogus   | Katalogus API                                                                     |
 | octopoes        | web         | Octopoes API                                                                      |
-| octopoes        | worker-beat | Celery worker running beat. There must only be exactly one container of this type |
-| octopoes        | worker      | Celery worker. Use this if you need to more than one work container for scaling   |
+
 
 ## Env variables
 


docs/source/technical_design/containers.md
```diff
@@ -102,8 +102,7 @@ We have two container images that are used to run multiple containers. What the
 | boefjes         | normalizer  | Normalizers runtime                                                               |
 | boefjes         | katalogus   | Katalogus API                                                                     |
 | octopoes        | web         | Octopoes API                                                                      |
-| octopoes        | worker-beat | Celery worker running beat. There must only be exactly one container of this type |
```
ammar92, line 105: I think this one still exists in the current Compose project setup

docs/source/technical_design/containers.md
```diff
@@ -102,8 +102,7 @@ We have two container images that are used to run multiple containers. What the
 | boefjes         | normalizer  | Normalizers runtime                                                               |
 | boefjes         | katalogus   | Katalogus API                                                                     |
 | octopoes        | web         | Octopoes API                                                                      |
```
praseodym, line 104: This container is called `octopoes_api` now

docs/source/technical_design/containers.md
praseodym: The container is called `octopoes_api_worker` and it processes events from RabbitMQ. We no longer use Celery.

praseodym (changes requested):

docs/source/technical_design/containers.md
ammar92: Yes the service has been renamed but there are still references to celery (as a dependency and code)

ring-ring-ring:
closed after short talk - we solve this in a different way

PR #978, opened 2023-05-16
Scheduler tests clean up

jpbruinsslot:
Prior refactoring consisted out of removing a lot of repeated `patch` decorators on test methods by leveraging them on the class instead of the methods. In order to keep the method argument short `*args` were useed. However this resulted in referencing those argument with their index within the method, which was confusing since instead of an identifier a index needed to be used.

With this pr the mocks are added as class attributes and as such are easier to reference them in the test.


underdarknl:
Nice, much more readable!

Donnype (approved):

underdarknl (approved):

Issue #570, opened 2023-03-24
include used signing provider in boefje meta file.


underdarknl:
We should include the used signing provider in the boefje Meta file. currently we only include the hash and the url, which works if you know the provider, and we do not switch providers.
Adding the used provider url / source / key inside the meta would allow us to use more than one provider during the installation lifetime and would make raw files and meta files combined be complete bundles of information not reliant on the availability of the config of the used kat install.

PR #994, opened 2023-05-17
Feature/add signing provider for raw file

Donnype:
### Changes
Add optional signing_provider field, since we cannot infer what previous providers were when we switch providers, while we do need to have this information to complete verification of old raw files.

### Issue link
Fixes https://github.com/minvws/nl-kat-coordination/issues/570

### Proof

No signing provider set:

image

Using freetsa.org:

image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


bytes/bytes/database/db_models.py
```diff
@@ -26,8 +26,10 @@ class RawFileInDB(SQL_BASE):  # type: ignore
     __tablename__ = "raw_file"
 
     id = Column(UUID, primary_key=True)
+
     secure_hash = Column(String(length=256), nullable=True)
     hash_retrieval_link = Column(String(length=2048), nullable=True)
+    signing_provider = Column(String(length=256), nullable=True)
```
praseodym, line 32: Maybe we should normalise this, i.e. create a signing provider table and make this a foreign key?

Donnype: Perhaps we should.. This would mainly save space given the append-only nature of these entities, at the cost of a join every time we fetch the meta data.

bytes/bytes/timestamping/.rfc3161.py.swp
dekkers, line 1: This file shouldn't have been committed.

bytes/bytes/database/db_models.py
dekkers: My guess is the join would be very cheap to do because there are only a very limited number of signing providers. It will save space compared to storing the whole url in each row because the primary key of the table can just be a 32-bit int.

dekkers (changes requested):

bytes/bytes/database/db_models.py
underdarknl: As longs as we make it reasonably easy for a user to have the signing endpoint available next to the raw data for validation, all is well.

Darwinkel:
## Checklist for QA:

### What works:
* Signing provider shows up in the logs

dekkers:
Moved this to QA to have it tested with the DB changes.

praseodym (approved):

Darwinkel:
>Moved this to QA to have it tested with the DB changes.

Still works without issues.

Darwinkel (approved):

underdarknl (approved):

PR #1002, opened 2023-05-19
Make two-factor authentication (2fa) optional

### LINKED ISSUE #57: two factor authentication cannot be disabled
Hello,

I run openKAT on an internal IP and made the settings in settings.py less restrictive. However setting twofactor authentication to False does not work. Even if I run the makefile again. Any suggestions?

### LINKED ISSUE #762: Move 2FA checks to middleware and make it possible to disable
The 2FA check should be moved to middleware so we don't have to this check for every view. This also makes it possible to disable 2FA for development/testing or when authentication would be done using single sign-on.

### LINKED ISSUE #968: Password reset link redirects to login
**Describe the bug**
Opening the password reset link (in an incognito tab) results in the `/en/login/` page. 

**To Reproduce**
1. Create a new user
3. Login once as the new user
4. Logout, open an incognito tab, and attempt password reset
5. Click the link in the console and notice you are redirected to the login page.

**Expected behavior**
Getting the password reset form.

**Screenshots**
image

**OpenKAT version**
Noticed on `main` https://github.com/minvws/nl-kat-coordination/commit/483df6a3b508087749e6a433cde1b8a7123228f6

**Desktop (please complete the following information if relevant):**
 - OS: Arch Linux
 - Browser: Firefox

**Additional context**
Not sure how long this bug has been present. Been a while since I last tested it.

praseodym:
### Changes
Make two-factor authentication (2fa) optional. Only enable it if `TWOFACTOR_ENABLED` is set to `True` in the environment.

This is done by moving all checks from decorators to the `AuthRequiredMiddleware`.

Furthermore the password reset functionality is fixed by checking for the reset URL including i18n prefix.

### Issue link
Fixes #762
Fixes #57
Fixes #968

### Proof
n/a

---

### Code Checklist

### Communication

---

---
## Checklist for QA:

**Please thoroughly check this with both `TWOFACTOR_ENABLED=True` and `TWOFACTOR_ENABLED=False`.** 


underdarknl:
Hm, can we invert the logic here to make sure existing installations keep 2fa on, even if they forget to add a new env-var?

praseodym:
> Hm, can we invert the logic here to make sure existing installations keep 2fa on, even if they forget to add a new env-var?

Done. The environment variable wasn't new, but was indeed to keep 2fa enabled after this PR. Now, 2fa is enabled when the `TWOFACTOR_ENABLED` environment variable is not set to `false` (case-insensitive).

Darwinkel:
## Checklist for QA:

### What works:
* Proper authentication and authorization (admin, redteamer, client) with `TWOFACTOR_ENABLED=True`
* Proper authentication and authorization (admin, redteamer, client) with `TWOFACTOR_ENABLED=False`
* Prompts for adding 2FA correctly when going from `False` to `True`
* Password reset link works again
* No apparent regressions

### Bug or feature?:
* When 2FA has been enabled for a user, setting `False` doesn't disable it. The user is still prompted. Users which never added 2FA aren't prompted.

praseodym:
Thanks for the QA! 

> * When 2FA has been enabled for a user, setting `False` doesn't disable it. The user is still prompted. Users which never added 2FA aren't prompted.

I'd say this behaviour is a good compromise. It would be nice if the 2fa setup UI (`/account/two_factor/`) is reachable through normal navigation, I'll create a new issue for that.

> * Going to `/admin` always prompts users to create 2FA if not already set.

This is a bit of a weird one: with 2FA disabled, as a superuser I can go to the Django admin without 2fa prompt. However, if I go to the Django admin as a regular admin user (which doesn't have access to the Django admin!), I get a 2fa prompt, even though having 2fa wouldn't grant permission.

I tried disabling the [`TWO_FACTOR_PATCH_ADMIN` setting](https://django-two-factor-auth.readthedocs.io/en/1.14.0/configuration.html?highlight=TWO_FACTOR_PATCH_ADMIN#general-settings) when 2fa is disabled to avoid this confusion, but while that results in a clearer error message for the admin user (no permission, and offer to log in as a different user) it circumvents the 2fa check for superusers if they have 2fa set up.

Going back to `TWOFACTOR_ENABLED=True` and our `main` branch, it looks like "Two factor authentication is not enabled for your account." is currently also our error message when the user simply doesn't have access to the Django admin. I'd say we create a new issue for this.

ammar92 (approved):

ammar92:
Looks good to me! This should work and the other two issues can indeed be implemented separately.

Thanks for the tests @Darwinkel. Do you agree on what @praseodym mentioned? If so, we can merge this very soon

Darwinkel:
> Looks good to me! This should work and the other two issues can indeed be implemented separately.
> 
> Thanks for the tests @Darwinkel. Do you agree on what @praseodym mentioned? If so, we can merge this very soon

Looks good. Nice follow-up issues. In principle, users should not need to access `/admin` (bad ux) at all, and I don't think we need to pay attention to that specific situation too much. Not very high priority imho.

Darwinkel (approved):

dekkers (approved):

PR #1004, opened 2023-05-22
Upgrade to Django 4.2

### LINKED ISSUE #958: Upgrade to Django 4
> We will have to update dependencies first, for example Django only has Python 3.11 compatibility since Django 4.1 (we're at Django 3.2.18).

_Originally posted by @praseodym in https://github.com/minvws/nl-kat-coordination/issues/627#issuecomment-1491871139_

https://docs.djangoproject.com/en/4.2/howto/upgrade-version/
https://docs.djangoproject.com/en/4.2/releases/4.0/
https://docs.djangoproject.com/en/4.2/releases/4.1/
https://docs.djangoproject.com/en/4.2/releases/4.2/

praseodym:
### Changes
Upgrade to Django 4.2.1

### Issue link
Closes #958 

### Proof
n/a

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


rocky/account/migrations/0003_alter_katuser_id.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("account", "0002_remove_first_last_name"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="katuser",
+            name="id",
+            field=models.AutoField(primary_key=True, serialize=False, verbose_name="ID"),
```
praseodym, line 15: I'm not sure why this migration was created, the only difference seems to be that `serialize=False` was added. I couldn't find anything about this in the Django 4.0-4.2 release notes.

Is there any downside in keeping this migration?

dekkers: As far as I understand it `primary_key=True` causes `serialize=False`. If you normally create migrations you get `primary_key=True` and `serialize=False`, but the katuser is created by a custom written migration to migrate off the standard Django User model and didn't have `serialize=False` . I think it's just some internal thing that works differently in 4.2 which causes the migration code to think something has changed while it didn't think that before.

There is no downside in keeping the migration except that every migration has a tiny cost when running `manage.py migrate` etc. The other option is to just add `serialize=False` in the 0001 migration, given that was already implicitly the case because of `primary_key=True` adding that shouldn't cause any problems.

praseodym: Thanks for the insight, I think it's likely to be caused by the [migrations autodetector changes](https://docs.djangoproject.com/en/4.2/releases/4.0/#migrations-autodetector-changes) in Django 4.0.

I've changed the initial migration like you suggested so we can avoid the additional migration.

ammar92 (approved):

Darwinkel:
## Checklist for QA:

Interesting to see that such a major upgrade doesn't appear to cause any issues.

### What works:
* User management works without issues
* No apparent regressions

underdarknl (approved):

Issue #735, opened 2023-04-14
Unified approach to dependency management (`requirements.txt`)


Darwinkel:
**Is your feature request related to a problem? Please describe.**
Currently, we have no standardized approach to dependency management in the various modules, aside from that we use `pip`'s `requirements.txt`.

**Describe the solution you'd like**
Some things worth considering:

- Using Poetry and a lock file (implemented in Keiko and Octopoes by @Lisser)
    - Has some nice features for managing `venv`s as well   
- A single top-level `requirements.txt` instead of per module
    - How would this impact the ability to package and run the modules in this repository individually?
- How to distinguish between production/development/CI dependencies (e.g. linters)?
    - Currently, we use seperate requirements files
- How can we do this in such a way that we have minimal extra work for Debian packaging?
- Integration with dependabot is definitely nice to have

**Additional context**
https://github.com/minvws/nl-kat-coordination/pull/729

praseodym:
I'd be in favour of one single list of dependencies so that we can share a single venv across all modules. This would also greatly speed up builds (e.g. we could base all Docker images off a single Python+venv base image).

I think this should also enable us to work with a single pyproject.toml for Poetry. Poetry also supports [dependency groups](https://python-poetry.org/docs/master/managing-dependencies/) so we can split out dev/ci/formatter/docs dependencies if we'd like.

ammar92:
I would opt for poetry too and exporting of `requirements.txt` formats to be used for building the images.

> I'd be in favour of one single list of dependencies so that we can share a single venv across all modules. This would also greatly speed up builds (e.g. we could base all Docker images off a single Python+venv base image).

I'm not (yet) in favor of having a single virtual environment for the entire project, because we still consider them as isolated "microservices". And even if we don't see them as such, but as e.g. standalone packages, it would still be easier to have separate environments for both development and deployment. Although your suggestion would indeed improve build speeds.

dekkers:
I'm working on moving our Debian packages to properly use Debian packages for the python dependencies. In that case it would be impossible to install the Debian packages if they have conflicting requirements, so I don't see the value of specifying for example the FastAPI version per service if they have to be the same.

dekkers:
[...]

PR #1038, opened 2023-05-24
Replace poetry with pip-compile for keiko and octopoes

Darwinkel:
### Changes
Replaces Poetry with pip-compile for keiko and octopoes. We should probably have a new discussion about it though, as we discovered that pip-compile does not actually take multiple Python versions into consideration. It generates requirements purely for the platform and interpreter used to run the command. This could become problematic if we get more complex requirements. 

### Issue link
https://github.com/minvws/nl-kat-coordination/issues/735

### Proof
See CI.

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


Darwinkel:
Notably, `pytz` is not needed on >=3.9, and `exceptiongroup` on >=3.11

.github/workflows/check-pip-dependencies.yml
```diff
@@ -0,0 +1,33 @@
+name: Check pip dependencies
+
+on:
+  push:
+    paths:
+      - requirements**.txt
```
dekkers, line 6: Path filtering is done on the full path, so this should be `**/requirements*.txt`.

pyproject.toml
```diff
@@ -87,3 +81,93 @@ task-tags = ["Example", "todo", "TODO", "FIXME"]
...
+
+[tool.setuptools]
+packages = ["rocky", "octopoes", "keiko", "boefjes", "mula", "bytes"]
+
+[tool.setuptools_scm]
+write_to = "_version.py"
+
+[project]
+name = "openkat"
+version = "0.0.1.dev1"
```
dekkers, line 96: We shouldn't specify version because we use setuptools-scm, instead `dynamic = ["version"]` should be added under `[project]`, see also the documentation on https://github.com/pypa/setuptools_scm.

pyproject.toml
```diff
@@ -87,3 +81,93 @@ task-tags = ["Example", "todo", "TODO", "FIXME"]
...
+description = "OpenKAT coordination package"
+authors = [
+    { name = "MinVWS", email = "maintainer@openkat.nl" }
+]
+license = { text = "EUPL" }
+readme = "README.rst"
+requires-python = ">=3.8"
+
+[project.optional-dependencies]
```
dekkers, line 105: I think we should not pin the versions in pyproject.toml, because we already do that in the requirements.txt. That way we can easily upgrade everything to the latest version by using `pip-compile -U` or upgrade a specific package with `pip-compile -P`. We can also make it clear that we don't want to upgrade a certain package to newer version, for example specifying `django<5` for rocky.

pyproject.toml
```diff
@@ -87,3 +81,93 @@ task-tags = ["Example", "todo", "TODO", "FIXME"]
...
+    "sphinxcontrib-mermaid==0.8.1",
+    "myst-parser==1.0.0"
+]
+
+common = []
+common-dev = [
+    "pytest==7.2.2",
+    "exceptiongroup==1.1.1", # needed for pytest on Python <3.11
+    "httpx==0.23.3",
+]
```
dekkers, line 119: I think we should only put direct dependencies in pyproject. An easy way to figure out direct dependencies is the pipreqs tool, but that might miss some things such as pytest fixtures.

The following aren't a direct dependency and can be removed from pyproject as far as I can see:
- certifi
- dnspython
- httpx
- typing-extension

pyproject.toml
```diff
@@ -87,3 +81,93 @@ task-tags = ["Example", "todo", "TODO", "FIXME"]
...
+    "pyparsing==3.0.9",
+    "packaging==23.0",
+    "tldextract==3.4.0",
+    "opentelemetry-sdk==1.17.0",
+    "opentelemetry-exporter-otlp-proto-grpc==1.17.0",
+    "opentelemetry-instrumentation-fastapi==0.38b0",
+    "opentelemetry-instrumentation-requests==0.38b0",
+    "certifi==2023.5.7",
+    "sqlalchemy==1.4.48",
+    "typing-extensions==4.5.0"
```
dekkers, line 141: Maybe we should also sort the list of dependencies ?

dekkers (changes requested):

ammar92 (changes requested):
I can see how this replaces Poetry, but it doesn't look like what we discussed earlier this week. I was under the assumption that:
- We align the module dependencies setup (which means getting rid of Poetry and using requirements.txt format and pyproject files like what we already do with the other modules)
- We will use `pip-tools` to manage the dependencies for each module
- We still use the same (virtual) environment setup
  - A unified environment was up for debate, but not a to-do for now

dekkers:
> I can see how this replaces Poetry, but it doesn't look like what we discussed earlier this week. I was under the assumption that:
> 
>     * We align the module dependencies setup (which means getting rid of Poetry and using requirements.txt format and pyproject files like what we already do with the other modules)
> 
>     * We will use `pip-tools` to manage the dependencies for each module
> 
>     * We still use the same (virtual) environment setup
>       
>       * A unified environment was up for debate, but not a to-do for now

The PR specifies dependencies for each module and uses a different environment/container image for each module?

ammar92:
> The PR specifies dependencies for each module and uses a different environment/container image for each module?

Yes, but we currently already do that. But this PR does a bit more than what the title states and already takes the "unified" approach that we've agreed not to do for now until it's discussed. I just spoke to @Darwinkel and explained my remarks

dekkers:
> > The PR specifies dependencies for each module and uses a different environment/container image for each module?
> 
> Yes, but we currently already do that. But this PR does a bit more than what the title states and already takes the "unified" approach that we've agreed not to do for now until it's discussed. I just spoke to @Darwinkel and explained my remarks

The unified approach that we wouldn't do for now was having a single list of dependencies and one environment. This PR doesn't do that.

pyproject.toml
praseodym: I think we should still pin packages on major versions for stable projects (Django 4.x), and minor versions for projects that have more breakage (FastAPI 0.95.x). This is what the [caret (^) in Poetry](https://python-poetry.org/docs/dependency-specification/#caret-requirements) does. I don't think the [PEP-508](https://peps.python.org/pep-0508/) that `pip-compile` supports in `pyproject.toml` support this, but we can write version ranges to to the same.

dekkers: Preventing breakages is already done by pinning the versions in requirements.txt. Pinning minor versions in pyproject.toml only make sense if the intentation is to stay on those versions, which in not the case of FastAPI as those minor versions don't receive security updates as far as I know. I think it is even more important to regularly upgrade to the latest version because of the breakage, because if we are a lot of behind and there is a security issue it's harder to upgrade to the latest version with the fix because more things need to be fixed.

Makefile
```diff
@@ -91,3 +91,14 @@ debian-build-image:
 # Build Ubuntu build image
 ubuntu-build-image:
 	docker build -t kat-ubuntu-build-image packaging/ubuntu
+
+# Note about `pip-compile`: exported requirements.txt's target the Python version and platform used to run the command
```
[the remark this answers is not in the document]
praseodym: That's because pip currently does the dependency resolution, whereas with `pip-compile` the resulting `requirements.txt` file will already include all transitive dependencies. See https://github.com/jazzband/pip-tools/issues/1326 and https://github.com/pypa/pip/issues/9644 for the limitations of this.

pyproject.toml
praseodym: Disagree, if running `pip-compile -U` often results in breaking the build then we simply won't run it as often. I think it would be good to often upgrade packages to minor versions within compatible ranges, and manually go through breaking changes (like FastAPI updates) on a regular basis when there is time and expertise to fix breaking changes.

Makefile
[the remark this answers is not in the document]
dekkers: You still don't explain how these limitations make it very difficult for us to maintain. With regards to the last issue we should run pip install with `--no-deps`, given that pip-compile already resolved the dependencies it's not necessary that pip does that again anyway.

With regards to the limitations, if we use python 3.11 to generate the requirements (and also have a check in CI for that in case somebody would use a different Python version) and install with Python 3.11 in the containers, what would make it very difficult to maintain? For the Debian packages that use older Python versions there could be an issue, but that is still theoretical for now. In my opinion we should use Debian packages for Python dependencies, but even if we continue with the current packaged virtualenvs we could easily fix any issue in the Debian packaging scripts. We already give different options to pip install there anyway because of the mismatch between installing wheels from PyPI and what Debian packaging expects.

pyproject.toml
dekkers: We already have automated updating dependencies with dependabot and dependabot is not going to run less often because things break. I don't think it is a good idea to skip updates for things that are likely to break. I would rather see a dependabot PR with failing CI. The alternatives is having to remember that we have to do this manually which we wil llikely forget until it is urgent because dependabot notifies about a security issue.

Makefile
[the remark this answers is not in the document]
praseodym: As far as I understand, the `requirements.txt` with resolved dependencies that someone generates on an Apple Silicon Mac (Darwin AArch64) can be different from the resolved dependencies on a x86-64 Linux machine. This means that we'd have to keep track of a separate `requirements.txt` per pair of platform and architecture.

For other, older Python versions we could indeed re-run `pip-compile` when needed, but when we don't pin exact dependency versions in `pyproject.toml` we will probably end up with slightly different versions which is also not desirable.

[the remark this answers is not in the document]
dekkers: It can be in theory, but with our dependencies, that doesn't seem to be the case if you look at what poetry generates for keiko/octopoes. I think it is unlikely that we will ever run into this given that what we do doesn't have any specific dependencies on architecture or platform.

pyproject.toml
underdarknl: Do we still version pin dependencies of our dependencies some way if we only pin our direct dependencies?

Makefile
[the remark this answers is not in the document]
dekkers: We discussed this in the office last Thursday. While the architecture is unlikely to be a problem for OpenKAT, that pip-compile does not take Python versions into account will likely be an issue. An example is pytest only depending on the backported exceptiongroup package for Python 3.10 and earlier. There might be workarounds for this, but the whole reason for going with pip-compile was that it was simpler and having to workaround these kind issues would not make it so simple anymore. So it would better to use Poetry which does support dependencies for only specific Python versions.

Issue #1025, opened 2023-05-23
Django upgrade to 4.2 breaks CSRF


tobiasBDO:
**Describe the bug**
Just tried to upgrade to the latest version with https://github.com/minvws/nl-kat-coordination/commit/6a2ca64d5c3d6f8fe84657accba48d931db81677 and CSRF seems broken.

**To Reproduce**
1. pull main
2. make kat
3. login

**Expected behavior**
I expect to be able to login and the octopoes_api to work.

**logs**

***rocky logs***
```
rocky_1                | [23/May/2023 12:44:12] "GET / HTTP/1.0" 302 0
rocky_1                | [23/May/2023 12:44:12] "GET /nl/ HTTP/1.0" 200 5087
rocky_1                | [23/May/2023 12:44:12] "GET /static/img/kat_logo.png HTTP/1.0" 200 46258
rocky_1                | [23/May/2023 12:44:12] "GET /static/js/dropdown.js HTTP/1.0" 200 818
rocky_1                | WARNING:opentelemetry.exporter.otlp.proto.grpc.exporter:Transient error StatusCode.UNAVAILABLE encountered while exporting traces, retrying in 8s.
rocky_1                | [23/May/2023 12:44:14] "GET /account/login/ HTTP/1.0" 302 0
rocky_1                | [23/May/2023 12:44:14] "GET /nl/login/ HTTP/1.0" 200 6027
rocky_1                | Forbidden (Origin checking failed - https://public-ip does not match any trusted origins.): /nl/login/
rocky_1                | WARNING:django.security.csrf:Forbidden (Origin checking failed - https://public-ip does not match any trusted origins.): /nl/login/
rocky_1                | [23/May/2023 12:44:15] "POST /nl/login/ HTTP/1.0" 403 2572
rocky_1                | WARNING:opentelemetry.exporter.otlp.proto.grpc.exporter:Transient error StatusCode.UNAVAILABLE encountered while exporting traces, retrying in 16s.
```

***octpoes_api logs***
```
octopoes_api_1         | [2023-05-23 12:39:00 +0000] [7] [ERROR] [io_services_utils] Socket failed to connect: <socket.socket fd=18, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('172.22.0.12', 35046)>; error=111 (Connection refused)
[...]

PR #1039, opened 2023-05-24
Add external auth support

### LINKED ISSUE #963: External authentication using REMOTE_USER
Add support to rocky to use external authentication using REMOTE_USER: https://docs.djangoproject.com/en/4.1/howto/auth-remote-user/

This will make it possible to do single sign-on using software such as oauth2-proxy. We also need #762 to be able to disable 2FA for this to work properly.

dekkers:
This adds external auth support using Django [remote user middleware](https://docs.djangoproject.com/en/4.2/howto/auth-remote-user/). The header to be used can be configured using the REMOTE_USER_HEADER environment variable. The REMOTE_USER_FALLBACK variable can be set to enable the fallback to the normal model based auth backend.

I also changed the settings to only enable the OpenTelemetry middleware when SPAN_EXPORT_GRPC_ENDPOINT is set.

Fixes #963


underdarknl:
> When I log in and try to change the language I get the following:
> 
> image
I'm still having issues with the notion that we need to do a POST for changing a language. But that's unrelated to this issue.
I'm guessing somewhere along the way the cookies get lost due to some https / secure cookie flags or allowed domains not being set correctly.

dekkers:
> > When I log in and try to change the language I get the following:
> > image
> 
> I'm still having issues with the notion that we need to do a POST for changing a language. But that's unrelated to this issue. I'm guessing somewhere along the way the cookies get lost due to some https / secure cookie flags or allowed domains not being set correctly.

The problem is #1025 and unrelated to this PR. I added CSRF_TRUSTED_ORIGINS to the settings on the test server and that fixes the problem.

ammar92 (approved):

Darwinkel:
## Checklist for QA:

I personally have no external authentication services running, so I can't actually test remote user support.

### What works:
* Non-remote (model) user management still works

### What doesn't work:
* kattest01.openkat.dev throws a 500 after attempting to login ;(

PR #1056, opened 2023-05-25
Finding Types Boefjes

### LINKED ISSUE #837: Change KATFindingType source
Should become a boefje instead of rocky method

### LINKED ISSUE #838: Change CVEFindingType source

### LINKED ISSUE #839: Change SnykFindingType source
Should become a boefje instead of rocky method

### LINKED ISSUE #841: Change RetureJSFindingType source
Should become a boefje instead of rocky method

### LINKED ISSUE #842: Add normalizer helper to translate severity to score and vice versa

### LINKED ISSUE #1032: add Boefjes/Normalizers to enrich FindingType data

Lisser:
### Changes
Implements the Boefjes and Normalizers to enrich FindingType metadata in Octopoes

### Issue link
Closes #837 
Closes #838
Closes #839
Closes #841 
Closes #842 
Closes #1032

### Proof
N/A

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


Lisser:
@noamblitz 
For when you have time: can you go through the situation where the CVE API (MITRE?) does not provide a CVSS score for a CVE yet? The severity should be set to RiskLevelSeverity.UNKOWN.

Let's write a test for this?

noamblitz:
Good idea, will do tomorrow!!

noamblitz:
> @noamblitz For when you have time: can you go through the situation where the CVE API (MITRE?) does not provide a CVSS score for a CVE yet? The severity should be set to RiskLevelSeverity.UNKOWN.
> 
> Let's write a test for this?

Done

kat_finding_types.json
```diff
@@ -0,0 +1,278 @@
...
+    },
+    "KAT-TLS-1.0-SUPPORT": {
+        "description": "The server supports TLS version 1.0. This is a protocol that encrypts data traffic through a deprecated and discouraged protocol which has security risks.",
+        "risk": "low"
+    },
+    "KAT-TLS-1.1-SUPPORT": {
+        "description": "The server supports TLS version 1.1. This is a protocol that encrypts data traffic through a deprecated and discouraged protocol which has security risks.",
+        "risk": "recommendation"
+    },
+    "KAT-TLS-1.1-AND-1.2-SUPPORT": {
```
praseodym, line 58: Name is incorrect, see https://github.com/minvws/nl-kat-coordination/pull/1091

praseodym (changes requested):

Darwinkel:
## Checklist for QA:

### What works:
* New boefjes show up correctly in the katalogus
* New boefjes can be enabled and disabled
* Findings are still correctly created
* No apparent regressions

image
image

### Bug or feature?:
* Finding Types boefjes are not enabled automatically. This is probably desirable for most KAT installations.
* Where do all the new cat pictures come from? Most important question.

Lisser:
> Finding Types boefjes are not enabled automatically. This is probably desirable for most KAT installations.

Maybe we should add a flag to plugins to indicate a default on state for a plugin?

Lisser:
> Finding Types boefjes are not enabled automatically. This is probably desirable for most KAT installations.

> Maybe we should add a flag to plugins to indicate a default on state for a plugin?

Actually, I found out that the Katalogus already picks up a default value for `enabled` through the JSON definition. I commited a default `true` value for all finding type plugins. 

I don't think we were aware of this behaviour, but now we know...

Lisser:
Fixes are verified by Patrick. I simplified the FindingType type definitions, so that the JSON validation is a little less strict (this needs to be implemented differently eventually anyway).

boefjes/boefjes/plugins/kat_adr_finding_types/main.py
```diff
@@ -0,0 +1,15 @@
+from typing import List, Tuple, Union
+
+import requests
+
+from boefjes.job_models import BoefjeMeta
+
+FINDING_TYPES_JSON_LOCATION = (
+    "https://raw.githubusercontent.com/minvws/nl-kat-coordination/finding-types-in-octopoes/adr_finding_types.json"
```
praseodym, line 8: This URL points to this branch, I don't think that's what we want (after this PR the branch will be deleted). Maybe point to `main`, or the current release branch?

boefjes/boefjes/plugins/kat_kat_finding_types/main.py
```diff
@@ -0,0 +1,15 @@
+from typing import List, Tuple, Union
+
+import requests
+
+from boefjes.job_models import BoefjeMeta
+
+FINDING_TYPES_JSON_LOCATION = (
+    "https://raw.githubusercontent.com/minvws/nl-kat-coordination/finding-types-in-octopoes/kat_finding_types.json"
```
praseodym, line 8: Same for this URL

boefjes/boefjes/plugins/kat_retirejs_finding_types/main.py
```diff
@@ -0,0 +1,11 @@
+from typing import List, Tuple, Union
+
+import requests
+
+from boefjes.job_models import BoefjeMeta
+
+
+def run(boefje_meta: BoefjeMeta) -> List[Tuple[set, Union[bytes, str]]]:
+    response = requests.get("https://raw.githubusercontent.com/RetireJS/retire.js/v3/repository/jsrepository.json")
```
praseodym, line 9: If we fetch this URL, what do we need the `octopoes/bits/retire_js/retirejs.json` and `rocky/data/retirejs.json` files for?

praseodym (changes requested):

boefjes/boefjes/plugins/kat_adr_finding_types/main.py
noamblitz: Yeah you're right, but for testing purposes its a little bit a chicken-and-egg verhaal haha

dekkers: This is a public interface now, so we should also think about backwards compatibility. I think there are two options:

- We can point to the release branch or tagged release to make sure we don't  break older versions of OpenKAT, but then what's the point in fetching this from github? We could just use the version included in OpenKAT.
- We can point to main, but then we have to make sure that any changes are backwards compatible. The change we just did with renaming KAT-TLS-1.1-AND-1.2-SUPPORT would break existing OpenKAT installations that would still use the older name. We can also never delete/rename fields. If we go this route I think we shoud include a version number so we have the option to do breaking changes or clean up cruft whilekeeping support for older OpenKAT versions for the same reason an API should also always include a version.

I think it would also be better to move the json files into a subdirectory instead of cluttering the root with them.

I think this data doesn't change that much in a breaking way that we need to point to a release branch or tag, so main would be fine if we version it so that we have the option to do breaking changes. So I suggest we we move it to `data/adr_finding_types/v1.json` and `data/kat_finding_types.json/v1.json`

underdarknl: Ideally, in the future This list of Findings would be something that lives on a community maintained location (albeit with moderation), like openkat.nl. Findings that show up in your tree from Boefjes that we run can have various locations which provide them with up-to-date remedies and scores. Providing those scores only inside the application makes it harder to stay up to date with those changes.
I don't think we should bind the url to a specific version of OpenKAT, as packaged. But we should bind it to a specific version of this 'api'.
Eg, link to "https://raw.githubusercontent.com/minvws/nl-kat-coordination/finding-types-in-octopoes/adr_finding_types-v1json"
Which we can eventually supersede by:
"https://raw.githubusercontent.com/minvws/nl-kat-coordination/finding-types-in-octopoes/adr_finding_types-v2.json"
Or 
"https://openkat.nl/sources/adr_finding_types-v2.json"

noamblitz: Since finding definitions indeed change alot per release, i agree with @praseodym suggestion. Doing it per api version would cause a lot of breaks.

boefjes/boefjes/plugins/kat_retirejs_finding_types/main.py
Lisser: > the one in rocky can be removed after removing finding types from knowledge base

Let's do that in: https://github.com/minvws/nl-kat-coordination/issues/1113

boefjes/boefjes/plugins/kat_adr_finding_types/main.py
dekkers: Created #1120 for this

dekkers (approved):

PR #1057, opened 2023-05-25
Add Reverse DNS boefje

Robinzor:
Creating RDNS Boefje

Also adding DNSPTRRecord functionality.

### Changes
Simple boefje adding the OOI object DNSPTRRecord to the DNS records maybe Noam can check this to make it more efficient.

### Issue link
_Please add a link to the issue. If there is no issue for this PR, please add it to the project board directly._

### Proof
_Please add some proof of your working change here, unless this is not required (e.g. this PR is trivial)._
image

---

### Code Checklist
- [X] All the commits in this PR are properly PGP-signed and verified;
- [X] This PR only contains functionality relevant to the issue; tickets have been created for newly discovered issues.
- [X] I have performed a self-review of my code and refactored it to the best of my abilities.

### Communication
- [X] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.

---

---
## Checklist for QA:


ammar92 (commented):
Thanks, looks good! Only a few remarks to make the CI happy (https://github.com/minvws/nl-kat-coordination/actions/runs/5091301776/jobs/9188454251?pr=1057):
- Trim trailing whitespaces in `kat_rdns/main.py` (it's probably having issues due to an empty first line)
- About the `print` statement; if this is a real error you might want to replace it with a (reraised) exception, if not, replace it with a `return None` instead (it's better to be explicit about the return value of a function). I would go for the latter since it's quite obvious that a `None` result means no result or failure in this case.
- Run `black` for the correct formatting

Robinzor:
I just realised the value before .in-addr.arpa should be backwards going to implement this now.

boefjes/boefjes/plugins/kat_rdns/boefje.json
```diff
@@ -0,0 +1,15 @@
+{
+    "id": "rdns",
+    "name": "Reverse DNS",
+    "description": "Resolve IP-addreses to a hostname",
+    "consumes": [
+        "IPAddressV4",
+        "IPAddressV6",
+        "IPV4NetBlock",
+        "IPV6NetBlock"
```
dekkers, line 9: This shouldn't include IPV6NetBlock because resolving all hosts in an IPv6 netblock isn't possible.

Robinzor:
removed ipv6 block in boefje.json

boefjes/boefjes/plugins/kat_rdns/boefje.json
Robinzor: I removed it from boefje.json as suggested.

boefjes/boefjes/plugins/kat_rdns/main.py
```diff
@@ -0,0 +1,41 @@
...
+from typing import List, Tuple, Union
+
+from boefjes.job_models import BoefjeMeta
+
+
+def run_rdns(cidr):
+    network = ip_network(cidr)
+    results = []
+
+    for ip in network.hosts():
```
underdarknl, line 13: I'm having some doubts about this functionality.
If someone asks this boefje to scan a Range, two options are possible:
Scan all ip's in this range (which would duplicate the data already present if all ip's are available in the graph anyway), or do a lookup on the PTR record for that specific rage.
The first obviously could be more efficient if you want PTR info on all Ip's, than scanning all IP's one by one, but we do not have a way of limiting both of these jobs to be made.
The second is currently done (because the start IP is in the range of Ip's) but will be drowned out by all the Other less related recrods.

boefjes/boefjes/plugins/kat_rdns/boefje.json
underdarknl: I'd like to ervert this change, ipv6 ranges should be included, but similar to ipv4 ranges, they should only return the rdns directly related to the range itself, not all the included ip's

Issue #228, opened 2022-12-29
Support batched operations for saving scan_profiles and deletion


Donnype:
Making a call for every object if there are many selected in the new multi-select functionality is rather inefficient of course.

Donnype:
A setup with more objects seemed to hang quite long (having not profiled this in a production setting though)

PR #1067, opened 2023-05-30
Add endpoints in Octopoes for bulk operations in the object list page

Donnype:
### Changes
This PR adds two endpoints in Octopoes (that are also being used in Rocky):
- POST `/{client}/objects/delete_many`
- POST `/{client}/scan_profiles/save_many`

These are now used for the object list page where there is a multiselect for OOIs, where we first made an HTTP call for each selected object. In Octopoes these are wrapped in one unit of work for as well, so for N objects this save N - 1 API calls to Octopoes and N - 1 calls from Octopoes to XTDB. The bad performance is currently noticeable even when just the first page of objects is filled and you did a select all on both operations. 

I also added a new integration test layer that tests the OctopoesAPIConnector against an Octopoes api instance (see tests).

### Issue link
Fixes https://github.com/minvws/nl-kat-coordination/issues/228

### Proof
I also added some tags to the Swagger docs:
image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


Donnype:
The tests in the pipeline are way to flaky with the current xtdb-multinode bug.. Perhaps we should disable the tests for now and trust that they work for me locally until the bug has been fixed?

Donnype:
Issue has been fixed!

underdarknl:
> I definitely agree with batch processing such calls. But I don't really see the need for new endpoints. Since the resources are already plurally oriented, we could easily support single/ multiple creation and deletion with the same endpoint. E.g.
> 
> ```
> DELETE /{client}/objects with 
> POST /{client}/scan_profiles
> ```
> 
> The body format (for mostly "create" endpoints) could be `ndjson` to enable passing multiple objects and at the same time keep backward compatibility. Query strings could also work, but many clients and servers implement different maximum sizes which could cause trouble.

I would advice against query string for anything other than single items (and read operations).
We could specify any send data format by sending the Content-Type header with the request, and default to something that is plain/text like we have now, expecting only a single item.

octopoes/.ci/docker-compose.yml
```diff
@@ -38,11 +38,40 @@ services:
     command: pytest -x tests/integration
     depends_on:
       - xtdb_multinode
+      - ci_octopoes
     env_file:
       - ".ci/ci-multinode.env"
 
   xtdb_multinode:
-    image: "ghcr.io/dekkers/xtdb-http-multinode:v1.0.4"
+    image: "ghcr.io/dekkers/xtdb-http-multinode:pr-11"
```
Darwinkel, line 46: Shouldn't this become `ghcr.io/dekkers/xtdb-http-multinode:v1.0.6` now that the PR has been merged?

Darwinkel:
## Checklist for QA:

I haven't personally tested the endpoints.

### What works:
* No apparent regressions

Donnype:
@ammar92 @underdarknl To do this the nicest way I think the endpoints regarding a single object should have the primary key as a path parameter instead of a query parameter, but if my memory serves me well this is not possible with the current primary key encoding of the objects (?) and hence the scan profiles. But even in that case it is not advised to either put a lot of primary keys in query strings or a add a body to a DELETE request.

Since there was no POST endpoint for the scan profiles yet we would indeed be able to leverage that one, the risk being that we are obscuring functionality that differs from that what might be assumed for a typical set of REST endpoints (i.e. using POST to save a single object). 

All in all I don't see the benefit of less endpoints if it's at the cost of consistency in the API semantics. I _do_ see the benefit in standardizing bulk operations (see [this](https://backendless.com/docs/rest/data_multiple_objects_bulk_create.html) example) or deciding that every POST endpoint should accept ndjson and support bulk by default. But in the latter case you will into issues with DELETE requests not having request bodies.

praseodym:
+1 for using ndjson as a bulk request content type. Because we don't really follow REST anyway, we might as well switch all DELETE requests to POST too so we can use a request body there too?

dekkers:
What is the benefits of using ndjson instead of just using a list in json? I don't really see them, but see some downsides. We would create inconsistency in the APIs or need to do a lot of work to convert every API to ndjson. Dumping and parsing would be more complicated as we can't use use the standard fastapi / request json support. It would probably also be slower because the jsonlines library is in pure Python while json is implemented as a C extension.

As discussed in the standup yesterday this is also something we easily refactor later because the APIs are only called by other OpenKAT code so I am moving this to ready to merge.

praseodym:
Lack of support in fastapi is indeed a big issue. I see that there has been [some discussion about it](https://github.com/tiangolo/fastapi/discussions/7292), but it doesn't look like support is coming anytime soon. Sticking with a list in JSON sounds like a better plan then.

underdarknl:
Further work in:
https://github.com/minvws/nl-kat-coordination/issues/1125

underdarknl (approved):

PR #1083, opened 2023-05-31
Fix leaking environment variables in locally running boefjes

Darwinkel:
### Changes
I discovered that locally running boefjes have full access to all of the interpreter's environment variables, including credentials and secrets. This is obviously not desirable, as boefjes should only be able to access their explicitly set environment variables.

### Issue link
Added directly.

### Proof
Below screenshot demonstrates how the `os.environ` in the nmap boefje no longer returns all environment variables.
image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


ammar92 (approved):
Nice work. Keep in mind that it might break plugins that e.g. depend on system-wide settings such as `HTTP_PROXY`. I would still be in favor of this solution since such settings can also be added manually (plugin settings) if needed. In the future we could also filter a few of these settings if needed (e.g. clear the environment namespace except for a few settings such as proxy, locale, etc.)

underdarknl:
> Nice work. Keep in mind that it might break plugins that e.g. depend on system-wide settings such as `HTTP_PROXY`. I would still be in favor of this solution since such settings can also be added manually (plugin settings) if needed. In the future we could also filter a few of these settings if needed (e.g. clear the environment namespace except for a few settings such as proxy, locale, etc.)

I propose to Keep any ENV global vars that are set with BOEFJE_ , and make them available without that preamble if the boefje has them mentioned as such in the manifest.
This makes sure boefjes cannot claim to need just any ENV var, and allows the system admin to pass selected globals at will by setting them as correctly named duplicates of already existing vars:
Eg, by setting:
BOEFJE_HTTP_PROXY=HTTP_PROXY
The system admin would signal to KAT that BOEFJE_HTTP_PROXY is allowed to be read by the boefjes. By removing the BOEFJE_ preamble we make sure that we don't need to change most already available images and scripts, as they would be expecting HTTP_PROXY to be available and not BOEFJ_HTTP_PROXY.

Darwinkel:
This PR is ready for review in principle, but there was some discussion about the implications of the environment inheritance. The PR currently implements the translation layer described by @underdarknl. Depending on the stakeholder's requirements, we might need to reconsider this.

Some design considerations:
- Should `.env` variables override the katalogus variables?
- Should boefjes need to specify that they need certain system variables?
- What does deleting/unsetting a variable in the katalogus mean? Does it mean that a user wants the boefje's default, or inherit it from the host?
- The current prefix-based approach causes some potential redundancy and `.env` bloat. Perhaps change the name or pull it from a different place?

praseodym (approved):

Darwinkel:
Actually, before merging, maybe we should add a changelog notice that users who insert boefje settings through dotenv should manually prefix it with `BOEFJE_`

dekkers (approved):

Issue #1119, opened 2023-06-06
Rocky global user groups are broken


praseodym:
Functionality of `account_katuser_groups`, i.e. global groups independent of any organisations, is currently broken. While a global group can be assigned to a user, it wil not actually grant any permissions because the user will not be considered a member of any organisation.

The code for getting a user's organisations doesn't take the global groups into account at all:
https://github.com/minvws/nl-kat-coordination/blob/e1f8090e7d3115327c66951ff228b666ead5a517/rocky/account/models.py#L99-L107

Before we fix this bug, we might want to reconsider how this functionality is used and whether we should redesign it.

PR #1122, opened 2023-06-06
Add external auth users to default organizations

### LINKED ISSUE #1058: Ability to add user to group/organisation after external auth
#1039 introduces external authentication. After a user is authenticated, they are not yet part of any groups or organisations. We should add the ability to automatically add users to groups and organisations after they have been authenticated using the external authentication method.

We can probably extend [Django's RemoteUserBackend](https://docs.djangoproject.com/en/4.2/ref/contrib/auth/#django.contrib.auth.backends.RemoteUserBackend) with some code that adds the user to organisations/groups based on configuration in environment variables.

praseodym:
### Changes
Add setting and custom remote user backend that adds users from external authentication methods (`REMOTE_USER` header) to a list of default organisations and associated groups.

### Issue link
Closes #1058

### Proof
n/a

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


praseodym:
Note that the initial plan for this PR a was to use global groups, but these don't work as expected: #1119.

underdarknl (commented):
Default behavior for people already enlisted in one org?

rocky/rocky/auth/remote_user.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class RemoteUserBackend(BaseRemoteUserBackend):
+    """
+    Custom RemoteUserBackend that adds users to default organizations and groups.
+    """
+
+    def configure_user(self, request, user, created=True):
+        if not user.is_superuser and settings.REMOTE_USER_DEFAULT_ORGANIZATIONS:
+            user_orgs = [o.name for o in user.organizations]
```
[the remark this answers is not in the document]
praseodym: That would mean it isn't possible to add users to a new organisation later, which I think would be useful to have (e.g. on our testserver, which is the primary use case for this PR).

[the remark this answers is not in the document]
Donnype: I think the main thing that is being handled here is adding new organizations to `settings.REMOTE_USER_DEFAULT_ORGANIZATIONS`. Which makes sense I think since I already logged into our test environment but still want to be added to these organizations when I log in again.

rocky/rocky/auth/remote_user.py
```diff
@@ -0,0 +1,34 @@
...
+            for item in settings.REMOTE_USER_DEFAULT_ORGANIZATIONS:
+                organization_name, group_name = item.split(":")
```
Donnype, line 20: There are no real safeguards here now; perhaps we want to add a little check on the setting format somewhere to avoid a small configuration mistake causing no one to be able to log in anymore?

rocky/rocky/auth/remote_user.py
```diff
@@ -0,0 +1,34 @@
...
+                if organization_name not in user_orgs:
+                    logger.info("Adding user '%s' to organization '%s'", user, organization_name)
+                    organization = Organization.objects.get(name=organization_name)
```
Donnype, line 23: When are the organizations created? Or do we only add organizations to this setting after they have been added in the interface?

praseodym: They need to be created by a superuser through the KAT UI before this setting can be used.

rocky/rocky/auth/remote_user.py
praseodym: I've added exception handling so errors will be logged without breaking the login flow, do you think that's sufficient?

rocky/rocky/auth/remote_user.py
```diff
@@ -0,0 +1,37 @@
...
+                        member = OrganizationMember.objects.create(
+                            user=user,
+                            organization=organization,
+                            status=OrganizationMember.STATUSES.ACTIVE,
+                            blocked=False,
+                            trusted_clearance_level=4,
+                            acknowledged_clearance_level=0,
+                            onboarded=False,
+                        )
+                        member.groups.set([Group.objects.get(name=group_name)])
```
dekkers, line 34: After adding a new organization we should do a `del user.organizations` to invalidate the cached organization list.

praseodym: Good one, thanks! Added this.

underdarknl:
Would a different feature (where the organization name in the userOrganization relatoin table is set to a wildcard be something that could also cover a bunch of these usecases, while not having as many side-effects?

rocky/rocky/auth/remote_user.py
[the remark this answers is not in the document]
dekkers: I think that if you configure that the users and organizations are externally defined then the expectation is that the external source defines this and not a mix of an external source and what is defined in the Rocky database. If you delete a user account this will also be recreated by the Django remote user backend. If you are going to mix this it going to be hairy how it should work when things change.

We do need to document that this is how it works, I have created #1180 for this.

dekkers (approved):

PR #1132, opened 2023-06-08
Clear cache before wp-scan

noamblitz:
Clear the cache before scanning wp-scan


noamblitz:
Probably needs to be backported to 1.8 and 1.9?

underdarknl:
I am confused as to why the container seems to be storing caches between runs. Wouldn't it be reset to default every time we re-run the container if we configure it correctly regardless of what the code in there does?

noamblitz:
I would also think that, but after seeing behavior that can only be explained by caching and finding this flag in their docker docs, this should be the fix. It is currently being tested, will update it here asap.

ammar92:
This is indeed highly unusual when running containers without volumes attached. Therefore I suspect it's not actually data that is causing this, but a side effect caused by the `--clear-cache` option that results in different behavior. If we're sure this is a real solution, we should accept it with some reluctance for now.

noamblitz:
It seems that this did not resolve the problem.

Issue #859, opened 2023-05-02
Add CVE links to OOI detail


Rieven:
We want to be able to add CVE links and show it at the OOI detail page.

These links are attached to an OOI-type and can be different for each organization.

When showing an OOI detail, it will fetch all links from the OOI-type that it is attached to. 

A user can decide whether to show these links in the frontend 

De links are stored in rocky-db as follows: ID, OOI-Type, link, content, name

Example:
OOI-Type: CVEFinding
link: https://cve.mitre.org/cgi-bin/cvename.cgi?name=[cvecode]
status: Mitre details for [cvecode]

`[cvecode]` is an OOI property

PR #1139, opened 2023-06-09
A new plugin called Deep Links for adding hyperlinks to outside sources

Rieven:
### Changes
- Deep links plugin model.
- Listing through KAT-alogus, new plugin category, plugin navigation
- Create, add, edit plugin
- Enable/ disable plugon for each organization.
- Deep links added at OOI detail to generate for example cvescore links

### Issue link
Closes https://github.com/minvws/nl-kat-coordination/issues/859

### Proof
Plugins navigation
image

Listing Deep Links for an organization (enabled/disabled)
image

Create a new Deep Link
image

Generate llink using OOI id 
image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


rocky/katalogus/forms/plugin_deep_link.py
```diff
@@ -0,0 +1,30 @@
...
+
+class PluginDeepLinkForm(BaseRockyModelForm):
+    enable = forms.BooleanField(required=False)
+
+    class Meta:
+        model = PluginDeepLink
+        fields = "__all__"
+        widgets = {"ooi_type": forms.Select(choices=SORTED_OOI_TYPES)}
+        help_texts = {
+            "ooi_type": _("Choose an OOI-type where this plugin will be bounded to."),
```
underdarknl, line 22: "will be bound to", or "will be shown for"

rocky/katalogus/forms/plugin_deep_link.py
```diff
@@ -0,0 +1,30 @@
...
+            "name": _("Give your plugin an unique name."),
```
underdarknl, line 23: "a unique name"

rocky/katalogus/views/plugin_deep_link_list.py
```diff
@@ -0,0 +1,75 @@
...
+from django.views.generic import ListView
+from tools.view_helpers import Breadcrumb, get_ooi_url
+
+from katalogus.models import OrganizationPlugin
+from rocky.views.mixins import OctopoesView
+from rocky.views.ooi_view import BaseOOIBreadcrumbs
+
+
+def replace_link_id_param(link: str, key: str) -> str:
+    return link.replace("[id]", key)
```
underdarknl, line 15: I was hoping for dynamic fieldnames based on the currently loaded Object-Model.

rocky/katalogus/models.py
```diff
@@ -0,0 +1,40 @@
...
+    content = models.CharField(max_length=128, blank=False)
+    link = models.URLField(max_length=256, blank=False)
+
+    class Meta:
+        unique_together = ["ooi_type", "name"]
+
+    def unique_error_message(self, model_class, unique_check):
+        if model_class == type(self) and unique_check == ("ooi_type", "name"):
+            return _("This plugin already exists. Choose another name or OOI-Type.")
+        else:
```
underdarknl, line 18: redundant Else, you can just return

rocky/katalogus/models.py
```diff
@@ -0,0 +1,40 @@
...
+    plugin = models.ForeignKey(PluginDeepLink, on_delete=models.CASCADE)
+    enabled = models.BooleanField()
+
+    class Meta:
+        unique_together = ["organization", "plugin"]
+
+    def unique_error_message(self, model_class, unique_check):
+        if model_class == type(self) and unique_check == ("organization", "plugin"):
+            return _("This plugin already exists. Choose another organization or plugin.")
+        else:
```
underdarknl, line 36: redundant else

rocky/katalogus/templates/plugin_deep_link_detailed_list.html
```diff
@@ -0,0 +1,53 @@
...
+            {{ ooi }}
+            <div>
+                {% include "partials/ooi_head.html" with ooi=ooi view="deep_links" %}
+
+                <h2>{% translate "Deep Links" %}</h2>
+                <p class="emphasized">
+                    {% translate "Showing a list of all deep links" %}
+                </p>
+                <a class="button"
+                   href="{% url 'plugin_deep_link_add' organization.code %}">{% translate "Create deep link" %}</a>
```
underdarknl, line 20: We could add the current OOIT_type in this url and preselect it on the add form

rocky/katalogus/templates/plugin_deep_link_detailed_list.html
```diff
@@ -0,0 +1,53 @@
...
+        <section>
+            {{ ooi }}
+            <div>
+                {% include "partials/ooi_head.html" with ooi=ooi view="deep_links" %}
+
+                <h2>{% translate "Deep Links" %}</h2>
+                <p class="emphasized">
+                    {% translate "Showing a list of all deep links" %}
+                </p>
+                <a class="button"
```
underdarknl, line 19: We should check for a new permisson to see if the user has these rights as it affects the whole organization

rocky/katalogus/templates/plugin_deep_link_list.html
```diff
@@ -0,0 +1,79 @@
...
+                                <tr>
+                                    <th>{% translate "Name" %}</th>
+                                    <th>{% translate "OOI-Type" %}</th>
+                                    <th>{% translate "Actions" %}</th>
+                                </tr>
+                            </thead>
+                            <tbody>
+                                {% for deep_link in deep_links %}
+                                    <tr>
+                                        <td>{{ deep_link.plugin.name }}</td>
```
underdarknl, line 47: This can also link to the edit form for this link.

underdarknl:
As these deeplinks can be enabled by any orga, we should limit the edit, create, and delete rights to those with sufficient rights to change things on an installation level. A future version of this might install deeplinks In the organization from a remote katalogus, which would limit the scope of changes made to a deeplink to only that organization.

Rieven:
> As these deeplinks can be enabled by any orga, we should limit the edit, create, and delete rights to those with sufficient rights to change things on an installation level. A future version of this might install deeplinks In the organization from a remote katalogus, which would limit the scope of changes made to a deeplink to only that organization.

I will also add permissions

dekkers:
I don't understand the design behind this. We already have plugins and a plugin model in Katalogus, but this adds an extra plugin type in the Katalogus Django app that does not store its data in Katalogus but instead in Rocky. So the Katalogus Django app would be partly a frontend for the Katalogus service with it owns plugins and partly has other plugins stored in the rocky database. I think it is very confusing and don't really understand how this all fits in the OpenKAT architecture. It feels like instant technical debt to me.

It also think it is bad UX that an admin would need to do all this configuration work for every link type for every organisation.

rocky/katalogus/templates/plugin_deep_link_list.html
Rieven: Added!
image

Donnype (commented):
I find it hard to review this PR as it uses the name "Plugin" for something completely different than we have been doing before. Having said that, I don't think these entities should be managed in the Katalogus as has relatively little to do with the boefjes/normalizers (or, `Plugins` ;) ) given the use case. Adding (security) domain specific information to OOIs has been the responsibility of the knowledge base as far as I'm aware, so I would expect this to live in perhaps a new app, and perhaps a new tab. (Good opportunity to start a knowledge base refactor!)

I also had a hard time understanding from the ticket what the current issue for users or the added benefit of:
 - Managing this per organization (I agree with @dekkers that per-organization configuration probably is bad UX)
 - Having multiple links vs. the current `source` fields on FindingTypes

Which could also be a matter of communication in the ticket. If these are settings users actually always configure once for the whole install, perhaps sporadically tweak, I think the argument could be made that this could simply be a configuration file or something of the sorts, for instance.

In either case, we should not call these "plugins" :grin:

underdarknl:
> I find it hard to review this PR as it uses the name "Plugin" for something completely different than we have been doing before. Having said that, I don't think these entities should be managed in the Katalogus as has relatively little to do with the boefjes/normalizers (or, `Plugins` ;) ) given the use case. Adding (security) domain specific information to OOIs has been the responsibility of the knowledge base as far as I'm aware, so I would expect this to live in perhaps a new app, and perhaps a new tab. (Good opportunity to start a knowledge base refactor!)
> 
> I also had a hard time understanding from the ticket what the current issue for users or the added benefit of:
> 
>     * Managing this per organization (I agree with @dekkers that per-organization configuration probably is bad UX)
> 
>     * Having multiple links vs. the current `source` fields on FindingTypes
> 
> 
> Which could also be a matter of communication in the ticket. If these are settings users actually always configure once for the whole install, perhaps sporadically tweak, I think the argument could be made that this could simply be a configuration file or something of the sorts, for instance.
> 
> In either case, we should not call these "plugins" 😁

These links are not meant as Sources, boefjes or knowledge base items. They are meant to be quick access links for more (external) information sources.
Since they are just links, its easier for people to create them, and add them to OpenKAT. This in turn creates user-buy in, as now they have added something useful to this project without needing to overcome the steep learning curve of actual plugins.
The reason they should be 'distributed' via the Katalogus is because that is our route to enable/disable organization specific changes, and also where we offer users extra functionality. How we distribute user creates 'deeplinks' is up for discussion however. As they are too small to be zip files or something like OCI-images.
I'm thinking they are along the lines of bookmarklets in browsers, but then installed for all users in an organization, and with a central appstore for distribution.

ammar92:
Closed for now to better refine this and discuss architecture and implementation

PR #1156, opened 2023-06-13
Fix expiring dict while flushing cache in scheduler

jpbruinsslot:
### Changes

It seems when the cache needs to be flushed, the time it takes to fill the cache the ttl could potentially pass and subsequently trigger an ExpiredError again which it then cannot handle anymore. With this change it would allow you to turn off the raising of an ExpiredError when you're trying to rebuild the cache. With this PR a additional boolean flag is added to an expiring dict, to signifiy to not raise an `ExpiredError` when using the the `ExpiringDict`.


ammar92:
It's not immediately clear why you need the toggle, could you explain it a bit?

underdarknl:
Is a more specified value not a more useful signature for this method?
Eg, set to True or False as an argument to this method, so as to not get in a position where everything gets inverted due to one missed call somewhere?

jpbruinsslot:
> Is a more specified value not a more useful signature for this method? Eg, set to True or False as an argument to this method, so as to not get in a position where everything gets inverted due to one missed call somewhere?

Sure

jpbruinsslot:
> It's not immediately clear why you need the toggle, could you explain it a bit?

It seems when the cache needs to be flushed, the time it takes to fill the cache the ttl could potentially pass and subsequently trigger an `ExpiredError` again which it then cannot handle anymore. With this change it would allow you to turn off the raising of an `ExpiredError` when you're trying to rebuild the cache.

mula/scheduler/utils/dict_utils.py
```diff
@@ -37,12 +38,26 @@ def reset(self) -> None:
             self.cache.clear()
             self.expiration_time = datetime.now(timezone.utc) + self.lifetime
 
+    def set_expiration_enabled(self, value: bool) -> None:
```
ammar92, line 41: This could also have been a property (e.g. using `@property.setter`)

ammar92 (commented):
Looks good to me. I only had one remark that could be considered

Darwinkel:
## Checklist for QA:

### What works:
* No aparrent regressions
* No `ExpiredError` present in the scheduler logs

ammar92 (approved):

dekkers (approved):

Issue #862, opened 2023-05-02
Aggregate port findings


zcrt:
**Is your feature request related to a problem? Please describe.**
When generating a .pdf report every open ports takes multiple lines such as:
```

IPPort—internet—<IP/>—tcp—<PORT/>
Port <PORT/> is not a common port and should possibly not be open

```
For multiple IPs and ports this can get cluttered. I would rather see a table with for every IP a list of open ports that is found by KAT.

**Describe the solution you'd like**
One solution is to implement a bit that aggregates all open port findingss into a single finding with a list of open ports for every IP. That would already clean up a lot of lines in the `.pdf` report.

underdarknl:
> 

agreed

noamblitz:
Im currently checking what implications it has for the Sources if we change input_ooi from port to ipaddress.

noamblitz:
So it will be problematic, new sources for the ipaddress will not be created. A solution could be to change the rerun bit endpoints to first delete and recreate all sources. That will solve this. Agreed?

noamblitz:
Draft PR: https://github.com/minvws/nl-kat-coordination/pull/1172

PR #1172, opened 2023-06-13
Change port classification bit to ip address

noamblitz:
### Changes
Move port classification bit which allows for aggregate findings when the correct config ooi exists. 

Closes: https://github.com/minvws/nl-kat-coordination/issues/862

This pr also introduces a fix for ooi add form when field type is Dict[str,str] which is the case for Config.config.

Still needs an update of the rerun bits endpoint and test fixes.


Lisser:
The recalculate bit logic, now also makes sure the Origins and OriginParameters exist in the graph.

This is done by:
- For each BitDefinition:
  - fetching all OOI's that are in the Bit's `consumes` field
  - For each OOI:
    - upsert Origin
    - For each Bit param:
      - query related objects according to the Path
        - For each related OOI
          - upsert OriginParameter

This is quite expensive..

noamblitz:
Tested functionality and works fine. Will work on tests

octopoes/octopoes/core/service.py
```diff
@@ -516,7 +523,42 @@ def get_scan_profile_inheritance(
...
+                for param_def in bit_definition.parameters:
+                    path = Path.parse(f"{param_def.ooi_type.get_object_type()}.{param_def.relation_path}").reverse()
+
+                    param_oois = self.ooi_repository.list_related(ooi, path, valid_time=valid_time)
+                    for param_ooi in param_oois:
+                        param = OriginParameter(origin_id=bit_instance.id, reference=param_ooi.reference)
+                        self.origin_parameter_repository.save(param, valid_time)
+
+        # remove all Origins and Origin Parameters, which are no longer in use
+
```
dekkers, line 557: Shouldn't there be code here that does this?

noamblitz: I think this is made to be only additive for now (all tough deleted bits will have no code and therefore return an empty list of OOIs which would me in work in practice).

Darwinkel:
## Checklist for QA:

### What works:
* Adding the below Config OOI related to Network|internet, and then re-running all bits results in aggregated findings in both the web UI as well as the pdf. 
image
image
image

How Config OOI's can be added is currently undocumented and we should do this soon. Would be nice if this PR came with updated docs?

underdarknl:
The Config OOI could be added via a Procedure.
To do this: We'd need a Bit that finds 'networks' without this specific Config available.
Then, this bit adds a QuestionOOI asking for the setting via a jsonschema defined form which in turn would result in a ConfigOOI on the related Network.
@Donnype this seems like a good initial usecase.

noamblitz:
Agree!! Question: on ports found on this network, do you want a finding per port or one finding for all open ports. Altough a bit abstract, a perfect usecase for questions.

dekkers (approved):

PR #1187, opened 2023-06-14
Fix hanging worker processes on a SIGKILL (e.g. when the container is out of memory)

Donnype:
### Issue

NOTE: I CONFIGURED MY SETUP TO LIMIT MEMORY USAGE TO 512M FOR THE BOEFJE CONTAINER

This PR is the result of some extensive investigation into multprocessing vs. Docker (sending system calls) behavior. Previously the setup was as follows:

```mermaid
graph LR

subgraph Process 0

SchedulerRuntimeManager -- "starmap(start_working, args)" --> multiprocessing.Pool
multiprocessing.Pool -- creates --> Worker-1["Worker 1<br/><i>target=start_working,<br/>args=args,<br/>daemon=True</i>"]
multiprocessing.Pool -- creates --> Worker-2["Worker 2<br/><i>target=start_working,<br/>args=args,<br/>daemon=True</i>"]

subgraph Process 1
    Worker-1
    Worker-1 -- runs --> Plugin1[" Plugin"]
end

subgraph Process 2
    Worker-2
    Worker-2 -- runs --> Plugin2[" Plugin"]
end
end

Worker-1 -- polls --> Scheduler1["Scheduler"]
Worker-2 -- polls --> Scheduler2["Scheduler"]
``` 

Whenever a worker runs out of memory (or in general: received a SIGKILL), the process terminates. The behavior of the Pool (as desired) is to create a new daemon process to keep the pool size constant. However, whenever the Pool creates a new daemon process, it does not recreate the same target/entrypoint (not desired):

```mermaid
graph LR

subgraph Process 0

SchedulerRuntimeManager -- "starmap(start_working, args)" --> multiprocessing.Pool
multiprocessing.Pool -- creates --> Worker-1["Worker 1<br/><i>target=None,<br/>daemon=True</i>"]
multiprocessing.Pool -- creates --> Worker-2["Worker 2<br/><i>target=start_working,<br/>args=args,<br/>daemon=True</i>"]

subgraph Process 1
    Worker-1
end

subgraph Process 2
    Worker-2
    Worker-2 -- runs --> Plugin2[" Plugin"]
end
end

Worker-2 -- polls --> Scheduler2["Scheduler"]
``` 

The fact that the parent process hangs indefinitely when all child processes are killed is [an issue of its own](https://bugs.python.org/issue9205), but for our use-case  we do not want our worker processes to crash in the first place when we run boefjes that generate SIGKILL's; this should be gracefully handled.

### Solution

To handle the SIGKILL's gracefully, the only way in the current setup is to run the plugins in a subprocess themselves again. Then this subprocess is targetted by the OS when memory limits are reached, which can be detected by the worker and handled. This means however that the process cannot be a daemon process, and hence we cannot use the processing pool to keep the number of actively running processes constant. This is not too much work to do manually however:

```mermaid
graph LR

subgraph Process 0

SchedulerRuntimeManager ----> multiprocessing.Process
multiprocessing.Process -- creates --> Worker-1["Worker 1<br/><i>target=start_working,<br/>args=args,<br/>daemon=False</i>"]
multiprocessing.Process -- creates --> Worker-2["Worker 2<br/><i>target=start_working,<br/>args=args, <br/>daemon=False</i>"]

subgraph Process 1
    Worker-1
    Worker-1 -- runs --> Plugin1[" Plugin"]
    subgraph Process 1.1
        Plugin1[" Plugin <br/><i>timeout=timeout</i>"]
    end
end

subgraph Process 2
    Worker-2
    Worker-2 -- runs --> Plugin2[" Plugin"]
    subgraph Process 2.1
        Plugin2[" Plugin <br/><i>timeout=timeout</i>"]
    end
end
end

Worker-1 -- polls --> Scheduler1["Scheduler"]
Worker-2 -- polls --> Scheduler2["Scheduler"]
```

An immediate benefit is that we can pass a _timeout_ to `Process.join()`. This means we can now start adding specific timeouts for boefjes if we want to without much effort. I added a global settings variable defaulting to None for the timeout. This way we could cap runtimes to e.g. 12 hours. Boefjes can also specify their own timeouts either in their definition file or in the settings.

---

### Up for discussion

- Should we create the subprocess in the LocalRunners instead, as we have more information such as the fields from the manifest file and access to the settings at that layer? This would introduce some complexity and refactoring work as this affects a big part of the tests and modules if we do it by default.
- Can't we recommend to defer replication and restart logic to a container orchestration tool by default (i.e. skip multiprocessing altogether for a specific entrypoint). Then we would only use this for debian packages and on-prem installs. 

### Before:
I ran a test boefje that eats up all your memory instantly:

```python3
from typing import List, Tuple, Union

    my_memory_eating_list = []

    while True:
        my_memory_eating_list.append(3.141592653589793238)

    return [(set(), b"")]
```

Which shows the following hanging logs:

before_memory_eater

The `docker stats` output showing the memory peak and the "flatlining" right after, as observed at Zcert:

boefje_died

### After:

The exception being handled and properly logged, including the received `SIGKILL` of the subprocess:

neat_memory_eater_logs

And the failure in the task list:

image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


praseodym:
The [parent process hanging issue](https://bugs.python.org/issue9205) was marked as fixed because `concurrent.futures.ProcessPoolExecutor` now detects killed child processes.

The rest of the issue thread has some more interesting tidbits:

> > Speaking of which, I wonder why we have both multiprocessing.Pool and concurrent.futures.ProcessPoolExecutor. They seem to fill the exact same needs, with different APIs...
> 
> Brian and I need to work on the consolidation we intend(ed) to occur
> as people got comfortable with the APIs. My eventual goal is to remove
> anything but the basic multiprocessing.Process/Queue stuff out of MP
> and into concurrent.* and support threading backends for it.

This all happened in 2011 and I'm not sure much as changed since then, the multiprocessing library looks mostly the same. A  [new issue was opened to actually fix multiprocessing](https://bugs.python.org/issue22393) but while there is a patch, it has not been merged yet.

This makes me doubt whether we can actually build a reliable system on top of multiprocessing and I think we should look into alternative solutions, such as the ones @Donnype mentioned in 'Up for discussion'.

dekkers:
Starting an extra process for every task execution seems to me we start more processes than necessary. You just need one main process to check the subprocesses and handle when they are getting killed. The problem is that we don't listen for tasks in the main process, but in the subprocess which is not how this is usually done. 

The usual way to do this is to have one main process which listens for new tasks or incoming requests and pass those on to a subprocess for execution. The main process is also responsible for the subprocesses, starting them, handle being killed, etc. This is how it is implemented in things like celery and gunicorn. That way you have just n+1 process running for n workers instead of (n*2)+1 with this PR. It would need some more refactoring and I am not sure whether we want to do this now.

Donnype:
@dekkers We can definitely do this with n + 1 processes, the caveat being that we will have to check whether a process is busy before passing a new task, or perhaps adding a `multiprocessing.Queue` that we only fill once the queue is empty, which may be a bit of extra management. I'll have another look today.

Donnype:
Update:

Now we have n + 1 processes while managing a multiprocessing.Queue and keeping track of the task that was being handled by the worker to set the status to failed when the worker was killed.

New setup, where new workers are created when a worker was killed:


```mermaid
graph LR

SchedulerRuntimeManager -- "pop()" --> Scheduler

subgraph Process 0

  multiprocessing.Queue
  SchedulerRuntimeManager -- "put(p_item)" --> multiprocessing.Queue["multiprocessing.Queue(maxsize=2)"]

  Worker-1["Worker 1<br/><i>target=_start_working,<br/>args=args</i>"] -- "get()" --> multiprocessing.Queue

  subgraph Process 1
    Worker-1
    Worker-1 -- runs --> Plugin1["Plugin"]
  end

  Worker-2["Worker 2<br/><i>target=_start_working,<br/>args=args</i>"] -- "get()" --> multiprocessing.Queue

  subgraph Process 2
      Worker-2
      Worker-2 -- runs --> Plugin2["Plugin"]
  end
end
``` 

Rough representation of the failure mode when a SIGKILL is sent to the worker:

```mermaid
sequenceDiagram
  participant SchedulerRuntimeManager
  participant handling_tasks
  participant Worker1
  participant Scheduler
  participant Worker2
  Worker1->>handling_tasks: set p_item.id for worker.pid
  SchedulerRuntimeManager->>Worker1: if not is_alive()
  SchedulerRuntimeManager->>handling_tasks: get p_item.id for worker1.pid
  SchedulerRuntimeManager->>Scheduler: set p_item.status to FAILED
  SchedulerRuntimeManager->>Worker1: close()
  SchedulerRuntimeManager->>Worker2: start()
```

---

Logs showing the handling of the failing memory_eater boefje:

handling_memory_eater

Task list for this boefjes:

set_status_to_failed

boefjes/boefjes/app.py
```diff
@@ -31,109 +28,192 @@ def __init__(
...
+                    worker.terminate()
+                    worker.join()
+                    worker.close()
+
+                raise
+
+    def _fill_queue(self, task_queue: mp.Queue, queue_type: WorkerManager.Queue):
+        if task_queue.full():
+            time.sleep(self.settings.poll_interval)
+            return
```
dekkers, line 70: I don't know what the supposed purpose of this sleep is (adding comments to code like this would help). Putting something on the queue already blocks if the queue is full, so we don't need to check if the queue is full and sleep for that. If the poll interval is higher sleeping for the poll interval will also mean the throughput is lower if tasks are executed faster than the poll interval.

If it is supposed to run `_check_workers` to make sure that we don't block when all the workers have died, this won't work reliably with multiple organisations because we can still block when we put multiple tasks on the queue later. I think it would be better to call `task_queue.put` with a timeout (which can be large, because this isn't supposed to happen) and run `_check_workers` if there is a timeout and then try to put the task on the queue again.

dekkers (changes requested):

boefjes/boefjes/app.py
Donnype: Good spot on the multiple organizations issue! I believe this is partly an artifact from a prior implementation and partly indeed to quickly spot broken workers. Checking the workers when the queue does not shrink for a while makes sense, but for large timeouts `t` we could theoretically block for `t` seconds in a worst case scenario.

Another possible solution would be to remove the `maxsize` and only fill the queue when `task_queue.qsize()` is below `n * #workers` for some `n`. Then we would check the workers every `self.settings.poll_interval` to quickly drop hanging workers and only lose some efficiency if all workers would finish `n` tasks in the interval (or, on average at least). @dekkers Thoughts?

I also realized that there is no graceful handling of dispatched tasks on the queue on shutdown.. (Which would be even more relevant if we increase the  queue size.) I Will have a look at that as well.

dekkers: I think not setting maxsize makes sense as that would prevent blocking when putting things on the queue.

If we want to sleep when all workers are busy I think it would better to introduce another setting for this as the length of poll interval is not relevant for this. For example if you have lots of small organisations you might set the poll interval higher to not hammer the scheduler, but that doesn't mean you want to sleep that long when all workers are busy. Maybe we should also sleep shorter intervals and check again if the queue is still full multiple times before checking the workers. E.g. sleep 1 sec, check queue, sleep 1 sec, check queue, ..., slept more than 10 times? check workers or something like that. That would prevent sleeping too long.

Donnype: A separate setting makes sense. Perhaps in the end we actually just want to be able to define a **rate limit** on the scheduler, a **heartbeat interval** on the workers and a **poll interval** on the queue, but that would perhaps require separate threads to keep things readable. Also, only checking the workers when the queue is full for 10 iterations would open up the possibility that we only detect a dead worker when the load is high, if I understand your suggestion correctly.

I'll create a separate setting and perhaps a feature request towards the scheduler to look into introducing an endpoint that would give exactly what these workers, at least in the common case, need: per `queue_type` at most `n` items that could immediately be dispatched to the multiprocessing queue ( E.g. `POST queues/pop?limit=3` vs. the current `POST queues/{queue_id}/pop`). That would reduce the calls to the scheduler significantly when you have a lot of organisations and also make this configuration value actually represent the real interval with which we ping the scheduler.

dekkers: I think the maxsize of the queue still needs to be removed to prevent potential blocking in calling `put`. The check in `_fill_queue` will then mean that the maximum number of tasks on the queue is pool_size+number of organisations.

Donnype: Ah indeed

Donnype: Done

Darwinkel:
## Checklist for QA:

### What works:
* No apparent regressions

underdarknl (approved):

PR #1202, opened 2023-06-16
Fix crisis room not correctly handling Octopoes error

dekkers:
It will show 0 findings and an error message. Also adds test case to make sure it doesn't happen again.


praseodym (approved):

praseodym:
My local environment also has this issue, fix works for me.

Lisser (approved):

underdarknl (approved):

Issue #1162, opened 2023-06-13
bootstrap_octopoes creates new AMQP connection for every API request


praseodym:
The `bootstrap_octopoes` function is used as a FastAPI dependency in the Octopoes API. This function creates a new AMQP (RabbitMQ) connection for every API request, even though connections can be shared and not every API call requires an AMQP connection.

Relates to #632, #1127, #1128

Lisser:
We could probably utilize the `startup` event and `app.state` to create a global connection.
```python
@app.on_event("startup")
async def startup():
    rabbit_connection = pika.BlockingConnection(pika.URLParameters(settings.queue_uri))
    channel = rabbit_connection.channel()
    channel.queue_declare(queue="create_events", durable=True)
    app.state.rabbit_connection = rabbit_connection
```

praseodym:
Let's make sure that reconnects also work properly, so that Octopoes doesn't have to be restarted when its connection to RabbitMQ breaks.

Donnype:
@Lisser If we ran FastAPI 0.93.0 we could also use [lifespan events](https://fastapi.tiangolo.com/advanced/events/).

PR #1204, opened 2023-06-16
Cache RabbitMQ connections and HTTP Sessions in Octopoes

Donnype:
### PLEASE QA THIS PROPERLY, I HAVE SEEN SOME ERRORS DUE TO RABBIT GOING DOWN IN THE LOGS BUT I AM NOT SURE IF THIS WAS DUE TO STARTUP RACE CONDITIONS.

- Cache the octopoes http session and use one rabbitmq connection per celery worker, and one rabbitmq connection for the api (one process).
- Use native support from FastAPI and Celery to spawn connections on startup of the worker/api. 
- Use native support from FastAPI and Celery to close RabbitMQ connections on shutdown of the worker/api. 

Some considerations and issues with the current approach:
- [Pika does not claim that the blocking connections are thread safe](https://pika.readthedocs.io/en/latest/faq.html), but the other services such as the scheduler has not experienced issues with this. For now I have added a `threading.local()` instance to resolve this.
- Running uvicorn with the `--reload` flag means that if we cannot connect to RabbitMQ on application startup (reasonable scenario when we rebuild kat for instance, or when we temporarily kill rabbitmq), the container does not exit. Else the container would kill itself when you code a syntax error during development. But this means that we have to build in retry logic for the initial connection ourselves, so I added RabbitMQ as a dependency that has to be healthy in the docker-compose file. There are still edge cases here where you start octopoes_api with `--no-deps` and RabbitMQ is down. You have to then manually restart Octopoes API since there is no retry logic at the application level implemented. A code change would however be detected by Uvicorn and restart the api successfully in that scenario. We could:
  * Ignore the exception and start the application anyway: then we might see the exception re-appear but at least the application is started so we would stay live and try creating connection until it succeeds [current solution]
  * Build in (longer) retry logic at startup as a lot of Octopoes is not really usable without RabbitMQ
  * (Interesting and ugly thing that I found out also works: if we catch the startup-connection-exception and kill the uvicorn process using `os.kill(os.getppid())`, the container actually properly dies and is restarted by docker due to our `restart: unless-stopped` policy! :nerd_face: )

Celery seems to have proper logic to respawn the process when we cannot connect workers on `worker.init`.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/1162

### Celery workers

Celery workers connecting on startup:

celerycon

Celery workers disconnecting on shutdown:

celdis

---

### FastAPI

FastAPI connecting on startup:

rabbitcon

FastAPI disconnecting on shutdown:

rabbitclose

FastAPI failing to connect but starting anyways:

image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


ammar92 (approved):

Darwinkel:
## Checklist for QA:

### What works:
* No apparent regressions
* Restarting containers sequentially one-by-one results in a working deployment

Donnype:
I added `pytest-timeout` to fix at least the sporadic hanging of the integration test.

octopoes/octopoes/api/router.py
```diff
@@ -76,21 +76,24 @@ def settings() -> Settings:
...
+def xtdb_session(
+    client: str = Depends(extract_client), settings_: Settings = Depends(settings)
+) -> Generator[XTDBSession, None, None]:
+    session = XTDBSession(get_xtdb_client(settings_.xtdb_uri, client, settings_.xtdb_type))
+
+    try:
+        yield session
+    finally:
+        session.commit()
+        logger.info("Committed XTDBSession")
```
dekkers, line 88: The old code also did not commit when an exception happened and I don't think that we want that, so we shouldn't do that inside finally.

dekkers (changes requested):

Darwinkel:
Re-tested. Still works fine! :)

dekkers (approved):

Issue #1195, opened 2023-06-15
Implement function to create a hash of the code used to run a Boefje (files/container)


Darwinkel:
Parent task: https://github.com/minvws/nl-kat-coordination/issues/1069

**Requirements**
- [ ] Should come with a working pipeline to zip & hash code files for local boefje runners
- [ ] Should be flexible enough to generalize to a container environment (put this logic at the runner-level?)

**Thoughts/design**
- What is the most logical place for this code? Should it be an interface of the BoefjeJobRunner, which is then implemented by e.g. LocalBoefjeJobRunner? Or perhaps the PluginRepository or BoefjeResource?

Issue #1196, opened 2023-06-15
Implement helper functions to collect Boefje job input information


Darwinkel:
Parent task: https://github.com/minvws/nl-kat-coordination/issues/1069

**Requirements**
- [ ] Should provide used environment variables
- [ ] Should provide used CLI arguments
- [ ] Should provide the input OOI 
- [ ] Should provide the stdin

Darwinkel:
In the `boefje_meta` table, we currently store the environment dictionary explicitly which is nice (#1227). The `arguments` column is a huge blob with input-related stuff. I think I'm going to have to deconstruct, clean, and split it where necessary.

Darwinkel:
I triple-checked; there does not appear to be any other input to take into consideration except for `boefje_meta.arguments`, which includes all input and arguments.

PR #1227, opened 2023-06-19
Retrieve and store env and code hashes of a Boefje job

Darwinkel:
### Changes
Adds functionality to retrieve and store the environment variables and hash of the code used as input for a Boefje job.

### Issue link
Closes https://github.com/minvws/nl-kat-coordination/issues/1195
(Partially implements https://github.com/minvws/nl-kat-coordination/issues/1196 and https://github.com/minvws/nl-kat-coordination/issues/1196 as well)

### Proof
image
image
image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


Darwinkel:
I do wonder whether `environment` should default to `null` or `{}` in the database. `null` would be more storage-efficient.

boefjes/boefjes/katalogus/local_repository.py
```diff
@@ -158,3 +160,25 @@ def _normalizer_to_plugin(normalizer: NormalizerResource) -> Normalizer:
...
 def get_local_repository():
     return LocalPluginRepository(BOEFJES_DIR)
+
+
+def get_runnable_hash(path: Path) -> str:
+    """Returns sha256(sha256(file1) + sha256(file2) + ...) of all files in the given path."""
+
+    file_hashes = []
+
+    for file in path.iterdir():
```
praseodym, line 170: `iterdir` doesn't iterate in a stable order, which can cause different hashes between runs.

ammar92: Indeed. You could sort this iteration which should give a pretty stable outcome

ammar92:
> I do wonder whether `environment` should default to `null` or `{}` in the database. `null` would be more storage-efficient.

If it's nullable you may want to use a null; with a null meaning not set. If it's not nullable, then you probably want to use an empty object `{}`

Given the number of excepted rows for a jobs table, I would go for this first option.

boefjes/boefjes/katalogus/local_repository.py
```diff
@@ -141,6 +142,7 @@ def _boefje_to_plugin(boefje: BoefjeResource) -> Boefje:
         def_file = boefje.path / "boefje.json"
         def_obj = json.loads(def_file.read_text())
         def_obj["repository_id"] = RESERVED_LOCAL_ID
+        def_obj["runnable_hash"] = get_runnable_hash(boefje.path)
```
Darwinkel, line 145: I was also thinking that the `BoefjeResource` is loaded every time `resolve_boefjes()` and `by_id` are called, which is every time a job is created. Meaning we iterate repeatedly over the filesystem, create importable packages of, and calculate hashes for, all Boefjes in the katalogus. Not the most efficient flow if we're only interested in running a single Boefje. Candidate for future refactoring?

ammar92: Great observation. Modern OSs and filesystem drivers actually do quite a good job at (partially) caching these kinds of situations, but there's still a lot of performance gain if we cache this ourselves. Listing directory content isn't a big deal, but extensively reading out files and calculating hashes is quite expensive. Definitely a candidate for future refactoring. But we need to select a good cache key which usually is a combination of the file name and modification time.

boefjes/boefjes/katalogus/local_repository.py
Darwinkel: Ah of course! Completely overlooked the sorting order stability.

boefjes/boefjes/katalogus/local_repository.py
```diff
@@ -158,3 +160,25 @@ def _normalizer_to_plugin(normalizer: NormalizerResource) -> Normalizer:
 
 def get_local_repository():
     return LocalPluginRepository(BOEFJES_DIR)
+
+
+def get_runnable_hash(path: Path) -> str:
+    """Returns sha256(sha256(file1) + sha256(file2) + ...) of all files in the given path."""
```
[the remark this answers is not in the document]
praseodym: Good point, a single `hashlib.sha256()` object to which all file contents are 'added' should be a better implementation. Example: https://docs.python.org/3/library/hashlib.html#usage

boefjes/boefjes/katalogus/local_repository.py
praseodym, line 170: Do we also want to hash `__pycache__` folders with `*.pyc` files in them? There's arguments for this (if the pyc files are changed, the actual running code changes) but also against (if the pyc files are created only after the first hash has been made, the hash will change even though the actual code did not).

ammar92: Ah, good point. It seems easier to ignore these files, although a cache or hash miss for the first run wouldn't hurt too much... I'd choose the first option, but either way is fine for me

boefjes/boefjes/katalogus/local_repository.py
praseodym: Only calculating the hash before a Boefje job is ran makes more sense to me. Can we change the API to make this possible?

Donnype: Rescanning the boefjes is quite nice in terms of quick feedback during development, so let's profile to verify the impact before creating some sort of caching layer.

bytes/bytes/database/sql_meta_repository.py
```diff
@@ -252,6 +252,8 @@ def to_boefje_meta_in_db(boefje_meta: BoefjeMeta) -> BoefjeMetaInDB:
         organization=boefje_meta.organization,
         started_at=boefje_meta.started_at,
         ended_at=boefje_meta.ended_at,
+        runnable_hash=boefje_meta.runnable_hash,
+        environment=boefje_meta.environment,
```
Donnype, line 256: These should be added in `to_boefje_meta` as well.

boefjes/boefjes/katalogus/local_repository.py
```diff
@@ -158,3 +160,25 @@ def _normalizer_to_plugin(normalizer: NormalizerResource) -> Normalizer:
...
+
+    file_hashes = []
+
+    for file in path.iterdir():
+        if file.is_file():
+            sha256 = hashlib.sha256()
+            with Path.open(file, "rb") as f:
```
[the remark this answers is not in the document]
Darwinkel: I've reworked it

boefjes/boefjes/katalogus/local_repository.py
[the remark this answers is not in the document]
Darwinkel: I've reworked it

Darwinkel:
> > I do wonder whether `environment` should default to `null` or `{}` in the database. `null` would be more storage-efficient.
> 
> If it's nullable you may want to use a null; with a null meaning not set. If it's not nullable, then you probably want to use an empty object `{}`
> 
> Given the number of excepted rows for a jobs table, I would go for this first option.

Did some experimenting with Ammar; actually nullable JSON columns are a pain and an empty dict is the path of least resistance here.

boefjes/boefjes/katalogus/local_repository.py
Darwinkel: As an alternative to guarantee reproducibility, we could also clear the folder/cache before we run each boefje (e.g. all boefjes are run from scratch), or make sure that the cache is always created prior to running the first job. But I can imagine that these approaches require additional engineering and design.

bytes/bytes/database/sql_meta_repository.py
Darwinkel: Done

boefjes/boefjes/katalogus/local_repository.py
Darwinkel: On the other hand: is it unreasonable to trust that the Python interpreter will automatically invalidate the cache if there is a mismatch between the two?

boefjes/boefjes/katalogus/local_repository.py
dekkers: Container images already have hashes so no need to calculate any hashes for that. What we do here is with every call to those methods we calculate the hashes of files shipped with OpenKAT. It would just be a waste of resources, we could even calculate the hashes once in CI for everybody or just record the OpenKAT version / commit id that was used.

And if a malicious user can change the code of an OpenKAT install I don't see how generating hashing is going to help anything because the malicious user can also change the code that does the hashing.

boefjes/boefjes/katalogus/local_repository.py
praseodym: > On the other hand: is it unreasonable to trust that the Python interpreter will automatically invalidate the cache if there is a mismatch between the two?

As far as I know, it doesn't. While there is [PEP 552 – Deterministic pycs](https://peps.python.org/pep-0552/), when it was introduced in Python 3.7 ["By default, Python still uses timestamp-based invalidation and does not generate hash-based .pyc files at runtime."](https://docs.python.org/3/whatsnew/3.7.html#pep-552-hash-based-pyc-files).

This means that a malicious actor could replace the pyc files with different code, make sure the timestamp is correct, and have their code run instead of the content of the py file.

Darwinkel: @praseodym I did some research into pyc cache poisoning, and while it doesn't appear that it is actively/frequently exploited, it also doesn't appear that the interpreter takes any serious precautions by default. 

https://www.reversinglabs.com/blog/when-python-bytecode-bites-back-who-checks-the-contents-of-compiled-python-files

https://community.broadcom.com/symantecenterprise/communities/community-home/librarydocuments/viewdocument?DocumentKey=f2d93dd5-00e8-4fd3-83a7-a547c42ac529&CommunityKey=1ecf5f55-9545-44d6-b0f4-4e4a7f5f5e68&tab=librarydocuments

In the linked PEP, it says this, which is not very encouraging:
>We will use a [SipHash](https://131002.net/siphash/) with a hardcoded key of the contents of the source file. Another fast hash like MD5 or [BLAKE2](https://blake2.net/) would also work. We choose SipHash because Python already has a builtin implementation of it from [PEP 456](https://peps.python.org/pep-0456), although an interface that allows picking the SipHash key must be exposed to Python. Security of the hash is not a concern, though we pass over completely-broken hashes like MD5 to ease auditing of Python in controlled environments.

boefjes/boefjes/katalogus/local_repository.py
underdarknl: We are currently not creating those containers, nor hashing them. 
This is a step in the right direction to do so for plain python code. Yes, there are ways around this, but doing things reasonably well, is better than not doing them at all.
The usecase for an auditor is to spot changes in code since it was ran. This hash will end up in the timestamped remote signed hash, and as such gives the auditor a reasonable guess that the code that was used to generate a result is the code he is seeing in front of them (and can hash to produce the same result). Since hashing the pyc files produces possibly changing hashes depending on the platform, runtime or time of day, those should not be part of that proof. 

If someone is willing to change their OpenKAT install, mock the hashing code with something they can influence in order to produce timestamped proofs of data that they are presenting to their auditor, there's not much we can do expect run a trusted OpenKAT build on a trusted computer. And guess what? Those users who indent to fool their auditors wont use that system, and instead will opt just to not run OpenKAT and not have and proofs for their auditing or compliance statements.

boefjes/boefjes/katalogus/local_repository.py
underdarknl: yay, its an easy win, adds very little overhead, and makes the whole setup a bit more hardened.
We can never guard against true malicious intent, but we can harden against small issues.
to make sure that we have reproducable hashes in our datastore, we should not include the pyc file hashes in the hash, as pyc files are not generated exactly the same for every run/runtime that builds them I'd guess.

ammar92 (dismissed):
Looks good to me 👍 Most remarks have already been tackled and a lot has been discussed offline too. I still wish there was an easier way of handling key-value data within the table, but this implementation also works fine.

ammar92:
Tested and seems to work
image

underdarknl (approved):

Issue #1231, opened 2023-06-20
CVE finding type normalizer fails if there is no CVSS3 score


dekkers:
The CVE finding type normalizer currently fails if there is no CVSS3 score. We should fallback to CVSS2 and if that also doesn't exist the severity level should be unknown.

noamblitz:
The only consideration is that then well fallback to cvss2 without notifying the user.

PR #1271, opened 2023-06-21
fallback to cvss2

noamblitz:
### Changes
Fallback to cvss2 when cvss3 not availbale

### Issue link
https://github.com/minvws/nl-kat-coordination/issues/1231

Closes https://github.com/minvws/nl-kat-coordination/issues/1231


dekkers:
I think there are also CVE's that don't have a CVSS score at all, so we probably also need to handle that both the cvss3 and cvss2 score doesn't exist?

noamblitz:
```python
    if data["impact"] == {}:
        risk_severity = RiskLevelSeverity.UNKNOWN
```

dekkers (approved):

Issue #1276, opened 2023-06-22
Unify (online) environment variable documentation


Darwinkel:
Parent: https://github.com/minvws/nl-kat-coordination/issues/1266

Community members have mentioned that expanding KAT's documentation would be nice. This looks like a step in the right direction.

I'm somewhat fond of what PHP does with `php.ini`. It's a self-explanatory config file thanks to the many comments. I think `.env-dist` could employ a similar strategy, which would save us the trouble of having to refer to other resources when discussing/describing config options.

This issue entails:
- [ ] Extending `.env-dist` with some additional comments and context
- [ ] Including `.env-dist` in the online documentation
- [ ] Making sure the online docs contain a single source of truth with regards to describing variables

ammar92:
I like the idea. I think we should also try to avoid using empty environment settings if they're somehow used for automatically generating a `.env` file (e.g. using `make env`).

Darwinkel:
After discussing, we decided to work on this ticket first: https://github.com/minvws/nl-kat-coordination/issues/1299

PR #1286, opened 2023-06-27
[Frozen] Unify (online) environment variable documentation

Darwinkel:
### Changes
Cleans up, extends, unifies, and documents the `.env` settings used throughout KAT. 

### Issue link
Closes [#1276](https://github.com/minvws/nl-kat-coordination/issues/1276)

### Proof
See online docs and `.env-dist`.

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


Darwinkel:
I think good next steps would be:

- to change some of the variable names (e.g. making the scheduler use QUEUE_URI)
- changing some of the defaults (e.g. octopoes' XTDB_TYPE from crux to xtdb_multinode)
- implementing stricter field validation (e.g. changing the pydantic fields from string to URI in some instances)

dekkers (changes requested):
I don't agree with the direction of this PR:
- .env-dist is meant for development and some values there are different than what people should use in production. Using that as base for production documentation seems wrong to me.
- I think just including .env-dist in the documentation create something that is very hard to read and understand. I think the tables that we had in the container documentation are  easier to understand.
- In production env variables (both configuration and secrets) need to be configured per service. This both the case for containers and Debian packages. I think it is better if we document per service which environments. With the changes users have to gather the envs that are needed for each service from different places which is prone to problems.
- Whether an env variable is required or not is completely removed from the documentation.
- For development I think we should actually move all the default values from .env-dist to .env-defaults. Then we can include both .env-defaults and .env with .env overriding .env-defaults. If we need to update/change some defaults we can just update .env-defaults and we don't break every development setup like we regurarly do now.

underdarknl:
Some remarks on @dekkers 's reply:

> I think just including .env-dist in the documentation create something that is very hard to read and understand. I think the tables that we had in the container documentation are easier to understand.
The specific env-vars per plugin should be listed in their manifest, and then on the boefje/normalizer detail pages in the kat-alogus. Tracking those in the docs is asking for trouble.
> Whether an env variable is required or not is completely removed from the documentation.
If this is required by a plugin, it should be listed on the plugin detail page / checked for its value when enabling this plugin.

Darwinkel:
@dekkers 

>.env-dist is meant for development and some values there are different than what people should use in production. Using that as base for production documentation seems wrong to me.

Might I then suggest that we update all `settings.py` and `config.py` files with sane production defaults, and configure `.env-dist` with settings suitable for development?

>I think just including .env-dist in the documentation create something that is very hard to read and understand. I think the tables that we had in the container documentation are easier to understand.

Unfortunately, time has proven that tables like those are rarely updated. I agree that the `.env-dist` is not pretty, but having a single source of truth, which is automatically updated in the online docs, is preferable. I'd rather have ugly but up-to-date information, than pretty documentation which is outdated.

>In production env variables (both configuration and secrets) need to be configured per service. This both the case for containers and Debian packages. I think it is better if we document per service which environments. With the changes users have to gather the envs that are needed for each service from different places which is prone to problems.

I tried to separate the variables in the file as much as possible per module. Guess we could make that more explicit. 

>Whether an env variable is required or not is completely removed from the documentation.

Technically, it's not a matter of a variable being required or not. (Almost) all of them have defaults in `settings.py` or `config.py`, which are used when a variable is not set in the environment. In our current architecture, setting `.env` variables is akin to overriding the defaults. The question is therefore which variables should deviate from the default in the `.env-dist`. 

>For development I think we should actually move all the default values from .env-dist to .env-defaults. Then we can include both .env-defaults and .env with .env overriding .env-defaults. If we need to update/change some defaults we can just update .env-defaults and we don't break every development setup like we regurarly do now.

See my previous points. Defaults are already configured in `settings.py` and `config.py`. I'd rather make sure that those defaults are production ready, than adding another file which we have to maintain.

ring-ring-ring:
> > I think just including .env-dist in the documentation create something that is very hard to read and understand. I think the tables that we had in the container documentation are easier to understand.
> 
> Unfortunately, time has proven that tables like those are rarely updated. I agree that the `.env-dist` is not pretty, but having a single source of truth, which is automatically updated in the online docs, is preferable. I'd rather have ugly but up-to-date information, than pretty documentation which is outdated.

From a conceptual point of view both of you are right - from a practical point of view having something that is automagically up to date is great news.

dekkers:
> Some remarks on @dekkers 's reply:
> 
> > I think just including .env-dist in the documentation create something that is very hard to read and understand. I think the tables that we had in the container documentation are easier to understand.
> 
> The specific env-vars per plugin should be listed in their manifest, and then on the boefje/normalizer detail pages in the kat-alogus. Tracking those in the docs is asking for trouble.
> 
> > Whether an env variable is required or not is completely removed from the documentation.
> 
> If this is required by a plugin, it should be listed on the plugin detail page / checked for its value when enabling this plugin.

This PR is not about plugin settings, but OpenKAT configuration such as the location of each service and the location and credentials of the database.

dekkers:
> Might I then suggest that we update all `settings.py` and `config.py` files with sane production defaults, and configure `.env-dist` with settings suitable for development?

Yes, I think it is a good idea to have good defaults and also remove everything from .env that doesn't change the default. We should also look into whether we need settings for some things at all, for example I don't understand why we need the SCHEDULER_BOEFJE_POPULATE setting.

> > I think just including .env-dist in the documentation create something that is very hard to read and understand. I think the tables that we had in the container documentation are easier to understand.
> 
> Unfortunately, time has proven that tables like those are rarely updated. I agree that the `.env-dist` is not pretty, but having a single source of truth, which is automatically updated in the online docs, is preferable. I'd rather have ugly but up-to-date information, than pretty documentation which is outdated.

Then we should make updating documentation an integral part of our workflow instead of making the documentation  worse. There is also more documentation than just environment variables that we should add or update when adding or changing things.

> > Whether an env variable is required or not is completely removed from the documentation.
> 
> Technically, it's not a matter of a variable being required or not. (Almost) all of them have defaults in `settings.py` or `config.py`, which are used when a variable is not set in the environment. In our current architecture, setting `.env` variables is akin to overriding the defaults. The question is therefore which variables should deviate from the default in the `.env-dist`.

The location of services, databases, RabbitMQ and all credentials can't have defaults in the code but are required. We should have clear documentation that all those env variables are required for having a functional OpenKAT installation because else people have to start things and wait for the next error message to figure that out.

> > For development I think we should actually move all the default values from .env-dist to .env-defaults. Then we can include both .env-defaults and .env with .env overriding .env-defaults. If we need to update/change some defaults we can just update .env-defaults and we don't break every development setup like we regurarly do now.
> 
> See my previous points. Defaults are already configured in `settings.py` and `config.py`. I'd rather make sure that those defaults are production ready, than adding another file which we have to maintain.

As I said that can't be done for the location of services, databases, RabbitMQ and all credentials, but we can configure those locations and usernames in a .env-defaults. Then the .env-dist will then only need to contain the secrets.

Darwinkel:
After discussing, we decided to work on this ticket first: https://github.com/minvws/nl-kat-coordination/issues/1299

I'll work a bit more on this PR to make a proof-of-concept on how we can split `.env-dist` into module/container specific `.env` files, which we could hopefully share with the Debian `/etc` config files.

PR #1295, opened 2023-06-28
Add Poetry configs for `bytes`, `boefjes`, `rocky`, and `mula`

Darwinkel:
### Changes
Implements Poetry dependency management for `bytes`, `boefjes`, `rocky`, and `mula`

Also removes an old requirements file in keiko and rocky, makes sure that the Poetry check workflow runs on changes in all relevant files and not just the project root, and slightly rewrites the Boefjes Dockerfile such that it doesn't intermix the hashed Poetry requirements and the unhashed module requirements. 

Unfortunately pip does not yet support hash checking mode when using git dependencies. This is a work in progress and should come later this year. Until then, we remove the hashes from rocky's requirements-dev.txt.

### Issue link
Closes https://github.com/minvws/nl-kat-coordination/issues/735

### Proof
See CI.

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


boefjes/Dockerfile
```diff
@@ -29,6 +29,7 @@ RUN --mount=type=cache,target=/root/.cache \
     pip install --upgrade pip \
     && pip install -r /tmp/boefjes-requirements.txt \
     && rm /tmp/boefjes-requirements.txt \
+    && pip install -r requirements.txt \
```
Donnype, line 32: Could also be an `else` the line below although I don't expect much performance difference

Donnype (approved):
This looks good, was also wondering if we should/could use the poetry version to generate the right version for the APIs (where we do a sed now I believe). Perhaps not though ;)

Darwinkel:
> This looks good, was also wondering if we should/could use the poetry version to generate the right version for the APIs (where we do a sed now I believe). Perhaps not though ;)

I looked into this, but it's non-trivial. https://pypi.org/project/poetry-dynamic-versioning/

I guess we could make a low-priority follow-up ticket to this.

boefjes/Dockerfile
Darwinkel: We could, but it's done this way in all of our Dockerfiles and looks readable this way.

ammar92 (approved):
Looks good. Most stuff already discussed offline

Donnype:
## Checklist for QA:

### What works:
Install still works fine, as expected given the changes.

### What doesn't work:
* _Likely unrelated_: I had some issues with the DNSReport where no Findings are in the list, but Findings are actually defined in the overview, but this could just be me having little experience creating DNS Reports:
image
image

* _Definitely unrelated_: I found a new bug with RabbitMQ in bytes that I will fix separately: we do not catch the issue of channels being closed

### Bug or feature?:
*

underdarknl (approved):

Issue #1166, opened 2023-06-13
Tasks get scheduled for organisations without indemnification


praseodym:
If an indemnification gets removed, tasks can still get scheduled.

jpbruinsslot:
There are no procedures in the scheduler for organization indemnification, only for objects. What should be the procedure for this?

underdarknl:
Organizations that do not have a global indemnification should not allow users to set clearances on ooi's, and as such no jobs could possibly exist.

Its probably cheaper to check for the organization indemnification than asking XTDB for objects that never probably don't have a clearance level. This would also allow the scheduler to stop adding jobs / clear the queue for that organization if an organization looses its indemnification.

jpbruinsslot:
Is the indemnification information of an organization from the katalogus? While monitoring organisations from the scheduler, it can check if the indemnification changed, and clear the queue.

TwistMeister:
After some consultation with @jpbruinsslot we decided it would be best that Rocky sends a request to a Mula endpoint when the last indemnification entry for that organization is removed. The scheduler should then proceed to disable the scheduler for that organization. An endpoint is in place already, but the logic has yet to be programmed.

jpbruinsslot:
Scheduler PR is in review: https://github.com/minvws/nl-kat-coordination/pull/1305

Next step is changes is in rocky @TwistMeister

PR #1305, opened 2023-07-03
Implement enabling/disabling schedulers

jpbruinsslot:
### Changes

Added the option to disable and enable schedulers. With a PATCH requests on the scheduler endpoint a scheduler can be enabled or disabled by setting the value of `enabled` in the request. The scheduler will stop all threads that populate the queue with items, and will set all the tasks that have the status `QUEUED` in the the database to `CANCELLED`.

We don't remove the scheduler (from memory) since this is reserved for when organisation are removed. We want to be able to re-enable the scheduler when this is requested.

### Issue link

https://github.com/minvws/nl-kat-coordination/issues/1166


Donnype (approved):

mula/tests/integration/test_boefje_scheduler.py
```diff
@@ -624,6 +625,57 @@ def test_post_pop(self):
         self.assertEqual(task_db.id, p_item.id)
         self.assertEqual(task_db.status, models.TaskStatus.DISPATCHED)
 
+    def test_disable_scheduler(self):
+        # Act
+        self.scheduler.disable()
```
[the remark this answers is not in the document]
jpbruinsslot: Yeah I initially created it for a BoefjeScheduler, however I wanted to make sure that the procedures could work for any particular scheduler. Then I was left with the decision to remove this, and thought it wouldn't be bad in order to leave it in for indeed more coverage. But it does introduce some duplication.

mula/scheduler/repositories/stores.py
```diff
@@ -18,8 +18,12 @@ def __init__(self) -> None:
     @abc.abstractmethod
     def get_tasks(
         self,
-        scheduler_id: str,
-        filters: Optional[List[models.Filter]],
+        scheduler_id: Optional[str] = None,
```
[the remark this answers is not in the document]
jpbruinsslot: Afaik, it won't break previous usages. This PR introduced starting using this method in the scheduler

mula/scheduler/models/scheduler.py
```diff
@@ -8,5 +8,5 @@ class Scheduler(BaseModel):
     unmarshalling of schedulers to a JSON representation."""
 
     id: Optional[str]
-    populate_queue_enabled: Optional[bool]
+    enabled: Optional[bool]
```
ammar92, line 11: I was wondering why this property is optional. What's the expected behavior if it's unset?

ammar92 (approved):
Looks good to me. Just a single remark/ question.

mula/scheduler/models/scheduler.py
jpbruinsslot: This optional is mainly used for the serialization of the model when patching. Might need to think about what models are used as a database representation and what are used for REST response models, and have a clearer definition and separation of those.

ammar92: If you're using this entire structure/ entity in the `PATCH` request, then you might want to consider having a default value instead.

jpbruinsslot: No not the entire structure. Think the only viable field for patching is `enabled` at the moment

Darwinkel:
## Checklist for QA:

### What works:
* Sending enable True/False PATCH request results in the container reporting a disabled scheduler:
image


(Rocky now gives the following error when trying to run a Boefje from the katalogus:)
image

mula/scheduler/models/scheduler.py
ammar92: Yes, but the current endpoint also allows `{"enabled": null}" which doesn't do anything which is unusual for a `PATCH` request.

Different representations are indeed a good idea. It's also common to have different request and response models. I'll leave it up to you for now.

jpbruinsslot: Ah I see. I'll have to look into it if this relates to the omission of the field (key/value) in a patch request or the omission of the value. But I'll up the suggestion of the different representations in another issue, which can be a nice addition to make the differentiation more clear.

jpbruinsslot: > Yes, but the current endpoint also allows
> 
> ```json
> {"enabled": null}
> ```
> 
> which doesn't do anything which is unusual for a `PATCH` request.
> 
> Different representations are indeed a good idea. It's also common to have different request and response models. I'll leave it up to you for now.

that json request will not be allowed by the current endpoint.

```
{
    "detail": [
        {
            "loc": [
                "body",
                "enabled"
            ],
            "msg": "value could not be parsed to a boolean",
            "type": "type_error.bool"
        }
    ]
}
```

underdarknl (approved):

Issue #1299, opened 2023-06-28
Set sane production defaults in the modules' `settings.py`/`config.py`


Darwinkel:
>Yes, I think it is a good idea to have good defaults and also remove everything from .env that doesn't change the default. We should also look into whether we need settings for some things at all, for example I don't understand why we need the SCHEDULER_BOEFJE_POPULATE setting.

_Originally posted by @dekkers in https://github.com/minvws/nl-kat-coordination/issues/1286#issuecomment-1611307324_

---

- [x] Consult with the team what values would be appropriate as production defaults
- [x] Remove any clutter/unnecessary config options
- [ ] Make sure `.env-dist` (and related files) only deviate from the production defaults where needed and appropriate

PR #1314, opened 2023-07-03
[Discussion] Redundant configuration options and production defaults

Darwinkel:
This PR serves as a starting point to decide:
- which configuration options can be removed;
- what should be the default for configurable options.

Relates to https://github.com/minvws/nl-kat-coordination/issues/1299

Feel free to add commits/code comments/reviews with your thoughts. I'll create a new PR when we have settled on the changes.


rocky/rocky/settings.py
```diff
@@ -283,7 +287,7 @@
 
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
 
-SESSION_EXPIRE_SECONDS = 60 * 60 * 2  # 2 hours
+SESSION_EXPIRE_SECONDS = 60 * 60 * 2  # Why 2 hours?
```
Darwinkel, line 290: If I recall correctly there's been some input from the community about users getting kicked out of the application after some inactivity. Maybe 2 hours is too short?

mula/scheduler/config/settings.py
```diff
@@ -36,10 +48,12 @@ class Settings(BaseSettings):
     span_export_grpc_endpoint: Optional[str] = Field(None, env="SPAN_EXPORT_GRPC_ENDPOINT")
```
Darwinkel, line 48: Also: maybe change `SCHEDULER_RABBITMQ_DSN` to be in line with `QUEUE_URI` which is used in the other modules??

dekkers:
My guess is that the prefetch of 100 should be more than enough, because as far as I understand it that would mean the scheduler would be processing 100 messages at the same time.

rocky/rocky/settings.py
underdarknl: this is something we'd need/want to check with our security testers. @sijmen-ruwhof @ITNerdbox

ITNerdbox: Before establishing an appropriate session timer, it is crucial to ascertain the implementation details. Specifically, we need to determine whether the time-out operates on a fixed duration irrespective of user activity or if it triggers after a period of user inactivity.

Additionally, it is important to consider the specific environment in question, such as distinguishing between production and staging environments. Different time-out settings may be applicable to each environment based on their respective requirements.

Furthermore, other factors can influence the session time-out, including company policies that may align with the principles of confidentiality, integrity, and availability (CIA) or business impact analysis (BIA) ratings. These dependencies should be taken into account to ensure the session time-out aligns with the overall security objectives and business needs.

<hr>

**Scenario**

Considering the scenario as a production environment within an organization where scanning and storing vulnerabilities and known weaknesses of internal systems or applications takes place, I would recommend the session **_idle_** time-out to be set at 15 minutes.

From a malicious perspective, gaining access to a system like this would be a primary objective during the execution phase of an attack.

<hr>

Since there are many dependencies on what the session time-out should be, it may be better to process this information during the installation process.

RobboRob: It would be great if this would be configarable...

mula/scheduler/config/settings.py
ammar92: It would make sense as a more generic name and if we later wish to replace the queue backend (currently `rabbitmq`)

rocky/rocky/settings.py
dekkers: The `SESSION_EXPIRE_SECONDS` defines the fixed duration, not the period after user activity. I think 2 hours has a good tradeoff between security and UX as default. I agree that this should be configurable using an env variable.

boefjes/boefjes/config.py
```diff
@@ -11,21 +11,21 @@ class RuntimeConfiguration(Enum):
 
 class Settings(BaseSettings):
     base_dir: Path = Path(__file__).parent.resolve()
-    log_cfg: Path = Path(__file__).parent / "logging.json"
+    log_cfg: Path = Path(__file__).parent / "logging.json"  # Follow-up ticket to make logging the same for all modules?
```
dekkers, line 14: Yes and I think we should also be able to set the log level and maybe other log parameters using environment variables.

boefjes/boefjes/config.py
```diff
@@ -11,21 +11,21 @@ class RuntimeConfiguration(Enum):
...
 
     # Worker configuration
-    pool_size: int = 2
-    poll_interval: float = 1.0
+    pool_size: int = 2  # Which value should be the default and why?
+    poll_interval: float = 1.0  # Which value should be the default and why?
```
dekkers, line 18: I think pool size of 2 is good for small installations.

I think we should be able to increase the poll interval to 10 without any problems and this lowers the number of requests to the scheduler.

boefjes/boefjes/config.py
```diff
@@ -11,21 +11,21 @@ class RuntimeConfiguration(Enum):
...
 
     # Queue configuration
-    queue_name_boefjes: str = "boefjes"
-    queue_name_normalizers: str = "normalizers"
+    queue_name_boefjes: str = "boefjes"  # Why would someone use another queue name?
+    queue_name_normalizers: str = "normalizers"  # Why would someone use another queue name?
```
dekkers, line 22: I can't think of a reason why we can't just hardcode this.

bytes/bytes/config.py
```diff
@@ -31,12 +31,14 @@ class Settings(BaseSettings):
     rfc3161_provider: Optional[str]
     rfc3161_cert_file: Optional[Path]
 
-    encryption_middleware: EncryptionMiddleware = EncryptionMiddleware.NACL_SEALBOX
-    kat_private_key_b64: str = ""
-    vws_public_key_b64: str = ""
+    encryption_middleware: EncryptionMiddleware = (
+        EncryptionMiddleware.NACL_SEALBOX
+    )  # Should we default to no encryption? Does NACL_SEALBOX require keys?
```
dekkers, line 36: Doing encryption requires keys, so not doing encryption as the default makes more sense because that works out of the box.

octopoes/octopoes/config/settings.py
```diff
@@ -20,13 +20,15 @@ class Settings(BaseSettings):
...
+    log_cfg: str = str(
+        Path(__file__).parent.parent.parent / "logging.yml"
+    )  # Follow-up ticket to make logging the same for all modules?
+    queue_name_octopoes: str = "octopoes"  # Why would someone use another queue name?
 
     # External services settings
     queue_uri: str = "amqp://guest:guest@rabbitmq:5672/%2fkat"
     xtdb_uri: str = "http://crux:3000"
-    xtdb_type: XTDBType = XTDBType.CRUX
+    xtdb_type: XTDBType = XTDBType.CRUX  # We only support xtdb-multinode right?
```
dekkers, line 31: Yes I think we can remove support for standard XTDB now.

boefjes/boefjes/config.py
```diff
@@ -11,21 +11,21 @@ class RuntimeConfiguration(Enum):
...
     queue_uri: str = "amqp://"
 
     # Runtime configuration
     runtime: RuntimeConfiguration = RuntimeConfiguration.LOCAL
 
-    enable_db: bool = True
+    enable_db: bool = True  # Why would someone disable the database?
```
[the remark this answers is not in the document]
Donnype: This is indeed an artifact that can be removed, dating back from when we had to phase this in carefully.

bytes/bytes/config.py
```diff
@@ -31,12 +31,12 @@ class Settings(BaseSettings):
     rfc3161_provider: Optional[str]
     rfc3161_cert_file: Optional[Path]
 
-    encryption_middleware: EncryptionMiddleware = EncryptionMiddleware.NACL_SEALBOX
-    kat_private_key_b64: str = ""
-    vws_public_key_b64: str = ""
+    encryption_middleware: EncryptionMiddleware = EncryptionMiddleware.IDENTITY
+    kat_private_key_b64: str = ""  # Should this be renamed?
+    vws_public_key_b64: str = ""  # Should this be renamed?
```
ammar92, line 36: Yep. Would `private_key` and `public_key` suffice?

mula/scheduler/config/settings.py
ammar92, line 1: I think we should use the `BaseSettings.Config.env_prefix` property instead of repeating `env="SCHEDULER_...` arguments.

octopoes/octopoes/config/settings.py
```diff
@@ -11,22 +11,24 @@
 
 
 class XTDBType(Enum):
-    CRUX = "crux"
-    XTDB = "xtdb"
     XTDB_MULTINODE = "xtdb-multinode"
```
ammar92, line 14: Todo: Refactor to a constant. A single-element enumeration is redundant

PR #1296, opened 2023-06-28
Add masscan boefje for scanning large IP ranges

zcrt:
### Changes
This pull request proposes a boefje for `masscan` to scan large IP-ranges more efficiently.

### Issue link
Related: https://github.com/minvws/nl-kat-coordination/issues/1281
(Not closes a Nmap may also be improved)

### Proof
image

---

### Code Checklist
- [x] All the commits in this PR are properly PGP-signed and verified;
- [x] This PR only contains functionality relevant to the issue; tickets have been created for newly discovered issues.
- [ ] I have written unit tests for the changes or fixes I made.
- [ ] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [x] I have performed a self-review of my code and refactored it to the best of my abilities.

### Communication
- [x] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [x] I have made corresponding changes to the documentation, if necessary.

---
## Checklist for code reviewers:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_code.md) into your comment.

---
## Checklist for QA:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_qa.md) into your comment.

PR #1340, opened 2023-07-05
Add masscan boefje

noamblitz:
### Changes
Based on:

https://github.com/minvws/nl-kat-coordination/pull/1296

But uses a container build in a github workflow instead of building the container on boefje run.

---

---
## Checklist for QA:


Darwinkel:
## Checklist for QA:

### What works:
* The new `masscan` boefje works appropriately on a  /31 netblock
image

boefjes/boefjes/plugins/kat_masscan/main.py
```diff
@@ -0,0 +1,110 @@
...
+import docker
+
+from boefjes.job_models import BoefjeMeta
+
+IMAGE = "ghcr.io/minvws/nl-kat-masscan-build-image:pr-1340"
+
+
+###############################################################################
+# Identical to Webpage Capture boefje.
+###############################################################################
```
zcrt, line 16: Perhaps a separate issue, but it would be nice if some kind of boefje utils would be supported so that code like this to fetch a file from a Docker container can be easily re-used in different boefjes

boefjes/boefjes/plugins/kat_masscan/boefje.json
```diff
@@ -0,0 +1,19 @@
+{
+    "id": "masscan",
+    "name": "masscan",
+    "description": "Quickly scan large amounts of IPs.",
+    "consumes": [
+        "IPV6NetBlock",
```
underdarknl, line 6: Is massscan fast enough to not die on ipv6 blocks? I'd reccon these would still take far too long

boefjes/boefjes/plugins/kat_masscan/main.py
noamblitz: Separate issue but indeed a very good idea. I'm actually looking at using this functionality in other tools that I want to integrate too.

boefjes/boefjes/plugins/kat_masscan/boefje.json
noamblitz: @zcrt Do you have any insights on this?

zcrt: To be honest, I have not tested the IPv6 functionality, it seems there is an open issue about on Masscan https://github.com/robertdavidgraham/masscan/issues/7

noamblitz: I also don't know if ipv6 works out of the box in docker

zcrt: A quick test indeed seems to result in some docker issues, maybe we should create a separate issue for IPv6 range scanning support?

noamblitz: Sounds good, in the meantime i have removed ipv6netblocks from consumes

boefjes/boefjes/plugins/kat_masscan/main.py
```diff
@@ -0,0 +1,110 @@
...
+    tarobject = f.next()
+    if tarobject.name == os.path.basename(path):
+        return f.extractfile(tarobject).read()
+
+
+###############################################################################
+
+
+def run_masscan(target_ip, tmp_path: str = "/tmp/output.json") -> bytes:
+    """Run Playwright in Docker."""
```
dekkers, line 77: Docstring doesn't match code.

boefjes/boefjes/plugins/kat_masscan/main.py
```diff
@@ -0,0 +1,110 @@
...
+    client = docker.from_env()
+
+    # Scan according to arguments.
+    port_range = os.getenv("PORTS", "53,80-82,443")
```
dekkers, line 81: I don't understand this default, port 81 and 82 are not well-known ports.

If we want to use the result to decide if we want the IP address added to OpenKAT and do further scanning, I think 25,53,80,443 might be a better default if we don't want to have a lot of ports as default.

boefjes/boefjes/plugins/kat_masscan/main.py
```diff
@@ -0,0 +1,110 @@
...
+    f = tarfile.open(mode="r|", fileobj=TarStream(stream).reader())
+    tarobject = f.next()
+    if tarobject.name == os.path.basename(path):
+        return f.extractfile(tarobject).read()
+
+
+###############################################################################
+
+
+def run_masscan(target_ip, tmp_path: str = "/tmp/output.json") -> bytes:
```
dekkers, line 76: run_masscan is never called with a `tmp_path` argument so I think we can better make this a module level variable like IMAGE.

boefjes/boefjes/plugins/kat_masscan/boefje.json
dekkers: IPv6 is never going to work, an /48 already has 2^80 IP addresses.

zcrt: https://github.com/minvws/nl-kat-coordination/issues/1342

boefjes/boefjes/plugins/kat_masscan/normalize.py
```diff
@@ -0,0 +1,61 @@
...
+    except json.decoder.JSONDecodeError:
+        # Masscan tends to forget to close the json with "]" if the wait window passed.
+        raw = json.loads(raw.decode() + "]") if raw else []
+
+    # Relevant network object is received from the normalizer_meta.
+    network = Network(name=normalizer_meta.raw_data.boefje_meta.arguments["input"]["network"]["name"])
+    yield network
+
+    netblock_ref = None
+    if "NetBlock" in normalizer_meta.raw_data.boefje_meta.arguments["input"]["object_type"]:
```
dekkers, line 56: The boefje will always have a netblock as input so this check is unnecessary.

.github/masscan_dockerfile/Dockerfile
dekkers, line 1: The .github location is for github configuration, we shouldn't put our own code there. I think it would better to put it under boefjes/images or something like that.

.github/workflows/masscan_container_image.yml
```diff
@@ -0,0 +1,59 @@
...
+on:
+  workflow_dispatch: {}
+  push:
+    branches: [ main ]
+    paths:
+      - '.github/masscan_dockerfile/Dockerfile'
+  pull_request:
+    branches: [ main ]
+    paths:
+      - '.github/masscan_dockerfile/Dockerfile'
```
dekkers, line 12: Also add `.github/workflows/masscan_container_image.yml` so the build is also triggerd when we update the workflow.

boefjes/boefjes/plugins/kat_masscan/main.py
zcrt: https://github.com/minvws/nl-kat-coordination/issues/1343

boefjes/boefjes/plugins/kat_masscan/boefje.json
```diff
@@ -0,0 +1,18 @@
...
+    "produces": [
+        "IPAddressV4",
+        "IPAddressV6",
+        "IPPort"
+    ],
+    "environment_keys": [
+        "PORTS",
+        "MAX_RATE"
+    ],
+    "scan_level": 3
```
dekkers, line 17: This can be scan level 2. Nmap IP range is also level 2.

dekkers (changes requested):

noamblitz:
Took the liberty to fix feedback

boefjes/boefjes/plugins/kat_masscan/main.py
zcrt: Agreed on 81 and 82, they were added to illustrate the functionality of `-`.

boefjes/boefjes/plugins/kat_masscan/normalize.py
noamblitz: Ah nice! Well add those checks back then when adding options for ips

boefjes/boefjes/plugins/kat_masscan/boefje.json
zcrt: I would not identify this tool as ‘touch at normal user level’ (https://docs.openkat.nl/introduction/howdoesitwork.html#indemnification). Maybe Nmap IP range needs a re-evaluation too. Potentially, the level should be made dependent on the configuration of the tool (I can imagine a difference between the top 10 ports and all 65535 ports).

boefjes/boefjes/plugins/kat_masscan/main.py
```diff
@@ -0,0 +1,111 @@
...
+import logging
+import os
+import tarfile
+from typing import ByteString, Generator, List, Tuple, Union
+
+import docker
+
+from boefjes.job_models import BoefjeMeta
+
+IMAGE = "ghcr.io/minvws/nl-kat-masscan-build-image:pr-1340"
```
praseodym, line 11: Shouldn't this refer to `main`?

noamblitz:
Yes, right before merge for QA

noamblitz:
> Yes, right before merge for QA

Or after merging to main

boefjes/boefjes/plugins/kat_masscan/boefje.json
```diff
@@ -0,0 +1,18 @@
+{
+    "id": "masscan",
+    "name": "masscan",
+    "description": "Quickly scan large amounts of IPs.",
+    "consumes": [
+        "IPV4NetBlock"
+    ],
+    "produces": [
+        "IPAddressV4",
+        "IPAddressV6",
```
dekkers, line 10: It can't produce an `IPAddressV6` when consuming an `IPV4NetBlock`. The code to return an IPv6 address can also be removed.

boefjes/boefjes/plugins/kat_masscan/schema.json
```diff
@@ -0,0 +1,19 @@
+{
+    "title": "Arguments",
+    "type": "object",
+    "properties": {
+        "PORTS": {
+            "title": "PORTS",
+            "type": "string",
+            "description": "Masscan ports argument, can use commas and hyphens, e.g. '80-82,433' (default)."
```
dekkers, line 8: This should be updated to `53,80,443`.

dekkers (changes requested):

Darwinkel (dismissed):
No regressions; works as advertised.

noamblitz:
Changes image to latest, ready for merge.

underdarknl (approved):

Issue #906, opened 2023-05-03
Filters on the Findings page do not seem to make sense


zcrt:
**Describe the bug**
Filter on the Findings page (clearance level/type) make no sense as findings have no associated clearance level/type.  The button `Set filters` is also shown twice.

**To Reproduce**
1. Check the Findings page filter

**Expected behavior**
Usable filters, maybe a filter on the finding types (as is the case on the objects page). Or it may be useful to add the clearance level of the object(s) from which the finding was created to a finding. In addition, I would not expect two `Set filters` buttons.

**Screenshots**
![image](https://user-images.githubusercontent.com/115991818/235883405-03696ada-5f0d-441e-b018-1c3fdf4f81bb.png)

**OpenKAT version**
1.8rc2

underdarknl:
n.b. Maybe in the future we do allow clearance levels on findings. For example. knowing that a certain CVE exists because of out of date software somewhere could be input for an exploiting level 4 boefje to see if it can be exploited / validated.

Issue #1292, opened 2023-06-27
[EPIC] Advanced Octopoes queries


Donnype:

originalsouth:
Blocked. Pending #2918.

brennodewinter:
Backlog consolidation: #1357 is being consolidated into this issue. Its only content is “See #1292”, and its requested overview/refinement is already an explicit TODO in this issue.

Issue #890, opened 2023-05-02
Mutefinding Filter/Select Improvements


zcrt:
**Is your feature request related to a problem? Please describe.**
It would be nice to select multiple findings at once on the findings page and (un)mute them, just like clearance level on the object page. In addition, a filter should be added on the findings page where you can select to show muted findings.

**Describe the solution you'd like**
- Multiselect to mute/unmute findings on the findings page, with a text box as reason for all mute actions.
- A filter to show/hide muted findings on the findings page

zcrt:
Additionally, quickly muting all DKIM/... findings for a selection of objects would be nice. This could be implemented by adding a filter on finding type on the Findings page (https://github.com/minvws/nl-kat-coordination/issues/906), when the selection option in this issue is implemented.

underdarknl:
Its probably easier to implement the Mute finding button as a mass action on the already present object list page, as this already allows for filtering, and for mass-actions.

underdarknl:
When muting multiple items the following scenarios come to mind:

* Not all selected objects are findings. (We could ignore those / present a info box to the user listing them)
* The user should be presented with (at least) one input field for a reason as to why the selected Findings are to be muted.
* Some findings might already be muted, which reason do we keep, the original, or the new? I'd opt for using the new reason.

NoemiOverdijk:
I made a separate ticket to update the findings page design that includes the issues mentioned here, and unassigned myself from this issue.

#1372

PR #1341, opened 2023-07-05
Finding List Filter extended

Rieven:
### Changes
Remove ooi based filters and create new filters for findings:

- Filter findings based on valid time (this was already there)
- Filter findings by severity 
- Filter findings to exclude muted findings
- Filter to show only muted findings
- Filtering based on FindingTypes
- Filtering based on risk score (min, max  and risk score range filters)

### How does the filtering work?
At first visiting the findings page it shows a list of all findings. At this stage all filters are blank, meaning that all is shown till the user decides to filter. So filtering means show me what I want (what filters is selected). 

### DEPENDS ON
https://github.com/minvws/nl-kat-coordination/issues/1292

### Issue link
https://github.com/minvws/nl-kat-coordination/issues/906
https://github.com/minvws/nl-kat-coordination/issues/890

### Proof
image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


rocky/rocky/templates/findings/findings_filter.html
```diff
@@ -0,0 +1,25 @@
+{% extends "partials/list_filters.html" %}
+
+{% load static %}
+{% load i18n %}
+{% load i18n %}
```
underdarknl, line 5: double

Rieven: oei

rocky/rocky/views/finding_list.py
```diff
@@ -46,13 +53,53 @@ def generate_findings_metadata(
...
+        if self.muted_findings == "show":
+            self.show_muted = True
+        if self.muted_findings == "exclude":
+            self.exclude_muted = True
+        return FindingList(
+            self.octopoes_api_connector,
+            self.valid_time,
+            self.severities,
+            exclude_muted=self.exclude_muted,
+            show_muted=self.show_muted,
```
underdarknl, line 78: `show_muted=(self.muted_findings in ("show", "exclude"))`

You can then remove the 4 lines above the return.

rocky/rocky/views/finding_list.py
```diff
@@ -46,13 +53,53 @@ def generate_findings_metadata(
...
-        return FindingList(self.octopoes_api_connector, self.get_observed_at(), severities)
+        if self.muted_findings == "show":
+            self.show_muted = True
+        if self.muted_findings == "exclude":
+            self.exclude_muted = True
+        return FindingList(
+            self.octopoes_api_connector,
+            self.valid_time,
+            self.severities,
+            exclude_muted=self.exclude_muted,
```
underdarknl, line 77: exclude_muted=(self.muted_findings == "exclude"),

rocky/rocky/views/finding_list.py
```diff
@@ -46,13 +53,53 @@ def generate_findings_metadata(
...
+        self.show_muted = False
+        self.exclude_muted = False
 
     def get_queryset(self) -> FindingList:
-        severities = self.get_severities()
-        return FindingList(self.octopoes_api_connector, self.get_observed_at(), severities)
+        if self.muted_findings == "show":
+            self.show_muted = True
+        if self.muted_findings == "exclude":
+            self.exclude_muted = True
```
underdarknl, line 72: Can be removed if evaluated directly in the return statement

rocky/rocky/views/finding_list.py
```diff
@@ -46,13 +53,53 @@ def generate_findings_metadata(
...
+    connector_form_class = ObservedAtForm
+
+    def setup(self, request, *args, **kwargs):
+        super().setup(request, *args, **kwargs)
+        self.severities = self.get_severities()
+        self.valid_time = self.get_observed_at()
+        self.muted_findings = request.GET.get("muted_findings", "all")
+        self.finding_types = request.GET.getlist("finding_types", [])
+        self.show_muted = False
+        self.exclude_muted = False
```
underdarknl, line 66: Can be removed if evaluated directly in the return statement

rocky/rocky/views/mixins.py
```diff
@@ -224,8 +243,22 @@ def __getitem__(self, key: Union[int, slice]) -> List[HydratedFinding]:
...
+
+            ooi_references = set()
+            finding_type_references = set()
+            filtered_findings = []
+            for finding in findings:
+                ooi_references.add(finding.ooi)
+                finding_type_references.add(finding.finding_type)
+                for finding_type in self.finding_types:
+                    if finding_type in str(finding.finding_type):
+                        filtered_findings.append(finding)
```
underdarknl, line 257: This can be a single list comprehension

rocky/rocky/views/mixins.py
```diff
@@ -224,8 +243,22 @@ def __getitem__(self, key: Union[int, slice]) -> List[HydratedFinding]:
                 offset=offset,
                 limit=limit,
             ).items
-            ooi_references = {finding.ooi for finding in findings}
-            finding_type_references = {finding.finding_type for finding in findings}
+            if self.show_muted:
+                findings = self.get_muted_findings(findings, offset, limit)
```
underdarknl, line 247: Dont you allready do a call for findings in general on line 239? This seems to just do another call to octopoes for the same?

rocky/rocky/views/mixins.py
underdarknl: This would also need some way of keeping track of the limit/offset numbers, as those pagination calls wont match the filtered result set anymore.

Rieven: Yeah I think I need to pour it again in a Paginated list

Rieven: I have tested it and the counting and limiting happens before already. I keep re-using and mutating what octopoes gave back as `listed_findings()`

rocky/rocky/views/mixins.py
underdarknl: Yes, but that depends on the flag exclude_muted=self.exclude_muted, in that call to self.octopoes_connector.list_findings, so lets try and fix that instead of adding another query.

rocky/rocky/views/finding_list.py
```diff
@@ -46,13 +53,47 @@ def generate_findings_metadata(
...
+        context["observed_at_form"] = self.get_connector_form()
+        context["severity_filter_form"] = FindingSeverityMultiSelectForm(
+            {"severity": self.request.GET.getlist("severity")}
+        )
+        context["risk_score_filter"] = FindingRiskScoreNumberForm(
+            {
+                "risk_score_greater_than": self.request.GET.get("risk_score_greater_than"),
+                "risk_score_smaller_than": self.request.GET.get("risk_score_smaller_than"),
+            }
+        )
```
ammar92, line 87:
```suggestion
        context["risk_score_filter"] = FindingRiskScoreNumberForm(self.request.GET)
```

rocky/tools/forms/findings.py
```diff
@@ -0,0 +1,77 @@
...
+class FindingSeverityMultiSelectForm(forms.Form):
+    severity = forms.MultipleChoiceField(
+        label=_("Filter by severity"),
+        required=False,
+        choices=FINDINGS_SEVERITIES_CHOICES,
+        widget=forms.CheckboxSelectMultiple,
+    )
+
+
+class FindingRiskScoreNumberForm(forms.Form):
```
ammar92, line 48: If the fields aren't required but do have boundaries, then they should have a default value right?

rocky/rocky/views/mixins.py
Rieven: Yes, the normal query is with `exclude_muted=False`
but I need also `exclude_muted=True`

To show only muted `Finding(exclude_muted=False)` - `Finding(exclude_muted=True)`

Rieven: I thought to filter the first query but it has no information about which finding is muted:
image

zcrt:
Not sure about the  radio button approach for the mute findings; I would like the table to be able to show:
- only muted findings
- only non-muted findings
- both muted and non-muted findings

Which seems not to be possible with a radio button. Two checkboxes (show muted, show non-muted) should probably work (except that disabling both should result in an empty table but that is okay for me).

zcrt:
Unfortunately, this solution would still not close https://github.com/minvws/nl-kat-coordination/issues/906 as it only introduces general type filters (`KATfinding`, `Snykfinding`,...). The intention of #906 is to (de)select  `KAT-NO-IPv6`, `KAT-NO-DNSSEC`, ...

Rieven:
> 

Filtering for both muted and non-muted = all findings

Rieven:
> Not sure about the radio button approach for the mute findings; I would like the table to be able to show:
> 
> * only muted findings
> * only non-muted findings
> * both muted and non-muted findings
> 
> Which seems not to be possible with a radio button. Two checkboxes (show muted, show non-muted) should probably work (except that disabling both should result in an empty table but that is okay for me).

Filtering for both muted and non-muted = all findings

rocky/tools/forms/findings.py
Rieven: So to clarify this more:

No filtering for risk scores (no values) = show all findings
image

one value inserted, in this case min risk score, filters everything from 3.5 up till max 10
image

max value inserted, filter till the max value, from 0 up
image

Filter all in between values:
image

underdarknl:
> Unfortunately, this solution would still not close #906 as it only introduces general type filters (`KATfinding`, `Snykfinding`,...). The intention of #906 is to (de)select `KAT-NO-IPv6`, `KAT-NO-DNSSEC`, ...

there's a new ticket for that:
https://github.com/minvws/nl-kat-coordination/issues/1349

zcrt:
> > Not sure about the radio button approach for the mute findings; I would like the table to be able to show:
> > 
> > * only muted findings
> > * only non-muted findings
> > * both muted and non-muted findings
> > 
> > Which seems not to be possible with a radio button. Two checkboxes (show muted, show non-muted) should probably work (except that disabling both should result in an empty table but that is okay for me).
> 
> Filtering for both muted and non-muted = all findings This is why those are radio checkboxes Having all findings back can be done by "Clear filters"

I don't think it is very user friendly to reset all filters if the user only wants to reset the muted filter and keep the others.

> > Unfortunately, this solution would still not close #906 as it only introduces general type filters (`KATfinding`, `Snykfinding`,...). The intention of #906 is to (de)select `KAT-NO-IPv6`, `KAT-NO-DNSSEC`, ...
> 
> there's a new ticket for that: #1349

#1349 is about disabling a finding type for an organisation. #906 is about being able to (dynamically) filter on those types. The difference can for example be seen when generating a report on an object; if the finding type is disabled it should not show up in the report. In contrast, if there is a filter active on the findings page hiding a finding type, that does not imply that that finding type should not be in the report for an object (outside that filter).

Rieven:
> > > Not sure about the radio button approach for the mute findings; I would like the table to be able to show:
> > > 
> > > * only muted findings
> > > * only non-muted findings
> > > * both muted and non-muted findings
> > > 
> > > Which seems not to be possible with a radio button. Two checkboxes (show muted, show non-muted) should probably work (except that disabling both should result in an empty table but that is okay for me).
> > 
> > 
> > Filtering for both muted and non-muted = all findings This is why those are radio checkboxes Having all findings back can be done by "Clear filters"
> 
> I don't think it is very user friendly to reset all filters if the user only wants to reset the muted filter and keep the others.
> 
> > > Unfortunately, this solution would still not close #906 as it only introduces general type filters (`KATfinding`, `Snykfinding`,...). The intention of #906 is to (de)select `KAT-NO-IPv6`, `KAT-NO-DNSSEC`, ...
> > 
> > 
> > there's a new ticket for that: #1349
> 
> #1349 is about disabling a finding type for an organisation. #906 is about being able to (dynamically) filter on those types. The difference can for example be seen when generating a report on an object; if the finding type is disabled it should not show up in the report. In contrast, if there is a filter active on the findings page hiding a finding type, that does not imply that that finding type should not be in the report for an object (outside that filter).

Ok, yes true, I will make them checkboxes instead and change the labels to make it more understandable.

ammar92 (approved):
No more remarks

zcrt:
For findings with a single OOI it would be nice to be able to filter on the clearance level of the OOI. (E.g. show me all findings on objects with at least level 2 clearance). That can be made a separate feature request though ;)

dekkers:
I have implemented the filter by muted findings in #1595. In my opinion the other filters still need more refinement. I think it make more sense to base further work on the new designs in #1554, so I propose that we close this PR and move discussion about the findings page to #1554.

PR #1348, opened 2023-07-06
Modify workflow to run `make poetry-dependencies` on Dependabot PR's and commit it

Darwinkel:
Small workflow file that should make sure that Dependabot PR's also export relevant `requirements.txt` files.


dekkers (approved):

ammar92 (approved):
Looks good. You could also set an env var on job level (e.g. `DEPENDABOT: ${{ github.actor == 'dependabot[bot]' }}` and then use this condition in the ifs: `if: env.DEPENDABOT`

Issue #1281, opened 2023-06-26
Scanning large IP-ranges


zcrt:
**Describe the bug**
Currently the solution for scanning IP ranges is done by `Nmap`. The current implementation of this tool in KAT does not scale very well for larger IP ranges.

**To Reproduce**
1. Add a fairly large IP-range (e.g. IPv4 /16)
2. Scan with KAT (e.g. TCP top 250)
3. Wait for days

**Expected behavior**
A first indication on which IPs have a reply in about a day for larger ranges. It may be useful to add a tool such as [Masscan](https://github.com/robertdavidgraham/masscan) or [Zmap](https://github.com/zmap/zmap).

**OpenKAT version**
1.9

underdarknl:
closed by https://github.com/minvws/nl-kat-coordination/pull/1340/files

Issue #447, opened 2023-03-07
Save and fetch settings at once now that Katalogus DB saves settings as JSON


Donnype:

underdarknl:
The settings form should be rendered in the same way as the questionOOI's are rendered. Using the jsonschema 2 form javascript code.

PR #1378, opened 2023-07-12
Add VLSM settings for Nmap IP-range

zcrt:
### Changes
Add variable-length subnet mask (VLSM) settings to the Nmap IP-range boefje. This make sure the user can decide which range size she wants to scan.

Also changes TOP_PORTS settings in json to allow 0 (this already works because checks do not seem to be implemented)
Some formatting has been changed by pre-commit.

### Issue link
https://github.com/minvws/nl-kat-coordination/issues/1281

### Proof
image

---

### Code Checklist
- [x]  This PR only contains functionality relevant to the issue; tickets have been created for newly discovered issues.

### Communication

---

---
## Checklist for QA:


boefjes/boefjes/plugins/kat_nmap_ip_range/schema.json
```diff
@@ -5,16 +5,30 @@
...
+            "minimum": 0,
+            "maximum": 32,
+            "description": "Minimum variable-length subnet mask for IPv4-ranges. Defaults to 0. Use this value to prevent scanning large ranges."
+        },
+        "MIN_VLSM_IPV6": {
+            "title": "MIN_VLSM_IPV6",
+            "type": "integer",
+            "minimum": 0,
+            "maximum": 128,
+            "description": "Minimum variable-length subnet mask for IPv6-ranges. Defaults to 0. Use this value to prevent scanning large ranges."
```
underdarknl, line 31: maybe we should introduce reasonable defaults. a /118 subnet for ipv6 seems manageable, and a /22 for ipv4.
Both are equal to 1024 addresses.

zcrt: It is possible to show a default value for a setting in the interface? Otherwise it would not be clear to an end-user what the default is, resulting in missed scans if we modify it here

underdarknl: https://json-schema.org/understanding-json-schema/reference/generic.html#annotations
You could use the 'default' value keyword. If we don't render that, we should.

zcrt: Added, but it indeed seems like its not rendered

dekkers:
I don't really like the UX that the boefje runs but silently does nothing. The IP range object page will show successful boefje runs but nothing will have been done.

What is the solution for larger IP ranges? If the solution is using the masscan boefje, then it doesn't really make sense to  scan smaller ranges with both masscan and nmap. Do we actually need the nmap ip range boefje given that we have the masscan boefje now and the masscan boefje can handle larger ranges? 

With regards to scanning /118 IPv6 ranges... do we have real life examples where this would be useful? Usually a /64 or /48 is handed out, so I am not really sure if this functionality is going to be useful in practice.

zcrt:
> I don't really like the UX that the boefje runs but silently does nothing. The IP range object page will show successful boefje runs but nothing will have been done.

Indeed https://github.com/minvws/nl-kat-coordination/issues/1440, the same would apply to Shodan skipping private IPs.

> What is the solution for larger IP ranges? If the solution is using the masscan boefje, then it doesn't really make sense to scan smaller ranges with both masscan and nmap. Do we actually need the nmap ip range boefje given that we have the masscan boefje now and the masscan boefje can handle larger ranges?

Have not been able to test Masscan in production as there is no release available yet. I can imagine that Masscan may replace Nmap in the future. However, I am not sure whether there may be any benefits from scanning with different tools and whether they may come up with a different response. In addition, as one has currently only one possible boefje config for all objects, I can imagine a scenario where Masscan is used for quick scanning of large ranges (e.g. small timeout) and Nmap for more intense scanning of smaller ranges.

> With regards to scanning /118 IPv6 ranges... do we have real life examples where this would be useful? Usually a /64 or /48 is handed out, so I am not really sure if this functionality is going to be useful in practice.

Yes (since organisations can decide to use smaller subnets in practice).

boefjes/boefjes/plugins/kat_nmap_ip_range/schema.json
underdarknl: https://github.com/minvws/nl-kat-coordination/pull/1808 implements the default rendering in the forms.

underdarknl: This form however is not a jsonschema form just yet. 
https://github.com/minvws/nl-kat-coordination/issues/447

stephanie0x00:
## Checklist for QA:

### What works:
The VLSM setting seems to work with maximum prefix sizes. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

underdarknl (approved):

PR #1369, opened 2023-07-11
Remove redundant `pyproject` config entries

Darwinkel:
### Changes
See title. These config options are redundant as they are configured in the root `pyproject.toml`. 

### Issue link
Linked directly.

### Proof
See CI.

---

### Code Checklist
- [x] All the commits in this PR are properly PGP-signed and verified;
- [x] This PR only contains functionality relevant to the issue; tickets have been created for newly discovered issues.
- [ ] I have written unit tests for the changes or fixes I made.
- [ ] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [x] I have performed a self-review of my code and refactored it to the best of my abilities.

### Communication
- [ ] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [ ] I have made corresponding changes to the documentation, if necessary.

---
## Checklist for code reviewers:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_code.md) into your comment.

---
## Checklist for QA:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_qa.md) into your comment.

PR #1393, opened 2023-07-13
Add pyproject.toml to black pre-commit config

Rieven:
### Issue
Black did not listen to pyrpoject.tml configs when running `pre-commit` with `git commit`

### Changes
Black did not listen to configuration of pyproject.toml, now added to the .pre-commit-config.yaml

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -48,6 +48,7 @@ repos:
     rev: "23.3.0"
     hooks:
       - id: black
+        files: pyproject.toml$
 
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v1.2.0


praseodym:
The files option is documented as "the pattern of files to run on", so I think this change will cause Black to only run on `pyproject.toml` which is not what we want.

Darwinkel:
I think we should focus on _why_ https://github.com/minvws/nl-kat-coordination/pull/1369 causes these pre-commit issues. I can't think of a reasonable explanation why pre-commit only appears to include/honor the root `pyproject.toml` when running explicitly `--all-files`, and why this was not? the case before that PR.

praseodym:
I did some tests with @dekkers yesterday, and it looks like the behaviour depends on from which directory Black is invoked. When run from the root of the repository, everything works fine. When run from a subdirectory, only the subdirectory's `pyproject.toml` is read. Because that file does not contain the correct Black configuration, Black runs with a different configuration (e.g. line length that is not 120).

ammar92 (changes requested):
This simply won't work as @praseodym already mentioned. See also [build output](https://github.com/minvws/nl-kat-coordination/actions/runs/5544509761/jobs/10122067294?pr=1393), the `black` gets skipped completely.

I suggest running `pre-commit` from the root directory (where the `pyproject.toml` is located) or if you insist on running `black` manually, make sure you point to the correct configuration file (using the `--config` option).

Rieven:
> I did some tests with @dekkers yesterday, and it looks like the behaviour depends on from which directory Black is invoked. When run from the root of the repository, everything works fine. When run from a subdirectory, only the subdirectory's `pyproject.toml` is read. Because that file does not contain the correct Black configuration, Black runs with a different configuration (e.g. line length that is not 120).

Running `pre-commit` at `nl-kat-coordination` level runs correctly, but not with `git commit`

praseodym:
In my quick test, the git hook always ran from the root of the repository which means that `pre-commit` should pick up the correct configuration. Not sure why that doesn't seem to work for you.

PR #1420, opened 2023-07-19
Fix robot test

Rieven:
### Changes
Fix for broken ci robot test

- I have also added `make local-ci-robot` to automate the process of onboarding, can be used for QA as well @Darwinkel


Makefile
```diff
@@ -112,3 +112,9 @@ poetry-dependencies:
 	# NOTE: pip does not yet support hash verification for git dependencies;
 	# rocky's requirements-dev.txt unfortunately has no hashing until then
 	sed -i '/--hash/d; s/ \\$$//' rocky/requirements-dev.txt
+
+
+local-ci-robot:
```
Donnype, line 117: Isn't this supposed to work?
```shell
$ cd rocky
$ make test-rf
```

It is not working for me I see so I suppose it hasn't been updated. I'd say we should remove that target (and `test-prepare`) and perhaps call this new make target in the github workflow?

Makefile
```diff
@@ -112,3 +112,9 @@ poetry-dependencies:
...
+	make reset
+	docker exec -it nl-kat-coordination-rocky-1 /bin/bash -c 'export DJANGO_SUPERUSER_PASSWORD=robotpassword && python manage.py createsuperuser --full_name robot --email robot@localhost --noinput'
```
praseodym, line 119: This should use `docker-compose run --rm rocky` so that it works in every environment and doesn't reuse the existing container.

Rieven: I have removed this part as this was an extra for this PR

Makefile
Rieven: I have removed it till we find a better solution in another PR

ammar92 (approved):

Darwinkel (approved):

dekkers (approved):

Issue #853, opened 2023-05-01
CSV upload for new users


TwistMeister:
We want to be able to upload new users to a KAT organisation in Bulk.

The button "add new member" should become a drop down button, like "add objects" on the objects list page. Then two options can be shown: Add member manually, Upload CSV.

The following screen should explain somethings about the CSV criteria. To be determined, but I guess at least: e-mailadres, full name, account_type. Assigned clearance level could be optional. And for the password there's a decision to be made: do we want to sent invite e-mails or does the CSV have to contain passwords for each user?

For already known e-mail addresses I thinkt it's best to create new organisation members and link them to the already known users with those e-mail addresses, leaving all other info on those user accounts untouched.

Add objects button, as example:
![Screenshot 2023-06-12 at 17 32 02](https://github.com/minvws/nl-kat-coordination/assets/16188579/ad1122d2-d037-4d21-911b-8fae629177c3)

Criteria overview example for importing an OOI CSV
![Screenshot 2023-06-12 at 17 31 18](https://github.com/minvws/nl-kat-coordination/assets/16188579/8a8c8285-baf9-401b-8a65-f22fd5b2cf06)

underdarknl:
Lets use the same flow we would use for adding multiple users to an organisation. So add users, or upload file, and then show the pending users in a list before saving them. Figma has designs for the manual flow for multiple users which can in turn send to the same CSV handler in the backend.

Donnype:
What do we do when we have duplicate rows with different account_types? In my current setup when we have two users with different account_types, they get added to both groups, e.g. here for c@c.nl:


![Image](https://github.com/minvws/nl-kat-coordination/assets/46660228/54459b57-df3f-4529-8f55-4544a53e6a03)

PR #1423, opened 2023-07-19
Upload members with a csv file

Donnype:
### Changes
Add a page where we can upload a csv file with users. Note: by default users get an empty password, meaning they cannot log in. However, we can leverage the password reset flow for this as well because their email adress is known at this stage. We can also consider sending an email to the users already instructing them to reset their password, and generate the link immediately, by creating a password reset form in the backend for all new users.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/853

### Proof

Add member(s) dropdown button added:

image

Upload page with the criteria:

image

New users:

image

Email sent to new users:

```
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit
Subject: Verify OpenKAT account on localhost:8000
From:
To: a@bbbb.dl
Date: Thu, 20 Jul 2023 13:34:32 -0000
Message-ID: <168986007241.76.14464090403674779824@af745d470510>




    Welcome to OpenKAT. You're receiving this email because you have been added to organization "test" at localhost:8000.
    Please go to the following page and choose a new password:

        http://localhost:8000/en/reset/MTY/brn1pk-914a9d550dbb2a5b0269c85f6b667e21/

    Sincerely,
    The OpenKAT team
```

UPDATE:

I also added the option to download a template csv file:
Schermafbeelding 2023-07-25 om 12 45 34

UPDATE:

Warnings:

image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


rocky/rocky/views/organization_member_add.py
```diff
@@ -54,3 +69,89 @@ def form_valid(self, form):
...
+                    member.trusted_clearance_level = trusted_clearance_level
+                    member.acknowledged_clearance_level = acknowledged_clearance_level
+
+                try:
+                    member.groups.add(Group.objects.get(name=account_type))
+                except ObjectDoesNotExist:
+                    logger.exception("Failed adding organization member to group")
+                    continue
+
+            messages.add_message(self.request, messages.SUCCESS, _("Successfully added csv."))
```
ammar92, line 149: The message is a bit misleading since errors are silenced/ not visible to the end user. Perhaps something like "CSV processed" suits better?

rocky/rocky/views/organization_member_add.py
```diff
@@ -54,3 +69,89 @@ def form_valid(self, form):
...
+    permission_required = "tools.add_organizationmember"
+
+    def get_success_url(self):
+        return reverse_lazy("organization_member_list", kwargs={"organization_code": self.organization.code})
+
+    def form_valid(self, form):
+        self.process_csv(form)
+        return super().form_valid(form)
+
+    def process_csv(self, form) -> None:
```
ammar92, line 86: To me most of these log statements here are more error logs than exception logs

ammar92 (commented):
Minimal remarks, looks good otherwise

rocky/rocky/views/organization_member_add.py
Donnype: Yess

rocky/rocky/views/organization_member_add.py
Donnype: I would always want the exception info to be printed in these cases as they would be probably be unexpected database exceptions?

Donnype:
@underdarknl Check. I also added a button to download a template.

rocky/rocky/views/organization_member_add.py
ammar92: > What do you mean when you say too broad? The exception clause or the message?

I'm referring to the exception clause indeed.

> I have looked into catching just Django related database errors here, but there is no real base exception for that to catch, so that would mean having a list of exceptions here. But then again the argument could be made that we do not want to handle unexpected database exceptions here and just crash, since such information would not be propagated to the user uploading the file and does not give information about existence of the users in the file. I think I actually prefer that and we can actually not continue on such errors.
> 
> (Still need to "roll back" the creation of the user though.)

Maybe I wasn't clear, but I meant to use `logger.error` more instead of the `logger.exception` (in this case and context). Catching the exceptions (and e.g. perform roll back actions) is obviously properly implemented.

Donnype: @ammar92 Can you have another look? I cleaned up the error handling

Donnype:
I created [a separate ticket](https://github.com/minvws/nl-kat-coordination/issues/1457) for writing new documentation to do this week.

Donnype:
I added the leveraging of transactions to fix the atomicity of users and organization members in each row.

Rieven (commented):
Generally speaking we need to also check if the organization of the members that are added also belongs to the same organization of the one that is adding it. Admin that only belongs to orgA cannot add a member to orgB.

Darwinkel:
## Checklist for QA:

### What works:
* Downloading the template, filling it in, and uploading it results in users being added in bulk.
* Password reset links are sent successfully and users are able to login

### What doesn't work:
* If a role is empty or invalid, no role is assigned to a user, with no ability to assign one later.
* If an e-mail already exists, the user is ignored
* It's possible to assign/accept a clearance level of >5

Looks like there's room for more validation :)

Donnype:
> If a role is empty or invalid, no role is assigned to a user, with no ability to assign one later.
> It's possible to assign/accept a clearance level of >5
> If an e-mail already exists, the user is ignored

@Darwinkel I am not able to reproduce these. Did you use the latest version of today? If the email adres exists we should not create a new user but we should add them to the organization.

Donnype:
> Generally speaking we need to also check if the organization of the members that are added also belongs to the same organization of the one that is adding it. Admin that only belongs to orgA cannot add a member to orgB.

I think this is handled by these view being an `OrganizationView` and the fact that we always use `self.organization` handles this, since we raise a 404 of no organization member exists and we check if the member has the right to edit organizations through the permissions mixin.

Darwinkel:
> > If a role is empty or invalid, no role is assigned to a user, with no ability to assign one later.
> > It's possible to assign/accept a clearance level of >5
> > If an e-mail already exists, the user is ignored
> 
> @Darwinkel I am not able to reproduce these. Did you use the latest version of today? If the email adres exists we should not create a new user but we should add them to the organization.

I re-checked and can confirm that it works now. It would probably still be nice to have some UI warning/feedback when an illegal role or clearance was encountered, instead of silently failing.

Donnype:
@Darwinkel Done!

Darwinkel:
Confirmed, nice.

ammar92 (approved):

underdarknl (approved):

Issue #1191, opened 2023-06-15
Save boefje Docker container stdout/stderr logs


praseodym:
When investigating #1190 I noticed that we don't seem to save Docker container stdout/stderr logs for boefjes running as a Docker container. I think it would aid debugging a lot if we save these logs in Bytes.

praseodym:
#1430 fixes this for the webpage capture boefje, the same technique can be used for other boefjes.

Darwinkel:
https://github.com/minvws/nl-kat-coordination/issues/1198

PR #1430, opened 2023-07-20
Fix webpage capture boefje

### LINKED ISSUE #1190: Webpage capture boefje is broken
The webpage capture boefje is broken on our testserver (https://kattest01.openkat.dev/en/mispoes/kat-alogus/plugins/boefje/webpage-capture/), we should investigate and fix any bugs we encounter. Might be related to #804.

#1191 would be helpful in debugging this.

praseodym:
### Changes
**Fix webpage capture boefje** 

Playwright container needed to be updated for it to run again, so let's
always use the latest container instead.

Also, improve error logging by returning the Playwright container log as
part of the Boefje exception.

**Harden Playwright container**

### Issue link

Fixes #1190

### Proof

screenshot

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


boefjes/boefjes/plugins/kat_webpage_capture/main.py
```diff
@@ -5,14 +5,27 @@
...
+    def __init__(self, message, container_log=None):
+        self.message = message
+        self.container_log = container_log
+
+    def __str__(self):
+        return str(self.message) + "\n\nContainer log:\n" + self.container_log
+
+    pass
+
+
```
ammar92, line 28:
```suggestion
```

boefjes/boefjes/plugins/kat_webpage_capture/main.py
```diff
@@ -68,7 +81,7 @@ def get_file_from_container(container: docker.models.containers.Container, path:
 
 def build_playwright_command(webpage: str, browser: str, tmp_path: str) -> str:
     """Returns playwright command including webpage, browser and locations for image, har and storage."""
-    return " ".join(
+    x = " ".join(
```
ammar92, line 84:
```suggestion
    command = " ".join(
```

boefjes/boefjes/plugins/kat_webpage_capture/main.py
```diff
@@ -5,14 +5,27 @@
 from typing import ByteString, Generator, List, Tuple, Union
 
 import docker
-from PIL import Image
+from PIL import Image, UnidentifiedImageError
 
 from boefjes.job_models import BoefjeMeta
 
-PLAYWRIGHT_IMAGE = "mcr.microsoft.com/playwright:v1.33.0-jammy"
+PLAYWRIGHT_IMAGE = "mcr.microsoft.com/playwright:latest"
```
ammar92, line 12: Why not just pin the latest version? We had a few discussions and bugs before when using the `latest` tag and therefore decided not to use that tag (as an unwritten rule) and always use a fixed version or hash. Also when using `latest` it doesn't guarantee that this plugin in production will use the latest image, whilst pinning the version will guarantee that a new image is downloaded if the version changes.

```suggestion
PLAYWRIGHT_IMAGE = "mcr.microsoft.com/playwright:v1.34.0-jammy"
```

praseodym: Because the error we got with the old version was that the container needed to be updated. I'm not sure how the containers are built, but if you run an old container version, an error message will appear saying that there is a newer version available that should be used instead. So there's no use in pinning a specific version, because it will break once a new version has been released.

boefjes/boefjes/plugins/kat_webpage_capture/main.py
praseodym: I'll remove this, was a remnant of my debugging session.

boefjes/boefjes/plugins/kat_webpage_capture/main.py
praseodym: I've included logs and more background information here: https://github.com/minvws/nl-kat-coordination/issues/1190#issuecomment-1645573638

ammar92 (dismissed):
Still not in favor of using this `latest` tag, but I don't have a better solution (for now) that doesn't require too much boilerplate code and heuristics to prevent this. Let's just see how this works out

boefjes/boefjes/plugins/kat_webpage_capture/main.py
underdarknl: So this container has two options? Either run 'latests', or don't run at all?
Seeing that we'd need to be able to show which version of each container was used for a given job, and be able to reproduce that, maybe this specific distribution is not a workable solution for us.
It also opens up the doors to supply line attacks, which is another reason to version pin dependencies.

praseodym: We can use the hash of the container to log which version was ran.

I'd say the bigger risk is that the container installs the Playwright NPM module when it starts, the NPM registry is presumably a bigger supply chain risk than Microsoft's container registry. The fact that this NPM module is into preinstalled is also what causes older container images to break in the first place.

Darwinkel:
## Checklist for QA:

### What works:
* The boefje now generally doesn't fail. 

### What doesn't work:
* It's unclear how the resulting screenshot can be viewed. The 3.5MB `raw` file cannot be easily opened in an image viewer. 
* Playwright doesn't like websites with a TLS certificate error:

```
  File "/usr/local/lib/python3.11/site-packages/docker/api/client.py", line 268, in _raise_for_status
    response.raise_for_status()
  File "/usr/local/lib/python3.11/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http+docker://localhost/v1.43/containers/e467e64ed50b545a86717bfecc6920380e05a62de036e6eedf3bd8c86bae42b2/archive?path=%2Ftmp%2Ftmp.png

The above exception was the direct cause of the following exception:

  File "/app/boefjes/boefjes/plugins/kat_webpage_capture/main.py", line 117, in run_playwright
    image = Image.open(io.BytesIO(get_file_from_container(container=container, path=f"{tmp_path}.png")))
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/boefjes/boefjes/plugins/kat_webpage_capture/main.py", line 72, in get_file_from_container
    raise e
  File "/app/boefjes/boefjes/plugins/kat_webpage_capture/main.py", line 67, in get_file_from_container
    stream, _ = container.get_archive(path)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docker/models/containers.py", line 262, in get_archive
    return self.client.api.get_archive(self.id, path,
  File "/usr/local/lib/python3.11/site-packages/docker/utils/decorators.py", line 19, in wrapped
    return f(self, resource_id, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docker/api/container.py", line 761, in get_archive
    self._raise_for_status(res)
  File "/usr/local/lib/python3.11/site-packages/docker/api/client.py", line 270, in _raise_for_status
    raise create_api_error_from_http_exception(e) from e
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/docker/errors.py", line 39, in create_api_error_from_http_exception
    raise cls(e, response=response, explanation=explanation) from e
docker.errors.NotFound: 404 Client Error for http+docker://localhost/v1.43/containers/e467e64ed50b545a86717bfecc6920380e05a62de036e6eedf3bd8c86bae42b2/archive?path=%2Ftmp%2Ftmp.png: Not Found ("Could not find the file /tmp/tmp.png in container e467e64ed50b545a86717bfecc6920380e05a62de036e6eedf3bd8c86bae42b2")

During handling of the above exception, another exception occurred:

  File "/app/boefjes/boefjes/local.py", line 53, in run
    return boefje_resource.module.run(boefje_meta)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/boefjes/boefjes/plugins/kat_webpage_capture/main.py", line 135, in run
    image_bytes, har_zip, storage_json = run_playwright(webpage=webpage, browser=BROWSER)
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/boefjes/boefjes/plugins/kat_webpage_capture/main.py", line 122, in run_playwright
    raise WebpageCaptureException(
boefjes.plugins.kat_webpage_capture.main.WebpageCaptureException: Error while running Playwright container, command was: /usr/bin/npx playwright screenshot -b chromium --full-page --save-har=/tmp/tmp.har.zip --save-storage=/tmp/tmp.json https://mispo.es/ /tmp/tmp.png

Container log:
2023-07-24T07:43:13.787786722Z npm WARN exec The following package was not found and will be installed: playwright@1.36.1
2023-07-24T07:43:15.131327488Z Navigating to https://mispo.es/
2023-07-24T07:43:15.326070014Z [Error: net::ERR_CERT_DATE_INVALID at https://mispo.es/
2023-07-24T07:43:15.326093035Z =========================== logs ===========================
2023-07-24T07:43:15.326096141Z navigating to "https://mispo.es/", waiting until "load"
2023-07-24T07:43:15.326098594Z ============================================================]
2023-07-24T07:43:15.380506060Z npm notice 
2023-07-24T07:43:15.380519372Z npm notice New minor version of npm available! 9.5.1 -> 9.8.1
2023-07-24T07:43:15.380609237Z npm notice Changelog: <https://github.com/npm/cli/releases/tag/v9.8.1>
2023-07-24T07:43:15.380621517Z npm notice Run `npm install -g npm@9.8.1` to update!
2023-07-24T07:43:15.380624833Z npm notice 


The above exception was the direct cause of the following exception:

  File "/app/boefjes/boefjes/job_handler.py", line 156, in handle
    boefje_results = self.job_runner.run(boefje_meta, boefje_meta.environment)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/boefjes/boefjes/local.py", line 55, in run
    raise JobRuntimeError("Boefje failed") from e
boefjes.runtime_interfaces.JobRuntimeError: Boefje failed
```

underdarknl:
https://github.com/microsoft/playwright/issues/24362

praseodym:
> * It's unclear how the resulting screenshot can be viewed. The 3.5MB `raw` file cannot be easily opened in an image viewer.

I've changed it to return PNG files instead. I thought this boefje also had a normaliser that parsed based on the current output format (a Pillow bytes stream) but I think I got confused with the webpage analysis normaliser. Returning a PNG file conforms with the `image/png` MIME type that is currently promised and seems like a better standard anyway.

> * Playwright doesn't like websites with a TLS certificate error

Not sure if that's something we should fix, any opinions?

boefjes/boefjes/plugins/kat_webpage_capture/main.py
```diff
@@ -68,51 +78,63 @@ def get_file_from_container(container: docker.models.containers.Container, path:
...
+        container.wait()
+        image = get_file_from_container(container=container, path=f"{tmp_path}.png")
+        har = get_file_from_container(container=container, path=f"{tmp_path}.har.zip")
+        storage = get_file_from_container(container=container, path=f"{tmp_path}.json")
+        return image, har, storage
+    except docker.errors.NotFound:
+        raise WebpageCaptureException(
+            "Error while running Playwright container, command was: " + " ".join(command),
+            container.logs(stdout=True, stderr=True, timestamps=True).decode(),
+        )
```
zcrt, line 124: Raising this error will result in no files saved when one of the files is actually present. Suppose the screenshot was succesfully taken but for some reason the `har.zip` was not created. Then the screenshot is also not returned anymore. (That is why it was designed to only raise a warning for each missing file)

zcrt:
A workaround for the `latest` issues could be to make it a setting of the boefje, so that the end-user can decide to upgrade the container to another version (if the default version is not updated yet in this repository).

zcrt:
>>        Playwright doesn't like websites with a TLS certificate error

> Not sure if that's something we should fix, any opinions?

We can add `--ignore-https-errors` to the command. It think it would be nice to have the screenshots even though there are TLS errors

```bash
# /usr/bin/npx playwright screenshot -b chromium --full-page --save-har=/tmp/tmp.har.zip --save-storage=/tmp/tmp.json --ignore-https-errors https://mispo.es /tmp/tmp.png
Navigating to https://mispo.es
Capturing screenshot into /tmp/tmp.png
# ls /tmp
tmp.har.zip  tmp.json  tmp.png

```

boefjes/boefjes/plugins/kat_webpage_capture/main.py
zcrt: Showing the container logs is a very nice addition by the way. It would be great if that can also be added to other boefjes using containers for their commands (e.g. Nmap).

praseodym:
> We can add `--ignore-https-errors` to the command. It think it would be nice to have the screenshots even though there are TLS errors

Added!

boefjes/boefjes/plugins/kat_webpage_capture/main.py
praseodym: Without this PR the container crashes when the screenshot is not present, it doesn't just log a warning.

Because we don't have a "partially successful" status code for Boefjes, I'd say the only reasonable way to handle this is to fail when any file is missing.

praseodym: > Showing the container logs is a very nice addition by the way. It would be great if that can also be added to other boefjes using containers for their commands (e.g. Nmap).

#1191 already exists for that.

underdarknl:
> > > ```
> > >    Playwright doesn't like websites with a TLS certificate error
> > > ```
> 
> > Not sure if that's something we should fix, any opinions?
> 
> We can add `--ignore-https-errors` to the command. It think it would be nice to have the screenshots even though there are TLS errors
> 
> ```shell
> # /usr/bin/npx playwright screenshot -b chromium --full-page --save-har=/tmp/tmp.har.zip --save-storage=/tmp/tmp.json --ignore-https-errors https://mispo.es /tmp/tmp.png
> Navigating to https://mispo.es
> Capturing screenshot into /tmp/tmp.png
> # ls /tmp
> tmp.har.zip  tmp.json  tmp.png
> ```

Ignoring https errors seems valid, other tooling will check the ssl/tls anyway.

underdarknl:
> A workaround for the `latest` issues could be to make it a setting of the boefje, so that the end-user can decide to upgrade the container to another version (if the default version is not updated yet in this repository).
> 
> Indeed a normalizer and Octopoes object should still be created for this boefje ([minvws/nl-kat-boefjes#90](https://github.com/minvws/nl-kat-boefjes/pull/90)) .(And after that the option to show screenshots in the interface and reports.) In addition the `har.zip` and `json` can be normalized for more information.

The har file and other artifacts can already be added to this boefje's output, and will be included in the future job pages. They could also function as inputs for other normlizers.

zcrt:
> The har file and other artifacts can already be added to this boefje's output, and will be included in the future job pages. They could also function as inputs for other normlizers.

Yes they are already added to the output

boefjes/boefjes/plugins/kat_webpage_capture/main.py
zcrt: > Because we don't have a "partially successful" status code for Boefjes, I'd say the only reasonable way to handle this is to fail when any file is missing.

I created a separate issue https://github.com/minvws/nl-kat-coordination/issues/1440

ammar92:
> Ah check, I missed that. Maybe we can add a fictional mimetype next to the "application/zip+json" mimetype to designate that this is a har-file? Following the schema used by XML we could do: application/har+json ? Same goes for the local-storage file, This would make it much easier to create a normalizer which targets all these har-files, or storage files.

A HAR file is already a JSON file, so we might as well extract this archive and provide a `application/text+json` type instead. I think it's already possible for normalizers to target this specific file type by expecting a combination of `application/text+json` and the boefje 'mimetype'? I'm not really sure right now, but I'll figure it out

boefjes/boefjes/plugins/kat_webpage_capture/main.py
zcrt: This one got probably lost in the comments, but why not make the version a setting of the boefje? (https://github.com/minvws/nl-kat-coordination/pull/1430#issuecomment-1647543346)

boefjes/boefjes/plugins/kat_webpage_capture/main.py
```diff
@@ -68,51 +78,64 @@ def get_file_from_container(container: docker.models.containers.Container, path:
...
-        ],
+        command=command,
         detach=True,
+        ipc_mode="host",
+        user="pwuser",
+        security_opt=[
+            (
+                'seccomp={"comment": "Allow create user namespaces", "names": ["clone", "setns", "unshare"], '
+                '"action": "SCMP_ACT_ALLOW", "args": [], "includes": {}, "excludes": {}}'
+            )
```
zcrt, line 112: Maybe a comment explaining why these settings are important would be useful

praseodym: There is a comment right above the container creation: https://playwright.dev/docs/docker#crawling-and-scraping

boefjes/boefjes/plugins/kat_webpage_capture/main.py
zcrt: Prevention of running a version of the container that was not tested with the release of OpenKAT

dekkers: With this container image there seem to be two options here as far as I can see:
- Use latest version that might break when a new version is released
- Use a pinned version, either in code or in settings, that is guaranteed to break when a new version is released.

I don't really understand why making sure it is always broken should be preferred above having the risk it might get broken.

Another solution might be to use a different container image / build our own.

underdarknl:
> > Ah check, I missed that. Maybe we can add a fictional mimetype next to the "application/zip+json" mimetype to designate that this is a har-file? Following the schema used by XML we could do: application/har+json ? Same goes for the local-storage file, This would make it much easier to create a normalizer which targets all these har-files, or storage files.
> 
> A HAR file is already a JSON file, so we might as well extract this archive and provide a `application/text+json` type instead. I think it's already possible for normalizers to target this specific file type by expecting a combination of `application/text+json` and the boefje 'mimetype'? I'm not really sure right now, but I'll figure it out

There is no true har file mime-type, and I'd think we are going to see many more of those situations where we want to target for example http-streams or partial files with a normalizer.
Currently, normalizers do not have the option to target specific raw files that have a combination of mime-types present. While an interesting idea, it would negate the flexibility of the normalizer to raw-file coupling if the boefje id is also required. A har-file normalizer should also work on har files imported from other sources (eg, uploads by users, or some system providing har files as its output)

underdarknl:
@zcrt versioning should be handled at another level in the future, a new boefje version should be published and checked on various levels before being used.

ammar92:
> There is no true har file mime-type, and I'd think we are going to see many more of those situations where we want to target for example http-streams or partial files with a normalizer.

Yep, it's just a JSON file with a defined schema, that's why I mentioned the `application/text+json` mime-type.

> While an interesting idea, it would negate the flexibility of the normalizer to raw-file coupling if the boefje id is also required. A har-file normalizer should also work on har files imported from other sources (eg, uploads by users, or some system providing har files as its output). Currently, normalizers do not have the option to target specific raw files that have a combination of mime-types present.

It wouldn't negate the flexibility of the normalizer, since my idea is not to make this boefje-id coupling required but optional. Some time ago I mentioned the idea of using tags next to mime-types for this purpose. It would make a lot of sense if, for example, a boefje plugin that includes e.g. the `HAR` tag would throw an output of mime-type `application/text+json` to be consumed by normalizers that can consume HAR archives. Tags will not only be making this easier (and not misusing mime-types for this), it will also allow categorizing plugins and grouping them.

underdarknl:
that would work better if the tag where applied to the raw-file, and not the whole boefje, but I think that would mean we adhere to strictly to what mime-types are, and if we cannot extend that to also mean mime-type+tag's for our purpose.

ammar92:
> that would work better if the tag where applied to the raw-file, and not the whole boefje, but I think that would mean we adhere to strictly to what mime-types are, and if we cannot extend that to also mean mime-type+tag's for our purpose.

Yep but you can still have tags for plugins and tags for output, they're easy to distinguish if we use an appropriate format

underdarknl (approved):

PR #1476, opened 2023-07-27
Update scheduler documentation

jpbruinsslot:
Updating the scheduler documentation to correctly resemble the current state of the scheduler.


ring-ring-ring (commented):
Added one or two small items, otherwise its just fine. One improvement might be to use relative links to documents.

ammar92 (approved):
No remarks

underdarknl (approved):

Issue #1477, opened 2023-07-27
JS files not loaded in Django admin because of missing CSP headers


Rieven:
**Describe the bug**
JS files are not being executed because they are not seen as same origin files.

**To Reproduce**
Go to the admin panel and use the browser's developers tools and inspect the missing loaded files.

**Expected behavior**
JS should load.

**Screenshots**
<img width="1424" alt="image" src="https://github.com/minvws/nl-kat-coordination/assets/30832687/d8f5e10e-7225-4597-9554-7b3d16f5c121">

PR #1480, opened 2023-07-27
Fix loading javascript in Django admin

Rieven:
### Changes
- Js files were not loaded in Django admin, making for example checkbox bulk selection not working. This has to do with that the JS files were not seen as same origin files. Added django admin templates and introduced new template tag to fix this.
 
 ### Issue link
Closes https://github.com/minvws/nl-kat-coordination/issues/1477


dekkers:
We use nonce because that is the most secure option. The django admin has 7 templates with script tags and if we would want to keep using nonce CSP we have to copy those templates and add the nonce attribute. We would have to update the templates when we upgrade to a new major Django version, but if we stick to the LTS versions the next time we would have to do that is April 2025 so I don't really see that as a big problem.

Rieven:
> We use nonce because that is the most secure option. The django admin has 7 templates with script tags and if we would want to keep using nonce CSP we have to copy those templates and add the nonce attribute. We would have to update the templates when we upgrade to a new major Django version, but if we stick to the LTS versions the next time we would have to do that is April 2025 so I don't really see that as a big problem.

Is it also and idea to add these 7 files to the list instead of "self" or it that as well less secured?

dekkers:
> > We use nonce because that is the most secure option. The django admin has 7 templates with script tags and if we would want to keep using nonce CSP we have to copy those templates and add the nonce attribute. We would have to update the templates when we upgrade to a new major Django version, but if we stick to the LTS versions the next time we would have to do that is April 2025 so I don't really see that as a big problem.
> 
> Is it also and idea to add these 7 files to the list instead of "self" or it that as well less secured?

You can't add django templates to the CSP header because the browser doesn't know about django templates. You can use a decorator in to specify a different CSP header in django-csp, but that also doesn't help us with the admin views.

underdarknl:
Could we upstream this to the django project? It seems like more people would and are going to run into this problem?

dekkers:
> Could we upstream this to the django project? It seems like more people would and are going to run into this problem?

https://code.djangoproject.com/ticket/15727

My guess is that most people don't use the nonce method so don't run into this problem.

ammar92:
If this isn't a huge problem and not getting a lot of attention from Django, perhaps we should close/ archive this PR?

dekkers:
Reopening because we are missing nice to have things in the admin because the javascript doesn't work. The PR already partly fixes the problem, there are few more templates in the django admin that uses javascript however. These are the templates from grepping for script in the django git repository:

- django/contrib/admin/templates/admin/base.html
- django/contrib/admin/templates/admin/change_form.html
- django/contrib/admin/templates/admin/change_list.html
- django/contrib/admin/templates/admin/delete_confirmation.html
- django/contrib/admin/templates/admin/delete_selected_confirmation.html
- django/contrib/admin/templates/admin/popup_response.html
- django/contrib/admin/templates/admin/prepopulated_fields_js.html

I think it would also be nice to add a comment describing why the templates are there and/or linking to this PR.

underdarknl:
While this fixes the 'issue' for now, it introduces a complex update/upgrade path for us. Ideally we still make sure we fix this in the Django project itself, and not just in the scope of our project.
Code beyond that remark looks as I'd expect it to look.

Rieven:
I have tested it and it works as intended.

underdarknl (approved):

PR #1484, opened 2023-07-27
allow subject to be None

noamblitz:
### Changes
There are certificates that do not have a common name. KAT did not support those certificates which would make the normalizer fail, thus yielding no certificates and thus creating findings.

Maybe has to be backported to 1.10.


octopoes/bits/ssl_certificate_hostname/ssl_certificate_hostname.py
```diff
@@ -34,7 +34,7 @@ def run(
     additional_oois: List[Union[Website, SubjectAlternativeNameHostname]],
     config: Dict[str, str],
 ) -> Iterator[OOI]:
-    subject = input_ooi.subject.rstrip(".")
+    subject = input_ooi.subject.rstrip(".") if input_ooi.subject else ""
```
ammar92, line 37:
```suggestion
    subject = input_ooi.subject.rstrip(".") if input_ooi.subject is not None else ""
```

octopoes/bits/ssl_certificate_hostname/ssl_certificate_hostname.py
```diff
@@ -52,7 +52,7 @@ def run(
     for website in websites:
         hostname = website.hostname.tokenized.name.rstrip(".")
 
-        if subject_valid_for_hostname(subject, hostname):
+        if subject and subject_valid_for_hostname(subject, hostname):
```
ammar92, line 55:
```suggestion
        if subject is not None and subject_valid_for_hostname(subject, hostname):
```

octopoes/bits/ssl_certificate_hostname/ssl_certificate_hostname.py
```diff
@@ -52,7 +52,7 @@ def run(
     for website in websites:
         hostname = website.hostname.tokenized.name.rstrip(".")
 
-        if subject_valid_for_hostname(subject, hostname):
+        if subject is not None and subject_valid_for_hostname(subject, hostname):
```
praseodym, line 55: Our validation currently doesn't match the method specified in [RFC2818](https://datatracker.ietf.org/doc/html/rfc2818#section-3.1), which specifies that only `subjectAltName` should be checked when it is present:

> If a subjectAltName extension of type dNSName is present, that MUST
> be used as the identity. Otherwise, the (most specific) Common Name
> field in the Subject field of the certificate MUST be used. Although
> the use of the Common Name is existing practice, it is deprecated and
> Certification Authorities are encouraged to use the dNSName instead. 

Modern browsers remove validation of the Common Name field entirely, e.g. [Chrome removed support in 2017](https://bugs.chromium.org/p/chromium/issues/detail?id=308330).

praseodym (changes requested):

octopoes/bits/ssl_certificate_hostname/ssl_certificate_hostname.py
noamblitz: Agree.

praseodym (approved):

underdarknl (approved):

PR #1490, opened 2023-07-27
remove expect-ct finding

noamblitz:
Expect-CT is deprecated and now removed from port-classification

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect-CT

### DIFF
--- a/octopoes/bits/missing_headers/missing_headers.py
+++ b/octopoes/bits/missing_headers/missing_headers.py
@@ -77,16 +77,6 @@ def run(
         yield ft
         yield finding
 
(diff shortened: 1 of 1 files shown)


ammar92 (approved):

praseodym:
Don't we need to remove `KAT-NO-EXCPECT-CT` (nice typo btw) from `kat_finding_types.json` and `OOI_database_seed.json`?

underdarknl:
Since Except-CT is now deprecated, should we not move it to the deprecated headers list so we trigger that warning instead?

dekkers (approved):

Issue #1288, opened 2023-06-27
Add normalizer detail page


TwistMeister:
Since normalizers are part of the KAT-alogus they also need a detail page. We can then link to this from various places, like the task list and the "task history list" on the boefje detail page.

PR #1506, opened 2023-07-28
Feature: Add task detail pages and show objects yielded by normalizer

### LINKED ISSUE #398: Show yielded objects of normalize
_Please add one or more of the following labels to your issue:_
`frontend`

The link `Download meta and raw data` on the Task page for a normalizer does not return the yielded normalizer output, nor its logs. Instead the input (boefje output) is shown.

I would expect both the input, (current) output and the logs of the normalizer to be available.

Read the logs in the containers, not very user friendly, not possible on every system for all users.

**Additional context**
Also when yielding an object returns an Octopoes API error, this should also be visible in the frontend (and possibly not result in a green 'completed').

TwistMeister:
### Changes
Feature: Add task detail pages and show objects yielded by normalizer.

### Issue link

Closes #398 and #1288 

### Proof
image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


octopoes/octopoes/repositories/origin_repository.py
```diff
@@ -114,9 +114,6 @@ def save(self, origin: Origin, valid_time: datetime) -> None:
         except ObjectNotFoundException:
             old_origin = None
 
-        if old_origin == origin:
-            return
-
         self.session.add((XTDBOperationType.PUT, self.serialize(origin), valid_time))
```
underdarknl, line 117: Since we are not adding a new origin to the database, Should we not also change the previous origin (or old_origin) to be deleted from this moment in time?

rocky/rocky/views/bytes_raw.py
```diff
@@ -37,6 +37,15 @@ def get(self, request, **kwargs):
             return redirect(request.META["HTTP_REFERER"])
 
 
+# class NormalizerOriginRawView(OrganizationView):
```
underdarknl, line 40: Stale edit?

zcrt:
Nice! Would it be possible to (also) add (part of) this detailed information by expanding rows of the task table? (Just as currently with rows of the findings table.)

rocky/rocky/views/task_detail.py
```diff
@@ -0,0 +1,79 @@
...
+
+        context["breadcrumbs"] = [
+            {
+                "url": reverse("task_list", kwargs={"organization_code": self.organization.code}),
+                "text": _("Tasks"),
+            },
+            {
+                "url": reverse(
+                    "boefje_task_view",
+                    kwargs={"organization_code": self.organization.code, "task_id": uuid.UUID(context["task"].id).hex},
```
ammar92, line 40: The key `task_id` already exists in the super's context, so you can use that instead of recreating a UUID object

rocky/rocky/views/task_detail.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class NormalizerTaskDetailView(NormalizerMixin, TaskDetailView):
+    template_name = "tasks/normalizer_task_detail.html"
+    plugin_type = "normalizer"
+
+    def get_output_oois(self, task):
+        try:
+            return self.octopoes_api_connector.list_origins(
+                valid_time=task.p_item.data.raw_data.boefje_meta.ended_at, task_id=uuid.UUID(task.id).hex
```
ammar92, line 56: See my previous point about reusing vars

rocky/rocky/views/task_detail.py
```diff
@@ -0,0 +1,79 @@
...
+        context["output_oois"] = self.get_output_oois(context["task"])
+        context["breadcrumbs"] = [
+            {
+                "url": reverse("task_list", kwargs={"organization_code": self.organization.code}),
+                "text": _("Tasks"),
+            },
+            {
+                "url": reverse(
+                    "normalizer_task_view",
+                    kwargs={"organization_code": self.organization.code, "task_id": uuid.UUID(context["task"].id).hex},
```
ammar92, line 73: See my previous points

rocky/rocky/views/bytes_raw.py
TwistMeister: Yes, will remove

rocky/rocky/views/task_detail.py
TwistMeister: Obviously. Thanks for pointing this out.

rocky/rocky/views/task_detail.py
TwistMeister: How can I reuse the context variable here? Or should I store the value in another way so I'm able to access it here as well?

ammar92 (approved):

Darwinkel:
## Checklist for QA:

First of all, awesome feature!

### What works:
* The new pages
* No apparent regressions

### What doesn't work:
- `Download meta and raw data` button on the detail page for normalizers results in a 500:
image
image

- Links can refer to non-existent boefjes
image

- The column called "RAW Produced by Boefje" actually contains links to the boefjes' katalogus page, and not the raw output. 
image


### Bug or feature?:

- From a design perspective, there are now two links to  `Download meta and raw data`, one on the tasks list, and one on the detail page. Isn't this a bit redundant / clutter?

image
image

In addition, normalizers also have two links to download the raw input. Same point applies here.

TwistMeister:
@Darwinkel the link to the 'manual' boefje was broken already, so no regression there. I will create a fix for it on another branch, pointing at main.

The column "RAW created by boefje" isn't meant to include the RAW file, it's meant to state which boefje yielded said RAW file. I will put the column headers up for discussion with the UX-team so this will be more clear.

Having two download links, both in the list as on the detail page doesn't seem to troubling to me from a UX perspective. For design purposes it would be neat if we could remove it as a column from the list page, that I agree on. I will put this up for discussion as wel.

TwistMeister:
@zcrt While I think that's certainly possible and I do think something like that will be beneficial I would like to merge this as is, and not make the PR any bigger. I will take up the whole (re)design of the task lists and the detail pages with the UX-team as a whole instead of fixing this now and having to fix it again after the redesign.

octopoes/octopoes/repositories/origin_repository.py
underdarknl: Currently an OOI is discovered by a normalizer and added to the database. At that point the origin record is also created, and points to the ooi and the normalizer job that yielded that object.
On subsequent runs on the same boefje/normalizer combination no update would be done to the database, the object would simply keep existing as no valid_till was present. 

To be able to show what that second run 'proofed' we need to make sure we either store the output of each normalizer run somewhere (bytes perhaps), or update the origin record to now point at this most recent proof.
By doing the latter we end up with one of two  things depending on the specific implementation:
* An origin record that when seen trough time/history contains a list of all jobs ever involved in proofing this ooi's existence
* A set of origin records related to this OOI, which all exists end to end, and collectively are the historical record of which jobs are involved in proofing this OOI.
I don't know how origin records are currently implemented, so which one it is needs to be checked, and implemented for accordingly

Darwinkel:
## Checklist for QA:

Nice UX, I like the expandable menu.  

### What works:
- Rescheduling
- Clicking yielded objects
- Downloading completed raw and meta

### What doesn't work:
- Downloading dispatched meta
image

- Rescheduling dispatched boefjes
image

- Rescheduling dispatched normalizers
image

- When on a second page, a JavaScript popup occurs when clicking the expand button on the normalizers page
image

- When on a second page, I get a JSON response when clicking a normalizer task

rocky/assets/js/renderNormalizerOutputOOIs.js
```diff
@@ -0,0 +1,45 @@
...
+        callback(status, xhr.response);
+      }
+    };
+    xhr.send();
+  };
+
+
+  button.addEventListener("click", () => getJson(json_url, function(err, data) {
+    if (err !== null) {
+      alert('Somthing went wrong: ' + err);
```
underdarknl, line 27: This could just be moved to the getJson function, no need to pass along the err variable, and then act on it here.
Also, maybe do a console.log of the error, and some visual feedback for the user telling him the 'fold open' did not go as planned?

rocky/assets/js/renderNormalizerOutputOOIs.js
```diff
@@ -0,0 +1,45 @@
+
+const buttons = document.querySelectorAll(".expando-button.normalizer-list-table-row");
+
+buttons.forEach((button) => {
+  const raw_task_id = button.closest('tr').getAttribute('data-task-id');
+  const task_id = button.closest('tr').getAttribute('data-task-id').replace(/-/g, "");
+  const json_url = window.location.href + "/" + task_id;
```
underdarknl, line 7: If this JS is going to be used on other pages too, this maybe just needs to be a constant?

rocky/assets/js/renderNormalizerOutputOOIs.js
```diff
@@ -0,0 +1,45 @@
...
+    if (err !== null) {
+      alert('Somthing went wrong: ' + err);
+    } else {
+      if(data.length > 0) {
+        data.forEach(object => {
+          const element = document.createElement('p')
+          const url = window.location.href.replace('/tasks/normalizers', '')
+          element.innerHTML = "<a href='" + url + "/objects/detail/?ooi_id=" + object +"'>" + object + "</a>"
+
+          button.closest('tr').nextElementSibling.querySelector('#yielded-objects-'+raw_task_id).appendChild(element);
```
underdarknl, line 35: What happens if I open, and then close, and then open the same row again? Im guessing we'd end up with the same yielded objects listed twice?

TwistMeister: Thanks for pointing this out. I had already fixed this, but had not yet pushed my changes :)

rocky/assets/js/renderNormalizerOutputOOIs.js
TwistMeister: Not sure what you mean. The task_id is only known within the context of the button, right?

rocky/assets/js/renderNormalizerOutputOOIs.js
TwistMeister: Yes, this definitely needs another fallback then to just show an alert.

TwistMeister:
@Darwinkel I have addressed the points you raised. The points I didn't fix in this PR, since it wasn't regression are:
- Downloading dispatched meta results in an error. Follow-up ticket: https://github.com/minvws/nl-kat-coordination/issues/1523
- Rescheduling task "twice" results in an error. Follow-up ticket: https://github.com/minvws/nl-kat-coordination/issues/1589

What I did fix is that a task that isn't yet completed doesn't show the "reschedule" button. But if you click "reschedule" on an earlier instance of that task, while a new one is already running the error you observed persists, hence ticket https://github.com/minvws/nl-kat-coordination/issues/1589.

octopoes/octopoes/repositories/origin_repository.py
dekkers: We have some documentation at https://docs.openkat.nl/developer_documentation/octopoes.html#origin
The current Origin model:
https://github.com/minvws/nl-kat-coordination/blob/c51122df60d4556214f898fbdb12072f24fdb6a6/octopoes/octopoes/models/origin.py#L9-L40

Looking at the documentation and the code:
- The Origin model  is used for declarations (objects created through rocky), observations (normalizer output) and inferences (bit output)
- The Origin model has an optional task_id that is only used by normalizers. The `__eq__` of the Origin method currently ignores task_id which causes it to not add a new Origin object if the other fields are the same. The purpose of the Origin objects is to keep track of the origins of objects in the graph and for this we don't need to track every normalizer task, so that hasn't been a problem up to now.
- This proposed change will cause a lot of duplicate inference origins for every bit run that has the same result, so we should look at another way to achiee this.

If we want to track all normalizer tasks in the origins/graph I think we have two options to do that:

1. We change the `__eq__` to include the task_id which will do what the current change intends to do. This will create a new Origin for every normalizer run, but won't cause duplication of inference Origins. It will create a lot more Origins object than we have now and also duplicate observation Origins in the sense that they have the same source/result. I'm not sure whether there isn't code that assumes that that doesn't happen which we would have to fix. Given that it wasn't possible before we also don't have unit tests that would test this, so we should investigate that if implement it this way. There might also be performance issues when we have a lot more Origin objects. 
2. Change the Origin object to include a list of tasks instead of a single task. If the result is the same we just add the task to the lists of tasks of the already existing Origin object. The challenge here is how to change this without breaking existing installations. Given that the task_id currently just points to the first normalizer task that returned this result, I am not sure whether we are really using this information anywhere, so we might be able to just change this without a problems.

Personally I have a preference for option 2, because it fits better with how we use the Origin object and we don't risk performance issues by creatings lots of Origin objects in the database.

dekkers: This has been implemented

Darwinkel:
Issue has been fixed! 

Unrelated: I took the liberty to add `--use-fuzzy` to some of our `compilemessages` pipelines as I discovered that we have been forgetting/losing translations for a while now. Django sometimes tries to automatically infer/fix new translation strings, which cause them to be annotated as fuzzy. By default, `compilemessages` doesn't include these, which has the subtle effect of effectively discarding some of our previous translation work.

Rieven (changes requested):

underdarknl (approved):

PR #1517, opened 2023-07-31
Refactor environment settings, names, and documentation

Darwinkel:
Draft to check docs pipeline.

## Changes

### Issue link

Closes ...

### Proof
_Please add some proof of your working change here, unless this is not required (e.g. this PR is trivial)._

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


.env-defaults
```diff
@@ -0,0 +1,32 @@
+# Container entrypoints will run database migrations if set to "true"
+DATABASE_MIGRATION=true
+
+# SECURITY WARNING: don't run with debug turned on in production!
+DEBUG=True
+
+# SECURITY WARNING: enable two factor authentication in production!
+TWOFACTOR_ENABLED=False
+
+# --- OpenTracing --- #
```
praseodym, line 10: OpenTelemetry, not OpenTracing (its precursor)

bytes/bytes/config.py
```diff
@@ -2,41 +2,83 @@
...
-
-    span_export_grpc_endpoint: Optional[str]
-    bytes_metrics_ttl_seconds: int = 300
+    secret: str = Field(..., description="Secret key used for generating Bytes' API JWT")
+    username: str = Field(..., description="Username used for generating Bytes' API JWT")
+    password: str = Field(..., description="Password used for generating Bytes' API JWT")
+    queue_uri: AmqpDsn = Field("amqp://", description="KAT queue URI", env="QUEUE_URI")
+    log_cfg: Path = Field(BASE_DIR / "dev.logging.conf", description="Path to the logging configuration file")
+
+    db_uri: PostgresDsn = Field("postgresql://xx:xx@host:5432/bytes", description="Bytes Postgres DB URI")
```
praseodym, line 19: Why would we set invalid defaults for a field? Wouldn't it be better to get an error that the configuration field is not set, rather than an error that the database connection failed because of some invalid default?

pyproject.toml
```diff
@@ -105,6 +105,8 @@ sphinx = "6.1.3"
 sphinx_rtd_theme = "1.2.2"
 sphinxcontrib-mermaid = "0.8.1"
 myst-parser = "2.0.0"
+settings-doc = "2.0.0"
+colorama = "0.4.6" # Required on all platforms, not just win32
```
praseodym, line 109: Why is it required on all platforms? The [Colorama PyPI page](https://pypi.org/project/colorama/) clearly says "Makes ANSI escape character sequences (for producing colored terminal text and cursor positioning) work under MS Windows.".

bytes/bytes/config.py
Darwinkel: Fair point. Maybe a middle ground would be to have empty but required defaults, and provide an example? 

E.g. `db_uri: PostgresDsn = Field(..., description="Bytes Postgres DB URI", example="postgresql://xx:xx@host:5432/bytes")`

.env-defaults
Darwinkel: Fixed!

bytes/bytes/config.py
Darwinkel: I've changed it. Let me know what you think :)

pyproject.toml
underdarknl: We could upstream an issue to them, but im guessing that dependency management does not have the correct handles to make this work based on OS.

ammar92 (commented):
I have no more remarks! We just need to fix the unit tests, but it shouldn't be too hard.

pyproject.toml
praseodym: Python dependency management can certainly do this, `colorama` is currently only required on win32:
https://github.com/minvws/nl-kat-coordination/blob/3da1c9aa97fb84e710cd74b89e6225727cb30637/requirements.txt#L86-L88

If the `typer` package would specify its dependencies like this I think it'd work fine.

Darwinkel:
Tested:

- dnssec, manyportsopen, and dns-records boefjes (these directly use config settings)
- Enabling/disabling opentelemetry
- Regular PR QA flow
- Question -> Config -> Findings 
- WPScan boefje

Not tested:

- Debian packages
- Katalogus encryption
- Timestamping service

pyproject.toml
dekkers: Click and sphinx only depend on colorama on win32, but typer depends on colorama on all platforms. So I think there are two issues here that we should report upstream:
- Typer should only depend on colorama on win32
- Poetry should not add the sys_platform to the colorama dependency in requirements.txt because types doesn't limited its dependency to win32.

Makefile
```diff
@@ -12,6 +12,15 @@ UNAME := $(shell uname)
 export DOCKER_BUILDKIT=1
 export COMPOSE_DOCKER_CLI_BUILD=1
 
+define build-settings-doc
+	PYTHONPATH=./$(1) settings-doc generate \
+	-f markdown -m $(2) \
+	-u docs/source/technical_design/environment_settings/$(3).md \
+	--heading-offset 1 \
+	--between "<!-- generated env. vars. start -->" "<!-- generated env. vars. end -->"
+endef
```
dekkers, line 21: I don't really like that settings-doc updates files that are in git and you have to run `git restore` after running `make docs` and make sure you don't accidently commit those files when working on documentation.

I would prefer to have some like `boefje.md.in` in git that is used to generate a `boefje.md` that is .gitignore.

boefjes/boefjes/config.py
```diff
@@ -2,43 +2,58 @@
...
 
 
 class RuntimeConfiguration(Enum):
     LOCAL = "local"
 
 
 class Settings(BaseSettings):
-    base_dir: Path = Path(__file__).parent.resolve()
-    log_cfg: Path = Path(__file__).parent / "logging.json"
+    base_dir: Path = Field(Path(__file__).parent.resolve())
```
dekkers, line 13: base_dir is not really a setting. It also only used in the tests, so we should use a normal variable for that.

docs/source/technical_design/environment_settings/index.rst
```diff
@@ -0,0 +1,15 @@
...
+.. toctree::
+   :maxdepth: 2
+   :caption: Contents
+
+   rocky
+   mula
+   octopoes
+   bytes
+   boefjes
+   keiko
```
dekkers, line 15: Can we sort this alphabetically?

boefjes/boefjes/config.py
```diff
@@ -2,43 +2,58 @@
...
+    log_cfg: Path = Field(Path(__file__).parent / "logging.json", description="Path to the logging configuration file")
```
dekkers, line 14: This will list the build directory of the documentation as default value which isn't really nice. There are also other places where the default value isn't even correct, such as when it is `set()` or with enums. Is there a way to override the default value in the documentation?

dekkers (changes requested):
I think we should provide backwards compatibility for the old names. We should either keep them permanently or for one or two versions and add warnings when the old names are used.

Unfortunately pydantic settings doesn't have built-in support for that, but we can do it ourself in the following way:

```python
import os

from typing import Any, Dict, Tuple
from pydantic import BaseSettings, PostgresDsn
from pydantic.env_settings import SettingsSourceCallable


class BackwardsCompatibleEnvSettings:
    backwards_compatibility_mapping = {
        'OLD_DATABASE_DSN': 'NEW_DATABASE_DSN',
    }

    def __call__(self, settings: BaseSettings) -> Dict[str, Any]:
        d: Dict[str, Any] = {}
        env_vars = {k.lower(): v for k, v in os.environ.items()}

        for old_name, new_name in self.backwards_compatibility_mapping.items():
            old_name, new_name = old_name.lower(), new_name.lower()

            if old_name in env_vars:
                d[new_name] = env_vars[old_name]

        return d


class Settings(BaseSettings):
    new_database_dsn: PostgresDsn

    class Config:
        @classmethod
        def customise_sources(
            cls,
            init_settings: SettingsSourceCallable,
            env_settings: SettingsSourceCallable,
            file_secret_settings: SettingsSourceCallable,
        ) -> Tuple[SettingsSourceCallable, ...]:
            backwards_compatible_settings = BackwardsCompatibleEnvSettings()
            return env_settings, init_settings, file_secret_settings, backwards_compatible_settings
```

boefjes/boefjes/config.py
```diff
@@ -2,43 +2,58 @@
...
-    bytes_api: AnyHttpUrl = "http://localhost:8002"
-    bytes_username: str = "test"
-    bytes_password: str = "secret"
-
-    span_export_grpc_endpoint: Optional[str] = None
-
-    remote_ns: str = "1.1.1.1"
-
-
-settings = Settings()
```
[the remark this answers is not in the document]
Darwinkel: 100% agree that this is a mediocre solution. The problem is that `settings-doc` attempts to load the module `config.py`, which results in a pydantic validation error given that `Settings()` is run. The only workaround that I can think of for this would be to explicitly provide some required environment variables when running `make docs`. But that's also not a clean solution. 

I haven't been able to find a way to cleanly inject environment variables, or how we can add an if statement to not load `Settings()` if run through `settings-doc`. Do you have any ideas for a solution? Now that I'm writing this, it occurs to me that an alright-ish middle ground could be to manually check (outside of Pydantic) if `CI=1` is set.

boefjes/boefjes/config.py
```diff
@@ -2,43 +2,58 @@
...
+    runtime: RuntimeConfiguration = Field(
+        RuntimeConfiguration.LOCAL, description="Type of runtime to use", possible_values=["local"]
+    )
```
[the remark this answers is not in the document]
Darwinkel: I've kept it in with future support for alternate Boefje runtimes in mind. We could remove the setting for now and add it when the design for the containerized runtime is ready, I guess.

.github/workflows/boefjes_tests.yml
```diff
@@ -47,3 +47,13 @@ jobs:
...
+        env:
+          KATALOGUS_DB_URI: "postgresql://placeholder@host:5432/katalogus"
+          QUEUE_URI: "amqp://placeholder"
+          XTDB_URI: "http://placeholder:3000"
+          KATALOGUS_API: "http://placeholder:8000"
+          OCTOPOES_API: "http://placeholder:8001"
+          SCHEDULER_API: "http://placeholder:8002"
+          BYTES_API: "http://placeholder:8003"
+          BYTES_USERNAME: "placeholder"
+          BYTES_PASSWORD: "placeholder"
```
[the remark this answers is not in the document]
Darwinkel: I'll look into it.

docs/source/technical_design/environment_settings/rocky.md
```diff
@@ -0,0 +1,10 @@
+# Rocky
+
+Note that Rocky does not support auto-generated environment documentation.
```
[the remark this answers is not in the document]
Darwinkel: Looks like a great option for a follow-up PR :)

Makefile
Darwinkel: I'll look into it

boefjes/boefjes/config.py
Darwinkel: I'll look into refactoring it for all modules.

boefjes/boefjes/config.py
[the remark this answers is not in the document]
Darwinkel: I fixed it!

Makefile
Darwinkel: @dekkers I've rewritten the `make docs` command such that it generates files from scratch through jinja templates. What do you think?

boefjes/boefjes/config.py
Darwinkel: I've changed it such that the BASE_DIR is fixed when compiling documentation. As for the other variables: I don't think there's a good way to do this except for either modifying the jinja template or manually specifying an if-statement alternative for DOCS=1. We could do either, but given that it's not critical I'd suggest moving this to future work.

Darwinkel:
@dekkers @ammar92 I'd like to merge this PR and do the pending improvements in this follow-up: https://github.com/minvws/nl-kat-coordination/issues/1580

ammar92 (dismissed):

dekkers:
> > @dekkers @ammar92 I'd like to merge this PR and do the pending improvements in this follow-up: #1580
> 
> I want to see the backwards compatibility implemented before merging the PR. I don't think it is a lot of effort to add the mapping of old names to new names to the the snippet I provided compared to how much trouble it will cause for all our users if we don't implement backwards compatibility.

To better explain why I think we should keep backwards compatibility:
- From personal experience doing ops for the last 20 years, projects that break their configuration by renaming things without providing backwards compatibility just cause a lot of work/problems which could have easily be prevented by simple backwards compatibility. On the other hand there are also projects where the old names still work for years because providing this kind of backwards compatibility isn't hard.
- The lack of forward compatibility also means you need to change configuration and upgrade at the same time. In case you have multiple environment that share configuration (e.g. develop, test and production) there is also not really a good/easy way to put this in your configuration management system. If you upgrade develop, it won't work with the old configuration, while your test/production won't work with the new one. This means you have to duplicate all the renamed variables and then later on spend more time removing the old ones.
- Based on the snippet I provided the only work that is left to do is defining the mapping of old to new variable names and putting it in the code. This shouldn't really be a lot of effort compared to the difficulties all our users will have when we don't provide backwards compatibility.

Darwinkel:
I've implemented and tested backwards compatibility for some of the more important variables:

image

boefjes/boefjes/config.py
```diff
@@ -1,44 +1,94 @@
...
 
-    katalogus_db_uri: PostgresDsn = "postgresql://xx:xx@host:5432/katalogus"
+        for old_name, new_name in self.backwards_compatibility_mapping.items():
+            old_name, new_name = old_name.lower(), new_name.lower()
 
-    scheduler_api: AnyHttpUrl = "http://localhost:8004"
+            # New variable not explicitly set through env,
+            # ...but old variable has been explicitly set through env
+            if new_name not in env_vars and old_name in env_vars:
+                logging.warning("Deprecation: %s is not valid, use %s instead", old_name.upper(), new_name.upper())
```
dekkers, line 32: Small nitpick, that isn't a blocker for merging: I would say "is deprecrated" instead of "is not valid", because it will still be valid as long was we keep the backwards compatibility.

Darwinkel:
Now reflects JP's updated scheduler documentation.

mula/.ci/.env.test
```diff
@@ -5,13 +5,13 @@ POSTGRES_DB=scheduler
 SCHEDULER_API_HOST="0.0.0.0"
 SCHEDULER_API_PORT="8000"
 SCHEDULER_PQ_MAXSIZE=0
-SCHEDULER_PQ_INTERVAL=60
-SCHEDULER_PQ_GRACE=1
-SCHEDULER_RABBITMQ_DSN=amqp://guest:guest@ci_rabbitmq:5672/%2fkat
+SCHEDULER_PQ_POPULATE_INTERVAL=60
+SCHEDULER_PQ_POPULATE_GRACE_PERIOD=1
+QUEUE_URI=amqp://guest:guest@ci_rabbitmq:5672/%2fkat
```
jpbruinsslot, line 10: might be a bit confusing naming this `QUEUE_URI` since it mainly directs to the rabbitmq host 'messaging queue' from which the scheduler pulls messages from. It could be misinterpreted that this might be the queue of the scheduler itself.

jpbruinsslot (approved):

dekkers (approved):

PR #1538, opened 2023-08-02
Add first version of new normalisers runner design

### LINKED ISSUE #1136: Design new Normalizer runner
Design the new Normalizer runner

Considerations:
- Runtime
- Image distribution
- I/O protocol specification
- Configuration within KAT

praseodym:
Closes #1136


ammar92 (approved):

boefjes/docs/normalisers-runner.md
```diff
@@ -0,0 +1,128 @@
...
+* Cache normaliser dependencies.
+
+## Design
+
+### Runtime
+
+| Technology | Security | Startup overhead | Ease of Use | Distribution |
+| --- | --- | --- | --- | --- |
+| MicroVM (Firecracker) | Excellent isolation | High | Hard | VM image |
+| Containers (Docker) | Medium isolation | Medium-high | Moderate | OCI image |
```
dekkers, line 60: The isolation depends on how the containers are deployed. With Kata Containers or gVisor you can also have excellent isolation of containers.

I think this is an important fact, because it means that if we create OCI images people with not so high security requirements that want a simple setup can just use docker or default Kubernetes setup with CRI-O or containerd, while people with high security requirements can create a more complex setup with hardware virtualization to use Kata Container or gVisor. 

With kubernetes using Kata Containers or gVisor is also pretty easy, we would only have to provide a way to let user configure which RuntimeClass to use.

praseodym: Good point, we discussed this earlier but this was not written down. I've added a separate paragraph on hardened runtimes.

dekkers (approved):

Issue #908, opened 2023-05-03
Some users may not be able to finish the onboarding due to an insufficient clearance level on their account


Rieven:
When onboarding a new redteam member, a clearance level of L2 is at least needed to scan a URL object during the onboarding. An extension of https://github.com/minvws/nl-kat-coordination/issues/880

**Issues:**
1. There is no place where a redteamer can acknowledge for L2 clearance during onboarding.
2. Admins can give clearance to redteamers, so admins must also at least give redteamers a clearance of L2, otherwise onboarding a redteamer cannot continue or at least proceed further than scanning.

**Solutions and discussions:**
- Admins cannot create redteamers that has not at least a clearance of L2 (accepted clearance level)
- Admins get a warning that a clearance of L2 is at least needed for a proper redteam onboarding.

What will happen if admin did not give clearance for L2? 
- What is the user flow when onboading? Do we proceed till scanning or at the beginning we say that there is no clearance.

**Feature needed:**
- A screen or split wizard flow to include clearance level acknowledgement. @HeleenSG and @NoemiOverdijk is going to look into this. Show clearance level admins gave to redteamer (accepted clearance level)

HeleenSG:
The first part @Rieven  mentioned will be picked up in a different ticket.
"There is no place where a redteamer can acknowledge for L2 clearance during onboarding." By adding it to the onboarding flow. 

**The part we would like to discuss here is:** 
The current onboarding flow asks the user to assign clearance level 2 to an object. But what if the user doesn't have permission to do so? This results in the user being unable to finish onboarding.

**Possible solutions that we've discussed so far are:**
- When creating a red teamer, an admin must give a clearance level of at least level 2. 
- Change the example in the onboarding so a lower clearance level is needed. (0, cause level 1 would just move the problem without actually solving it) 
- Grant the user temporary clearance for the duration of the onboarding (This will be confusing to users so wouldn't recommend from an ux perspective. And it might have legal/responsibility issues since the user chooses the url to scan and might not have permission to do so.) 
- Admins get a warning that a clearance of L2 is at least needed for a proper redteam onboarding.

Please add suggestions, thoughts etc.

PR #1474, opened 2023-07-26
Clearance level choice for redteam member creation + warning for onboarding redteamers with clearance level below L1

Rieven:
### Changes
- Member creation has now an added flow (wizard type like), first to select the account type.
- Based on the account type, redteamers will have a different form, having the clearance level (paws) added.
- Trusted clearance level selection is not shown in the form when adding admins/clients.
- Creation of admins will automatically get a L4 (This was the idea from the beginning, as admins have the most clearance as well)
- Creations of clients will have the default clearance level of -1
- Onboarding redteam creation have now the trusted clearance level added.
- Cleanup/refactoring of code necessary, example the use of double forms and logic handling between the onboarding flow and member creation flow.

### Issue link (partially solved)
https://github.com/minvws/nl-kat-coordination/issues/908

### Proof

Frist select account type:
image

Adding redteamer:
image

Warning at user creation, trusted clearance level below L2
image

Adding admins/client same form without trusted clearance level:
image

---

### Code Checklist
- [x] All the commits in this PR are properly PGP-signed and verified;
- [x] This PR only contains functionality relevant to the issue; tickets have been created for newly discovered issues.
- [x] I have written unit tests for the changes or fixes I made.
- [x] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [x] I have performed a self-review of my code and refactored it to the best of my abilities.

### Communication
- [x] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [x] I have made corresponding changes to the documentation, if necessary.

---
## Checklist for code reviewers:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_code.md) into your comment.

---
## Checklist for QA:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_qa.md) into your comment.

PR #1549, opened 2023-08-02
Redteamer can now acknowledge clearance level during onboarding

Rieven:
### Changes
- Adding screen in onboarding to acknowledge clearance level
- This is a follow up of: https://github.com/minvws/nl-kat-coordination/pull/1474

### Issue link
Closes https://github.com/minvws/nl-kat-coordination/issues/908

### Proof
This step has been splitted now (introduction about clearance levels) 
image

NEW added screen that talks about the clearance level permissions and acknowledging
Here a redteamer does not have enough clearance to continue onboarding
image

Screen to accept onboarding or just skip 
image

After accepting, now one can continue onboarding
image

Setting clearance level comes afterwards, only focusses on setting the clearance level:
image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


underdarknl:
The required clearance level for the onboarding is a function of the boefjes that the DNSReport requires, and as such is dynamic. We should let the user know which level of trust he needs from his Admin in order to complete this report. If the admin has not given the user enough permissions or a high enough trusted clearance level, he needs to be directed to talk to his admin.

In the last screen, we should change to wording to make sure that the user knows he is setting the level for the URL object he entered earlier.

Rieven:
> The required clearance level for the onboarding is a function of the boefjes that the DNSReport requires, and as such is dynamic. 

Yes, if the clearance level is at least 1 or above, there are no Boefjes showed when it has a clearance of for example L0 and then the onboarding for making a DNS Report does not make any sense. I have tested also a max level of L4, the same Boefjes are populated in the selection. 

image

Because making a DNS Report is very specific and uses only 3 Boefjes, see screenshot, we are limiting the choice here. the dynamic part comes in when any report can be generated according to the max scan level. (trusted + acknowledges clearance). I think we need to make an apart PR for this having the clearance level decides which Boefjes is needed and also which Report can be generated.

Rieven:
> We should let the user know which level of trust he needs from his Admin in order to complete this report. If the admin has not given the user enough permissions or a high enough trusted clearance level, he needs to be directed to talk to his admin.

Yes, in this case "contacting an admin" will happen outside the onboarding as the user at the moment cannot continue the onboarding. Skipping the onboarding is here an option.

Rieven:
> In the last screen, we should change to wording to make sure that the user knows he is setting the level for the URL object he entered earlier.

Yes, forgot to add it :)
image

rocky/account/views/account.py
```diff
@@ -13,31 +15,36 @@ class PageActions(Enum):
...
-                organizationmember.acknowledged_clearance_level = organizationmember.trusted_clearance_level
+                self.organization_member.acknowledged_clearance_level = self.organization_member.trusted_clearance_level
             elif action == PageActions.WITHDRAW_ACCEPTANCE.value:
-                organizationmember.acknowledged_clearance_level = -1
+                self.organization_member.acknowledged_clearance_level = -1
             else:
                 raise Exception(f"Unhandled allowed action: {action}")
-            organizationmember.save()
+            self.organization_member.save()
         except RequestException as exception:
```
dekkers, line 33: I know that this is already there, but the catching of `RequestException` doesn't really make sense here because we don't do a request. I think we remove the try/except.

rocky/account/views/account.py
```diff
@@ -13,31 +15,36 @@ class PageActions(Enum):
...
 
 
-class AccountView(OrganizationView, DetailView):
-    template_name = "account_detail.html"
-    context_object_name = "member"
-
-    def get_object(self):
-        return OrganizationMember.objects.get(user=self.request.user, organization=self.organization)
-
+class OOIClearanceManager(OrganizationView, TemplateView):
```
dekkers, line 18: This is a Mixin, not a Manager. That also means it shouldn't inherit from the other views.

rocky/onboarding/templates/step_3e_trusted_acknowledge_clearance_level.html
```diff
@@ -0,0 +1,76 @@
...
+                {% include "partials/stepper.html" %}
+
+                <h2>{% translate "Setup scan - OOI clearance for" %} {{ ooi }}</h2>
+                <h3>{% translate "Trusted clearance level" %}</h3>
+                {% blocktranslate %}
+                    The administrators have implemented a trusted clearance level permission,
+                    ensuring that scanning capabilities are restricted to
+                    objects within this trusted clearance level. This precautionary measure helps
+                    maintain a secure environment by limiting their access to sensitive information
+                    beyond the designated threshold.
```
dekkers, line 22:
```suggestion
                    OpenKAT has a permission system that allows administrators to 
                    configure which users can set a certain clearance level. The will make sure
                    that only users that are trusted can start the more aggresive scans.
```

rocky/account/views/account.py
```diff
@@ -13,30 +10,22 @@ class PageActions(Enum):
...
 
 
-class AccountView(OrganizationView, DetailView):
-    template_name = "account_detail.html"
-    context_object_name = "member"
-
-    def get_object(self):
-        return OrganizationMember.objects.get(user=self.request.user, organization=self.organization)
-
+class OOIClearanceMixin(OrganizationView, TemplateView):
```
dekkers, line 13: A mixin should not inherit from View classes. So in this case `OOIClearanceMixin` should inherit from no other class. The `AccountView` and `OnboardingAcknowledgeClearanceLevelView` should inherit from  `OOIClearanceMixin`, `OrganizationView` and `TemplateView`. 

This will still inherit from two View classes which is also not correct, but was already the case and is something we should fix later. We still shouldn't make it worse however.

Rieven:
> * Contact the administrator for receiving higher clearance

Added more context:
image

Darwinkel:
Re-tested. Fix confirmed! :)

dekkers (approved):

Issue #1471, opened 2023-07-26
Documentation about question OOIs


Donnype:
### Issue links with info and schematics:
- https://github.com/minvws/nl-kat-coordination/issues/288

### PRs
- https://github.com/minvws/nl-kat-coordination/pull/1407
- https://github.com/minvws/nl-kat-coordination/pull/1408
- https://github.com/minvws/nl-kat-coordination/pull/1409

### Diagrams and technical explanation

Config flow and cardinality:

```mermaid
classDiagram

direction RL
    class Question
    class Config
    class JsonAnswer
    class JsonAnswer2["JsonAnswer"]

    Question : ooi[OOI]
    Question : schema[str]

    Config : ooi[OOI]
    Config : config[str]
    Config : bit_id[str]

    OOI <|-- Question
    Question <|-- JsonAnswer
    Question <|-- JsonAnswer2
    JsonAnswer <|-- Config
    JsonAnswer2 <|-- Config
```

Resulting ERD:

```mermaid
classDiagram

direction RL
    class Question
    class Config
    class OOI

    Question : ooi[OOI]
    Question : schema[str]

    Config : ooi[OOI]
    Config : config[str]
    Config : bit_id[str]

    OOI <|-- Question
    OOI <|-- Config
```

#### The Flow of creating Configs through Questions

```mermaid
sequenceDiagram
    participant User
    participant Rocky
    participant Normalizer
    participant Octopoes
    participant Bits
    participant Bytes

    Normalizer->>Octopoes: Add Network
    Bits->>Octopoes: Add Question["What ports may be open for this Network?"]
    Rocky->>Octopoes: Get Question
    Rocky->>User: Prompt Question to user
    User->>Rocky: Give answer (form)
    Rocky->>Bytes: Add answer (json) to Bytes
    Bytes->>Normalizer: Read answer
    Normalizer->>Octopoes: Create Config
    Bits->>Octopoes: Read Config
```

---

To create a question OOI in Rocky:
- Rebuild Rocky and the boefjes/normalizers (Or: make kat)
- Add a Network object
- Increase its clearance level to at least L2
- Wait a few seconds and refresh the objects page (perhaps filter on Question objects)
- Open the Question detail page and see the json form rendered.

PR #1569, opened 2023-08-07
add question objects / config bits to the manual

ring-ring-ring:
### Changes
add question oois to the manual with some basic explanation

related to this issue
https://github.com/minvws/nl-kat-coordination/issues/1471


docs/source/introduction/img/dataflowquestionobjects.png
praseodym, line 1: I'm not sure how you made this diagram, but would it be possible to add the source or make the diagram in Mermaid so it's possible to edit it later?

ring-ring-ring: for sure; its a mermaid diagram I stole from @Donnype but when using restructuredtext it does not render automagically, hence the png.

ring-ring-ring: https://github.com/minvws/nl-kat-coordination/issues/1471 < Donny kindly pasted them here - I tried to use the main one but failed to make it work with rst.

praseodym: Maybe we should figure out why Mermaid doesn't work in reStructuredText? It seems to work fine on docs pages written in Markdown, e.g.:

https://github.com/minvws/nl-kat-coordination/blob/59b12f379211a53efb08706ad2e566cc4fe939b1/docs/source/developer_documentation/rocky.md?plain=1#L94-L119

is rendered on https://docs.openkat.nl/developer_documentation/rocky.html#running-a-boefje.

ammar92: Strange indeed, it should also work without embedding it in Markdown. I've updated `sphinxcontrib-mermaid` (0.8.1 -> 0.9.2) in this branch, perhaps you could try this again @ring-ring-ring?

ring-ring-ring: image

Computer says no. Any ideas?

ammar92: I think there's an indentation problem with your example, but also an error in the syntax (`end` shouldn't be there). Anyways I managed to get this documentation built with Sphinx, but it look like `rstcheck` doesn't know the `mermaid` directive which causes the hook to fail. 

image

@Darwinkel any ideas on how to make this work with `rstcheck` as well?

ammar92: > @Darwinkel any ideas on how to make this work with `rstcheck` as well?

I was thinking if there's no good solution we could [ignore](https://rstcheck.readthedocs.io/en/latest/usage/config/#ignore-directives) the `mermaid` directive and optionally use external mermaid files (possible with the current extension) and use a pre-commit hook to check these mermaid files instead.

praseodym: For completeness: this `rstcheck` change was done in https://github.com/minvws/nl-kat-coordination/commit/627ebacc222de08a88498e7213800de78358f18b

ammar92: > For completeness: this `rstcheck` change was done in [627ebac](https://github.com/minvws/nl-kat-coordination/commit/627ebacc222de08a88498e7213800de78358f18b)

@praseodym Thanks for noticing this, I forgot to update this thread

Darwinkel: @ammar92 I'm not aware of standalone mermaid checkers nor good ways to modify rstcheck. I think using the `ignore` directive is appropriate here.

Darwinkel (approved):

praseodym (approved):

dekkers (approved):

Issue #1472, opened 2023-07-26
Documentation about CSV uploads


Donnype:
### PR
- https://github.com/minvws/nl-kat-coordination/pull/1423

### Issue
- https://github.com/minvws/nl-kat-coordination/issues/853

PR #1605, opened 2023-08-13
csv upload documentation

ring-ring-ring:
Added the CSV upload to the documentation: https://github.com/minvws/nl-kat-coordination/issues/1472


docs/source/manual/usermanual.rst
```diff
@@ -149,7 +149,80 @@ User management
...
+
+*Warnings*
+
+If the CSV file contains data that cannot be parsed OpenKAT will give a warning with the data concerned.
+
+User notification
+*****************
+
+After the CSV file has been uploaded the users receive a welcome email on their account. The link in this email allows them to create a password for their account.
+
```
underdarknl, line 207: We should make sure we tell the user this only works when their OpenKAT install can actually send emails.
@Rieven has at some point written a piece of detection code that can check if the required variables are filled in in the environment. If not, we should probably tell the user adding these accounts as such.
We no outbound Email server is configured the emails are printed to std-out (I think)

Darwinkel: Just a thought I had, but maybe a docs page about emailing in general would be useful.

Rieven: Yes, there is a check when the user uses the link to reset password. https://github.com/minvws/nl-kat-coordination/blob/6cb366b424b3cf0dae97e0ff185394a7f66d07f3/rocky/account/views/password_reset.py#L45

Also the server must be setup to send emails, if not it will use the std-out 
image

dekkers: I think we should disable CSV upload when e-mail isn't configured and DEBUG isn't set with a message that e-mail needs to be configured first. The console email backend should only be used for development and testing.

docs/source/manual/usermanual.rst
```diff
@@ -392,97 +492,3 @@ Automation of the verification process
 **************************************
 
 OpenKAT has been created to automate tedious tasks such as this one. We like to include an automated verification process for objects that includes the entire chain of information, with nice green checkmarks. It is on the roadmap, if you want to contribute to it you are most welcome! Get in touch through meedoen@openkat.nl.
-
-What are Config OOI's?
```
praseodym, line 396: This was only added last month (cc78ec6ced32967ef594e5fddbe8fae18e8c9b50), so I'd say it deserves a better location rather than being deleted.

Darwinkel: Most of it is rewritten in this PR: https://github.com/minvws/nl-kat-coordination/pull/1569

praseodym (approved):

dekkers (approved):

PR #1611, opened 2023-08-14
Scheduler rate limit functionality

### LINKED ISSUE #1317: Rate limiting by interval and group as set in boefje manifest.
We need some way of doing rate-limiting when accessing external services. The easiest way for now is to keep track of which jobs are given out to the runtimes and make sure we dont hit the various services too often.

Implement the following (low hanging fruit) rate-limiter:

Describe in the boefjes-manifest (if required) two settings:
* rate limit interval
* rate limit group target:
** Either a string (eg, shodan.com)
** or a placeholder that we can replace with data from the referenced inputOOI. (eg, [ipaddress])

When a worker requests a job, check the boefje manifest and set a timer (if needed) for the duration interval, and key it to the given grouping target.
If a subsequent worker requests a job, make sure to see if the grouping target already has a timer, and if it is still within the duration interval, if so, select the next job. If not give it out, and reset the timer to the current time.

Some cleaning of obsolete timers can be done periodically.
We could probably do this filtering in the job query to avoid repeated database queries when asking for jobs for a given boefje worker.

None currently spring to mind.

**Additional context**

jpbruinsslot:
### Changes

This PR adds rate-limiting capabilities to the scheduler for tasks. Boefjes that are subject to rate limiting (defined in the boefje manifest) are being tracked by an in-memory rate limiter, leveraging the package [limits](https://limits.readthedocs.io/en/stable/index.html). When a task is subject to a rate-limit it will get the `DELAYED` status and will be postponed until it is allowed again. A thread will check for delayed tasks and put them on the queue when they're ready.

This will expect some changes in the katalogus mainly exposing a grouping/namespacing on which to rate limit and the [parseable rate limit](https://limits.readthedocs.io/en/stable/quickstart.html#rate-limit-string-notation).

Example:

- grouping/namepace: `org-1/external-api-key`
- rate limit: `10/hour;100/day;2000 per year`

### Issue link
Closes #1317 

### Processess

For a boefje scheduler when a task gets created it will get check whether the boefje is rate limited or not:

```mermaid
flowchart TD
    A[Task Creation] --> B{is_task_rate_limited}
    B -->|Yes| C[set Task to status <br>DELAYED]
    B ---->|No| E[End]
```

Checking if a task is rate limited:

```mermaid
graph TD
    A[Start] --> B{rate_limit is None?}
    B -- Yes --> C[Return False]
    B -- No --> D{Try to parse rate_limit}
    D --> E{parsed_rate_limit is None?}
    E -- Yes --> F[Raise ValueError]
    E -- No --> G[Lock rate_limiter]
    G --> H[Get identifier_template]
    H --> I[Render identifier from template]
    I --> J[Test rate_limiter with identifier]
    J --> K{can_consume?}
    K -- No --> L[Return True]
    K -- Yes --> M{hit?}
    M -- Yes --> N[Hit rate_limiter]
    N --> O[Return False]
    M -- No --> O
    D --> P[Caught ValueError]
    P --> Q[Log warning]
    Q --> R[Raise exc]
```

Within the boefje scheduler a new process (`push_tasks_for_delayed_tasks`) is running next to:

* check for scan profile changes
* check for enabled boefjes
* check for rescheduling (random ooi endpoint octopoes)

This `push_tasks_for_delayed_tasks` method gets all tasks that have the status of `DELAYED`. And will be checked whether or not task can be pushed onto the queue.


jpbruinsslot:
@ammar92 I've implemented organisational support would you mind taking a look?

praseodym:
Originally posted at https://github.com/minvws/nl-kat-coordination/pull/1413#pullrequestreview-1569235551, but still valid:

As far as I understand, the rate limiter is a property of the `BoefjeScheduler` class. This class has a separate `rate_limiter` instance per organisation, which means that rate limiting only works per organisation.

In larger deployments, API keys such as for Shodan will likely be shared by multiple KAT organisations. How do we implement rate limiting for that scenario?

jpbruinsslot:
> Originally posted at [#1413 (review)](https://github.com/minvws/nl-kat-coordination/pull/1413#pullrequestreview-1569235551), but still valid:
> 
> As far as I understand, the rate limiter is a property of the `BoefjeScheduler` class. This class has a separate `rate_limiter` instance per organisation, which means that rate limiting only works per organisation.
> 
> In larger deployments, API keys such as for Shodan will likely be shared by multiple KAT organisations. How do we implement rate limiting for that scenario?

I've updated the code so that the rate limiter memory is now on the application level. The identifier of the rate limiter can then be formatted with a key for instance so that it will be shared across multiple organisations.

underdarknl:
> In larger deployments, API keys such as for Shodan will likely be shared by multiple KAT organisations. How do we implement rate limiting for that scenario?

This also means that we should probably include both 'shodan' and the actual key in the 'group' to make sure we don't limit shodan requests across the whole install even though we might have multiple keys available.

mula/scheduler/schedulers/boefje.py
```diff
@@ -379,6 +435,49 @@ def is_task_allowed_to_run(self, boefje: Plugin, ooi: OOI) -> bool:
...
+                self.organisation.id,
+                self.scheduler_id,
+            )
+            raise exc
+
+        with self.ctx.rate_limiter_lock:
+            # Get the identifier for the rate limiter
+            identifier_template = rate_limit.identifier
+            environment = Environment(loader=BaseLoader())
+            identifier = environment.from_string(identifier_template).render(task=task)
```
underdarknl, line 469: I feel like using Jinja here is probably too much of a good thing. Woudn't fstring be more than capable enough?

jpbruinsslot: I think the main reason was to give more flexibility to the user to construct a rate limit identifier that can use the information embedded in the task.

@ammar92 am I correct in that assumption?

jpbruinsslot:
Superseded by #4126

PR #1615, opened 2023-08-15
Updated cryptography

ammar92:
### Changes
Updates `cryptography` package.


dekkers (dismissed):

rocky/pyproject.toml
```diff
@@ -53,6 +53,7 @@ opentelemetry-exporter-otlp-proto-grpc = "^1.18.0"
 opentelemetry-instrumentation-django = "^0.39b0"
 opentelemetry-instrumentation-psycopg2 = "^0.39b0"
 opentelemetry-instrumentation-requests = "^0.39b0"
+cryptography = "^41.0.3"
```
praseodym, line 56: Do we really need to include it as an explicit dependency?

PR #1618, opened 2023-08-15
Update GitHub Actions

Darwinkel:
Closes https://github.com/minvws/nl-kat-coordination/pull/1599
Closes https://github.com/minvws/nl-kat-coordination/pull/1597
Closes https://github.com/minvws/nl-kat-coordination/pull/1596
Closes https://github.com/minvws/nl-kat-coordination/pull/1600


.github/workflows/boefjes_container_image.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         id: buildx
 
       - name: Login to GitHub Container Registry
-        uses: docker/login-action@v2
+        uses: docker/login-action@v2.2.0
```
dekkers, line 38: This will only cause work to bump version numbers. Given that this hasn't caused us any problems in the past and is the normal way to specify actions as far as I know, I think we should just keep `@v2`. I think dependabot only changed it because in another place we specified a specific version.

dekkers (approved):

PR #1627, opened 2023-08-16
Fix `poetry-dependencies` target in Makefile

praseodym:
macOS does not like newlines in the for loop, and needs `sed` to have an
(empty) extension for the backup file for the `-i` option.

~~Additionally, I think we can run `poetry lock --no-update` instead of
`poetry lock --check` to update the Poetry lockfile if required, but
without updating dependencies.~~

---

---
## Checklist for QA:


Darwinkel:
Nice! Using `--no-update` is fine with me. If it causes issues we can always revert it.

Unfortunately, and perhaps typically, making `sed` macOS-compatible breaks GNU `sed` :roll_eyes: 

image

Looks like we'll have to use something like this:


```
ifeq ($(UNAME),Darwin)
        sed -i '' '/--hash/d; s/ \\$$//' rocky/requirements-dev.txt
else
        sed -i '/--hash/d; s/ \\$$//' rocky/requirements-dev.txt
endif
```

praseodym:
> Nice! Using `--no-update` is fine with me. If it causes issues we can always revert it.

Looks like it already causes changes for me. `poetry lock --check -C rocky` says that the lockfile is fine, but `poetry lock --no-update -C rocky` creates a small change:

```diff
diff --git a/rocky/poetry.lock b/rocky/poetry.lock
index 4b9fc11f..5a5a1a70 100644
--- a/rocky/poetry.lock
+++ b/rocky/poetry.lock
@@ -2030,7 +2030,7 @@ name = "prompt-toolkit"
 version = "2.0.10"
 description = "Library for building powerful interactive command lines in Python"
 optional = false
-python-versions = ">=2.6,<3.0.dev0 || >=3.3.dev0"
+python-versions = ">=2.6,<3.0.0 || >=3.3.0"
 files = [
     {file = "prompt_toolkit-2.0.10-py2-none-any.whl", hash = "sha256:e7f8af9e3d70f514373bf41aa51bc33af12a6db3f71461ea47fea985defb2c31"},
     {file = "prompt_toolkit-2.0.10-py3-none-any.whl", hash = "sha256:46642344ce457641f28fc9d1c9ca939b63dadf8df128b86f1b9860e59c73a5e4"},
```

So I'll revert that change.

> Unfortunately, and perhaps typically, making `sed` macOS-compatible breaks GNU `sed` 🙄

There are [some options for portability](https://unix.stackexchange.com/questions/92895/how-can-i-achieve-portability-with-sed-i-in-place-editing) that should work, I think they are a bit nicer.

ammar92 (approved):
Nice one 👍 I was implementing and testing another approach based on using `find` and its `-exec/-execdir` option, which I think is a bit more convenient because it will work for every module with a `pyproject.toml` file (instead of manually listing them, which is why we probably forgot to add the `cveapi/` folder). I hadn't figured out what to do with `sed` yet, nice to see you got it working. Let's merge this and perhaps we can discuss my idea later.

Darwinkel (approved):

dekkers (approved):

Issue #1580, opened 2023-08-08
Further environment (documentation) improvements


Darwinkel:
This is a follow-up ticket to https://github.com/minvws/nl-kat-coordination/pull/1517. Remaining work:

- [x] Remove and clean up old documentation about environment settings
- [ ] #1830
- [x] <s>Perhaps some aesthetic changes (maybe a nice table instead of a list of variables?)</s>
  - [x] Current look is sufficient imho
- [x] Prepare upgrade/migration instructions for the release notes and the community chats
  - [x] Double-check required migrations/instructions for the Debian packages and environments
  - [x] https://github.com/minvws/nl-kat-coordination/pull/1636
  - [x] https://docs.openkat.nl/release_notes/1.11.html
- [x] Update community scripts (https://github.com/minvws/nl-kat-coordination/pull/1309) 
  - [x] https://github.com/minvws/nl-kat-coordination/pull/1655
- [x] Update example code in the docs
- [x] Make sure Mula's documentation is correctly updated as well (https://github.com/minvws/nl-kat-coordination/pull/1476/)

Darwinkel:
Closing for now in favour of follow-up ticket: https://github.com/minvws/nl-kat-coordination/issues/1830

PR #1655, opened 2023-08-23
Update install script with new env vars

Darwinkel:
### Changes
Minor changes to two required env variables in the openkat install script since https://github.com/minvws/nl-kat-coordination/pull/1517 was merged.

`SCHEDULER_DSP_BROKER_URL` has been unused for a year and has thus been removed from the script.

### Issue link
Related to https://github.com/minvws/nl-kat-coordination/issues/1580

### Proof
Should be a trivial change (name changes only).

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


dekkers:
I tested it on 1.11.0rc1 and it worked. I think it would be better to keep the `SCHEDULER_RABBITMQ_DSN` and `SCHEDULER_DB_DSN` lines so the script still works for 1.10 which is still the current version and the version it will download by default. If the config file doesn't have the variable sed will just do nothing, so having both in the script shouldn't be a problem.

Darwinkel:
> I tested it on 1.11.0rc1 and it worked. I think it would be better to keep the `SCHEDULER_RABBITMQ_DSN` and `SCHEDULER_DB_DSN` lines so the script still works for 1.10 which is still the current version and the version it will download by default. If the config file doesn't have the variable sed will just do nothing, so having both in the script shouldn't be a problem.

Ah yes, good point, hadn't thought of that. Will add them back in. Thanks!

dekkers (approved):

PR #1678, opened 2023-08-25
Fix restarting services

dekkers:
The boefje/normalizer runner didn't call sys.exit() in the signal handler so sending SIGTERM would never exit the main process.

Systemd also sends the SIGTERM to all processes. This confuses multiprocessing/gunicorn/etc. because all the subprocesses are killed while it also being killed itself. Changing `KillMode` to `mixed` means only the main process gets a SIGTERM so it can kill the subprocesses itself.

---

---
## Checklist for QA:


ammar92 (approved):
Nice find. No remarks

Darwinkel:
## Checklist for QA:

Don't have a suitable Debian environment at hand, but the changes to the service files look trivial enough.

### What works:
* No apparent regressions
* Workers report that they receive SIGTERM when `docker compose down` is called

### Bug or feature?:
* `docker compose down` doesn't appear to be much faster

underdarknl (approved):

PR #1688, opened 2023-08-29
:lock: hide sensitive secrets

zcrt:
### Changes
Adds functionality to hide sensitive secrets from the boefje detail page in the katalogus.

### Issue link
Was mentioned on Figma

### Proof
image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:

### DIFF
--- a/rocky/katalogus/templates/plugin_settings_list.html
+++ b/rocky/katalogus/templates/plugin_settings_list.html
@@ -22,11 +22,15 @@
                             <tr>
                                 <td>{{ setting.name }}</td>
                                 <td>
-                                    {% if setting.value is None %}
-                                        {% translate "Unset" %}
-                                    {% else %}
-                                        {{ setting.value }}
-                                    {% endif %}
+                                    {% with sname=setting.name.lower %}
+                                        {% if setting.value is None %}
+                                            {% translate "Unset" %}
+                                        {% elif "token" in sname or "key" in sname or "api" in sname or "secret" in sname %}
+                                            •••••••••••••
+                                        {% else %}
+                                            {{ setting.value }}
+                                        {% endif %}
+                                    {% endwith %}
                                 </td>
                                 <td>
                                     {% if setting.required %}


dekkers (changes requested):
I think we should mark environment settings that have secrets so that we can properly do this is in all places where it is necessary. Boefje task data will also show the environment and thus the name. We might show this is in more places. We shouldn't do guessing based on the name because that will give both false positives and false negatives. For example a setting `api_url` would also be masked with this PR and that would be wrong.

zcrt:
Agreed that would be a cleaner solution, although requiring significantly more work. However, I rather start hiding false positives such as `api_url` now than showing potential secrets in plain text in the interface. I would classify false positives as only a minor inconvenience in this case.

Alternatively, we can rename variables such as `SHODAN_API` to `SHODAN_API_KEY` since I think it would be relatively safe to assume `secret`, `token` and `key` to be sensitive information. (I would not solve all false positives and negatives, but would solve a large part of this problem.)

underdarknl:
Its a bit more work to signal which settings are secrets, and which are not an a per boefje basis, but I do want to do this via that route.
How we signal this specifically is to be discussed. In the current jsonschema there is room for 'required' fields, and I'd suggest we do the same for 'secret' fields, where we would just list all variables (env, parameters, arguments) by name that we don't want regular users to see in the user-interface.

zcrt:
I have modified the PR to include a specific property in the schema.json to be set to `true`.

rocky/katalogus/forms/plugin_settings.py
```diff
@@ -25,6 +25,7 @@ def populate_fields(self):
         for field_name, field_props in self.plugin_schema["properties"].items():
             kwargs = {
                 "required": "required" in self.plugin_schema and field_name in self.plugin_schema["required"],
+                "secret": "secret" in self.plugin_schema and field_name in self.plugin_schema["secret"],
```
dekkers, line 28: These are arguments to django form field object (https://docs.djangoproject.com/en/4.2/ref/forms/fields/), so adding "secret" here doesn't work. I think this is also the cause of the test errors.

I think what should be changed here is that if the value is a secret we don't set `kwargs["initial"]` (line 38) so we don't show the secret value in the form.

zcrt: @underdarknl can you look into this? I think this change was a result of the change to store the secrets in a list

underdarknl: Hm, yes. I did not see we where pushing these things into the django 'form-field' object, and instead expected it to be a regular object.
If that's the case we could set the disabled flag, and remove the value if the user does not have viewing rights, even further this does not need to be a form, if the user does not have the correct rights at all.
This also means we should change the template, as it does not render a form, but just  list of values, (where the secrets are replaced by something unreadable)

dekkers: I don't understand your comment. This is the form for adding and changing settings. Setting it to disabled or removing the form means it would be no longer possible to add or change the settings.

What should happen is that for secrets we don't show the value if you edit the settings. This was also not implemented before the settings were stored in a list. The suggestion I did to just remove the secret from the initial is not complete however. It doesn't work correctly when there are multiple fields and you don't want to change the secret value, because you would overwrite the secret field with an empty value when you save the form. We need to implement more logic in the form to handle that case and add units tests to make sure it works correctly.

underdarknl: Users who can edit boefje settings can see secrets.
Users who cannot see secrets also cannot edit settings.
Settings should by themselves be treated as single post calls including *all* fields described in the boefje schema, and can as such be validated against the schema when changed / created. There's a separate ticket for that. https://github.com/minvws/nl-kat-coordination/issues/447

dekkers: This PR hides the secret on the boefjes page for all users, also for users who can edit settings.

I also don't think it is a good idea to show secrets when a user has the permission to edit settings. This is also not what I am used to with other systems like secrets in github or gitlab. The secrets there are write only fields, so you can change it to a new value if you have the right permission, but never see the existing value.

zcrt: Alternatively, we could use the `writeOnly` keyword, which is part of the JSON Schema. For the boefjes page this can be used as indicator to hide the value. Maybe a separate PR is then required for the edit page. I agree that it would be good to hide the secrets, even from the accounts that have the rights to overwrite them.

dekkers: I discussed this with @underdarknl and agreed that we should do the edit page in a different PR if we want to do that. For now it means that if you have permission to edit settings you can also see the secrets. I removed the line in the form to fix the issue.

dekkers:
I tested the PR and found no issues.

dekkers (approved):

PR #1693, opened 2023-08-29
Add boefje name to Boefje Task

jpbruinsslot:
For the branch fix/task-list-plugin-name-change it was requested to add additional information to the `BoefjeTask` such that this could be relayed in the task list of rocky.


Rieven (approved):
This is working as intended

PR #1703, opened 2023-08-31
Add more descriptive error messages when pushing on scheduler queues

### LINKED ISSUE #1700: Extend `/queues/{queue}/push` endpoint with better error messages
For the `/queues/{queue}/push` endpoint we would like to see better error messages that e.g. indicate that the client has made a bad request or rescheduling is not allowed for a reason. This allows the requester to do better error handling (and for example, produce a nicer error output in the frontend)

The `detail` value in `{"detail": "..."}` should contain a meaningful message about why the request failed.

**Additional context**
For example, in #1689 assumptions are made about the possible error, which is not ideal.

jpbruinsslot:
### Changes

Added more descriptive error messages when pushing to the queue from the api. This allows the services that push to the scheduler queues to relay and/or interpret those messages. Additionally, http status codes have been updated to represent the actual error being produced.

### Issue link

Closes #1700


mula/scheduler/server/server.py
```diff
@@ -409,26 +409,26 @@ def push_queue(self, queue_id: str, item: models.PrioritizedItem) -> Any:
...
+        except queues.QueueFullError as exc_full:
+            raise fastapi.HTTPException(
+                status_code=429,
+                detail="queue is full",
+            ) from exc_full
         except queues.errors.NotAllowedError as exc_not_allowed:
             raise fastapi.HTTPException(
-                status_code=400,
-                detail="not allowed",
+                status_code=409,
```
ammar92, line 430: This is indeed the appropriate status code, but please choose the constant from `fastapi.status` instead

mula/scheduler/server/server.py
```diff
@@ -409,26 +409,26 @@ def push_queue(self, queue_id: str, item: models.PrioritizedItem) -> Any:
...
-                detail="queue is full",
-            ) from exc_full
         except ValueError as exc_value:
             raise fastapi.HTTPException(
                 status_code=400,
                 detail="invalid item",
             ) from exc_value
+        except queues.QueueFullError as exc_full:
+            raise fastapi.HTTPException(
+                status_code=429,
```
ammar92, line 425: Although it would be convenient to have this one, strictly speaking, 429s are used for rate-limiting requests. They can be applied to single endpoints or on a middleware layer (that e.g. rate limits based on a token or resource id). Having a full queue is more of a resource problem. I'll leave it up to you.

See also point below about status constants.

ammar92 (commented):
Nice and descriptive messages 👍 I only had a few remarks/ discussion points, but should be ready to merge soon

mula/scheduler/server/server.py
jpbruinsslot: Yeah, I noticed as well that it is used specifically for rate-limiting requests, it is a bit of a toss-up. My consideration was as follows: The HTTP 409 status code indicates that the request could not be completed due to a conflict with the current state of the target resource. In this case, the conflict arises from the fact that the item we're trying to push onto the queue is already present, and therefore the operation cannot be completed as requested. In that regard, it could be considered a rate-limiting problem, and we could even relay a rate-limiting header to the client who wants to wait for a set amount of time.

mula/scheduler/server/server.py
jpbruinsslot: Yeah will update it for all the endpoints

mula/scheduler/server/server.py
ammar92: In that case, it would indeed be more justified if we add a rate-limiting header

ammar92 (dismissed):
Looks good to me 👍

Donnype (approved):
Cool 👍

praseodym (approved):

underdarknl (approved):

PR #1773, opened 2023-09-14
Add support to set stalled tasks to failed in scheduler

### LINKED ISSUE #1766: In the scheduler set tasks that haven't got a status update (to complete, failed, cancelled) after a set time to failed
In the event that the task is being rescheduled and the task has not finished, remaining dispatched, and the grace period (24h) has passed we want to set the task as failed, such that the rescheduled task is able to reschedule the task again and that 'hanging tasks' are being cleaned.

jpbruinsslot:
### Changes

This PR adds support for marking the tasks that have stalled (tasks that have the status 'dispatched' for longer than the grace period) to failed.

### Issue link

Closes #1766


mula/scheduler/schedulers/boefje.py
```diff
@@ -413,6 +414,17 @@ def is_task_running(self, task: BoefjeTask) -> bool:
...
+        if (
+            task_db is not None
+            and task_db.status == TaskStatus.DISPATCHED
+            and (
+                task_db.modified_at is not None
+                and datetime.now(timezone.utc)
+                > task_db.modified_at + timedelta(seconds=self.ctx.config.pq_grace_period)
+            )
+        ):
+            raise TaskStalledError()
```
dekkers, line 426: If find this confusing. The function is changed to also be able to return whether a task is stalled, but instead of changing the function signature to reflect that it can return three different states an exception is thrown.

underdarknl: Im ok with this change. The task is running or not, that seems to make sense. A stalled task is however an unexpected situation, and as such warrants an exception in my book.

dekkers: It is just one of the things the scheduler needs to handle in the normal course of doing its job. It is not an unexpected situation like failing to connect to something, inconsistent state or a connection that gets reset.

In my opinion you create the most maintainable code over the long run when methods work how someone new to the code would expect them to work. If such a person would have to write some new code where there is a need to check if a task is still running, then the expectation would in my eyes be that `is_task_running` returns just that and not also return other states via exceptions that would have to be explicitely handled. And then if you don't handle the exception everything just works even in QA, but then you deploy in production and it fails after some time when it encouters a stalled task.

In my opinion the goal should not be that we just write code that works, but also code that can still be maintained years in the future, when we either have forgotten how we implemented things years earlier, or maybe we have moved on and other people are working on the project.

Darwinkel:
## Checklist for QA:

### What works:
* Dispatched tasks which are hanging (due to the boefje runner dying) are correctly set to failed after passing the grace period and/or restarting the scheduler
* There are no hanging Dispatched tasks on the entire task page
* No apparent regressions

dekkers (approved):

Issue #1183, opened 2023-06-14
uWSGI can cause bad gateway error when scaling down number of workers


dekkers:
We are currently using uWSGI in Rocky to handle requests. There seems to be a problem that when scales down the number of workers it might kill a worker that is still handling a request. In the uWSGI log you will see

```
Jun 14 14:30:57 kattest02 kat-rocky[14762]: [busyness] 10s average busyness is at 0%, cheap one of 2 running workers
Jun 14 14:31:05 kattest02 kat-rocky[14762]: worker 1 killed successfully (pid: 819791)
Jun 14 14:31:05 kattest02 kat-rocky[14762]: uWSGI worker 1 cheaped.
```

And there is a 502 in the nginx log:

```
[14/Jun/2023:14:31:05 +0200] "GET /en/test/ HTTP/2.0" 502 150 "https://kattest02.openkat.dev/en/test/kat-alogus/" "Mozilla/5.0 (X11; Linux x86_64; rv:109.0)  Gecko/20100101 Firefox/114.0"
```

There seems to be a problem with threading and/or cheaping as described in https://github.com/unbit/uwsgi/issues/1702 and https://github.com/unbit/uwsgi/issues/1804. We could disable both, but then there is not much advantage of using uWSGI compared to other servers like gunicorn and uWSGI is also in maintance-only mode / not much development is going on.

We could switch to gunicorn what we also use for the other services or maybe switch to granian which is a new Python http server in rust and seems to be faster than gunicorn: https://github.com/emmett-framework/granian

underdarknl:
related:
https://github.com/minvws/nl-kat-coordination/pull/1784

PR #1784, opened 2023-09-15
Add support for using granian instead of uWSGI

dekkers:
### Changes
This adds support for using granian (https://github.com/emmett-framework/granian). We keep uWSGI so people can change from uWSGI to granian and we don't break existing installations. New installation of the Debian package will already use granian.

Granian is relativly new python application server written is rust. This means it is both secure (compared to uWSGI that is written in C) and fast. It also seems to be a lot better maintained than uWSGI, with a responsive maintainer and regular releases.

### Issue link
Part of https://github.com/minvws/nl-kat-coordination/issues/1183

---

---
## Checklist for QA:


rocky/Dockerfile
praseodym, line 1: Don't we want to use Granian instead of uWSGI by default in the containers?

dekkers: Yes, but only after we have given users time to migrate. It is possible that people have configured uwsgi in the container using environment variables and just switching from uwsgi to granian would break that. Using granian is opt-in with this PR and the release after the next one we will switch to it.

ammar92 (approved):
Looks good to me. I haven't tried Granian yet, but seems promising and since there's a migration plan we should definitely try this. I hope this does allow us to write/ stream responses in chunks, which wasn't an option with uWSGI.

Darwinkel:
(QA is in progress. I've encountered some potential migration issues but we're double checking if that's an issue on my end or in Debian's upgrade path)

Darwinkel:
Did some pretty thorough testing and appears to work fine in both the Docker containers as well as the Debian packages (including migration paths). I did note, however, that the Debian variant no longer automatically provides https/8443 by default. This is fine in principle (people should probably use a proper reverse proxy anyway), but this does mean it is not a 1:1 drop-in replacement for our current uWSGI setup. We can choose to drop https/8443 support, but then we should very explicitly mention this in the release notes, as well as update our scripts and documentation where necessary.

dekkers:
Granian can't listen on both http/8000 and https/8443. I also think you should put a reverse proxy in front of OpenKAT, so I documented how to use setup Caddy as reverse proxy. I also documented how to configure https/8443.

docs/source/release_notes/1.13.rst
```diff
@@ -0,0 +1,67 @@
+============
+OpenKAT 1.13
+============
+
+We are replacing the uWSGI server in rocky with Granian. uWSGI doesn't seem to
+be well maintained anymore and we run into a bug that exists for a long time and
```
Darwinkel, line 6: (...) into a bug that _has existed_ (...)

docs/source/release_notes/1.13.rst
```diff
@@ -0,0 +1,67 @@
...
+doesn't get fixed. Granian is new application server that is better maintained
+and also has good performance. Granian is also written in Rust instead of C so
+should also have better security.
+
+If you have made any configuration changes or are using rocky on port 8443 you
+need to make sure to configure Granian correctly as it will listen on 127.0.0.1
+port 8000 by default. To not break anything in 1.13 we will support using both
```
Darwinkel, line 13: Maybe add explicitly that we listen on `0.0.0.0` in the old situation, which is no longer the case with Granian (if I remember correctly).

Darwinkel (approved):
Minor textual comments aside, looks good and ready for merge.

underdarknl (approved):

PR #1794, opened 2023-09-18
Backup scripts from TobiasBDO

ring-ring-ring:
Added a reference to the backupscripts from @tobiasBDO to the docs


underdarknl:
ideally, I'd have these in the scripts folder, or even added to the makefile. But linking to them works for now. it does only work on Dockerized setups though, so maybe we'd need to make that clear.
Clean Database backups in production should be performed by backing up the database that is used to store xtdb's indices and documents. (rocksdb, postgres of kafka)

underdarknl:
@dekkers you agree with the location, or do we want to merge them into the scripts folder?

underdarknl:
@tobiasBDO gave his blessing for us to create a PR with these scripts ourselves.

scripts/backup/create-volume-from-backup.sh
```diff
@@ -0,0 +1,90 @@
...
+uuid="$(cat /proc/sys/kernel/random/uuid)"
+cwd="$(pwd)"
+
+IMAGE=alpine:latest
+docker run \
+    --mount "type=volume,src=${volume_name},dst=/data" \
+    --name "$uuid" \
+    "$IMAGE"
+
+mkdir "/tmp/$uuid"
```
ammar92, line 84: Although the return UUID is unique, you could also use `mktemp -d -t $uuid` to create a unique directory; this will ensure the directory exists and may be a bit more convenient to use since you don't have to deal with path variables. Just a suggestion, I'll leave it up to you

scripts/backup/create-volume-from-backup.sh
```diff
@@ -0,0 +1,90 @@
...
+tar -xf "$backup_path/$volume/$snapshot" -C "/tmp/$uuid"
+cd "/tmp/$uuid"
+
+docker cp -a . "$uuid:/data"
+docker rm "$uuid"
+cd "$cwd"
```
ammar92, line 90: Is it necessary to go back to the previous working directory? Unless you explicitly allow to source this script rather than only execute it, I would leave it out

docs/source/technical_design/scripts.rst
```diff
@@ -82,3 +82,87 @@ Empty queue
 `openkat-empty-job-queue.sh
 <https://raw.githubusercontent.com/minvws/nl-kat-coordination/main/scripts/installation/openkat-empty-job-queue.sh>`__
 stops your OpenKAT processes, empties the job queue and starts all processes.
+
+Backup
+------
+
+TobiasBDO contributed `two backup and restorescripts <https://github.com/tobiasBDO/backup-openkat/tree/master>`_ included as part of his answer to `the question how to backup XTDB <https://github.com/minvws/nl-kat-coordination/issues/1757>`_ properly.
```
ammar92, line 89:
```suggestion
TobiasBDO contributed `two backup and restore scripts <https://github.com/tobiasBDO/backup-openkat/tree/master>`_ included as part of his answer to `the question how to backup XTDB <https://github.com/minvws/nl-kat-coordination/issues/1757>`_ properly.
```

docs/source/technical_design/scripts.rst
```diff
@@ -82,3 +82,87 @@ Empty queue
...
+
+We have slightly adjusted the scripts to prevent some potential issues from occurring (globbing and word splitting: shellcheck SC2086) and to clear some things up in the code. These scripts are maintained on a ‘best effort’ basis, thus no guarantees are provided. You are responsible for your own backups, OpenKAT is not responsible nor liable in case your backups are broken.
+
+Below is a description on how these backup scripts can be used.
+
+How to backup your volume
+-------------------------
+
+In your OpenKAT directory go to the scripts/backup folder:
```
ammar92, line 98:
```suggestion
In your OpenKAT directory go to the ``scripts/backup`` folder:
```

docs/source/technical_design/scripts.rst
```diff
@@ -82,3 +82,87 @@ Empty queue
...
+Run the script with the chosen backup path:
+
+``$ sudo ./backup-volume.sh -p <backup_path>``
+
+This directory will contain multiple folders each containing the backup file for that specific docker container as archived files (.tar.gz). If you run the command again it will create new archived files into those subdirectories. Your old backup will remain, as each backup name contains the timestamp of moment of creation. An example of such a file is: ``2024-03-28_173258_nl-kat-coordination_bytes-data.tar.gz``.
+
+Restoring your docker volume
+----------------------------
+
+In your OpenKAT directory go to the scripts/backup folder:
```
ammar92, line 117:
```suggestion
In your OpenKAT directory go to the ``scripts/backup`` folder:
```

ammar92 (approved):
Looks good to me. Just a few suggestions, but consider it approved

scripts/backup/create-volume-from-backup.sh
stephanie0x00: True, will pick this up. Didn't want to change the script too much as it was written by someone else.

stephanie0x00:
@ammar92 Could you QA this for me please?

ammar92:
So I've been playing around with this, but I couldn't directly get this to work. So this is what I did, and some additional remarks:
- The script assumes that the volume names are prefixed with `nl-kat-`, but this is a dangerous assumption.
  - Volume names are generated using the directory name, in my case I've checked out the repository to a local folder called `openkat`, and therefore the volume names are prefixed with `openkat-`
  - There probably is some way using e.g. `dirname` or `basename` to get the name of the repository directory two folders up, or maybe even `docker compose ls --quiet`
  - The `name` label in the `docker volume ls` is a "contains" filter, so the `*` is not needed
- `/proc/sys/kernel/random/uuid` doesn't exist on all systems, a nicer approach would be to use `uuidgen` instead
- You might want to use a prefix before the UUID to prevent confusion if removal of the container goes wrong (e.g. `backup-$uuid`)
- A possible faster way for backup would be to mount the destination folder and directly archive the the data to the destination folder
- Now that I read it, I think these are better names:
  - `backup-volume.sh` -> `backup-volumes.sh`
  - `create-volume-from-backup.sh` -> `restore-volumes.sh`

underdarknl (approved):

Issue #1630, opened 2023-08-17
Add initial version of Docker boefjes runner


praseodym:
As designed in #1620.

PR #1811, opened 2023-09-20
Add container I/O API endpoints to boefjes runner

### LINKED ISSUE #1629: Add container I/O API endpoints to boefjes runner
As designed in #1620.

praseodym:
### Changes

Add container I/O API endpoints to boefjes runner. This will be by boefjes running in containers as part of the design in #1620. The design is edited to align the `boefje_meta` schema in the input with the existing object used in KAT.

The next step for the new boefjes runner is #1630, which uses these I/O API endpoints from Docker containers.

### Issue link

Closes #1629

### Proof
n/a

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


boefjes/boefjes/api.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+class File(BaseModel):
+    name: Optional[str]
+    content: str = Field(..., contentEncoding="base64")
+    tags: Optional[List[str]]
+
+    class Config:
+        schema_extra = {
+            "required": ["content", "tags"],
```
dekkers, line 79: Tags is defined as optional, but required here?

boefjes/boefjes/api.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def create_boefje_meta(task):
+    boefje = task.p_item.data.boefje
+    boefje_resource = local_repository.by_id(boefje.id)
+    env_keys = boefje_resource.environment_keys
+    environment = get_environment_settings(task.p_item.data, env_keys) if env_keys else {}
+
+    organization = task.p_item.data.organization
+    input_ooi_pk = task.p_item.data.input_ooi
```
dekkers, line 159: Now that it is changed to the same boefjemeta the variable can also be named `input_ooi`.

boefjes/boefjes/__main__.py
```diff
@@ -27,6 +27,12 @@ def cli(worker_type: str, log_level: str):
 
     queue = WorkerManager.Queue(worker_type)
     runtime = get_runtime_manager(settings, queue, log_level)
+
+    if worker_type == "boefje":
+        import boefjes.api
+
+        boefjes.api.run()
```
Donnype, line 34: I guess it will never happen that this returns without crashing, but does a return after this line make sense to ensure we don't end up starting a worker on line 36?

Donnype (commented):
I'm a bit late with this question as I've read the design docs before (but probably not that thoroughly), but what are the considerations regarding handing the containers a "callback/webhook" versus having the runners fetch data from the container?

The reason I ask is mainly that given that the runners have access to some orchestrator to run the container, it is likely that they have access to other orchestrator api's to communicate e.g. files from the container into your current environment. With callbacks your containers should always have access to the network of your Boefje API, which might not always be the case in more involved production setups (or with boefjes that are not allowed access to the internet for some reason?). We would perhaps have to think about proper authentication too?

I can see this model making sense with more asynchronous systems that have open API's, but since we also explicitly wait for the container to finish, isn't it easier to handle all handling or errors, the saving of data etc. for a task to bytes right away, so it stays in one place (i.e. the runner)? It would also save us an extra API to manage?

praseodym:
> I'm a bit late with this question as I've read the design docs before (but probably not that thoroughly), but what are the considerations regarding handing the containers a "callback/webhook" versus having the runners fetch data from the container?

Because fetching data from the container is an anti-pattern in orchestrators like Kubernetes. Using stdin/stdout has very few guarantees and is mostly meant for debugging. Copying files even more so, for example the `kubectl cp` to copy files from a container actually executes `tar` in the container using `kubectl exec`. There is also no guarantee that the container will be around when it's done, because after the container exits it is usually removed right away. Alternatively, if we keep the container alive, we'd need to signal the runner that the container is done so that files can be copied out before it is removed, but the signalling would require something like a REST API which brings us back to square one.

> The reason I ask is mainly that given that the runners have access to some orchestrator to run the container, it is likely that they have access to other orchestrator api's to communicate e.g. files from the container into your current environment. With callbacks your containers should always have access to the network of your Boefje API, which might not always be the case in more involved production setups (or with boefjes that are not allowed access to the internet for some reason?). We would perhaps have to think about proper authentication too?

In Kubernetes, [network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) can be used to allow/deny certain network connections to the boefje containers, for example allow boefjes to only call out to the internet and the boefje API. 

I don't see why we'd need authentication for the boefje API: the only 'malicious' thing that a boefje could do with that API is to call the endpoints with a different task id, but the task id being a uuid makes it impossible to guess.

> I can see this model making sense with more asynchronous systems that have open API's, but since we also explicitly wait for the container to finish, isn't it easier to handle all handling or errors, the saving of data etc. for a task to bytes right away, so it stays in one place (i.e. the runner)? It would also save us an extra API to manage?

The wait for the container to finish is actually the result of implementing the Docker runner within the current job runner. I'd much rather have written a single runner that can manage multiple containers in parallel, but because that is incompatible with the current Python boefjes that would mean the introduction of a completely new runner with its own scheduler queue which would then need to be ran in parallel to the current runner. That's a lot of extra complexity, I went for this implantation instead.

boefjes/boefjes/__main__.py
praseodym: This `run()` actually returns right away, because we run the API within the current Boefje runner.

.env-defaults
```diff
@@ -13,6 +13,7 @@ SCHEDULER_API=http://scheduler:8000
 KEIKO_API=http://keiko:8000
 KATALOGUS_API=http://katalogus:8000
 XTDB_URI=http://crux:3000
+BOEFJE_API=http://boefje
```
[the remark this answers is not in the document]
praseodym: It is consistent with the octopoes_api that runs on port 80 🙃but changed to port 8000 now

boefjes/boefjes/api.py
praseodym: Good point, it shouldn't be required as we have default tags (MIME types) for each boefje. I removed the `schema_extra.required` entirely because the typing (`Optional` or not) already covers what fields are required.

boefjes/boefjes/api.py
praseodym: Changed

boefjes/boefjes/api.py
```diff
@@ -0,0 +1,177 @@
...
+
+app = FastAPI()
+logger = logging.getLogger(__name__)
+local_repository = get_local_repository()
+scheduler_client = SchedulerAPIClient(settings.scheduler_api)
+bytes_api_client = BytesAPIClient(
+    settings.bytes_api,
+    username=settings.bytes_username,
+    password=settings.bytes_password,
+)
```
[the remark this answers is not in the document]
praseodym: Thanks, changed to use FastAPI dependency injection.

boefjes/tests/conftest.py
```diff
@@ -124,3 +126,18 @@ def manager(item_handler: MockHandler, tmp_path: Path) -> SchedulerWorkerManager
...
+    from boefjes.api import app
+
+    # test task without OOI to prevent call enrichment call to Octopoes
+    scheduler_client = MockSchedulerClient(
+        get_dummy_data("scheduler/queues_response.json"),
+        [get_dummy_data("scheduler/pop_response_boefje_no_ooi.json")],
+        [],
+        tmp_path / "patch_task_log",
+    )
+    boefjes.api.scheduler_client = scheduler_client
```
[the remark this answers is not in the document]
praseodym: Changed!

boefjes/boefjes/api.py
```diff
@@ -0,0 +1,182 @@
...
+    return "OK"
+
+
+@app.get("/api/v0/tasks/{task_id}", response_model=BoefjeInput)
+async def boefje_input(
+    task_id: str,
+    scheduler_client: SchedulerAPIClient = Depends(get_scheduler_client),
+    local_repository: LocalPluginRepository = Depends(get_local_repository),
+):
+    task = get_task(task_id, scheduler_client)
```
dekkers, line 99: I didn't realize this when doing the first review, but doing sync requests like this will block the asyncio loop as far as I understand it. I don't think this is a blocker, but we should probably make the scheduler client also async in the future.

boefjes/tests/conftest.py
```diff
@@ -124,3 +125,16 @@ def manager(item_handler: MockHandler, tmp_path: Path) -> SchedulerWorkerManager
...
+
+
+@pytest.fixture
+def api(tmp_path):
+    from boefjes.api import app
+
+    # test task without OOI to prevent call enrichment call to Octopoes
+    # def _mocked_scheduler_client():
+    #     return MockSchedulerClient(
+    #         tmp_path / "patch_task_log",
```
dekkers, line 137: Commented out code should be removed.

boefjes/boefjes/api.py
```diff
@@ -0,0 +1,182 @@
...
+def get_scheduler_client():
+    return SchedulerAPIClient(settings.scheduler_api)
+
+
+def get_bytes_client():
+    return BytesAPIClient(
+        settings.bytes_api,
+        username=settings.bytes_username,
+        password=settings.bytes_password,
+    )
```
dekkers, line 85: I don't like that we create a new object with every request. We can refactor to optimize this later. I think the best pattern for this would be defining the client objects as singleton in the files that define the classes.

Donnype (dismissed):

boefjes/tests/conftest.py
praseodym: Thanks, done

boefjes/boefjes/api.py
praseodym: This is indeed not ideal, but we'd have to create new versions of the scheduler, bytes, octopoes and katalogus API clients to make them async. If we refactor the current API clients, we'd also have to change all existing usages to await the async calls.

dekkers:
Looks good to me so I moved this to QA.

I think it makes more sense to do QA of this together with #1812 instead of trying to just do QA on the API implemented in this PR. So my proposal would be to only check if all the existing functionality still works and QA the new functionality together with QA of #1812.

Darwinkel:
## Checklist for QA:

### What works:
* Tested thoroughly, no apparent regressions
* Scheduling and running all sorts of boefjes and normalizers works as usual

dekkers (approved):

PR #1844, opened 2023-09-28
Delegate creation of `.po` files externally

Darwinkel:
During the experiments with Weblate (and with our `make lang` pipeline), we found that we have a significant amount of merge conflicts that stem either from:

- Weblate's PR attempting to make changes on `.po` files which have already been changed in the main repo; and/or
- A PR getting merge conflicts when trying to update the branch from `main`, after another recently merged PR touched the `.po` files.

With this PR, `make lang` will now only update the extracted source strings to an English `.pot` file. This means that the actual translation of this source file will be delegated to Weblate, which then also performs `msgmerge`. This should save us a significant amount of headaches. 

Note that we may still encounter headaches if we update nl/it/pap `.po` files manually through a PR. But since that is also the case right now, I guess this is still an improvement.


.github/workflows/rocky_makelang.yml
```diff
@@ -36,9 +36,14 @@ jobs:
...
-      - name: Generate the .po files if they changed
+      - name: Generate the .pot file if source strings changed
         run: SECRET_KEY="whatever" make lang
         working-directory: ./rocky
 
-      - name: Check if there are any changed files
-        run: git diff --exit-code
+      - name: Check if at least one source string changed
+        run: |
+          changed_lines=$(git diff --unified=0 rocky/rocky/locale/django.pot | grep -o '^[-+]' | wc -l)
```
ammar92, line 45: This is a bit hard to understand, you could perhaps also grep against strings that match `msgid` or `msgstr`?

rocky/Makefile
```diff
@@ -54,8 +54,9 @@ export_migrations:
 	python manage.py export_migrations accounts 0001
 
 languages:
-	# Extracts strings to .po files which should be translated
-	python manage.py makemessages -i "venv/*" -i "build/*" -i "octopoes/*" -i "node_modules/*" --verbosity 2 --add-location file -a ; \
+	# Extracts strings to `.pot` file which should be translated
+  # Note that the creation of `.po` files is delegated to another tool (Weblate)
+	python manage.py makemessages -i "venv/*" -i "build/*" -i "octopoes/*" -i "node_modules/*" --verbosity 2 --add-location file -a --keep-pot
```
ammar92, line 59: I think you can also run this only on `rocky/locale/` or run make messages from within that directory (using `manage.py` or `django-admin` command). This way you can get rid of all the exclude options

Darwinkel: Tested this, but it's a limitation of makemessages sadly :/

.github/workflows/rocky_makelang.yml
Darwinkel: I've updated the regex, should be more explicit now

underdarknl:
We should really get our hosting agreement fixed with weblate before we do this, as currently our project there is in maintenance mode, and I expect this to be because of the license-limbo.

Darwinkel:
> We should really get our hosting agreement fixed with weblate before we do this, as currently our project there is in maintenance mode, and I expect this to be because of the license-limbo.

I put the project in maintenance mode to prevent merge conflicts until this PR is merged :)

dekkers (approved):

Issue #1768, opened 2023-09-13
Show more task related metrics for scheduler


jpbruinsslot:
Create endpoint that gives overview of task metrics over the past 24 hours. That can be rendered in the front end. Additionally, we want this information to be available  at the metrics endpoint.

jpbruinsslot:
### Backgrounnd

This issue has 2 parts:
1. statistics / aggregation endpoint that can be referenced by rocky
2. task / queue statistics added to the metrics endpoint to be referenced by prometheus

Writing the queries for 1. can also re-used for point 2.

1. How many items are currently on the queue for a scheduler (`qsize`)
2. Aggregation on date range with buckets of every hour for every task status how many items are there

### Discussion

By the nature of how tasks are persisted in the scheduler. Any information on state changes of a task is lost when it's updated. Specifically, the nature of how tasks are managed can result in the loss of information regarding their state changes, making it difficult to perform aggregations over date ranges. This would require us to keep track of those events. 

To address this issue and answer questions such as the current state of the scheduler, the number of items in the queue, the count of schedulers in operation, and the distribution of tasks by their status, we can do this on a request basis, at that moment in time. This information can then be referenced by Rocky. Additionally, we can expose this data via the metrics endpoint that is compatible with Prometheus so this can be leveraged for aggregations on date ranges.

jpbruinsslot:
@underdarknl

underdarknl:
Could you do a Group by on the last update date grouped by hour?
Then we'd be able to see for each of the last 24 hours (eg, 1 default cycle of the scheduler)  how many tasks ended up in the various status groups per hour. Obviously this result would have more pending jobs in the most recent hour, which as you mention would then turn into completed or failed jobs if the user would refresh.

jpbruinsslot:
Metrics endpoint response

```javascript
# HELP app_settings_info Scheduler configuration settings
# TYPE app_settings_info gauge
app_settings_info{katalogus_cache_ttl="30",monitor_organisations_interval="60",pq_grace_period="86400",pq_max_random_objects="50",pq_maxsize="1000"} 1.0
# HELP scheduler_qsize Size of the scheduler queue
# TYPE scheduler_qsize gauge
scheduler_qsize{scheduler_id="boefje-dev"} 0.0
scheduler_qsize{scheduler_id="normalizer-dev"} 0.0
# HELP scheduler_task_status_counts Total number of tasks
# TYPE scheduler_task_status_counts gauge
[...]

PR #1846, opened 2023-09-28
Create task stats endpoint for scheduler

jpbruinsslot:
### Changes

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/1768

### DIFF
--- a/mula/scheduler/server/server.py
+++ b/mula/scheduler/server/server.py
@@ -130,6 +130,20 @@ class Server:
         )
 
         self.api.add_api_route(
+            path="/tasks/stats",
+            endpoint=self.get_task_stats,
+            methods=["GET"],
+            status_code=status.HTTP_200_OK,
+        )
+
+        self.api.add_api_route(
+            path="/tasks/stats/{scheduler_id}",
+            endpoint=self.get_task_stats,
+            methods=["GET"],
+            status_code=status.HTTP_200_OK,
+        )
+
+        self.api.add_api_route(
             path="/tasks/{task_id}",
             endpoint=self.get_task,
             methods=["GET"],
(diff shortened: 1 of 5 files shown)


mula/scheduler/server/server.py
```diff
@@ -350,6 +364,18 @@ def patch_task(self, task_id: str, item: Dict) -> Any:
 
         return updated_task
 
+    def get_task_stats(self, scheduler_id: Optional[str] = None) -> Any:
```
ammar92, line 367:
```suggestion
    def get_task_stats(self, scheduler_id: Optional[str] = None) -> Optional[Dict[str, Dict[str, int]]]:
```

mula/tests/utils/functions.py
```diff
@@ -35,3 +35,9 @@ def create_task(p_item: models.PrioritizedItem) -> models.Task:
         p_item=p_item,
         status=models.TaskStatus.QUEUED,
     )
+
+
+def compile_query(query):
```
ammar92, line 40:
```suggestion
def compile_query(query: Query) -> str:
```

mula/tests/integration/test_task_store.py
```diff
@@ -0,0 +1,196 @@
...
+                id=p_item.id,
+                hash=p_item.hash,
+                type=functions.TestModel.type,
+                scheduler_id=p_item.scheduler_id,
+                p_item=p_item,
+                status=models.TaskStatus.DISPATCHED,
+                modified_at=datetime.now(timezone.utc) - timedelta(hours=23),
+            )
+
+            self.mock_ctx.datastores.task_store.create_task(task)
```
ammar92, line 104: I think this piece can be optimized a bit. The only difference here I see are the ranges, `status` and `modified_at` field. So you could loop over e.g.:
```python3
for r, status, modified_at in zip((range(0, 2), range(2, 4), ...), (TaskStatus.COMPLETED, TaskStatus.FAILED, ...), (some time delta, some time delta, ...)): ...
```
(this usually looks a bit better when using pytests parametrize utils)
It's also better to use constants for the `timedelta` objects, since calling `datetime.now()` on different time timestamps could produce different results (the above example already does that)

ammar92 (commented):
Looks good to me! Just a few remarks to improve it even more. Let me know what you think

Darwinkel:
## Checklist for QA:

### What works:
* Enabling/disabling the metrics through `.env`
* `/tasks/stats`
* `/tasks/stats/<scheduler id>`
* `/metrics`
* No apparent regressions

underdarknl (approved):

PR #1877, opened 2023-10-05
Add more traces to scheduler

jpbruinsslot:
### Changes

Adding more traces to the scheduler to increase observability.

### DIFF
--- a/mula/scheduler/app.py
+++ b/mula/scheduler/app.py
@@ -169,6 +169,7 @@ class App:
                 additions,
             )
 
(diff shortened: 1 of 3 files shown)


underdarknl:
Looks good, did we also already have traces on updates from the katalogus, or session refreshes / logins to other services like bytes?

jpbruinsslot:
Yeah bytes and katalogus both have traces for their endpoints enabled

dekkers:
Looks good to me, I don't think there is anything to QA here?

Issue #1884, opened 2023-10-09
Scheduler not able to reach katalogus


jpbruinsslot:
`[WARNING] [Katalogus] Exception: Expecting value: line 1 column 1 (char 0)`

is the result of the scheduler getting a 404 response on the health endpoint of  the katalogus, and is not able to parse the contents of that response. The 404 is the result of  `urljoin` having unexpected results when joining urls that contain a path.

```
>>> urljoin("http://katalogus:8000/katalogus", "health")
'http://katalogus:8000/health'

>>> urljoin("http://katalogus:8000/katalogus/", "health")
'http://katalogus:8000/katalogus/health'

>>> urljoin("http://katalogus:8000/katalogus/", "/health")
'http://katalogus:8000/health'
```

jpbruinsslot:
Additionally the `AnyHttpUrl` from pydantic adds a `/` when you enter a host name, e.g. `http://localhost:8000` it changes it to `http://localhost:8000/`, however if you enter a path is removes the trailing slash. `http://localhost:8000/path/` becomes `http://localhost:8000/path`

Darwinkel:
## Checklist for QA:
- [x] I have checked out this branch, and successfully ran a fresh `make reset`.
- [x] I confirmed that there are no unintended functional regressions in this branch:
    - [x] I have managed to pass the onboarding flow
    - [x] Objects and Findings are created properly
    - [x] Tasks are created and completed properly
- [ ] I confirmed that the PR's advertised `feature` or `hotfix` works as intended.

I tried very hard to replicate the situation from the issue, but I couldn't get it to work even after changing some endpoint prefixes in the katalogus' router. Regardless, the PR doesn't seem to introduce any regressions either. I think we can merge this and await feedback from the original reporter.

PR #1886, opened 2023-10-09
Normalize host urls in scheduler v1.12

jpbruinsslot:
### Changes

Fixes how configurable host urls are handled in the scheduler. Trailing slashes are removed and formatted strings are used instead of `urljoin()`.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/1884

### DIFF
--- a/mula/scheduler/context/context.py
+++ b/mula/scheduler/context/context.py
@@ -42,20 +42,20 @@ class AppContext:
 
         # Services
         katalogus_service = services.Katalogus(
-            host=str(self.config.host_katalogus),
+            host=self._remove_trailing_slash(str(self.config.host_katalogus)),
             source=f"scheduler/{scheduler.__version__}",
             cache_ttl=self.config.katalogus_cache_ttl,
         )
 
         bytes_service = services.Bytes(
-            host=str(self.config.host_bytes),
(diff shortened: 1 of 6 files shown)


mula/scheduler/context/context.py
```diff
@@ -103,3 +103,9 @@ def __init__(self) -> None:
             registry=self.metrics_registry,
             labelnames=["scheduler_id"],
         )
+
+    def _remove_trailing_slash(self, url: str) -> str:
```
ammar92, line 107: Two things:
- This is a more common "problem" and I'd expect this in some kind of utils module rather than this class specifically
- It could've been a (static) function

ammar92 (approved):
One remark tiny remark, but considered it approved as it does the job for now. Since I'm also dealing with this with the Pydantic package upgrade, I propose to come up with a more common solution together soon?

Darwinkel:
## Checklist for QA:

I tried very hard to replicate the situation from the issue, but I couldn't get it to work even after changing some endpoint prefixes in the katalogus' router. Regardless, the PR doesn't seem to introduce any regressions either. I think we can merge this and await feedback from the original reporter.

underdarknl (approved):

PR #1888, opened 2023-10-09
First draft of cookie crawler "lapje"

floort:


boefjes/images/lapje/sitecrawler/crawloutput/2023-10-08_134122.mproxy
underdarknl, line 1: Is this file included by accident?

boefjes/images/lapje/sitecrawler/main.go
```diff
@@ -0,0 +1,71 @@
...
+
+	conf.Logger.Println("Reading crawl list")
+	if err := conf.ReadUrls(*crawlUrl); err != nil {
+		panic(err)
+	}
+
+	// Start proxy
+	conf.Logger.Println("Starting mitmproxy")
+	network_dumpfile := path.Join(conf.OutputDir, "network.mproxy")
+	proxy := exec.Command("mitmdump", "-p", fmt.Sprintf("%d", conf.Port), "-w", network_dumpfile)
```
underdarknl, line 57: I believe a HAR file contains more raw data that we could use / normalizer, Its still not completely implemented in mitmproxy, but seems more complete than the mproxy output file.
see: https://docs.mitmproxy.org/stable/concepts-options/#hardump
https://www.mitmproxy.org/posts/har-support/

boefjes/boefjes/plugins/kat_lapje/main.py
```diff
@@ -0,0 +1,48 @@
...
+from typing import List, Tuple, Union
+
+import docker
+
+from boefjes.job_models import BoefjeMeta
+from boefjes.plugins.helpers import get_file_from_container
+
+logger = logging.getLogger(__name__)
+
+OCI_IMAGE = "lapje"
```
Donnype, line 11: Looking at this tag versus e.g. the masscan boefje and its custom github workflow: I suppose this should just work by building the image with the `lapje` tag locally?

UPDATE:
Yep this works:
```shell
docker build -t lapje boefjes/images/lapje
```

image

Donnype (approved):
I haven't looked at the Go code, but given that this tool has been used already is probably a good indicator it works properly!

Darwinkel:
## Checklist for QA:

I just want to say I think it's awesome to see that the work we did for the containerized boefjes and the the new report generation is paying off! :tada: 

### What works:
- Boefje and normalizer output looks good
- Report generation works as intended
- Tested on a bunch of ugly websites, looks like the retrieved information is correct (plus no crashes)

### What doesn't work:
- Image does not exist yet and requires running `docker build -t lapje boefjes/images/lapje`. I think we should modify and generalize masscan's workflow before we merge this (shouldn't be to much work). 

- Minor overflow bug on the reports page when dealing with long cookie values:
(not a bug introduced by this PR though, we should fix this in a separate PR)
image


### Bug or feature?:
- Chrome is currently not pinned. To prevent unexpected regressions like the ones we've seen in the past, we probably should. I don't think this is possible through Google's repository though. Why don't we use Debian's chromium package?

- From the report page, it is not entirely clear why some records have no values/content, and only a hostname. Some more info (maybe in Lapje's katalogus detail page) about what Lapje does exactly, and how its objects should be interpreted, would be nice.
image

boefjes/boefjes/plugins/kat_lapje/main.py
```diff
@@ -0,0 +1,48 @@
...
+        logging.debug("Container output: %s", line)
+
+    output = get_file_from_container(container, "/app/output/dump.har")
+
+    try:
+        container.remove()
+    except Exception as e:
+        logging.warning(e)
+
+    return [(set("har"), output)]
```
Donnype, line 48: Just noticed, but this does not work as it results in the set `{'h','a','r'}`

rocky/reports/report_types/crawl_report/report.html
```diff
@@ -0,0 +1,50 @@
+{% load i18n %}
+
+<h3>Crawl information</h3>
+{% if data.hostnames_and_cookies %}
+    <table>
```
Rieven, line 5: Can you put an extra div wrapping the table?
`<div class="horizontal-scroll">`

When the values are too long it will overflow the whole page instead of the table. We have to fix this also for DNS and TLS reports.

boefjes/boefjes/plugins/kat_lapje/main.py
underdarknl: correct, it needs an extra ,

underdarknl: We should harden against this in the runner to warn met single char mimetypes are returned.

dekkers:
I renamed the boefje and normalizer.

> ### What doesn't work:
> 
>     * Image does not exist yet and requires running `docker build -t lapje boefjes/images/lapje`. I think we should modify and generalize masscan's workflow before we merge this (shouldn't be to much work).

I added the github workflow, but that currently fails because external contributors PRs can't push to the container registry. Should be fixed when the PR is merged.

>     * Minor overflow bug on the reports page when dealing with long cookie values:
>       (not a bug introduced by this PR though, we should fix this in a separate PR)

This has been fixed.

> ### Bug or feature?:
> 
>     * Chrome is currently not pinned. To prevent unexpected regressions like the ones we've seen in the past, we probably should. I don't think this is possible through Google's repository though. Why don't we use Debian's chromium package?

Chrome is used because we are not sure that Debian's chromium doesn't have any modifications that impact the results or might get those in the future. I think getting accurate results using the browser that most people use is more important than the risk of breakage from newer versions.

>     * From the report page, it is not entirely clear why some records have no values/content, and only a hostname. Some more info (maybe in Lapje's katalogus detail page) about what Lapje does exactly, and how its objects should be interpreted, would be nice.

The records have no value because they are external requests that don't set a cookie. The fact that there is an external request is already useful information.

underdarknl:
Multiple pages inside a HAR file can be dealt with in a separate issue:
https://github.com/minvws/nl-kat-coordination/issues/1996

boefjes/boefjes/plugins/kat_chrome_mitmproxy/main.py
```diff
@@ -0,0 +1,48 @@
...
+from typing import List, Tuple, Union
+
+import docker
+
+from boefjes.job_models import BoefjeMeta
+from boefjes.plugins.helpers import get_file_from_container
+
+logger = logging.getLogger(__name__)
+
+OCI_IMAGE = "ghcr.io/minvws/nl-kat-chrome-crawler-mitmproxy:latest"
```
underdarknl, line 11: Could we pin this to a version once we have this PR merged, and a version build and uploaded? While I understand the flexibility in building the image from the latest upstream chome version, I'd want to avoid a situation where the day to day usage of a boefje changes based on a new version of the used container if we can.

underdarknl:
Lets create a HAR file boefje from this PR, and move the normalisers to a separate issues / PR once we figure out what kind of objects we want in the graph (specifically, how fine grained we want the cookies)

Issue #1685, opened 2023-08-28
Add more tests for scheduler: external services


jpbruinsslot:
Integrated test with the external services that the scheduler communicates with. First up: bytes

jpbruinsslot:
There is something weird that the test organisations are getting created, when in the tests I haven't explicitly asked to create those organisation in the katalogus. It isn't a mocking issue, since the organisations are actually created in the running katalogus container.

jpbruinsslot:
factory boy has a sqlalchemy integration where it seems that calling `OrganisationFactory()` would create the organisations. However it is strange that it would then create this in the katalogus db, since it is called from the scheduler that has connections to its own database.

PR #1891, opened 2023-10-10
Add more tests for scheduler: external services

jpbruinsslot:
### Changes

Adding test cases that specifically test the integrations with external service api's.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/1685


mula/.ci/docker-compose.yml
```diff
@@ -24,20 +24,23 @@ services:
...
+      - ci_bytes
+      - ci_katalogus
+      - ci_rabbitmq
 
   ci_postgres:
     image: postgres:15
     healthcheck:
-      test: ["CMD", "pg_isready"]
+      test: ["CMD", "gosu", "postgres", "pg_isready"]
       interval: 10s
```
ammar92, line 36: Since the timeout is removed, I suggest using a lower interval (with a possibly higher retry count)

mula/.ci/docker-compose.yml
```diff
@@ -24,20 +24,23 @@ services:
...
-      timeout: 5s
-      retries: 5
+      retries: 10
     env_file:
       - .ci/.env.test
+    environment:
+      APPS: "BYTES KATALOGUS SCHEDULER"
+    volumes:
+      - ../init-user-db.sh:/docker-entrypoint-initdb.d/init-user-db.sh
```
ammar92, line 43: If I'm not mistaken it should be one directory up?

mula/.ci/docker-compose.yml
```diff
@@ -47,3 +50,44 @@ services:
...
+        condition: service_healthy
+
+  ci_katalogus:
+    build:
+      context: ..
+      dockerfile: boefjes/Dockerfile
+      target: dev
+      args:
+        ENVIRONMENT: dev
+    command: uvicorn boefjes.katalogus.api:app --host 0.0.0.0 --port 8000 --reload --reload-dir /app/boefjes/boefjes/katalogus
```
ammar92, line 80: - `8000` is already the default port, which makes the option unnecessary
- The `--reload` and `--reload-dir` options don't make sense within this (pipeline) context and potentially slow down this service while being tested

```suggestion
    command: uvicorn boefjes.katalogus.api:app --host 0.0.0.0
```

mula/scheduler/context/context.py
```diff
@@ -42,20 +43,20 @@ def __init__(self) -> None:
 
         # Services
         katalogus_service = services.Katalogus(
-            host=self._remove_trailing_slash(str(self.config.host_katalogus)),
+            host=remove_trailing_slash(str(self.config.host_katalogus)),
```
ammar92, line 46: I recommend considering the new Pydantic v2 branch for my proposed solution. In this implementation, clients take on the responsibility of handling base URL manipulation by stripping or preparing it. This approach reduces verbosity, as the callee is relieved of the task of dealing with these details. The expectation is that clients will automatically append endpoints to base URLs, assuming they lack a trailing slash. This simplifies the caller's code, making it more intuitive and concise.

mula/tests/integration/test_services.py
```diff
@@ -0,0 +1,194 @@
...
+
+from scheduler import config, models, storage
+from scheduler.connectors import services
+from scheduler.utils import remove_trailing_slash
+
+from tests.factories import PluginFactory
+
+
+class BytesTestCase(unittest.TestCase):
+    def setUp(self) -> None:
```
ammar92, line 14: Since there's no teardown of the tests (mainly "logging" the user out or clearing the session), I assume creating the new settings and client instance during test setup results in new login sessions?

mula/tests/integration/test_services.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def test_expired_token_refresh(self):
+        self.service_bytes.get_last_run_boefje(
+            boefje_id="boefje-1",
+            input_ooi="ooi-1",
+            organization_id="org-1",
+        )
+        initial_token = copy.deepcopy(self.service_bytes.headers.get("Authorization"))
+
+        time.sleep(70)
```
ammar92, line 37: This is quite some time that gets wasted, is there any possible solution to avoid this sleep?

mula/tests/integration/test_services.py
```diff
@@ -0,0 +1,194 @@
...
+
+        self.service_bytes.get_last_run_boefje(
+            boefje_id="boefje-1",
+            input_ooi="ooi-1",
+            organization_id="org-1",
+        )
+        refresh_token = copy.deepcopy(self.service_bytes.headers.get("Authorization"))
```
ammar92, line 44: I'm a bit surprised that we need a deep copy for this. In what cases are different calls to `id(self.service_bytes.headers.get("Authorization"))` result in the same references when the value changes?

mula/tests/integration/test_services.py
```diff
@@ -0,0 +1,194 @@
...
+            models.Organisation(id="org-1", name="org-1"),
+            models.Organisation(id="org-2", name="org-2"),
+        ]
+
+        # Act
+        self.service_katalogus.flush_organisations_plugin_cache()
+
+        # Assert
+        self.assertEqual(len(self.service_katalogus.organisations_plugin_cache), 2)
+        self.assertIsNotNone(self.service_katalogus.organisations_plugin_cache.get("org-1"))
```
ammar92, line 79: I think it's better to test something like: `self.assertCountEqual(self.service_....cache.keys(), ("org-1", "org-2"))` (untested pseudocode)

This way you test the content and number of elements.

mula/.ci/.env.test
```diff
@@ -1,22 +1,43 @@
...
 SCHEDULER_API_PORT="8000"
 SCHEDULER_PQ_MAXSIZE=0
 SCHEDULER_PQ_POPULATE_INTERVAL=60
 SCHEDULER_PQ_POPULATE_GRACE_PERIOD=1
-SCHEDULER_DB_USER=postgres
-SCHEDULER_DB_PASSWORD=mysecretpassword
+SCHEDULER_DB_USER=scheduler_app
+SCHEDULER_DB_PASSWORD=${POSTGRES_PASSWORD}
 SCHEDULER_DB=scheduler
 SCHEDULER_DB_URI=postgresql://${SCHEDULER_DB_USER}:${SCHEDULER_DB_PASSWORD}@ci_postgres:5432/${SCHEDULER_DB}
```
ammar92, line 21: Does this still work if `SCHEDULER_DB_USER` and `SCHEDULER_DB_PASSWORD` are removed?

ammar92 (commented):
Looks good in general, just a few remarks to consider

mula/.ci/docker-compose.yml
jpbruinsslot: I think since the make file is being called from the mula folder and as such the docker-compose command from that context it needs to go one folder up in order to add the init-user-db.sh script

mula/scheduler/context/context.py
jpbruinsslot: Would it be advisable to implement that change during this pr or after it has been merged?

mula/tests/integration/test_services.py
jpbruinsslot: Indeed, have to think about how to do this with an integration with bytes. Since bytes has a minimum token expiration of 1 minute, in order to hit that I have to wait for at least a minute

mula/tests/integration/test_services.py
jpbruinsslot: Yeah `id()` is the better choice here

mula/scheduler/context/context.py
ammar92: It doesn't matter too much, but I think it's preferable to align as soon as possible to keep the consistency

mula/tests/integration/test_services.py
jpbruinsslot: Yes, that was my thinking as well

mula/tests/integration/test_services.py
jpbruinsslot: For now, I will skip the unit test and request a feature for bytes to be able to set the token expiration to that less of a minute

PR #1893, opened 2023-10-10
can_scan cannot be called

Rieven:
### Changes
can_scan cannot be accessed by class object Boefje in this case

### DIFF
--- a/rocky/katalogus/client.py
+++ b/rocky/katalogus/client.py
@@ -38,6 +38,7 @@ class Plugin(BaseModel):
         plugin_dict["produces"] = {ooi_class.get_ooi_type() for ooi_class in plugin_dict["produces"]}
         return plugin_dict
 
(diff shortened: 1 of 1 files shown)


rocky/katalogus/client.py
```diff
@@ -38,6 +38,7 @@ def dict(self, *args, **kwargs):
         plugin_dict["produces"] = {ooi_class.get_ooi_type() for ooi_class in plugin_dict["produces"]}
         return plugin_dict
 
+    @classmethod
```
dekkers, line 41: This should actually be a static method, not a class method, because it doesn't use the class object in the method. Changing it to staticmethod also means you need to remove the unused self argument from the def.

Rieven: I could not reproduce this, I think it was like a difference in versions of code

PR #1896, opened 2023-10-10
Set correct headers and increase timeout for scheduler

jpbruinsslot:
### Changes

Set correct headers and increase timeout for scheduler


mula/scheduler/connectors/services/bytes.py
```diff
@@ -63,7 +63,7 @@ def _get_token(self) -> str:
         response = self.post(
             url=url,
             payload=self.credentials,
-            headers={"Content-Type": "application/x-www-form-urlendcoded"},
+            headers={"Content-Type": "application/x-www-form-urlencoded"},
```
ammar92, line 66: When you use a dict-like object for the `data` argument in `requests`, it will automatically set this header.

```suggestion
```

mula/scheduler/connectors/services/services.py
```diff
@@ -100,9 +100,12 @@ def get(
         Returns:
             A request.Response object
         """
+        if headers:
+            self.headers.update(headers)
```
ammar92, line 104: Personally I think this is an unwanted side effect. This will change the header of every next request? If this is what you want, why not make the callee change the public property `headers` instead?

mula/scheduler/connectors/services/services.py
```diff
@@ -134,9 +137,12 @@ def post(
...
```
ammar92, line 141: See previous point

ammar92 (commented):
Just a few remarks

PR #1902, opened 2023-10-12
CVE-2023-44487 boefje

ammar92:
### Changes
Boefje and normalizer for CVE-2023-44487

### Proof
image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


dekkers (commented):
This code will create a CVE finding for every HTTP/2 server. I don't think this is very useful. As far as I understand it every compliant HTTP/2 support RST_STREAM because that is part of the HTTP/2 specification. `SAFE` is only returned when the server doesn't support HTTP/2, and `LIKELY` and `POSSIBLE` are returned when an error is encountered.

The readme of the original exploit has also been updated to reflect this, see also https://github.com/bcdannyboy/CVE-2023-44487/issues/7#issuecomment-1760265385 and https://github.com/bcdannyboy/CVE-2023-44487/issues/1#issuecomment-1760268937

Basically the code checks whether the server is susceptible to attack, e.g. if it implements RST_STREAM as specified in the HTTP/2 spec. It doesn't check how it handles a larger number of RST_STREAM requests and I don't think we can check this without actually doing the DoS attack.

I don't think we should create a CVE finding for every HTTP/2 server and keep that finding forever even if the server is patched or was never vulnerable.

underdarknl:
@dekkers correct, that is what it does for now, and I'd rather *have* the warning, than not have the warning. Once we get better signatures available, or once we get a more precise test (without hammering the server until the boefje trips a firewall rule) this is an OK path to follow for a zero day.  Obviously, we could also wait till everyone has patched their servers, and there's a surgical test available, but that would be more or less a meaningless addition to OpenKAT at that point.
Tldr, this boefje will get improved capabilities once the community at large developers better tests / signatures.

dekkers:
It is currently already meaningless in my opinion because it would just add a CVE finding to every web server that is not so ancient that it doesn't implement HTTP/2. And it doesn't add a warning, it adds a finding that the server is vulnerable to CVE-2023-44487. And we already know that this wrong for widely used webservers such as nginx and haproxy which don't seem to be impacted in the default configuration and for webservers that are patched. 

In my opinion we should aim for higher quality than this and we should make sure that we don't generate high number of false positives.

underdarknl:
Ok, then we'll not warn people.

PR #1925, opened 2023-10-17
Create general feature flag context processor

dekkers:
### Changes
The feature flag is currently broken because the template tag doesn't work as I thought it did and my code review feedback was not correct. I implemented it differently with a general context processor that can be used make a context variable of every Django setting that starts with `FEATURE_`

---

---
## Checklist for QA:


underdarknl:
Looks good to me.

rocky/tools/context_processors.py
```diff
@@ -4,6 +4,14 @@
 from rocky.version import __version__
 
 
+def feature_flags(request):
+    context = {}
+    for name in dir(settings):
+        if name.startswith("FEATURE_"):
+            context[name.lower()] = getattr(settings, name)
```
ammar92, line 11: I was just wondering why not just leave the variable/ context name as it is. I'd prefer to see these context vars (which are settings) unmodified in templates as constants. There's no real benefit to changing the case

ammar92 (commented):
Great addition 👍 Just one remark

rocky/tools/context_processors.py
underdarknl: Agreed, this would also make sure you could do the same for css classes without polluting the css namespace. Keeping them prefixed helps to keep the reasons clear.

dekkers: Because the convention is that context variables are lowercase and most of variables are that, but looking at the Django context processor I see that django isn't really consistent with that either:

https://docs.djangoproject.com/en/4.2/ref/templates/api/#built-in-template-context-processors

`debug` is lowercase for DEBUG, but the others are upper case. I'm going to change it to upper case because that's seems to be used most of the time.

ammar92 (approved):

PR #1930, opened 2023-10-18
Add mime-types to boefje definition

### LINKED ISSUE #1833: Boefjes should specify which raw file types it produces instead of OOIs
Currently boefjes don't specify which raw file types it produces. The boefje metadata lists OOIs in the produces field, but those aren't produced directly by the boefje, but by the normalizers. This makes it impossible to link to the normalizers that consume the raw files produced by the boefje on the boefje detail page. 

We should list the raw file types in boefje produces. We might also consider adding a check to make sure that a boefje can only return types that are listed in the metadata.

Donnype:
### Changes
Add a `mime_types` field to the boefje definition (JSON file and Katalogus API response). We could rename this back to `produces` with the follow up changes in Rocky, but the idea was to keep it backward compatible for now in case this PR grew too big because of many changes to `boefjes.json` files.

### Issue link
Closes #1833 

### Proof

image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


underdarknl:
nice. next step is removing the Produces field \o/

underdarknl:
Would now also be a good moment to drop the mime-types that are unprefixed? Eg, keep the `boefje/$name` variant, and remove the `$name` variant?  I feel we should have people target the `boefje/$name` variant from their normalizer as it is less ambiguous and keeps our databases cleaner.

boefjes/boefjes/katalogus/models.py
```diff
@@ -39,6 +39,7 @@ class Boefje(Plugin):
     scan_level: int = 1
     consumes: Set[str] = Field(default_factory=set)
     produces: List[str] = Field(default_factory=list)
+    mime_types: List[str] = Field(default_factory=list)
```
ammar92, line 42: I think this should be a set, as the elements are unique and order doesn't matter

ammar92 (commented):
It looks good to me. I do agree with @Donnype; we might as well reuse the term *produces* for this as soon as possible (and preferably with this PR), as currently having these two properties coexist together is imo  misleading/ illogical and there's not really an backward compatibility issue (yet)

Donnype:
> […] and there's not really an backward compatibility issue (yet)

@ammar92 There is a backward compatibility issue in Rocky though, because we still render the produces field in the boefje detail page (possibly at more places)

Donnype:
@underdarknl Wrt dropping un-prefixed mime-types: if we drop them let's do that in https://github.com/minvws/nl-kat-coordination/pull/1945. This however probably affects boefje-normalizer mapping logic as most normalizers consume plain boefje ids and I am not sure what the scheduler does. @jpbruinsslot ?

jpbruinsslot:
> @underdarknl Wrt dropping un-prefixed mime-types: if we drop them let's do that in #1945. This however probably affects boefje-normalizer mapping logic as most normalizers consume plain boefje ids and I am not sure what the scheduler does. @jpbruinsslot ?

Based on the mime-type of the created raw data from bytes the scheduler retrieves normalizers from the katalogus. There is a cache in the scheduler that maps mime-type to normalizers. It builds this cache looking at the plugins' `consumes` field..

Donnype:
@jpbruinsslot So if the normalizer's `consumes` field would contain prefixed values, everything should just work as expected? Then we could just update all `normalizer.json` files @underdarknl.

jpbruinsslot:
> @jpbruinsslot So if the normalizer's `consumes` field would contain prefixed values, everything should just work as expected? Then we could just update all `normalizer.json` files @underdarknl.

Yes if that that maps to the same mime type then this would work as expected. If possible let's test it.

dekkers:
> > […] and there's not really an backward compatibility issue (yet)
> 
> @ammar92 There is a backward compatibility issue in Rocky though, because we still render the produces field in the boefje detail page (possibly at more places)

This is the only place where we render it as far as I remember looking into this. It is currently wrong because the boefje doesn't produce those objects. If we render the raw file types there it would fix the problem of wrong information, so I don't really see the problem with using `produces`. We can also do this in a later PR.

dekkers:
> Would now also be a good moment to drop the mime-types that are unprefixed? Eg, keep the `boefje/$name` variant, and remove the `$name` variant? I feel we should have people target the `boefje/$name` variant from their normalizer as it is less ambiguous and keeps our databases cleaner.

Ambiguous compared to what? Only boefjes produce raw files, so what is the benefit of having to specify that raw file type  is produced by a boefje? I don't really see the added value of the `boefje/` prefix and I'd rather remove the `boefje/$name` variant and keep `$name`.

Donnype:
@dekkers The produces field I believe is rendered with a link for the OOIType, so this would require some changes in rocky where we either drop or replace the link. My main objective now is keeping this PR from growing too big, given that a follow up PR will either drop or replace the produces field

Donnype:
@dekkers The produces field I believe is rendered with a link for the OOIType, so this would require some changes in rocky where we either drop or replace the link. My main objective now is keeping this PR from growing too big, given that a follow up PR will either drop or replace the produces field

TwistMeister:
## Checklist for QA:

### What works:
* Boefjes now list their output mime types

### Bug or feature?:
* The only question I had was why the `mime-types` was added alongside `produces` instead of replacing it as suggested in the initial ticket. But that was answered in the PR description; to keep this PR scoped and to keep things backwards compatible until we follow this up in Rocky as well. Which seems perfectly reasonable.

TwistMeister (approved):

PR #1949, opened 2023-10-20
Fix installing rocky packages in RDO workflow (1.13)

dekkers:
Backport of #1946 to release 1.13

### DIFF
--- a/.github/workflows/build-rdo-package.yml
+++ b/.github/workflows/build-rdo-package.yml
@@ -8,6 +8,7 @@ on:
   push:
     tags:
       - v*
+  workflow_dispatch:
 
 env:
   PKGDIR: /home/runner/work/nl-kat-coordination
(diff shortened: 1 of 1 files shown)


.github/workflows/build-rdo-package.yml
```diff
@@ -8,6 +8,7 @@ on:
   push:
     tags:
       - v*
+  workflow_dispatch:
```
ammar92, line 11: Just to be sure; it's supposed to be an empty object, right?

ammar92 (approved):

.github/workflows/build-rdo-package.yml
dekkers: Yes, it makes it possible manually schedule the workflow which should make it easier to test things.

underdarknl (approved):

PR #1954, opened 2023-10-23
Add test for task download and reschedule task other org

Rieven:
### Changes
Add test to check if:
- Member from org A can **reschedule** task from orgB
- Member from org A can **download** task from orgB

### DIFF
--- a/rocky/tests/test_boefjes_tasks.py
+++ b/rocky/tests/test_boefjes_tasks.py
@@ -1,9 +1,12 @@
 from http import HTTPStatus
 from unittest.mock import call
 
+import pytest
+from django.http.response import Http404
 from pytest_django.asserts import assertContains
 from requests import HTTPError
 
+from rocky.views.bytes_raw import BytesRawView
 from rocky.views.tasks import BoefjesTaskListView
 from tests.conftest import setup_request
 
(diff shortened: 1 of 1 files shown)


rocky/tests/test_boefjes_tasks.py
```diff
@@ -91,6 +94,34 @@ def test_reschedule_task(rf, client_member, mocker, task):
     )
 
 
+def test_reschedule_task_from_other_org(rf, client_member, client_member_b, mocker, task):
+    mock_scheduler_client = mocker.patch("rocky.views.tasks.client")
+    mock_scheduler_client.get_task_details.return_value = task
+    mocker.patch("tools.view_helpers.client.session")
+    mocker.MagicMock()
```
ammar92, line 101: Calling this doesn't have any effect?

rocky/tests/test_boefjes_tasks.py
```diff
@@ -91,6 +94,34 @@ def test_reschedule_task(rf, client_member, mocker, task):
...
+
+    request = setup_request(
+        rf.post(
+            f"/en/{client_member.organization.code}/tasks/boefjes/?task_id={task.id}",
+            data={"action": "reschedule_task"},
+        ),
```
ammar92, line 107: @underdarknl I was just thinking: should this view within this context yield a permission error instead, or is that too much information disclosure?

ammar92:
I'm still hesitant. The tests are good for now, but ideally, this should be fixed on a higher level. Otherwise, the tests would explode with tests for permission checks.

rocky/tests/test_boefjes_tasks.py
Rieven: @ammar92 Clients can not start the task (for example another redteam member) but can view and download task details.

underdarknl: We should make sure there is no difference between asking for a nonexistent Job, OOI, task or organisation to be rescheduled, and for one of those that does exists but is not available to you as a user.

rocky/tests/test_boefjes_tasks.py
Rieven: You are right, this was redundant. Removed!

underdarknl:
> I'm still hesitant. The tests are good for now, but ideally, this should be fixed on a higher level. Otherwise, the tests would explode with tests for permission checks.

I think most of these functions should not be in the views, they should be on their respective models. So for this perhaps a Job model that has a getraw method which takes the organization, and the uuid from the from the job instance, but the user trying to perform this action as an argument.

ammar92:
## Checklist for QA:

### What works:
* It seems to be working on my end; however, I'm hesitant about how bulletproof it is

Rieven:
> For the next PR:
> 
> * Fill in the complete PR form
> * It would be easier for others to test if you describe a test plan/ path/ how to test
> 
> ## Checklist for QA:
> * [ ]  I have checked out this branch, and successfully ran a fresh `make reset`.
> * [x]  I confirmed that there are no unintended functional regressions in this branch:
>   
>   * [ ]  I have managed to pass the onboarding flow
>   * [x]  Objects and Findings are created properly
>   * [x]  Tasks are created and completed properly
> * [x]  I confirmed that the PR's advertised `feature` or `hotfix` works as intended.
> 
> ### What works:
> * It seems to be working on my end; however, I'm hesitant about how bulletproof it is

@Donnype made also a backend implementation that makes it more bulletproof. https://github.com/minvws/nl-kat-coordination/pull/1991

rocky/rocky/views/bytes_raw.py
```diff
@@ -8,33 +8,38 @@
...
-            client.login()
+            self.bytes_client.login()
             boefje_meta_id = kwargs["boefje_meta_id"]
-            raw_metas = client.get_raw_metas(boefje_meta_id)
+            raw_metas = self.bytes_client.get_raw_metas(boefje_meta_id)
+
+            raws = {
+                raw_meta["id"]: self.bytes_client.get_raw(raw_meta["id"])
+                for raw_meta in raw_metas
+                if raw_meta["boefje_meta"]["organization"] == self.organization.code
```
underdarknl, line 27: Ideally we should move this check / filtering into the bytes_client.get_raw() method. This would make sure you cannot forget to perform this filterin in any view.

underdarknl (approved):

Issue #1955, opened 2023-10-23
More Scheduler Error handling from Rocky with more task details


Rieven:
- Show task details in rocky messages (failed + success) 
- ConnectionError handler when Scheduler is unreachable. 

**Describe the bug**

1. When rocky tasks failed to be rescheduled or failing in general. Show plugin + OOI in rocky messages, giving better info of failed tasks. 
2. When the Scheduler is offline, throw an ConnectionError, everywhere in rocky where the Scheduler is needed.

**To Reproduce**

- Stop Scheduler container (stop the service)
- Try to reach tasks page, ooi detail page, plugin detail page.

**Expected behavior**
500 error

**OpenKAT version**
1.12.1

PR #1957, opened 2023-10-23
Refactor Task List and filters with error handlers for Scheduler 

Rieven:
### Changes

Task list can be found at: 
 - "Tasks" in main menu
 - OOI detail page
 - Boefje Detail page
 
- When scheduler client is offline, throw a ConnectionError and catch it to show message in rocky.
- Add more task details for when task fetching fails/success, or (re)scheduling task fails, or get task failures. (plugin + OOI)

### Issue link
https://github.com/minvws/nl-kat-coordination/issues/1955

Closes https://github.com/minvws/nl-kat-coordination/issues/1955

How's everything related?
```mermaid
sequenceDiagram
    SchedulerView->>+SchedulerClient: Get task list
    SchedulerClient->>+Scheduler: Get task list
    Scheduler->>+SchedulerClient: throws: ConnectError, HTTPError, etc..
    SchedulerClient->>+SchedulerView: Django message
    SchedulerView->>+SchedulerClient: Schedule task
    SchedulerClient->>+Scheduler: Schedule task
    Scheduler->>+SchedulerClient: throws: ConnectError, HTTPError, etc..
    SchedulerClient->>+SchedulerView: Django message
    SchedulerView->>+SchedulerClient: Re-schedule task
    SchedulerClient->>+Scheduler: Re-schedule task
    Scheduler->>+SchedulerClient: throws: ConnectError, HTTPError, etc..
    SchedulerClient->>+SchedulerView: Django message
    SchedulerView->>+SchedulerClient: Run Boefje
    SchedulerClient->>+Scheduler: Run Boefje
    Scheduler->>+SchedulerClient: throws: ConnectError, HTTPError, etc..
    SchedulerClient->>+SchedulerView: Django message
    SchedulerView->>+SchedulerClient: Run Normalizer
    SchedulerClient->>+Scheduler: Run Normalizer
    Scheduler->>+SchedulerClient: throws: ConnectError, HTTPError, etc..
    SchedulerClient->>+SchedulerView: Django message
    
```

### Proof

Errors will be redirected to the health page:
image

### Code Checklist

### Communication

---

---
## Checklist for QA:


rocky/rocky/scheduler.py
```diff
@@ -238,9 +255,62 @@ def push_task(self, queue_name: str, prioritized_item: QueuePrioritizedItem) ->
...
+
+def schedule_task(request: HttpRequest, organization_code: str, task: QueuePrioritizedItem) -> None:
+    plugin_name = ""
+    input_ooi = ""
+    if task.data.type == "boefje":
+        plugin_name = task.data.boefje.name
+        input_ooi = task.data.input_ooi
+    if task.data.type == "normalizer":
+        plugin_name = task.data.normalizer.id  # name not set yet, is None for name
+        input_ooi = task.data.raw_data.boefje_meta.input_ooi
```
ammar92, line 283: What about the else clause?

rocky/rocky/scheduler.py
```diff
@@ -238,9 +255,62 @@ def push_task(self, queue_name: str, prioritized_item: QueuePrioritizedItem) ->
...
+    try:
+        get_scheduler_client().push_task(f"{task.data.type}-{organization_code}", task)
+    except (BadRequestError, TooManyRequestsError, ConflictError) as task_error:
+        error_message = task_error.message + _(" Scheduling {} {} with input object {} failed.").format(
+            task.data.type.title(), plugin_name, input_ooi
+        )
```
ammar92, line 289: I would expect the message the other way around: first the notification of failure, and then the reason

rocky/rocky/scheduler.py
```diff
@@ -238,9 +255,62 @@ def push_task(self, queue_name: str, prioritized_item: QueuePrioritizedItem) ->
...
+            request,
+            _(
+                "Task '{} {} with input object {}' is scheduled and will soon be started in the background. "
+                "Results will be added to the object list when they are in. "
+                "It may take some time, a refresh of the page may be needed to show the results."
+            ).format(task.data.type.title(), plugin_name, input_ooi),
+        )
+
+
+def get_list_of_tasks_lazy(request: HttpRequest, **params) -> LazyTaskList | None:
```
ammar92, line 304: The return type doesn't make sense, either `None` should be a list, or the returned list should be a `None` instead

rocky/rocky/scheduler.py
```diff
@@ -238,9 +255,62 @@ def push_task(self, queue_name: str, prioritized_item: QueuePrioritizedItem) ->
...
+    try:
+        return get_scheduler_client().get_lazy_task_list(**params)
+    except SchedulerError as error:
+        messages.error(request, error.message)
+        return []
 
 
-client = SchedulerClient(settings.SCHEDULER_API)
+def get_details_of_task(request: HttpRequest, task_id: str | None) -> Task | None:
```
ammar92, line 312: This function just returns `Task`

rocky/rocky/scheduler.py
```diff
@@ -238,9 +255,62 @@ def push_task(self, queue_name: str, prioritized_item: QueuePrioritizedItem) ->
...
-        return ServiceHealth.parse_raw(health_endpoint.content)
+        try:
+            health_endpoint = self.session.get(f"{self._base_uri}/health")
+            health_endpoint.raise_for_status()
+            return ServiceHealth.parse_raw(health_endpoint.content)
+        except ConnectionError:
+            raise SchedulerError()
+
+
+def get_scheduler_client():
```
ammar92, line 266: Return type

rocky/rocky/scheduler.py
```diff
@@ -238,9 +255,62 @@ def push_task(self, queue_name: str, prioritized_item: QueuePrioritizedItem) ->
...
+    try:
+        client = SchedulerClient(settings.SCHEDULER_API)
+        client.health()
```
ammar92, line 269: This functionality doesn't belong here. If the caller intends to ensure system health, that responsibility should rest with the caller, not within this factory function.

ammar92 (changes requested):
Can be improved. Tests are failing, and there are errors in return types. Moreover, there is a mix of type annotation styles (e.g. `Optional[None]` vs ` ... | None`), so these should be rewritten to keep consistency.

rocky/rocky/scheduler.py
Rieven: When the client is offline, I want to first check if its healthy meaning running and not offline.  I have refactored it to return the client instance anyways but still cannot do anything with it till the Scheduler is back online.

ammar92: > When the client is offline, I want to first check if its healthy meaning do a simple request to the health page. The other thing is LazyTaskList will first do an internal count throwing a ConnectionError inside and I cannot get it out to rocky. My intention was here to even before that when the client is initiated to check if it is usable. I have refactored it to return the client instance anyways but still cannot do anything with it till the Scheduler is back online.

Factory methods, like the one here, shouldn't have side effects such as calling external services.

Performing health checks before using the `LazyTaskList` doesn't help you out, since it doesn't guarantee the service's health state won't change during the lifespan of this `LazyTaskList` instance.

And what do you mean by "cannot get it out to rocky"? I'm sure we can refactor this piece to achieve what we want.

rocky/rocky/scheduler.py
```diff
@@ -182,17 +185,25 @@ class ConflictError(SchedulerError):
     message = _("Task already queued.")
 
 
+class TaskNotFoundError(SchedulerError):
+    message = _("Task could not be found.")
+
+
 class SchedulerClient:
     def __init__(self, base_uri: str):
         self.session = requests.Session()
```
underdarknl, line 194: Could we add a small middleware class that offers the following method to inject the error handling as we use it now?

```
class SchedulerSession(requests.Session):
  def request(self, *args, **kwargs):
    try:
      return super().request(*args, **kwargs)
    except (requests.ConnectionError, requests.RequestException, requests.TooManyRedirects, requests.ConnectTimeout, requests.ReadTimeout, reuqests.Timeout) as error:
      raise SchedulerError(error)
```

Rieven: I've found a midway solution.

rocky/rocky/scheduler.py
```diff
@@ -189,21 +189,35 @@ class TaskNotFoundError(SchedulerError):
     message = _("Task could not be found.")
 
 
-class SchedulerClient:
+class SchedulerSession(requests.Session):
     def __init__(self, base_uri: str):
-        self.session = requests.Session()
+        super().__init__()
+        self.session = self
```
ammar92, line 195: This is very uncommon, why would you assign `self` to a bound property?

rocky/rocky/scheduler.py
```diff
@@ -189,21 +189,35 @@ class TaskNotFoundError(SchedulerError):
...
         self._base_uri = base_uri
-        self.is_alive()
 
+        try:
+            self.session.get(f"{self._base_uri}/health")
```
ammar92, line 199: These kind of calls do *not* belong in a constructor/ initializer. We've talked about doing these kind of calls before

rocky/rocky/scheduler.py
```diff
@@ -189,21 +189,35 @@ class TaskNotFoundError(SchedulerError):
...
+        except (
+            requests.ConnectionError,
+            requests.RequestException,
+            requests.TooManyRedirects,
+            requests.ConnectTimeout,
+            requests.ReadTimeout,
+            requests.Timeout,
+        ) as error:
```
ammar92, line 207: Since these are all subclassed from `RequestException`, you might as well use just that. Unless other exceptions are coming from `RequestException` that you don't care about?

rocky/rocky/scheduler.py
```diff
@@ -189,21 +189,35 @@ class TaskNotFoundError(SchedulerError):
...
+            raise SchedulerError(error)
+
+
+class SchedulerClient(SchedulerSession):
```
ammar92, line 211: A client and a session are two different things. In our codebase, we usually encapsulate a session object in a client.

ammar92 (changes requested):

stephanie0x00:
## Checklist for QA:

### What works:
Error message is shown on the Tasks, Health checks, when opening a finding object, and on the 'See Details' page in the Katalogus. When enabling the scheduler again the error message disappears again. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

rocky/rocky/scheduler.py
underdarknl: Yes, this is a specific subset of issues.

ammar92: > Yes, this is a specific subset of issues.

It seems a bit unusual to include `requests.RequestException` in this specific list. Doing so may make the other exceptions in this list redundant. On the other hand, if you'd remove `requests.RequestException` from the exception list, does that mean we don't care about the other exceptions such as SSL errors, JSON errors, proxy errors, etc.?

Rieven: I have checked indeed that `requests.RequestException` incorporates all the other errors, or we remove `requests.RequestException` and keep the listed ones, or we just use `requests.RequestException` and it catches the list + others mentioned.

rocky/rocky/scheduler.py
Rieven: We need to somehow talk to the Scheduler, before the lazyloading happens, it will not happen if its not checked in init. I think getting the /health endpoint or just / is enough to check if its online.

ammar92: What do you mean by we *need to*? Lazy loading must happen when it needs to happen, hence *lazy loading*. Health checking is for troubleshooting or startup requirements, **not a requirement to check for every request**

underdarknl: @Rieven means we need to figure out if the scheduler is available or not before we do a lazy loading call. A lazy loading call would end up being executed in the template context where we have a harder time handling any errors.
Since doing a health check before hand is an extra request, we should not do so, it just makes everything more expensive, and might still end up erroring out when we do an actual data request.

I think we should not make this (or any other restfull api request) lazy loading. If these requests are needed, they should be done in the model (or view), and we should handle any errors and exceptions in that scope too. Our targeted restfull api's dont offer anything that resembles lazy-loading anyway, so there's no gain in having these calls be lazy or deferred to when the template is rendering.

Rieven: Ok, then I need to refactor the lazyloader which means that I need to also refactor pagination for task items.

underdarknl:
Blocked by https://github.com/minvws/nl-kat-coordination/issues/2598

rocky/rocky/views/task_detail.py
```diff
@@ -1,32 +1,41 @@
...
-    def get_task(self, task_id):
-        try:
-            return client.get_task_details(self.organization.code, task_id)
-        except TaskNotFoundError as error:
-            messages.error(self.request, error.message)
+class DownloadTaskDetail(SchedulerView):
+    def get(self, request, *args, **kwargs):
+        task_id = kwargs["task_id"]
+        filename = "task_" + task_id + ".json"
+        task_details = self.get_task_details(task_id)
```
underdarknl, line 16: No try/catch around this get_task_details for bytes errors?

Rieven: Yes will be handled by the SchedulerView at: https://github.com/minvws/nl-kat-coordination/blob/ca3bf8c05933b7c3ecd1639734c62325aa93bcad/rocky/rocky/views/scheduler.py#L62

rocky/rocky/views/scheduler.py
```diff
@@ -0,0 +1,190 @@
...
+    task_filter_form = TaskFilterForm
+
+    def setup(self, request, *args, **kwargs):
+        super().setup(request, *args, **kwargs)
+        self.scheduler_client = scheduler_client(self.organization.code)
+
+    def get_task_filters(self) -> dict[str, Any]:
+        return {
+            "scheduler_id": f"{self.task_type}-{self.organization.code}",
+            "task_type": self.request.GET.get("type", self.task_type),
```
underdarknl, line 43: Should self.task_type not be used on line 43 too, or should we not read from the GET request for both?

rocky/rocky/templates/tasks/partials/task_status.html
```diff
@@ -0,0 +1,15 @@
...
+    <i class="icon running"></i>&nbsp;{{ task.status.value|capfirst }}
+{% elif task.status.value == "cancelled" %}
+    <i class="icon cancelled"></i>&nbsp;{{ task.status.value|capfirst }}
+{% elif task.status.value == "failed" %}
+    <i class="icon negative"></i>&nbsp;{{ task.status.value|capfirst }}
+{% elif task.status.value == "completed" %}
+    <i class="icon positive"></i>&nbsp;{{ task.status.value|capfirst }}
+{% else %}
+    <i class="icon neutral"></i>&nbsp;{{ task.status.value|capfirst }}
+{% endif %}
```
underdarknl, line 15:
```suggestion
{% if task.status.value == "pending" or task.status.value == "queued" %}
  {% set status = "queued" %}
{% elif task.status.value == "running" or task.status.value == "dispatched" %}
  {% set status = "running" %}
{% elif task.status.value == "cancelled" %}
  {% set status = "cancelled" %}
{% elif task.status.value == "failed" %}
  {% set status = "negative" %}
{% elif task.status.value == "completed" %}
  {% set status = "positive" %}
{% else %}
  {% set status = "neutral" %}
{% endif %}
<i class="icon {{ status }}"></i>&nbsp;{{ task.status.value|capfirst }}
```
Or better, map these values in the backend?

rocky/rocky/scheduler.py
```diff
@@ -147,7 +150,9 @@ def __len__(self):
     def __getitem__(self, key) -> list[Task]:
         if isinstance(key, slice):
             offset = key.start or 0
-            limit = key.stop - offset
+            limit = LazyTaskList.HARD_LIMIT
+            if key.stop:
+                limit = key.stop - offset
```
underdarknl, line 155: limit = min(LazyTaskList.HARD_LIMIT, key.stop - offset if key.stop or LazyTaskList.HARD_LIMIT)

rocky/rocky/scheduler.py
```diff
@@ -147,7 +150,9 @@ def __len__(self):
     def __getitem__(self, key) -> list[Task]:
         if isinstance(key, slice):
             offset = key.start or 0
-            limit = key.stop - offset
+            limit = LazyTaskList.HARD_LIMIT
```
underdarknl, line 153:
```suggestion
            limit = min(LazyTaskList.HARD_LIMIT, key.stop - offset if key.stop or LazyTaskList.HARD_LIMIT)
```

rocky/rocky/scheduler.py
```diff
@@ -123,6 +124,8 @@ class PaginatedTasksResponse(BaseModel):
 
 
 class LazyTaskList:
+    HARD_LIMIT = 99_999_999
```
underdarknl, line 127: In the context of Rocky, we are never going to need this many tasks, 500 is more than enough

rocky/rocky/views/health.py
```diff
@@ -52,10 +52,10 @@ def get_octopoes_health(octopoes_api_connector: OctopoesAPIConnector) -> Service
     return octopoes_health
 
 
-def get_scheduler_health() -> ServiceHealth:
+def get_scheduler_health(organization_code: str) -> ServiceHealth:
     try:
-        scheduler_health = client.health()
-    except HTTPError:
+        scheduler_health = scheduler_client(organization_code).health()
+    except (SchedulerError, HTTPError):
```
underdarknl, line 58: Are there still HTTPErrors expected to bubble up here? I'd suggest catching all those in the client and raising a SchedulerError (or a more specific subclass of that) to abstract away that we are dealing with a http endpoint.

rocky/rocky/views/tasks.py
```diff
@@ -1,117 +1,63 @@
...
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        try:
-            context["stats"] = client.get_task_stats(self.organization.code, self.plugin_type)
-        except SchedulerError:
-            context["stats_error"] = True
-        else:
-            context["stats_error"] = False
+        context["task_filter_form"] = self.get_task_filter_form()
```
ammar92, line 46: Evaluate lazy object
```suggestion
        context["task_list"] = list(context["task_list"]) 
        context["task_filter_form"] = self.get_task_filter_form()
```

rocky/rocky/views/health.py
Rieven: ahhh yes, I forgot to move this to the client

rocky/rocky/templates/tasks/partials/task_status.html
Rieven: even better, I have removed this partial and just use the value as class. Added these vars in sass

rocky/rocky/views/tasks.py
Rieven: this did not work, because the error is thrown in `super().get_context_data` of parent `ListView` 
I made a workaround though, check it out.

rocky/rocky/views/scheduler.py
Rieven: removed `self.request.GET.get` completely

stephanie0x00:
During QA the following issues were identified: 

- [ ] DnsRecords boefje details page in Katalogus misses some data. A similar thing is happening for Dnssec and nmap TCP boefjes. 
- [ ] Normalizer template file removed from PR, which prevents various detail pages from loading. 



#### DnsRecords boefje details page in this branch: 
image

On main: 
image


#### Template file removed from PR

image

stephanie0x00:
From QA perspective it looks good now. The pages that were not properly loading in the katalogus now work again.

Rieven:
> From QA perspective it looks good now. The pages that were not properly loading in the katalogus now work again. Can be merged after Review has approved the changes.

One little thing I am fixing now is that the normalizer detail page is not showing the task list, but this I am fixing now, thanks!

underdarknl (approved):

PR #1958, opened 2023-10-24
Implement new ranking strategy for boefje tasks

### LINKED ISSUE #937: Improve scheduler ranking
We can update the scheduler queue ranking based on prior tasks, and findings, and calculate its priority based on those metrics.

Related issues:
* https://github.com/minvws/nl-kat-coordination/issues/801

jpbruinsslot:
### Changes

Implement a new ranking strategy for the prioritization of boefje tasks, it takes into account the results of prior tasks to determine the priority. Weights are added to fine-tune what feature should weigh more in the priority calculation.

* Added calls to external services to gather information, e.g. get objects that have been created by prior task, and get findings created by prior task.

* Added a naive calculation of priority by combining:

1. duration of prior tasks
2. total of created objects of a task
3. total of created findings of a task
4. priorities of prior tasks
5. time since grace period

### Issue link

Closes #937


mula/scheduler/rankers/boefje.py
```diff
@@ -30,19 +34,109 @@ def rank(self, obj: Any) -> int:
         max_days_in_seconds = self.MAX_DAYS * (60 * 60 * 24)
         grace_period = timedelta(seconds=self.ctx.config.pq_grace_period)
 
+        prior_tasks: List[models.Task] = self.ctx.datastores.task_store.get_tasks_by_hash(obj.hash)
```
jpbruinsslot, line 37: Do we only want the tasks that are finished, or do we also need to count the failures?

mula/scheduler/rankers/boefje.py
```diff
@@ -30,19 +34,109 @@ def rank(self, obj: Any) -> int:
...
+            objects = self.ctx.services.octopoes.get_objects_by_ooi(
+                organisation_id=obj.organization,
+                reference=prior_task.p_item.data.get("input_ooi"),
+            )
+            freq_objects.append(len(objects))
+            self.logger.info(objects)
+
+            # How many objects where created by the task?
+            # How many findings were generated by the task?
+            findings = self.ctx.services.octopoes.get_findings_by_ooi(
```
jpbruinsslot, line 74: exception handling

mula/scheduler/rankers/boefje.py
```diff
@@ -30,19 +33,112 @@ def rank(self, obj: Any) -> int:
...
+
+        def division(a, b):
+            return a / b if b != 0 else 0
+
+        # Normalize the values
+        norm_duration: float = division((med_duration - min_duration), (max_duration - min_duration))
+        norm_objects: float = division((med_objects - min_objects), (max_objects - min_objects))
+        norm_findings: float = division((med_findings - min_findings), (max_findings - min_findings))
+        norm_priorities: float = division((med_priorities - min_priorities), (max_priorities - min_priorities))
+        norm_since_grace_period: float = min(division((time_since_grace_period - 0), (max_days_in_seconds - 0)), 1)
```
Donnype, line 115: I believe this form of normalization has some unwanted consequences for the ranking.

We now normalize each feature of the task based on a set of properties of the task. If we take `norm_duration` for some `task_1` as an example, we have that it is derived from the fact that the `task_1` ran for 100, 102 and 105 seconds respectively. Then the `med_duration = 102` and `norm_duration = 0.4`. For another task, `task_2`, we have a duration of 10 and 12 seconds. We derive a `norm_duration` of `0.5`. This is because we normalize based on the given data set for each task, not against a pre-defined measure. (Often, e.g. when normalizing training data for a model, you have the whole dataset at hand, so you can easily normalize by dividing by the maximum for a feature.) In short, `norm_duration` is a measure of variance in the data of each task, not a measure for how long a task actually takes relative to other tasks. That's probably not what we want for the ranking.

It would make more sense I think to define a global maximum against which we normalize these sequences. This is a bit coarse perhaps, but at least defines a measure that is consistent for independent tasks. For example, we could say that the maximum amount of findings per task is 100, based on a query in our test environment or gut feeling. We could correct outliers by setting every freq_findings larger than 100 equal to 100, take the mean of that set and normalize the mean by dividing by 100:

```python3
MAX_FINDINGS = 100
freq_findings = [min(x, MAX_FINDINGS) for x in freq_findings]
mean_findings = statistics.mean(freq_findings)
normalized_mean_findings = mean_findings / MAX_FINDINGS
```

Then `0 <= normalized_mean_findings <= 1`, so we can multiply the weights by these normalized_means. (This also resolves the division issue.)

mula/tests/integration/test_boefje_ranker.py
```diff
@@ -0,0 +1,99 @@
...
+            num_elements = random.randint(1, 10)
+            elements = [random.randint(1, 100) for _ in range(num_elements)]
+            return elements
+
+        # Mock
+        self.mock_get_tasks_by_hash.return_value = prior_tasks
+        self.mock_get_objects_by_ooi.side_effect = generate_random_elements
+        self.mock_get_findings_by_ooi.side_effect = generate_random_elements
+
+        self.ranker.rank(boefje_task)
```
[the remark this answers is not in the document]
jpbruinsslot: Yeah tests are a bit bare-boned, as far for the draft I wanted to make sure everything was working correctly

mula/scheduler/rankers/boefje.py
```diff
@@ -30,19 +33,112 @@ def rank(self, obj: Any) -> int:
...
+        # Calculate the score based on the duration, number of objects, number
+        # of findings and the priority of the prior tasks.
+        freq_duration, freq_objects, freq_findings, freq_priorities = [], [], [], []
+
+        # Iterate over the prior tasks
+        for prior_task in prior_tasks:
+            # How long did it take for the task to run?
+            duration = (prior_task.modified_at - prior_task.created_at).seconds
+            freq_duration.append(duration)
+            self.logger.info(duration)
```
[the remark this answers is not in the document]
jpbruinsslot: Yeah mainly left in for developing purposes, will remove them before merging.

mula/scheduler/rankers/boefje.py
```diff
@@ -30,19 +33,112 @@ def rank(self, obj: Any) -> int:
...
+
+        min_findings: int = min(freq_findings)
+        max_findings: int = max(freq_findings)
+        med_findings: float = statistics.median(freq_findings)
+
+        min_priorities: int = min(freq_priorities)
+        max_priorities: int = max(freq_priorities)
+        med_priorities: float = statistics.median(freq_priorities)
+
+        def division(a, b):
```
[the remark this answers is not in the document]
jpbruinsslot: Yeah, you are right, was mainly meant as a division by zero error protection. But I understand what you mean with constant sequences.

mula/scheduler/rankers/boefje.py
jpbruinsslot: Would it make sense when the ranker is initialized to gather this data from the current dataset (if not available to use a global maximum) and update these values at specified intervals as background tasks?

Donnype: That could be a tactic, but I am not sure what we would gain in accuracy is worth the extra complexity vs. overshooting the maximum for now. That is, if we gain any accuracy at all. Also note that statically defined weights are meaningless in that case since we cannot tune them in development; the "dynamic normalization" probably affects the scale of the data a lot. For example, if we suddenly "re-scale" an important parameter that had a high weight defined, we expect the parameter to stay important. But if an anomaly happened in production and the parameter suddenly has a maximum that is 10 times as high, the parameter would suddenly have almost no effect anymore as its normalized value drops by a factor of 10.

mula/scheduler/rankers/boefje.py
underdarknl: Lets assume we get the best measurement by looking at completed tasks. Failed tasks would skew the roi to a lower value even though it might have been a temporary issue.

mula/scheduler/rankers/boefje.py
```diff
@@ -30,19 +33,112 @@ def rank(self, obj: Any) -> int:
...
+
+        if not any([freq_duration, freq_objects, freq_findings, freq_priorities]):
+            """
+            return int(3 + (max_priority - 3) * (1 - time_since_grace_period / max_days_in_seconds))
+            """
+            return -1
+
+        max_duration: int = max(freq_duration)
+        min_duration: int = min(freq_duration)
+        med_duration: float = statistics.median(freq_duration)
```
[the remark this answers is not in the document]
Donnype: Hm perhaps not necessarily for sequences with frequent outliers (when I ignore the other suggestions I made for the ranking and normalization)

Issue #1662, opened 2023-08-23
'Wappalyzer - Software Scan' follows redirects and links software (injustice) to the object on which the scan originally started.


jordy-kennisnet:
**Describe the bug**
'Wappalyzer - Software Scan' follows redirects and links software (injustice) to the object on which the scan originally runs.

This can also lead to scanning a website where no clearance level has been given!

**To Reproduce**

1. Choose a website that offers a redirect to another website (with different software).
2. Scan this website using Wappalyzer.
3. Download the task files and observe where the redirect takes place.
4. Add the redirected URL as an object and perform a scan on it as well.
5. Observe that the same software is detected on both sites, despite the fact that they differ in reality.

**Expected behavior**
When Wappalyzer follows a redirect, it should lead to a new object (the redirected URL). The discovered software must be linked to this new object, and not to the original object where the scan was initially started.

**OpenKAT version**
- Version: 1.10rc2

**Additional context**
In the task file of Wappalyzer, you can see its followed redirects:
{"urls":{"https://web.site/":{"status":301},"https://redirected.url":{"status":200}}

underdarknl:
Seeing that this issue arises more often due to missing config flags, and since we are also seeing tools that we cannot control to access a specific IP I propose the following:
We implement a safeguard around these tools inside the boefjes runner. This could (optionally) replace the resolv.conf with an empty file for that specific container, and specify the correct hostname and IP combo in the /etc/hosts file as given by our input OOI.
This way a redirect would either run into an unresolveable dns entry, or stay on the same host (which probably has the correct indemnifications present)

jordy-kennisnet:
That would be one part of the solution. Because you also want to have a view of the redirects. In other words, which instance would creates a new object - where the redirect is located? Which can be scanned with the correct clearance level or not?

Donnype:
@underdarknl For this specific instance it was quite easy to detect the redirects and create an intermediate step. I suggested a fix in the linked PR.

PR #1962, opened 2023-10-25
Yield redirected urls from Wappalizer instead of wrong software instances

Donnype:
### Changes
Change the normalizer logic to return redirected urls when the HTTP status code for the original input url was 3**.

### Issue link
Closes https://github.com/minvws/nl-kat-coordination/issues/1662

### Proof
image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


boefjes/boefjes/plugins/kat_website_software/normalize.py
```diff
@@ -4,15 +4,33 @@
...
     hostname_reference = Reference.from_str(pk)
 
+    original_url_status = results["urls"][url]["status"]
+
+    if 300 <= original_url_status < 400:
+        # The requested url was redirected, so only return the new url instance. If needed we rescan the new url.
+        results["urls"].pop(url)
+
+        for redirected_url in results["urls"]:
+            yield URL(network=hostname_reference.tokenized.netloc.network.name, raw=redirected_url)
```
underdarknl, line 30: Should'nt we also create a new instance of the HTTPResource which updates/sets the redirects_to field to this new Url?

Donnype: I've looked into this but for an HTTPResource we need to attach a Website in this scope, which depends on an IPService, etc. So this probably not what we want to do here.

Perhaps wappalyzer should not run on `HostnameHTTPURL` objects but on `HTTPResource` objects, as the name `kat_website_software` also suggests a website should exists. In that case the initial HTTPResource has `redirects_to` set to `None`, but after the first run set to the actually redirected site. But then I would argue that we should not have two `HTTPResource` objects pointing to a website if one also points to the other through "redirects_to". We don't actually get the resource from the first URL.

But I noticed we do not set redirects_to for `HTTPResource` objects anywhere yet. And given this setup, wouldn't it make more sense to move that field to the `HostnameHTTPURL` object and only have the actual url that does not redirect added to the `HTTPResource`? Then this piece of code can simply set the `redirects_to` field on the input  `HostnameHTTPURL` and we can later configure that this boefje should only run on ``HostnameHTTPURL` where the `redirects_to` field is `None`. This would also perhaps tackle (part of) the issue @dekkers described where we redundantly spawn multiple boefjes for urls of which we know they all redirect to the same URL.

underdarknl: Hm, yes. Maybe looking at it in reverse would work better? These url's are pointing to this resource?
Or an intermediate Redirect Object which binds urls to other urls.

noamblitz: Just to keep in mind that in the way it is currently set up, we cannot force an IP on Wappalyzer, which means that for all HTTPResources, the boefje would run and would do exactly the same. For that reason, we initially had HostnameHTTPURL as input, since it is IPService unaware.

Donnype: @noamblitz The difference would be that the url field of one `HTTPResource` would not point to the resource directly, so if we run wappalyzer with my changes we only add new `HTTPResource`s for runs on resources with redirecting urls. (Sounds a bit strange, hence the suggested model change.) Then we should run this boefje only on resources that do not redirect. Then we still have redundancy, but we only have it once per instance.

underdarknl: Yes, forcing an IP trough resolve files, or host files is another issue, and can be tackled for all boefjes imho.
Maybe we should ignore this redirect issue for this PR, and create a new Issue for it specifically?

noamblitz: @Donnype Yeah okay makes a little bit more sense. I feel like all this complexity surpasses the goal of the boefje a little bit: "Which software is running on this website?"

Donnype: A schematic of the difference:

### Current (example) setup
```mermaid
classDiagram
direction RL

    IPService  <|--  Website
    Hostname  <|--  Website
    
    class HTTPResource1["HTTPResource"]
    class HostnameHTTPURL1["HostnameHTTPURL"]
    HTTPResource1: redirects_to[example.org]
    HostnameHTTPURL1 : www.example.org
    Website  <|--  HTTPResource1
    HTTPResource1 --|> HostnameHTTPURL1

    class HTTPResource2["HTTPResource"]
    class HostnameHTTPURL2["HostnameHTTPURL"]
    HTTPResource2: redirects_to[example.com]
    HostnameHTTPURL2 : example.org
    Website  <|--  HTTPResource2
    HTTPResource2 --|> HostnameHTTPURL2
    
    class HTTPResource3["HTTPResource"]
    class HostnameHTTPURL3["HostnameHTTPURL"]
    HostnameHTTPURL3 : example.com
    Website  <|--  HTTPResource3
    HTTPResource3 --|> HostnameHTTPURL3
    
    class HTTPResource4["HTTPResource"]
    class HostnameHTTPURL4["HostnameHTTPURL"]
    HTTPResource4: redirects_to[example.com]
    HostnameHTTPURL4 : www.example.com
    Website  <|--  HTTPResource4
    HTTPResource4 --|> HostnameHTTPURL4

    HTTPResource1 --|> HTTPResource2
    HTTPResource2 --|> HTTPResource3
    HTTPResource4 --|> HTTPResource3
```

### Suggested (example) setup
```mermaid
classDiagram
direction RL

    IPService  <|--  Website
    Hostname  <|--  Website
    
    class HostnameHTTPURL1["HostnameHTTPURL"]
    HostnameHTTPURL1 : www.example.org
    HostnameHTTPURL1: redirects_to[example.org]
    HostnameHTTPURL1 --|> HostnameHTTPURL2

    class HostnameHTTPURL2["HostnameHTTPURL"]
    HostnameHTTPURL2 : example.org
    HostnameHTTPURL2: redirects_to[example.com]
    HostnameHTTPURL2 --|> HostnameHTTPURL3
    
    class HostnameHTTPURL3["HostnameHTTPURL"]
    HostnameHTTPURL3 : example.com
    Website  <|--  HTTPResource
    HTTPResource --|> HostnameHTTPURL3
    
    class HostnameHTTPURL4["HostnameHTTPURL"]
    HostnameHTTPURL4 : www.example.com
    HostnameHTTPURL4: redirects_to[example.com]
    HostnameHTTPURL4 --|> HostnameHTTPURL3
```

(One could argue that there also is an `HTTPResource` returning the redirects. But I'd argue that there is no website "attached" to such an `HTTPResource`..)

Donnype: @noamblitz (@underdarknl) I'd say this is definitely a separate issue/feature. Still the discussion at least puts some perspective on how to perhaps improve wappalyzer later on, and in particular why we don't process the feedback here/now. I'll also link an issue.

Donnype: See https://github.com/minvws/nl-kat-coordination/issues/1979 for even another option.

noamblitz: I understand your suggestion and it makes sense. However, I think it is closer to reality that resources redirect instead of urls. Because different resources on different ipservices could in theory redirect to different urls.

originalsouth:
## Checklist for QA:

### What works:
On main:
image

On this PR:
image

### What doesn't work:
N/A

### Bug or feature?:
N/A

dekkers (approved):

PR #2025, opened 2023-11-09
Feat/more csp checks

### LINKED ISSUE #2022: csp normalizer, always check for base-uri
If not base_uri is set, the default-src does not provide a fallback, and as such leaves a security hole if someone can set the base tag's url

Check for the existance of a base-uri, and its value (as we'd proably already do if present)

**Additional context**
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/base-uri
default-src fallback | No. Not setting this allows any URL.

underdarknl:
### Changes
Various changes to tigthen up CSP parsing, and validation.

### Issue link
closes #2022  and various other improvements

### Code Checklist

### Communication


Donnype (approved):

stephanie0x00:
## Checklist for QA:

### What works:
Enabling the Webpage analysis boefje works and it creates findings for some domains that have the CSP header. If the main domain has a CSP header, but paths on the domain do not have the header this results in the 'No CSP' finding, eg. example.com has a CSP header, but example.com/images/image.png does not have the CSP header. 

### What doesn't work:
A domain without a CSP header does not get the 'No CSP' finding, I've tried this for the domain 'hackintheclass.org', which does not have a CSP header. The lack of this header doesn't result in a finding. 

### Bug or feature?:
See above.

underdarknl:
The following Bit should be responsible for missing spf records, it looks reasonable at first glance, but if its not working we should debug that one:

https://github.com/minvws/nl-kat-coordination/blob/main/octopoes/bits/missing_spf/bit.py

underdarknl:
the following code should be adding the no CSP finding, it does this when there are http-headers available, so the recource needs to be level 2 indemnified, and no specific CSP header is present.

https://github.com/minvws/nl-kat-coordination/blob/f73203361ac560b14d0a511499de8a172a2607fe/octopoes/bits/missing_headers/missing_headers.py#L27

Donnype (approved):

dekkers (approved):

PR #2038, opened 2023-11-16
Fix/2037 kat nmap normalizer

### LINKED ISSUE #2037: Service name  inference by port in kat_nmap normalizer
Question regarding the following lines of code: 
https://github.com/minvws/nl-kat-coordination/blob/56c8004ba41e9f6d3d7ba94920048e361278a78a/boefjes/boefjes/plugins/kat_nmap/normalize.py#L46-L50

Suppose one places a `ssh` server behind port 80 or port 443, do we really want to report `http` or `https` respectively?
Perhaps there are considerations for this behavior?

originalsouth:
### Changes
This PR fixes issue #2037 by using the provided tunnel information of the protocol.

### Issue link
Issue #2037 

Closes #2037 

### Demo
image
Implemented test case passes.

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


boefjes/tests/test_nmap.py
```diff
@@ -163,3 +169,18 @@ def test_nmap_arguments_udp_top250_ipv6(self):
...
+
+    def test_normalizer(self):
+        input_ooi = parse_obj_as(
+            OOIType,
+            {
+                "object_type": "IPAddressV4",
+                "network": "Network|internet",
+                "address": "134.209.85.72",
+            },
+        )
```
ammar92, line 181: It's more convenient to just create the necessary object:
```suggestion
        input_ooi = IPAddressV4(network=Network(name="internet").reference, address="134.209.85.72")
```

boefjes/tests/test_nmap.py
```diff
@@ -163,3 +169,18 @@ def test_nmap_arguments_udp_top250_ipv6(self):
...
+        boefje_meta = get_boefje_meta(input_ooi=input_ooi.reference)
+        boefje_meta.arguments["input"] = serialize_ooi(input_ooi)
+        output = [x for x in run(get_normalizer_meta(boefje_meta), get_dummy_data("raw/nmap_mispoes.xml"))]
+        self.assertEqual(17, len(output))
+        self.assertEqual("https", output[12].name)
```
ammar92, line 186: So rather than checking on an index, because it's harder to read/ maintain, you could just check for the existence of this object since we already know the list may contain only one instance of this time:
```suggestion
        self.assertEqual(1, len(list(filter(lambda x: x.name == "https", output)))
```

boefjes/tests/test_nmap.py
```diff
@@ -163,3 +169,18 @@ def test_nmap_arguments_udp_top250_ipv6(self):
...
+            OOIType,
+            {
+                "object_type": "IPAddressV4",
+                "network": "Network|internet",
+                "address": "134.209.85.72",
+            },
+        )
+        boefje_meta = get_boefje_meta(input_ooi=input_ooi.reference)
+        boefje_meta.arguments["input"] = serialize_ooi(input_ooi)
+        output = [x for x in run(get_normalizer_meta(boefje_meta), get_dummy_data("raw/nmap_mispoes.xml"))]
```
ammar92, line 184: The list comprehension is redundant, you can just create a list out of the iterable

```suggestion
        output = list(run(get_normalizer_meta(boefje_meta), get_dummy_data("raw/nmap_mispoes.xml")))
```

ammar92 (commented):
Looks good but can be improved even more

boefjes/tests/test_nmap.py
originalsouth: Thanks, I changed the testing a bit to make it more generic.

boefjes/tests/test_nmap.py
```diff
@@ -174,7 +174,7 @@ def test_normalizer(self):
         boefje_meta.arguments["input"] = serialize_ooi(input_ooi)
         output = list(run(get_normalizer_meta(boefje_meta), get_dummy_data("raw/nmap_mispoes.xml")))
         self.assertEqual(17, len(output))
-        for i in range(0, len(output)):
+        for i in range(0, len(output) - 1):
```
ammar92, line 177: Using `enumerate` would even be better. You'll never get out of bounds and you get a tuple with the index and value for each iteration

ammar92: Because it's usually writing less code and making the code more runtime and memory-efficient. But also in your case, it is to prevent getting out of bounds (I assume that was your intention based on the commit message)

originalsouth: Have you looked at the subsequent code? I sincerely fail to see how `enumarate` would prevent an out-of-bounds?
Also the difference will be 8 characters.

ammar92: Yes, I've carefully reviewed the subsequent code. It's true that the use of 'enumerate' has minimal impact on character count, but the primary focus here is on enhancing code clarity and preventing potential out-of-bounds issues. Dealing with not accessing the last element can generally be done using slices. I'll leave it up to you

ammar92 (approved):

boefjes/boefjes/plugins/kat_nmap/normalize.py
```diff
@@ -44,9 +44,7 @@ def get_ip_ports_and_service(host: NmapHost, network: Network, netblock: Referen
             yield ip_port
 
             service_name = service.service
-            if port == 80:
-                service_name = "http"
-            if port == 443:
+            if service.tunnel == "ssl":
```
underdarknl, line 47: Are we convinced that SSL always means https? As many other services could also be carried over SSL/TLS.
Nmap itself seems to return ssl/https (possibly with an extra ? at the end if its not sure).

originalsouth: Good point. Addressed in cd27d89a1cc982065fbb467ee1b35401850462f7.

underdarknl (approved):

Issue #1956, opened 2023-10-23
Task events for scheduler


jpbruinsslot:
Related issues:

* https://github.com/minvws/nl-kat-coordination/issues/1161

Currently, we only have `created_at` and `modified_at` timestamp information for tasks. Status changes can only be derived from the status and the `modified_at`, any other mutations and events are lost.

It would be helpful to know how long a task was queued, and how long this task took when it was picked up and finished.

The most straightforward solutions seems to be adding more datetime columns for these state changes or adding an events table with a 1:m relation between a task and task events that then can be referenced from a task.

underdarknl:
Future work on this could also include which runner-instance executed the job, and how many resources where consumed (cpu, ram, network-io), as those could be useful for the scheduler to rank on for future jobs on similar objects.

jpbruinsslot:
During a discussion with Donny about the [current implementation](https://github.com/minvws/nl-kat-coordination/pull/2043), we condensed the following considerations:

### Considerations

* If we want, we can still opt to use columns with the runtime, queued, and duration fields instead of this adding this to an events table, however this can limit the extension of using additional information like cpu, memory, network i/o load, unless adding additional database columns.
* Usage of postgres trigger, could be considered as a bit of obfuscation, and a bit of magic that is done in the 'background'. We probably want to us Python instead, on tasks updates from the api
* Additionally we probably want to move away from those triggers if the task runner creates events like cpu, memory load etc. 
* Scalability, when considering high loads we can end up with a large events table, and perhaps partitioning should be considered, or compacting on date.
* The queries for duration, queued, and runtime should only be done when requested. It should not be done for the task list endpoint
* The fields type, context, and events can superfluous, and could be combined into 1 field and perhaps with dot notation

@underdarknl thoughts?

underdarknl:
* Adding columns would allow us to more easily add indices for events, however I'm not sure we have a good usecase for them. Having a single json blob with extensible fields is more likely to result in flexibility in consuming metrics as deliverable by the various plugin runners (eg, clouds have other resource counters, compared to local execution).
* Database triggers should not be used in OpenKAT I'd say, as it splits the business logic over two domains. Deterministic view logic (eg, combining tables, or calculating averages), is something that I would consider to be something you'd want to do in database land since it can be obviously named columns.
* the main usecase for these metrics is twofold:
** Show the user what resources a specific tool has consumed. This would cover all historical data, but is incidental and does not suffer from scalability issue as much.
[...]

Issue #1578, opened 2023-08-08
Show run time of plugin


zcrt:
**Is your feature request related to a problem? Please describe.**
Currently it is not clear from the interface how long a plugin (boefje, normalizer, bit) took to complete its tasks.

**Describe the solution you'd like**
- [ ] Showing the time each plugin took for each task in the interface
- [ ] Being able to filter on this value (e.g. show me all boefjes that took more than 1 hour to finish over the last 10 days)

underdarknl:
duplicate of https://github.com/minvws/nl-kat-coordination/issues/1161

PR #2043, opened 2023-11-20
Add task events to the scheduler

jpbruinsslot:
### Changes

This PR add task events functionality to the scheduler.

1. Adds an event to the events table on inserts and updates on the tasks table. E.g. a task is created or a task status is updated. This is done by leveraging postgresql triggers.
2. Exposes the ability to add events to the events table by a rest api endpoint

This allows us to:

1. Add calculated attributes to a task like how long a task was queued, how long the task ran for in the task runner, how long the task took in total.
2. Create additional events that are related to a task, e.g: how much cpu, memory, network i/o the task runner used.

### Considerations

* If we want, we can still opt to use columns with the runtime, queued, and duration fields instead of this adding this to an events table, however this can limit the extension of using additional information like cpu, memory, network i/o load, unless adding additional database columns.
* Usage of postgres trigger, could be considered as a bit of obfuscation, and a bit of magic that is done in the 'background'. We probably want to us Python instead, on tasks updates from the api
* Additionally we probably want to move away from those triggers if the task runner creates events like cpu, memory load etc. 
* Scalability, when considering high loads we can end up with a large events table, and perhaps partitioning should be considered, or compacting on date.
* ~~The queries for duration, queued, and runtime should only be done when requested. It should not be done for the task list endpoint~~ has been implemented
* The fields type, context, and events can superfluous, and could be combined into 1 field and perhaps with dot notation
* 
### TODO

* [x] Create EventStore
* [x] Create postgresql trigger on task updates to create event in the events table
* [x] Fix serialization between sqlalchemy and pydantic models
* [x] Implement events endpoint
* [x] Write tests
* [x] Make sure that list task view doesn't call the individual queries, that should be done on the detail view

### Issue link

Closes [1956](https://github.com/minvws/nl-kat-coordination/issues/1956)


mula/scheduler/server/server.py
```diff
@@ -567,6 +593,85 @@ def push_queue(self, queue_id: str, item: models.PrioritizedItem) -> Any:
...
+        except storage.filters.errors.FilterError as exc:
+            raise fastapi.HTTPException(
+                status_code=fastapi.status.HTTP_400_BAD_REQUEST,
+                detail=str(exc),
+            ) from exc
+        except ValueError as exc:
+            raise fastapi.HTTPException(
+                status_code=fastapi.status.HTTP_400_BAD_REQUEST,
+                detail=str(exc),
+            ) from exc
```
ammar92, line 645: These two clauses can be combined (`except FilterError, ValueError as exc: ...`)

ammar92 (dismissed):
Looks good in general and nice tests 👍 Some general remarks to reconsider:
- Although it's a bit challenging since you need to define these upfront (as much as possible), you might want to use enum fields in the events table (e.g. for type and event). This will improve performance in db time
- Instead of storing the complete task in the event, consider referencing it by its ID. This not only saves storage, but also (slightly) improves query response times. Keep in mind though, that this could potentially lead to more queries if the task data is required later. However, in such cases, it's worth looking at materialized views for such specific queries.

mula/scheduler/schedulers/scheduler.py
```diff
@@ -127,10 +141,20 @@ def post_pop(self, p_item: models.PrioritizedItem) -> None:
...
         self.ctx.datastores.task_store.update_task(task)
 
-        return None
+        # Create event
+        event = models.Event(
+            task_id=task.id,
+            type="events.db",
+            context="task",
+            event="update",
+            data=task.model_dump(),
```
dekkers, line 154: This will result in a very big event table, because the task model is pretty big already and this will be duplicated unnecessarily multiple times in the events table. I don't think this is a good idea with regards to performance and resource usage.

mula/scheduler/models/tasks.py
```diff
@@ -57,13 +79,22 @@ class Task(BaseModel):
 
     status: TaskStatus
 
+    duration: Optional[float] = Field(None, alias="duration", readonly=True)
+
+    queued: Optional[float] = Field(None, alieas="queued", readonly=True)
+
+    runtime: Optional[float] = Field(None, alias="runtime", readonly=True)
```
dekkers, line 86: Type should be `timedelta` instead of `float`.

mula/scheduler/models/events.py
```diff
@@ -1,11 +1,60 @@
...
+
+    data: dict
+
+
+class EventDB(Base):
+    __tablename__ = "events"
+
+    id = Column(Integer, primary_key=True)
+
+    task_id = Column(GUID)
```
dekkers, line 43: This should be a foreign key to the task table.

dekkers (commented):
As far as I can see the events table solution will scale a lot worse compared to simply storing the data in the task table. The task will only be executed once and duration/queued/runtime only need to be stored once for a task, so there is no need for 1:n relation table. Storing it as simple fields in the task table means:

- We can give the fields proper timedelta/interval types
- We can easily index it. While you can also add indexes to jsonb, that will always have  more unnecessary indirection and will thus be slower.
- Smaller database. This is important for performance, because in my experience performance doesn't decrease linearly with database size, but it goes downhill a lot faster when the database gets bigger.
- Half the number of queries, because the events table needs an extra insert every time we do an insert and update. 

With regards to data is not generated yet for runtimes that do not exist yet, I'd say it would be better to follow the YAGNI principle (https://en.wikipedia.org/wiki/You_aren't_gonna_need_it). And even if we already want to implement it, it would be better to just add a jsonb field in the task table for it.

jpbruinsslot:
> As far as I can see the events table solution will scale a lot worse compared to simply storing the data in the task table. The task will only be executed once and duration/queued/runtime only need to be stored once for a task, so there is no need for 1:n relation table. Storing it as simple fields in the task table means:
> 
> * We can give the fields proper timedelta/interval types
> * We can easily index it. While you can also add indexes to jsonb, that will always have  more unnecessary indirection and will thus be slower.
> * Smaller database. This is important for performance, because in my experience performance doesn't decrease linearly with database size, but it goes downhill a lot faster when the database gets bigger.
> * Half the number of queries, because the events table needs an extra insert every time we do an insert and update.
> 
> With regards to data is not generated yet for runtimes that do not exist yet, I'd say it would be better to follow the YAGNI principle (https://en.wikipedia.org/wiki/You_aren't_gonna_need_it). And even if we already want to implement it, it would be better to just add a jsonb field in the task table for it.

Yes, you bring up valid points. My consideration with the suggested changes was that keeping an event log opens up more flexibility to do more complex queries regarding the tasks (https://github.com/minvws/nl-kat-coordination/issues/1578). However, I agree we can opt for a more simple approach and cross that bridge when we get there.

jpbruinsslot:
Closing, superseded by https://github.com/minvws/nl-kat-coordination/pull/2214

PR #2063, opened 2023-11-28
Scheduler threading fixes

### LINKED ISSUE #1992: Scheduler threading optimizations
The "cannot schedule new futures after interpreter shutdown" means we try to start a new thread using `ThreadPoolExecutor` after the main thread has shutdown. The exception listed does not directly indicate the problems described, that would be another exception that should have been logged earlier.

I tried understanding what happens from reading the code, I think the following happens:

- An unhandled exception is caught by `run_forever` or `run_once` of the `ThreadRunner` class. Stop_event is set here, then the exception is reraised, which isn't handled so `unhandled_exception` is called.

- Because `stop_event` is set, the main thread will exit. This will call the shutdown code, which means  `ThreadPoolExecutor` will refuse to create new threads while joining all existing threads before the interpreter exits (this is written in the documentation and you can also see that if you read the code raising the exception in `/usr/local/lib/python3.11/concurrent/futures/thread.py`). The threadpool executor is used to push tasks and if the queue is full it will block forever, so main thread exiting will also block forever in case the queue is full.  (I first thought it only blocked for 5 seconds, but max_tries is actually always set to -1 because the subclass __init__ doesn't even allow setting anything else. The code would be easier to understand if the `push_item_to_queue_with_timeout` didn't have the `max_tries` and `timeout` parameters that are always the same value.)

- The `unhandled_exception` docstrings says it gracefully shutdowns threads, but joining threads doesn't shut them down. The join just waits 5 seconds for every thread that is still running. If I understand it correctly there are already a lot of threads when there are a lot of organisations and when the queue is full there will be even more. For every task pushed to the queue we create a thread and when the queue is full that thread will just wait and try to push the tasks, so there might be a lot of threads trying to push tasks. This means it will take a long time for `_exit` to be called at the end, causing the scheduler to partly hang for a long time with stop_event being set while some other parts of the scheduler keep running. This results in the "cannot schedule new futures after interpreter shutdown" exception.

Other observations for reading the code and trying to understand what is going on:

- The `stop_threads` method of the Scheduler class doesn't stop any thread because joining threads doesn't stop them.

- I don't understand why `stop_event.wait()` is called with a timeout of 0.01 in a loop in the main thread. This will unnecessarily wake up the main thread 100 times per second and just waste resources. Why not just call `wait()` without timeout?

- Stop_event isn't checked everywhere, so some parts won't have a graceful shutdown. This stack trace is one example, the `basic_consume()` will just keep listening as far as I understand, then call the callback, which will just execute even if stop_event is set. In this case you get an exception because it tries to spawn a new thread after the main thread has shutdown. Another place where stop_event isn't checked is in the uvicorn server. There might be more places.

- `App.server` is set to a different `Server` instance than the instance than the one that is started in the thread. It also isn't used anywhere, so I think the variable can just be removed?

- `push_item_to_queue_with_timeout` has a race condition, because it checks whether the queue is full, but there is no guarantee that no other thread doesn't push tasks to the queue between this check and `push_item_to_queue` putting the task on the queue.

_Originally posted by @dekkers in https://github.com/minvws/nl-kat-coordination/issues/1967#issuecomment-1785324702_

jpbruinsslot:
### Changes

Implementing and updating code based on findings from #1992

Note: gracefully shutting down the uvicorn server seems to be more intricate than expected. Picked this up in another PR.

### Issue link

Closes #1992


ammar92 (approved):
No remarks

jpbruinsslot:
@jpbruinsslot TODO: fix tests

PR #2071, opened 2023-11-29
Add benchmarking script to the scheduler

### LINKED ISSUE #2012: Structured way to do benchmarking
It would be nice to have a more structured way to do benchmarking. For instance, I'd opt for adding an integration test that executes your load script. However, this might require setting up more service-wide integration tests of course..

But then we could just mark these tests using `pytest.mark.benchmark` or `pytest.mark.performance` and configure pytest to either always skip these unless explicitly targeted, or skip them when some file with a lot of test data (organizations, jobs etc.) is not present. Then we can start keeping track of the performance over time and see the impact of changes more consistently, especially if we always take the performance results of the github actions?

_Originally posted by @Donnype in https://github.com/minvws/nl-kat-coordination/pull/1999#pullrequestreview-1715267340_

jpbruinsslot:
### Changes

Added a script to measure cpu/memory usage, error logs counts of the load.py script. Allows us to run the load.py script and see how a scheduler test run performs.

### Issue link

Closes #2012


mula/tests/scripts/benchmark.py
```diff
@@ -0,0 +1,163 @@
...
+        help="The pid of the process to monitor.",
+    )
+
+    # Parse arguments
+    args = parser.parse_args()
+
+    # Configure logging level, if the -v (verbose) flag was given this will
+    # set the log-level to DEBUG (printing all debug messages and higher),
+    # if -v was not given it defaults to printing level warning and higher.
+    default_loglevel = logging.INFO
```
ammar92, line 154: I think *level* suffices, *default* doesn't add anything
```suggestion
    level = logging.INFO
```

mula/tests/scripts/benchmark.py
```diff
@@ -0,0 +1,163 @@
...
+
+    # Parse stats
+    parse_stats()
+
+    # Parse logs
+    parse_logs("logs.txt")
+
+
+if __name__ == "__main__":
+    TIMEOUT_FOR_LOG_CAPTURE = 5
```
ammar92, line 128: Unused here? Or should be defined as a constant outside of this clause

ammar92 (approved):

underdarknl (approved):

PR #2095, opened 2023-12-06
Task rescheduling checks now for organization

Rieven:
### Changes

- When rescheduling task, it did not check if the task is bounded to the organization making that a member could guess the task id and trigger rescheduling from another organization. 
- This PR implements organization checks when rescheduling a task.

### Issue link
(https://github.com/minvws/nl-kat-coordination/issues/2094)

Closes https://github.com/minvws/nl-kat-coordination/issues/2094

### Demo
image

---

### Code Checklist
- [x] All the commits in this PR are properly PGP-signed and verified;
- [x] This PR only contains functionality relevant to the issue; tickets have been created for newly discovered issues.
- [ ] I have written unit tests for the changes or fixes I made.
- [x] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [x] I have performed a self-review of my code and refactored it to the best of my abilities.

### Communication
- [x] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [x] I have made corresponding changes to the documentation, if necessary.

---
## Checklist for code reviewers:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_code.md) into your comment.

---
## Checklist for QA:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_qa.md) into your comment.

PR #2115, opened 2023-12-11
Fix manual scan regression

dekkers:
### Changes
#2095 changed `schedule_task` function so that functionality is actually rescheduling a task given a task id. This didn't take into account that this function was also used on the ooi detail and boefje pages to manually start a scan and broke that functionality. This restores the `schedule_task` function and creates a new `reschedule_task`

I also changed the try/except to only catch `SchedulerError` because all the other exceptions are subclasses of `SchedulerError` and there is no need to list them.

---

---
## Checklist for QA:


ammar92 (approved):
No remarks. +1 for cleaning up the exceptions, I've seen similar cases lately where we should apply this abstraction as well

underdarknl (approved):

Issue #2081, opened 2023-12-04
[EPIC] Basic Security aggregate report


noamblitz:
#### Overview table

![image](https://github.com/minvws/nl-kat-coordination/assets/46660228/93325dfc-abfa-435e-a270-827e32029205)

### Checks voor basic security.

Kan aanvulling gebruiken.

Doorgestreept (nog) niet mogelijk.

### DNS

- [ ] ~~Geen DNS zone transfers mogelijk    (niet mogelijk is goed)~~
- [ ] ~~Geen Cache poisoning mogelijk       (niet mogelijk is goed)~~
- [ ] ~~Beschikt over meest recente software versie~~
- [x] Geen onnodige poorten open
  - ~Everything except 53 UDP?~
  - Check for Uncommon port finding
- [x] DNSSEC aanwezig
  - No `Hostname.<ooi[is Finding].finding_type` with id "KAT-NO-DNSSEC"
- [x] Valide DNSSEC chain
  - The one above and no `Hostname.<ooi[is Finding].finding_type` with id "KAT-INVALID-DNSSEC"

### Web

IMPORTANT: we should ignore the duplicate Findings on Hostname that come from the internetnl bit.

![image](https://github.com/minvws/nl-kat-coordination/assets/46660228/5d9e62b2-4c5e-47aa-aa73-e9c190006f74)

- [ ] ~~Beschikt over meest recente software versie~~
- [ ] ~~Geen gebruik van standaard wachtwoorden op login panels~~
- [x] CSP header aanwezig
  - ~No `Hostname.<hostname[is Finding].finding_type` with id "KAT-NO-CSP"~ -> Internetnl bit
  - No `Hostname.<hostname[as Website].website[as HTTPResource].<ooi[is Finding].finding_type`  with id "KAT-NO-CSP"
- [x] Veilige configuratie CSP header
  - ~The one above and no `Hostname.<hostname[is Finding].finding_type` with id "KAT-CSP-VULNERABILITIES"~ -> Internetnl bit
  - The one above and no `Hostname.<hostname[as Website].website[as HTTPResource].resource[as HTTPHeader].<ooi[is Finding].finding_type` with id "KAT-CSP-VULNERABILITIES"
- [x] Valide CSP header
  - This will at most result in a "KAT-CSP-VULNERABILITIES" with the current normalizer (so: see above)
- [ ] ~~Cookies bevatten de 'httpOnly' flag~~
- [x] Redirect van HTTP naar HTTPS
[...]

PR #2123, opened 2023-12-12
Add the Web basic security report

Donnype:
### Changes
Adds a web system basic security report.

### Issue link
Closes part of #2081

### Demo
image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


stephanie0x00:
## Checklist for QA:
- [] I confirmed that the PR's advertised `feature` or `hotfix` works as intended.

### What works:
Onboarding flow works. 
It's possible to generate the basic security report and it shows the compliance checks for all hosts. 

### What doesn't work:
See below. 

### Bug or feature?:
- Not all hosts have the table with findings, my guess would be that no compliance data is present for these sites.
- The compliance issue table lines are shown twice, see the marking in the screenshot below. 
- Some of the compliance issues contain the text 'None' with the risk level 'Pending'. 
- It would be nice to have this list view sorted based on the risk level as that helps to filter out which require action first. 
- There might be a bug in the check or boefje for security.txt, as transip.nl seems to have a security.txt file, but the compliance check says 0/1. Maybe because it redirects to https://examplec.com/security.txt (without the .well-known folder). 

security-report

transip

transip-sec

Donnype:
@stephanie0x00 I resolved most issues, but I am not sure what you mean by the first one? And I think the last one is due to a boefje issue since when I manually add a SecurityTXT object, the compliance issue is resolved for me..

stephanie0x00:
So for some of the hosts in the report it doesnt show the compliance issue table below the table with compliance checks. So the table of the first screenshot isn't added to all hosts in the report. Does that make it more clear?

Donnype:
@stephanie0x00 Ah right, indeed: most compliance check are the absence of a specific Finding, except the `SecurityTXT` and certificate checks where we look for `SecurityTXT` and `X509Certificate` OOIs. If the table is empty that means no Findings have been found, even when not all compliance has been fixed, as we have no recommendation and risk level mechanism for the `SecurityTXT` and `X509Certificate` issues. I added a check that only adds the table if there actually are elements to be shown.

Looking into this now I see that we can actually just leverage the `"KAT-NO-CERTIFICATE"` finding type, which should fix the latter one. This is fixed now, although we now have the same check for "offers https"... Perhaps we should consider a `"KAT-NO-SECURITY-TXT"` as well @noamblitz?

rocky/reports/report_types/web_system_report/report.py
```diff
@@ -0,0 +1,208 @@
...
+    name = _("Web System Report")
+    description = _("Web system reports check web systems on basic security standards.")
+    plugins = {
+        "required": [
+            "nmap",
+            "dns-records",
+            "testssl-sh-ciphers",
+            "ssl-version",
+            "ssl-certificates",
+            "webpage-analysis",
```
underdarknl, line 85: Shouldn' this also list the security.txt boefje?

underdarknl (approved):

PR #2113, opened 2023-12-11
Mail report

madelondohmen:
### Changes
This PR adds a system specific mail report. It checks if the following finding types are found:
* KAT-NO-SPF
* KAT-NO-DMARC
* KAT-NO-DKIM

For each hostname, it returns the mail security measures that have been found. It also returns the number of hostnames, spf, dmarc and dkim, which are used in report.html.

### Issue link
Closes #2111 

### Demo
afbeelding

---

### Code Checklist
- [x] All the commits in this PR are properly PGP-signed and verified;
- [x] This PR only contains functionality relevant to the issue; tickets have been created for newly discovered issues.
- [x] I have written unit tests for the changes or fixes I made.
- [x] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [x] I have performed a self-review of my code and refactored it to the best of my abilities.

### Communication
- [x] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [x] I have made corresponding changes to the documentation, if necessary.

---
## Checklist for code reviewers:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_code.md) into your comment.

---
## Checklist for QA:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_qa.md) into your comment.

PR #2125, opened 2023-12-12
Rename nmap tcp boefje back

dekkers:
### Changes
This renames the `nmap-tcp` boefje back to `nmap` so it kept enabled on upgrading. Also renames the TOP_PORTS argument back to its original name so that configuration is also preserved. I tested that on upgrade from 1.13 both are kept.

I also cleaned up the boefjes code.

---

---
## Checklist for QA:


boefjes/boefjes/plugins/kat_nmap_udp/main.py
```diff
@@ -43,17 +33,15 @@ def build_nmap_arguments(host: str, protocol: Protocol, top_ports: Optional[int]
 
 def run(boefje_meta: BoefjeMeta) -> List[Tuple[set, Union[bytes, str]]]:
     """Build Nmap arguments and return results to normalizer."""
-    top_ports = int(getenv("TOP_PORTS_UDP", TOP_PORTS_DEFAULT))
-    assert (
-        TOP_PORTS_MIN <= top_ports <= TOP_PORTS_MAX
-    ), f'{TOP_PORTS_MIN} <= {top_ports} <= {TOP_PORTS_MAX} fails. Check "TOP_PORTS" argument.'
+    top_ports = int(getenv("TOP_PORTS", TOP_PORTS_DEFAULT))
```
underdarknl, line 36: Would it be wise to not collide this ENV var with the other TOP_PORTS var used in the TCP boefje? For now these vars are inserted via the schema, but maybe there's a future usecase for setting these vars directly on the boefje workers instead of via KAT's katalogus settings?

underdarknl: Yes, This would clash with the feature used in boefjes/boefjes/job_handler.py where we ingest env vars prefixed with BOEFJE_ if no katalogus setting is present.

```
            if key.startswith("BOEFJE_"):
                katalogus_key = key.split("BOEFJE_", 1)[1]
                # Only pass the environment variable if it is not explicitly set through the katalogus,
                # if and only if they are defined in boefje.json
```

underdarknl:
We should also now revert: 
https://github.com/minvws/nl-kat-coordination/pull/2113/commits/cca889a09ad80f9fc74ac2fb9448db2591b5f2d8

boefjes/boefjes/plugins/kat_nmap_udp/main.py
dekkers: I've renamed the UDP var back to TOP_PORTS_UDP.

dekkers:
> We should also now revert: [cca889a](https://github.com/minvws/nl-kat-coordination/commit/cca889a09ad80f9fc74ac2fb9448db2591b5f2d8)

That was later changed again and the report currently doesn't list nmap as required plugin.

underdarknl (approved):

PR #2137, opened 2023-12-15
more descriptive first use related to user

underdarknl:
### Changes
Add a bit more explanation for first time users about the created superuser account.

### Code Checklist

### Communication

---

---
## Checklist for QA:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ endef
 kat: env-if-empty build up
 	@echo
 	@echo "The KAT frontend is running at http://localhost:8000,"
-	@echo "An inital superuser has been created"
+	@echo "An initial superuser has been created"
 	@echo "The username is stored in DJANGO_SUPERUSER_EMAIL in the .env-default file."
         @echo "run `grep 'DJANGO_SUPERUSER_EMAIL' .env` to find it."
 	@echo "The related password can be found as DJANGO_SUPERUSER_PASSWORD in the .env file."
--- a/Makefile
+++ b/Makefile
@@ -25,9 +25,11 @@ endef
 kat: env-if-empty build up
 	@echo
 	@echo "The KAT frontend is running at http://localhost:8000,"
-	@echo "An inital superuser has been created: superuser@localhost"
-	@echo "The related password can be found as DJANGO_SUPERUSER_* in the .env file."
-        @echo "run `grep 'DJANGO_SUPERUSER_' .env` to find it."
+	@echo "An inital superuser has been created"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -25,7 +25,9 @@ endef
 kat: env-if-empty build up
 	@echo
 	@echo "The KAT frontend is running at http://localhost:8000,"
-	@echo "credentials can be found as DJANGO_SUPERUSER_* in the .env file."
+	@echo "An inital superuser has been created: superuser@localhost"
+	@echo "The related password can be found as DJANGO_SUPERUSER_* in the .env file."
+        @echo "run `grep 'DJANGO_SUPERUSER_' .env` to find it."
```
ammar92, line 30: Imho the suggested change isn't a big improvement because of two things:
- `superuser@localhost` isn't a fixed value, so you might not want to describe it as such. It might be better to mention that the email/ account name is stored in the `.env` file as `DJANGO_SUPERUSER_EMAIL`.
- `The related password can be found as DJANGO_SUPERUSER_*` could be clearer by specifying `DJANGO_SUPERUSER_PASSWORD` explicitly instead of using the wildcard `*`

I suggest we should just mention that the credentials are stored as `DJANGO_SUPERUSER_EMAIL` and `DJANGO_SUPERUSER_PASSWORD` in the `.env` file. The use of a fixed value like `superuser@localhost` may not be appropriate due to its variability and explicitly mentioning `DJANGO_SUPERUSER_PASSWORD` makes it more clear we're referencing a password variable

underdarknl: DJANGO_SUPERUSER_EMAIL. does not exists after running make kat as far as I can see?
Agreed on the other point, unless we do in fact get both vars out of that single command if Email does exists.

ammar92: > DJANGO_SUPERUSER_EMAIL. does not exists after running make kat as far as I can see? Agreed on the other point, unless we do in fact get both vars out of that single command if Email does exists.

`DJANGO_SUPERUSER_EMAIL` exists in the `.env-defaults` file. I'm not sure how the make command currently works, but if you're using the Docker Compose environment it will probably use that variable when building KAT. In any case, `createsuperuser --no-input` will use the environment variables `DJANGO_SUPERUSER_EMAIL` and `DJANGO_SUPERUSER_PASSWORD` to create the superuser account.

ammar92 (approved):

ammar92 (approved):

dekkers (approved):

PR #2152, opened 2023-12-17
Improve system identification

dekkers:
### Changes
This is partly some changes already done on the aggregate report branch. I've also pushed those changes there, but I'm opening a seperate PR so we can have discussion about it if necessary.

This adds some more service mappings. HTTP was missing, meaning that every website was classified as both "other" via services and as "web" via Website OOI. It also doesn't classify the system as "other" if it isn't listed in software mapping, because classifying everything that isn't dicom as "other" doesn't make much sense to me.

I've also removed some services that I either can't find in the nmap services file (tsdns, smtp-stats, smtp-proxy, mail-admin), can't find what it is so probably old and irrelevant (mailq) or is something different than normal DNS (mdns).

I also sort the list of services so they are always in the same order.

---

---
## Checklist for QA:

### DIFF
--- a/rocky/poetry.lock
+++ b/rocky/poetry.lock
@@ -3733,6 +3733,22 @@ doc = ["sphinx"]
 test = ["pytest", "pytest-cov"]
 
 [[package]]
+name = "strenum"
+version = "0.4.15"
+description = "An Enum that inherits from str."
+optional = false
+python-versions = "*"
+files = [
+    {file = "StrEnum-0.4.15-py3-none-any.whl", hash = "sha256:a30cda4af7cc6b5bf52c8055bc4bf4b2b6b14a93b574626da33df53cf7740659"},
+    {file = "StrEnum-0.4.15.tar.gz", hash = "sha256:878fb5ab705442070e4dd1929bb5e2249511c0bcf2b0eeacf3bcd80875c82eff"},
+]
+
+[package.extras]
+docs = ["myst-parser[linkify]", "sphinx", "sphinx-rtd-theme"]
+release = ["twine"]
+test = ["pylint", "pytest", "pytest-black", "pytest-cov", "pytest-pylint"]
+
+[[package]]
 name = "svglib"
 version = "1.5.1"
 description = "A pure-Python library for reading and converting SVG"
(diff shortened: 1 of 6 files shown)


Donnype (dismissed):
Unless we say that we should ignore the http services in favor of the Website OOI (@noamblitz ?): LGTM!

noamblitz:
Website OOI is better/more future proof. But this is more consistent. So im fine with this.

stephanie0x00:
## Checklist for QA:

### What works:
Onboarding flow works as expected. Can enable boefjes that scan as intended. Have generated the system report and it looks good. 

### What doesn't work:
- System type for IPv6 hosts.

### Bug or feature?:
N/a.

noamblitz:
Does your ipv6 have ipservices/ipports?

stephanie0x00:
Yep. I scanned various domains which all have IPv4 and IPv6 enabled. No output is shown for the IPv6 hosts.

noamblitz:
I mean, did nmap scan this ipv6 address and did it find open ports with services. System recognition works with ipservices. If the ipv6 is not scanned, or if there are no open ports found (for example because nmap runs in Docker), then it wont recognise any systems.

underdarknl:
> I mean, did nmap scan this ipv6 address and did it find open ports with services. System recognition works with ipservices. If the ipv6 is not scanned, or if there are no open ports found (for example because nmap runs in Docker), then it wont recognise any systems.

noamblitz:
We discussed, and this is indeed the case

stephanie0x00:
I know it's an docker connectivity issue. I have the required fixes locally to enable docker for main which I'll push.

dekkers:
The issue I had was that just giving the OpenKAT containers IPv6 via docker compose is not enough. The boefje will start its own container using API call to the docker socket and that container also needs IPv6. If I configure containers to have IPv6 by default using https://docs.docker.com/config/daemon/ipv6/#use-ipv6-for-the-default-bridge-network then nmap will work and I will also see system types for IPv6 hosts:

image

underdarknl (approved):

Issue #2161, opened 2023-12-18
Starting a manual scan doesn't work


dekkers:
**Describe the bug**
Starting a manual scan on boefje detail page doesn't work.

**To Reproduce**
Go to boefje page, select OOIs and start scan.

**Expected behavior**
The boefje is successfully started.

**Logs**
```
boefje_1  | [2023-12-18 21:35:34 +0000] [1] [ERROR] [app] Popping task from scheduler failed, sleeping 10 seconds
boefje_1  | Traceback (most recent call last):
boefje_1  |   File "/app/boefjes/boefjes/app.py", line 105, in _fill_queue
boefje_1  |     p_item = self.scheduler_client.pop_item(queue_type.id)
boefje_1  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
boefje_1  |   File "/app/boefjes/boefjes/clients/scheduler_client.py", line 104, in pop_item
boefje_1  |     return TypeAdapter(Optional[QueuePrioritizedItem]).validate_json(response.content)
boefje_1  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
boefje_1  |   File "/usr/local/lib/python3.11/site-packages/pydantic/type_adapter.py", line 256, in validate_json
boefje_1  |     return self.validator.validate_json(__data, strict=strict, context=context)
boefje_1  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
boefje_1  | pydantic_core._pydantic_core.ValidationError: 4 validation errors for nullable[QueuePrioritizedItem]
boefje_1  | data.BoefjeMeta.id
boefje_1  |   UUID input should be a string, bytes or UUID object [type=uuid_type, input_value=None, input_type=NoneType]
boefje_1  |     For further information visit https://errors.pydantic.dev/2.5/v/uuid_type
boefje_1  | data.NormalizerMeta.id
boefje_1  |   UUID input should be a string, bytes or UUID object [type=uuid_type, input_value=None, input_type=NoneType]
boefje_1  |     For further information visit https://errors.pydantic.dev/2.5/v/uuid_type
boefje_1  | data.NormalizerMeta.raw_data
[...]

PR #2166, opened 2023-12-20
Remove default setting of id to None

jpbruinsslot:
### Changes

Remove default setting of id to None. Since pydantic 2.0 setting an optional field that is not required, a None value is suggested as default however creating a new instance will create this default value.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/2161

### DIFF
--- a/mula/scheduler/server/server.py
+++ b/mula/scheduler/server/server.py
@@ -544,8 +544,10 @@ class Server:
 
             if s.queue.item_type == models.BoefjeTask:
                 p_item.data = models.BoefjeTask(**item.data).dict()
+                p_item.id = p_item.data["id"]
             elif s.queue.item_type == models.NormalizerTask:
                 p_item.data = models.NormalizerTask(**item.data).dict()
+                p_item.id = p_item.data["id"]
             else:
                 p_item.data = item.data
         except Exception as exc:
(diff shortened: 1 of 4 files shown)


ammar92 (approved):

ammar92 (approved):

underdarknl:
I think we use the Id's on various pages to allow the user to 'inspect' or 'download the raw files' for any given job, right? So removing them would be unwise. 
It would be more obvious if the save/create endpoints on the scheduler just don't accept these id's or clean them out. Or, if we do want to clean them in Rocky, we do so before sending the model out to the scheduler.

jpbruinsslot:
Yeah think we need to do the same as in the scheduler make seperate request and response models

underdarknl:
Im not sure. The scheduler picks up the job from Rocky Just fine. It creates it, and dispenses it to the worker. However, the worker receives this Job with an ID of None as opposed to a UUID as what the scheduler does for its own jobs. Maybe we just need to fill in the null value in the scheduler, regardless of where the job came from (rocky vs the scheduler itself)

underdarknl:
Getest: 
ik kan reschedulen, schedulen vanaf de object page, en schedulen vanaf de boefje page.  
Ook kan ik bij zowel normalisers als boefjes de raw file downloaden.

underdarknl (approved):

PR #2205, opened 2023-12-22
Download report JSON and ReportData OOI

### LINKED ISSUE #1970: Add Json output link to html report  page.
For each report, we should also give the user the option to download the raw data that went into the template as a Json file.

Allow the user to click on a Json download link. This json should contain the same data that the template itself consumed.

dekkers:
### Changes
This adds the functionality to download the report data as JSON. The Download JSON button is quick&dirty copy of the Download PDF button and will probably need some UX improvement.

It also adds a ReportData OOI and normalizer so we can upload the JSON as raw file with type `openkat/report-data`. 

The `str()` calls are necessary because the JSON encoder wants strings as keys and not the translation proxy objects.

Closes #1970

---

---
## Checklist for QA:

### DIFF
--- a/rocky/reports/report_types/aggregate_organisation_report/report.py
+++ b/rocky/reports/report_types/aggregate_organisation_report/report.py
@@ -143,7 +143,8 @@ class AggregateOrganisationReport(AggregateReport):
         for ip, compliance in rpki["rpki_ips"].items():
             ip_services = systems["services"][str(ip)]["services"]
 
-            for service in ip_services:
+            for service in services:
+                service = str(service)
                 if service not in basic_security["rpki"]:  # Set initial value
                     basic_security["rpki"][service] = {
                         "rpki_ips": {},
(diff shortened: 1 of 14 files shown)


rocky/reports/utils.py
```diff
@@ -0,0 +1,27 @@
...
+            logger.error("Key %s is type %s, path is %s", key, type(key), path)
+        if isinstance(value, dict):
+            debug_json_keys(value, path + [key])
+
+
+class JSONEncoder(DjangoJSONEncoder):
+    def default(self, o):
+        if isinstance(o, OOI):
+            return str(o)
+        elif dataclasses.is_dataclass(o):
```
underdarknl, line 24: can just be an IF

rocky/reports/utils.py
```diff
@@ -0,0 +1,27 @@
...
+            return dataclasses.asdict(o)
+        else:
```
underdarknl, line 26: Not needed due to returns on previous branches.

rocky/reports/utils.py
ammar92: The if-elif-else clause here is readable and a common pattern (maybe to be replaced soon by Python's newer structural pattern matching). I don't think it's redundant in this case and I don't have to think about the return strategy of this specific function.

rocky/reports/report_types/aggregate_organisation_report/report.py
```diff
@@ -338,16 +339,16 @@ def is_mail_compliant(result):
 
         summary = {
             # _("General recommendations"): "",
-            _("Critical vulnerabilities"): total_criticals,
-            _("IPs scanned"): total_ips,
-            _("Hostnames scanned"): total_hostnames,
+            str(_("Critical vulnerabilities")): total_criticals,
```
ammar92, line 342: Would it be possible to process translation proxy objects in the JSON serializer to not have to think about stringifying these objects?

underdarknl: Maybe, lets create a technical design ticket around these issues, as we'd like the json to be untranslated, where the html/pdf views would need to be translated.

underdarknl (approved):

rocky/reports/report_types/aggregate_organisation_report/report.py
dekkers: The Django encoder already does this, but processing in the JSON encoder that you can easily override is for values, not for keys. If we want to have everything multilingual we should indeed not have translated strings in the JSON and do the translation when we generate the multi organisation report.

Issue #2207, opened 2023-12-22
Preselection for OOIs for Reports


Rieven:
When making a report for Generate or Aggregate report, we want a pre-selected OOIs based on:
- OOIs with clearance levela L1, L2, L3, L4. (excluding L0)
- OOIs which has a declared clearance type

underdarknl:
We should check what the minimal required clearance level of required boefjes is. If that is level 2, then allowing the user to select level 1 ooi's doesn't make sense.

PR #2208, opened 2023-12-22
Preselected OOIs for Reporting

Rieven:
### Changes
When selecting OOIs for 'generate' or 'aggregate' reports a pre-selection of OOIs (filtered OOIs) is handy, that is:
- Filter for clearance levels L2- L4 (So excluding all L0's and L1's )
- Clearance types: only declared OOIs
- Fix bug in OOI List which did not count the total OOIs right, after filtering it must conserve the intital counts on all OOIs
- Remove MultipleOOIMixin which had only 1 method and having a redundant method to get OOIList. The OOI listing can be directly feched from the api. Called also onyl at 1 place.
- Currently filtered on OOI list of reports

### Issue link
https://github.com/minvws/nl-kat-coordination/issues/2207

Closes https://github.com/minvws/nl-kat-coordination/issues/2207

Generate report:
image

Aggregate report
image

More clearly when you open filter options which filters are set:
image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


rocky/reports/views/base.py
```diff
@@ -28,8 +28,11 @@ class ReportType(TypedDict):
 class ReportBreadcrumbs(OrganizationView, BreadcrumbsMixin):
     current_step: int = 1
 
-    def get_selection(self):
-        return "?" + urlencode(self.request.GET, True)
+    def get_selection(self, pre_selection: Optional[Dict[str, Union[str, List[str]]]] = None) -> str:
+        selection = self.request.GET.copy().dict()
+        if pre_selection:
+            selection.update(pre_selection)
```
Donnype, line 34: I think this will end up always overwriting the selection with the pre_selection, which is not what we want I suppose since it should act as a default when there is no selection

rocky/rocky/views/mixins.py
```diff
@@ -261,28 +261,6 @@ def __getitem__(self, key: Union[int, slice]) -> List[HydratedFinding]:
         raise NotImplementedError("FindingList only supports slicing")
 
 
-_EXCLUDED_OOI_TYPES = ("Finding", "FindingType")
-
-
-class MultipleOOIMixin(OctopoesView):
```
Donnype, line 267: This is unused (now)?

rocky/reports/views/aggregate_report.py
```diff
@@ -59,7 +59,13 @@ class LandingAggregateReportView(BreadcrumbsAggregateReportView, TemplateView):
...
 
     def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
-        return redirect(reverse("aggregate_report_select_oois", kwargs=self.get_kwargs()))
+        kwargs = self.get_kwargs()
+        pre_selection = {
+            "clearance_level": ["1", "2", "3", "4"],
+            "clearance_type": "declared",
+        }
+        selection = self.get_selection(pre_selection)
+        return redirect(reverse("aggregate_report_select_oois", kwargs=kwargs) + selection)
```
Donnype, line 68: We can avoid string manipulations and add the selection as a dict to the kwargs as well

rocky/reports/views/generate_report.py
```diff
@@ -58,7 +58,13 @@ class LandingGenerateReportView(BreadcrumbsGenerateReportView, TemplateView):
...
 
     def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
-        return redirect(reverse("generate_report_select_oois", kwargs=self.get_kwargs()) + self.get_selection())
+        kwargs = self.get_kwargs()
+        pre_selection = {
+            "clearance_level": ["1", "2", "3", "4"],
+            "clearance_type": "declared",
+        }
+        selection = self.get_selection(pre_selection)
+        return redirect(reverse("generate_report_select_oois", kwargs=kwargs) + selection)
```
Donnype, line 67: Same here

rocky/reports/views/aggregate_report.py
Donnype: Also I see that the default return value of `self.get_kwargs()` is just the organization code. Why is the organization a query parameter (as well)?

docs/source/developer_documentation/rocky.md
```diff
@@ -207,8 +207,6 @@ direction RL
     OOIAddView <|-- BaseOOIFormView
     FindingAddView <|-- BaseOOIFormView
 
-    MultipleOOIMixin <|-- OctopoesView
-    BaseOOIListView <|-- MultipleOOIMixin
```
Donnype, line 211: 👍

rocky/rocky/views/ooi_view.py
```diff
@@ -61,20 +56,42 @@ def get_ooi_profile_types(self) -> Set[ScanProfileType]:
...
+        # A filtered OOI list
+        return OOIList(
+            self.octopoes_api_connector,
+            ooi_types=self.get_ooi_types(),
+            valid_time=self.get_observed_at(),
             scan_level=self.get_ooi_scan_levels(),
             scan_profile_type=self.get_ooi_profile_types(),
         )
 
+    def get_all_oois_and_count(self) -> Tuple[int, OOIList]:
```
Donnype, line 77: I think we can just return `all_oois` here and do `len()` in the `setup` method. Though I was wondering: why can't we use the `get_queryset` method for that? Since we are using the `setup` method I suppose a lot of the used properties might not have been set versus the way we retrieve them in `get_queryset`, but I do not have the Django View execution order fresh in mind at this point.

Donnype (dismissed):

rocky/rocky/views/ooi_view.py
Rieven: Yes, the Django way, the setup method is called before get_queryset, we can off course move everything into the get_queryset, it was more of the way we use setup, to initiate things there only.

rocky/reports/views/aggregate_report.py
Rieven: The kwargs itself is not a query parameter, it needs the `organization_code` in `urls.py`. This is by design this way when we implemented the switch or orgs using this kwarg. The selection part will come after the kwarg.

Rieven: If we want to define a kwargs for oois selection is also possible, then in `urls.pu` we need to explicitely set a `<ooi_selection>` something like this so it can read it. We don't need to do this for query parameters.

rocky/rocky/views/ooi_view.py
Rieven: I have use `setup` here foremost of how the information is distributed across the View. The filtered items will also be shown as text and also for `get_context` to use in templates

Rieven: > Yes, the Django way, the setup method is called before `get_queryset`, we can off course move everything into the `get_queryset`, it was more of the way we use setup, to initiate things there only.

rocky/rocky/views/mixins.py
Rieven: Yes, it is not needed anymore as we can directly do an api call

underdarknl (approved):

PR #2214, opened 2023-12-28
Alternative: Add task events to the scheduler

jpbruinsslot:
### Changes

This PR add task events functionality to the scheduler.

1. Tracks status change events for tasks
2. Allows and adds information about the task lifetime with new fields: `queued` (how long was the task on the queue), `runtime` (how long did the taskrunner take), `duration` (how long total time of processing of the task from creation to finalization). These new fields are exposed by the api. Since these fields are `@computed_field` we can also opt for adding these to their own jsonb field if these values regularly need to be queried.
3. `events` field added to `Task`, and is also exposed by the api, such that events (status change events for now) can be inspected through the api.

Changes to initial PR (https://github.com/minvws/nl-kat-coordination/pull/2043)

1. Adds a task events table, instead of an abstracted events table
4. Only inserts in tasks events table on specific events, in this case only on status change events (`post_push`, `post_pop`, `patch_task`)
5. Entire task object is **not** persisted in a jsonb field, only specific field for the event, in case of status change `from_status`, `to_status`


jpbruinsslot:
Closing in favor of https://github.com/minvws/nl-kat-coordination/pull/2451

PR #2284, opened 2024-01-08
Fix missing env var prefix in documentation

dekkers:
### Changes
Currently the environment variables in the documentation are wrong because they all miss the prefix. I think this is since we updated pydantic, pydantic-settings and settings-doc. I've fixed settings-doc to include the prefix, this PR changed settings-doc to my fork while we wait for the PR to be merged and a new version released to PyPI.

---

---
## Checklist for QA:

### DIFF
--- a/.github/workflows/build_docs_on_pr.yml
+++ b/.github/workflows/build_docs_on_pr.yml
@@ -25,7 +25,7 @@ jobs:
           cache: 'pip' # caching pip dependencies
 
       - name: Install pip dependencies
-        run: pip install -r requirements.txt
+        run: grep -v git+https:// requirements.txt | pip install -r /dev/stdin ; grep git+https:// requirements.txt | pip install -r /dev/stdin
 
       - name: Compile static HTML
         run: make docs
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -36,7 +36,7 @@ jobs:
           cache: 'pip' # caching pip dependencies
 
       - name: Install pip dependencies
-        run: pip install -r requirements.txt
+        run: grep -v git+https:// requirements.txt | pip install -r /dev/stdin ; grep git+https:// requirements.txt | pip install -r /dev/stdin
 
       - name: Compile static HTML
         run: make docs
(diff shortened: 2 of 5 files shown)


ammar92 (commented):
I'm still a bit hesitant when it comes to including external hotfixes because of the obvious reasons (keeping stuff up to date, missing dependabot alerts, etc.), but I think this could work as a temporary solution until, as you mentioned, this modification gets approved and merged in the designated package.

I was also wondering and currently still looking into this if it's possible to achieve this with the template (as it worked before?), but you've probably already looked into that. 

Also, does this work for aliased fields? For example, the scheduler has a `DEBUG` variable, but based on the submitted PR I can infer that this would always produce prefixed fields like `SCHEDULER_DEBUG` and I'm not sure if that's even correct and also loaded by `pydantic-settings` as the expected environment variable.

ammar92:
> I was also wondering and currently still looking into this if it's possible to achieve this with the template (as it worked before?), but you've probably already looked into that.

After some reading and experimenting, I found out we can reach the `model_config.env_prefix` attribute using the `classes` context variable in the template. With some effort we might be able to produce the same using only template code.

dekkers:
I forgot to mention the PR I opened, but it is https://github.com/radeklat/settings-doc/pull/29

ammar92:
> In my experience maintainers rarely release new versions without merging open PRs or giving feedback.

While you might experience that, nothing indicates that this holds the same for this project, since the project doesn't have any real contributors expect for the owner (see issues and PRs).

> I couldn't figure it out quickly. Given that the env_prefix is part of pydantic-settings, in my opinion settings-doc should also support that and it is the right place to fix this.

Agree, this should be part of `settings-docs`. But it isn't part of it yet, and maybe never will be. While with the templates it can already be achieved.

> You would also need to make sure that aliased fields don't get the prefix. It might be possible, but it would also be double work, because there is already a fix for settings-doc.

Shouldn't be too hard to apply conditional formatting in Jinja templates: `{% if field.alias %}{{ field.alias }}{% else %}{{ prefix }}{{ env_var }}{% endif %}`

> And while we are dicussing this, everybody who goes to OpenKAT documentation gets the wrong information and potentially wastes a lot of time before figuring out that our documentation is currently just plain wrong.

Agree, therefore we should quickly do something. Ultimately it comes down to whether we want a "corrupted" dependency tree or simply to use the template. I'll leave it up to you and @underdarknl

Issue #1963, opened 2023-10-25
NXDOMAIN keeps findings in KAT


zcrt:
**Describe the bug**
Suppose a domain is forgotten and has some vulnerabilities. This domain is correctly identified as such and reported by KAT. The owner of the domain then continues to solve the problems. One easy solution is taking down the domain. However, KAT wil not remove its findings once a domain resolves to NXDOMAIN.

**To Reproduce**
1. Have a domain with findings
2. Take down the domain
3. Notice the findings do not dissappear

**Expected behavior**
No more findings as long as the domain resolves to NXDOMAIN.

**OpenKAT version**
1.12.1

underdarknl:
Are these findings bound directly to the domain, or are the present on objects related to this domain (like a webserver?)
In the later case, those related objects should be cleaned up once the responsible boefjes (dnsrecords) cannot proof their existence anymore.
If findings are directly related to the domain, but are created by boefjes that now don't longer have the ablity to run (due to the webserver being removed), we need to find another way to remove them. Ideally we'd make sure that boefjes only create objects related to their input OOI's, and not somewhere else in the graph to avoid this problem.

zcrt:
In this case we saw `dnsrecord` fetch `NXDOMAIN` multiple days ago, but still there exist findings about invalid certificates ('objects related to the domain') for this domain

underdarknl:
Yup, so that means at some point there where dns-records, and an ip-adres, ip-port and service which where the input for the Tls boefjes. Those plugins then coupled the findings no to the service (which has now probably been deleted), but to the Domain. Since there is now no way to invalidate these results they will never be proven as solved.
We should either make sure objects are not bound 'out of scope' into the graph (like what happened here), or we should make sure that the origin (the input-ooi for the tls tools) of these findings triggers a delete on them regardless of where these findings where added in the graph.
@originalsouth this seems like a good first issue for your discovery of how deletion propagation works in Octopoes, and what Origins are in the model.

originalsouth:
> Since there is now no way to invalidate these results they will never be proven as solved.
Right, this goes directly against the Octopoes' OWA...

originalsouth:
Steps to replicate:
1. Spin up a fresh install add an organization and the like
1. Disable all boefjes and solely turn on the "Wappalizer - Software Scan" boefje
1. Have the following OOIs in place on L2: 
    * Network[`internet`] 
    * Hostname[`mispo.es`]
[...]

PR #2310, opened 2024-01-10
Fix/1963 nxdomain keeps findings in kat

originalsouth:
### Changes
- Mock Octopoes' event manager for integration testing and write tests exploiting the mocks,
- Regulate output of Boefjes so that self proving OOI's get deleted in the case of an NXDOMAIN (tested manually since because there is no way to that with the current integration tests) and change tests accordingly,
- Turn output OOI's of hydrating Boefjes into declarations and change tests accordingly.

### Issue link
Adressess #1963 (issues with Wappalyzer output remains because they run on HostnameHTTPURL OOIs which is to be addressed in a later stage).

### Demo
URL: https://mispo.es
Boefjes: 

- DNS Records 
- Wappalyzer
- NMAP TCP 

Findings yield:

image

Then triggering an NXDOMAIN (simulated with a pihole after overriding the default DNS)  yields only Wappalyzer findings (as explained above):

image

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


ammar92 (commented):
In most of the submitted normalizers, their `yield` clashes with the expected return type `OOI`. I'm not sure anymore how this exactly works, but usually normalizers yield OOIs, and the job handler (or Rocky when creating OOIs manually) creates a `Declaration` from the given OOI and saves it to Octopoes.

dekkers:
Declarations are for objects declared by the user (as defined in https://docs.openkat.nl/developer_documentation/octopoes.html). Changing the findingtypes to declarations looks wrong to me, because the findingtypes are not declared by the user.

As far as I can see we change this so that we have something to filter on, but:
- Shouldn't we fix the normalizers that return the input OOI instead of creating a workaround that silently ignores certain normalizer output?
- If we want to throw an error when this happens for normalizers that are not supposed to do this we could accomplish that in other ways. For example add some flag to the normalizer metadata that it supposed to return the input OOI and only give an error when the flag isn't set or add a new hydration origin type that can be used. Not sure what the best way to do this is, but I think there is a better ways than pretending that the user declared the findingtypes.
- In my opinion having clear errors is better than silently ignoring normalizer output.

noamblitz:
I agree with most of your points @dekkers. However, some normalizer should return their input ooi by design... We could however choose to not add an origin in that case (which keeps the hydration but reduces issues).

Donnype:
@dekkers I think one issue here is fact that Declarations, being the only models in Octopoes not subject to deletion propagation, are tied to a "user" by definition. (To me this makes less sense at least, given that Octopoes has no notion of users.) But the practical issue with this, and reason to have added this in the first place, is the calvin normalizer: there we want to return an arbitrary amount of objects that should not be deleted after a new run. Here we decided to use declarations as well.

With respect to the input ooi: allowing normalizer to return an observation of the input_ooi for update purposes will have the side effect of the input_ooi becoming self-referencing (which is roughly the "technical definition" of a declaration). I was not sure if it makes more sense to crash than to do what users in most cases expect unless they are specifically looking for certain behavior - in which case I'd recon they'd read the documentation where we could inform them. But I guess the hydration (or "update") use-case can be quite common.

I do think in terms of Octopoes we might want to save the input ooi in a declaration separately so that all self-referencing objects are in Origins with type "declaration" and only have the input ooi in their result set. So the error can suggest to either not yield the input ooi or yield it as a Declaration?

Donnype:
After a talk with @underdarknl and @originalsouth, the conclusion was that there indeed is something between an `Observation` and a `Declaration` that should allow persistent OOIs, but using Declarations here is not optimal. So the first step would be to implement @dekkers suggestions of reverting the use of declarations for hydration and either:
- Add a way to communicate that yielding input oois is allowed for the plugin and crash otherwise
- Check the current plugins and only show warnings when this happens, until we have a way to yield objects that are not affected by deletion propagation.

underdarknl:
I see the following scenario:

An OOI gets created because it is observed by a Boefje+normalizer or BIT, This origin only has partial details about the OOI. However, this origin in the reason why the object Exists.
Another boefje+normaliser or Bit also observes the OOI and also observes some (possibly overlapping, possibly conflicting) information. We should in theory create a second OOI with the same PK, and retain both sets of information with their respective origins as we would need to be able to remove either one or both of the origins if their observation is no longer reproducible.

Now, for Hydration, these tools don't really observe that an OOI exists in our network, and thus are not tied to our deletion logic. They simply add information about an object that has already been proves to be relevant.  You might call this Hydration, or supplementing the dataset. But, being able to hydrate or supplement does not proof to us the OOIs needs to be part of our graph. Deletion should still hinge on the original origin(s). 

Two ways out of this as far as I see:
* Create multiple OOIs with multiple proofs, containing multiple datasets.
  * Some of these OOIs can be picked up by deletion propagation (eg, when their origin is an observation)
  * Some of these OOIs exists only when their sibling OOI exists. (maybe similar to deletion propagation, use the original OOI + hydration boefje + normalizer as the origin)
  * Somehow combine these data-sets into a coherent OOI for consumption by other api-users. (what rules would apply?)
* Create a hydrated-Finding next to the current finding and don't allow hydration or even data-changes by another normalizer than the one in the initial origin of any OOI.

boefjes/boefjes/job_handler.py
```diff
@@ -209,15 +209,26 @@ def handle(self, normalizer_meta: NormalizerMeta) -> None:
             results = self.job_runner.run(normalizer_meta, raw)
             connector = self.octopoes_factory(normalizer_meta.raw_data.boefje_meta.organization)
 
+            logger.info("Obtained results %s", str(results))
+
             for observation in results.observations:
+                for result in observation.results:
+                    if self._parse_ooi(result).primary_key == observation.input_ooi:
```
underdarknl, line 216: We are doing self._parse_ooi tree times here. lets store the results in a var, and reuse the data.

originalsouth: Addressed in d0f04700de0211e53ed572da84b08dcddc67cce5.

originalsouth:
Merge conflicts obtained with automatic merging with b42b21cfcf0aac3633f962ccad67df5ea5241efd.

octopoes/octopoes/api/models.py
```diff
@@ -65,3 +74,12 @@ class ValidatedDeclaration(BaseModel):
...
     task_id: Optional[uuid.UUID] = Field(default_factory=lambda: uuid.uuid4())
+
+
+class ValidatedAffirmation(BaseModel):
+    """Used by Octopoes API to validate and parse correctly"""
+
+    ooi: OOIType
+    valid_time: AwareDatetime
+    method: Optional[str] = "hydration"
+    task_id: Optional[uuid.UUID] = Field(default_factory=lambda: uuid.uuid4())
```
ammar92, line 85:
```suggestion
    task_id: Optional[uuid.UUID] = Field(default_factory=uuid.uuid4)
```

ammar92 (approved):
One small suggestion, but looks good otherwise

octopoes/octopoes/api/models.py
```diff
@@ -73,7 +73,7 @@ class ValidatedDeclaration(BaseModel):
     ooi: OOIType
     valid_time: AwareDatetime
     method: Optional[str] = "manual"
-    task_id: Optional[uuid.UUID] = Field(default_factory=lambda: uuid.uuid4())
+    task_id: Optional[uuid.UUID] = Field(default_factory=uuid.uuid4())
```
ammar92, line 76:
```suggestion
    task_id: Optional[uuid.UUID] = Field(default_factory=uuid.uuid4)
```

ammar92 (approved):
No more remarks

PR #2315, opened 2024-01-11
Adds CAA records to the model, boefje, normalizer, adds a check bit and a finding

underdarknl:
### Changes
Adds CAA records to OpenKAT

### Demo
image

### Code Checklist

### Communication

---

---
## Checklist for QA:


octopoes/octopoes/models/ooi/dns/records.py
```diff
@@ -155,3 +156,31 @@ class DNSPTRRecord(DNSRecord):
...
     def format_reference_human_readable(cls, reference: Reference) -> str:
         return f"{reference.tokenized.address.address} -> {reference.tokenized.hostname.name}"
+
+
+class CAATAGS(Enum):
+    ISSUE = "issue"
+    ISSUEWILD = "issuewild"
+    IODEF = "iodef"
+    CONTACTEMAIL = "contactemail"
+    CONACTPHONE = "contactphone"
```
dekkers, line 166: issuevmc and issuemail are missing, see https://www.iana.org/assignments/pkix-parameters/pkix-parameters.xhtml#caa-properties

boefjes/boefjes/plugins/kat_dns/normalize.py
```diff
@@ -152,6 +155,14 @@ def register_record(record: DNSRecord) -> DNSRecord:
                         )
                     )
 
+                if isinstance(rr, CAA):
+                    recordvalue = str(rr)
+                    recordvalue = recordvalue.split(" ", 2)
+                    default_args["flags"] = min(max(0, int(recordvalue[0])), 255)
```
dekkers, line 161: If the flag has some strange value I don't think we should silently change that.

I am also not sure if DNS guarentees that we get 3 record values back or that the DNS record can also be malformed?

boefjes/boefjes/plugins/kat_dns/main.py
```diff
@@ -21,40 +21,39 @@ def run(boefje_meta: BoefjeMeta) -> List[Tuple[set, Union[bytes, str]]]:
...
+    resolver.nameservers = [str(settings.remote_ns)]
+
+    zone_soa_record = get_parent_zone_soa(resolver, requested_dns_name)
 
     answers = [
         zone_soa_record,
     ]
 
-    dns_record_types = ["A", "AAAA", "TXT", "MX", "NS", "CNAME", "DNAME"]
+    dns_record_types = ["A", "AAAA", "CAA", "CERT", "RP", "SRV", "TXT", "MX", "NS", "CNAME", "DNAME"]
```
dekkers, line 33: I don't think the CERT and RP records are much in use, do we really want to fetch those? And if we fetch them shouldn't we also normalize them? Just fetching them to include them in the raw file doesn't feel very useful.

dekkers (commented):
I think we should also have a check whether `accounturi` is used for the CAs such as letsencrypt that support it, but we can also add that later.

boefjes/boefjes/plugins/kat_dns/normalize.py
underdarknl: The raw record can be malformed indeed, however dns python DNS class already does some formatting to fit it into its DNS answer class.
Ideally we should not do this, and instead don't parse the data from the DNS-server in the boefje, and just return it verbatim into the raw file. If then a record has 'other data' that doesn't fit the expected format we should yield a Finding informing the user of that record and its original malformed raw value string.

boefjes/boefjes/plugins/kat_dns/main.py
underdarknl: We don't need to normalize everything that we have, but those seemed to provide some additional useful information which I plan on including in the model at a later date.

octopoes/octopoes/models/ooi/dns/records.py
underdarknl: I used an old RFC it seems. Added.

underdarknl:
> I think we should also have a check whether `accounturi` is used for the CAs such as letsencrypt that support it, but we can also add that later.

Yes, I was thinking of adding the related url and email objects too once we get those parsed.

boefjes/boefjes/plugins/kat_dns/main.py
```diff
@@ -17,44 +18,50 @@ class ZoneNotFoundException(Exception):
...
+    return list(set(requested_record_types).intersection(set(dns_record_types))
+
+                
 def run(boefje_meta: BoefjeMeta) -> List[Tuple[set, Union[bytes, str]]]:
     hostname = boefje_meta.arguments["input"]["name"]
 
     requested_dns_name = dns.name.from_text(hostname)
-    zone_soa_record = get_parent_zone_soa(requested_dns_name)
+    resolver = dns.resolver.Resolver()
+    nameserver = getenv("REMOTE_NS", str(settings.remote_ns))
```
ammar92, line 35: `settings` doesn't reference anything?

boefjes/boefjes/plugins/kat_dns/main.py
```diff
@@ -17,44 +18,50 @@ class ZoneNotFoundException(Exception):
     pass
 
 
+def get_record_types() -> List[str]:
+    dns_record_types = ["A", "AAAA", "CAA", "CERT", "RP", "SRV", "TXT", "MX", "NS", "CNAME", "DNAME"]
```
ammar92, line 22: Can be a constant `DNS_RECORD_TYPES` (and as a `set` type) outside of the function

boefjes/boefjes/plugins/kat_dns/main.py
```diff
@@ -17,44 +18,50 @@ class ZoneNotFoundException(Exception):
...
+    requested_record_types = getenv("RECORD_TYPES", False)
```
ammar92, line 23: It's better to return an empty string instead of a boolean, this would make it type-safe (and thus more clear IMHO)

```suggestion
    requested_record_types = getenv("RECORD_TYPES", "")
```

boefjes/boefjes/plugins/kat_dns/main.py
```diff
@@ -17,44 +18,50 @@ class ZoneNotFoundException(Exception):
...
+    if not requested_record_types:
+        return dns_record_types
+    requested_record_types = re.sub("[^A-Za-z,]", requested_record_types.upper()).split(',')
```
ammar92, line 26: `re.sub` usually has three required positional arguments, here's only two

boefjes/boefjes/plugins/kat_dns/normalize.py
```diff
@@ -152,6 +155,14 @@ def register_record(record: DNSRecord) -> DNSRecord:
                         )
                     )
 
+                if isinstance(rr, CAA):
+                    recordvalue = str(rr)
```
ammar92, line 159: `record_value`, `value` (or `record`)

octopoes/octopoes/models/ooi/dns/records.py
```diff
@@ -155,3 +156,33 @@ class DNSPTRRecord(DNSRecord):
...
+class DNSCAARecord(DNSRecord):
+    object_type: Literal["DNSCAARecord"] = "DNSCAARecord"
+    dns_record_type: Literal["CAA"] = "CAA"
+
+    # https://datatracker.ietf.org/doc/html/rfc8659#name-canonical-presentation-form
+    # An unsigned integer between 0 and 255.
+    flags: Optional[int] = None
+
+    # A non-zero-length sequence of ASCII letters and numbers in lowercase.
+    tag: CAATAGS = None
```
ammar92, line 183: This field is unclear, is it required, or is it optional with a default, or is it a nullable field?
You probably want one of the following:
- `tags: CAATAGS`; this means required
- `tags: Optional[CAATAGS]`; this means required, but can be `None`
- `tags: Optional[CAATAGS] = None`; this means optional with a default value

octopoes/octopoes/models/ooi/dns/records.py
```diff
@@ -155,3 +156,33 @@ class DNSPTRRecord(DNSRecord):
...
+
+    # The Value field, expressed as either (1) a contiguous set of characters
+    # without interior spaces or (2) a quoted string.
+    value: str = None
```
ammar92, line 187: The same goes for this one. Based on the natural key attributes I can guess that this is a required field, so therefore it should be just `value: str`

ammar92 (commented):
Nice one, looks good in general. Just a few remarks to improve it even more

boefjes/boefjes/plugins/kat_dns/main.py
underdarknl: See line 11

ammar92: Ah my bad, missed this one. I noticed that `kat_dns`, `kat_rdns`, `kat_dnssec` and `kat_manyportsopen` have a reference to the settings. However, I assumed that these references were removed after we implemented KATalogus settings. The latter would also make more sense, in the context of isolated entities/ containers that eventually get their input from the environment instead. Let's discuss this soon if you're on the same line

underdarknl: I agree, there should be a few very clear input options for each defined Boefje Being:
* Input-OOI
* Env-vars
* parameters
* Config

Those should be configurable in a mix of katalogus settings, local env-vars and whatever the scheduler sends as the input-OOI.
There might be an overlap in the parameters, config and env to accomodate for global settings etc.

boefjes/boefjes/plugins/kat_dns/normalize.py
ammar92: I'm not hinting about the content of the variable, but simply the name based on PEP naming conventions. "Record value" is a two-word phrase and should therefore be `record_value`

stephanie0x00:
## Checklist for QA:

### What works:
Onboarding works. The onboarding DNS report also includes No-CAA as  findings in the table. 
For domains without CAA a finding is created. For domains that have CAA records no findings are created. Clicking through the object pages works as expected and nothing found by checking the raw files for the scanned tasks. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

dekkers (approved):

PR #2328, opened 2024-01-13
Ansible-playbook, met wat edits... nog niet 'klaar'

sigio:
Pull based on Jeroen Baten (kwoot)'s version, but with some fixed I ran into, but didn't get it working, so scrapped and replaced with #2329 as a from-scratch retry.


sigio:
Probably no longer needed if #2329 works better

PR #2428, opened 2024-02-03
Allow creation of declared scan profiles through normalizers

### LINKED ISSUE #131: Add indemnification on objects that are imported through a Boefje from an external source (e.g. DADB)
Currently, adding objects and setting their indemnification level are two separate actions.
We only allow boefjes to set objects, and have the graph / a user set indemnifications for objects because this should be a user initiated action.
However, in the case of using a CMDB like Zcerts DADB as a source of truth, we can safely allow the objects created from that source to be given an indemnification.
I guess we need an option (false by default) which allows a user to toggle wither or not a given boefje can set indemnifications for objects that it created), and we would need a way to store those indemnifications with the inserted data and process them.
Since the given indemnifications will not be set by a user, this means we need to make sure we keep track of its origin (as we are also going to add for manual declarations).

Donnype:
### Changes
This adds logic that handles `ScanProfiles` being returned from a normalizer, and a whitelist environment variable that makes sure this is not allowed by default. There is no need (at least at this point) to make this configurable by users.

Because this is built around a first use-case, it is not clear what users will need in terms of configurability later. Do we need to parse some config for the normalizer to change the level? Then we need to update the normalizer handler to propagate environment variables. Does this also have to be configurable through the API? Then we need to update Rocky as well to allow users to manage these configs for normalizers as well, perhaps adding a separate schema.json or use the same for all plugins in the same directory.

In short, even if we need more configurability I would opt for creating a separate issue and PR, preferably when there are more feature requests from users.

### Issue link

Closes #131

### Demo
-

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


boefjes/boefjes/config.py
```diff
@@ -52,6 +52,12 @@ class Settings(BaseSettings):
         "1.1.1.1", description="Name server used for remote DNS resolution in the boefje runner"
     )
 
+    scan_profile_whitelist: Dict[str, conint(strict=True, ge=0, le=4)] = Field(
+        {},
```
ammar92, line 56: Please use a default factory for this

boefjes/boefjes/job_handler.py
```diff
@@ -172,19 +185,28 @@ def handle(self, boefje_meta: BoefjeMeta) -> None:
...
 
 class NormalizerHandler(Handler):
-    def __init__(self, job_runner):
-        self.job_runner: NormalizerJobRunner = job_runner
+    def __init__(
+        self,
+        job_runner: NormalizerJobRunner,
+        bytes_client: BytesAPIClient,
+        octopoes_factory=get_octopoes_api_connector,
+        whitelist: Dict[str, int] = None,
```
ammar92, line 193: Should contain Optional in the type hint

boefjes/boefjes/job_handler.py
```diff
@@ -207,16 +229,31 @@ def handle(self, normalizer_meta: NormalizerMeta) -> None:
                         valid_time=normalizer_meta.raw_data.boefje_meta.ended_at,
                     )
                 )
+
+            validated_scan_profiles = [
+                profile
+                for profile in results.scan_profiles
+                if self.whitelist and profile.level <= self.whitelist.get(normalizer_meta.normalizer.id, -1)
```
underdarknl, line 236: Maybe we should just clamp the requested scan-level to the max allowed scan level instead of ignoring them. 

This could be as simple as:
` profile.level  = min(self.whitelist.get(normalizer_meta.normalizer.id, -1), profile.level)`
if whitelisted, and if it returns higher than -1 we can proceed.

Donnype: Not sure what makes the most sense, but for now we could definitely just do this

ammar92 (approved):
No more remarks

underdarknl (approved):

stephanie0x00:
Shall we not try to merge stuff without a real QA? ;) 

## Checklist for QA:

### What works:
...

### What doesn't work:
- Onboarding report doesn't show the expected findings. After it completed, all findings are marked as 'Pending'. 
- Tried to scan with an additional boefje 'TCP' but this is not being scheduled. Enabling/disabling didn't help either. 

### Bug or feature?:
n/a

Issue #2336, opened 2024-01-16
Fix `make -C octopoes rtest` FAILS


originalsouth:
**Describe the bug**
`make -C octopoes rtest` FAILS

**To Reproduce**
Run `make -C octopoes rtest` in project root

**Expected behavior**
`make -C octopoes rtest` PASSES

**OpenKAT version**
main

noamblitz:
Did you install robot in your local virtualenv?

noamblitz:
Because in make rtest, robot tests are run locally i think.

originalsouth:
You would get an _error_ like this:
```
robot -d reports --variablefile tests/robot/variables.py:xtdb tests/robot
make: robot: No such file or directory
make: *** [Makefile:66: rtest] Error 127
```
rather than a _FAIL_ like this:
```
==============================================================================
Robot
==============================================================================
Robot.01 Scan Profiles
==============================================================================
Inheritance Of Two Declared Scan Profiles                             | PASS |
------------------------------------------------------------------------------
Recalculate Inheritance After Modification                            | PASS |
------------------------------------------------------------------------------
Empty Scan Profiles                                                   | PASS |
------------------------------------------------------------------------------
Robot.01 Scan Profiles                                                | PASS |
3 tests, 3 passed, 0 failed
==============================================================================
Robot.02 List Objects
==============================================================================
List Objects                                                          | PASS |
------------------------------------------------------------------------------
List Objects With Filter                                              | PASS |
------------------------------------------------------------------------------
List Random Objects With Filter                                       | PASS |
------------------------------------------------------------------------------
Load Bulk                                                             | PASS |
------------------------------------------------------------------------------
Robot.02 List Objects                                                 | PASS |
4 tests, 4 passed, 0 failed
==============================================================================
Robot.03 Deletion Propagation
==============================================================================
Propagate Deletion                                                    | PASS |
------------------------------------------------------------------------------
Robot.03 Deletion Propagation                                         | PASS |
1 test, 1 passed, 0 failed
==============================================================================
Robot.04 Save Declaration
==============================================================================
Add Several Append Origins                                            | PASS |
------------------------------------------------------------------------------
Robot.04 Save Declaration                                             | PASS |
1 test, 1 passed, 0 failed
==============================================================================
Robot.05 Bits
==============================================================================
Bit With Scan Level 1                                                 | PASS |
------------------------------------------------------------------------------
Robot.05 Bits                                                         | PASS |
[...]

Issue #2083, opened 2023-12-04
Octopoes' integration tests unstable


originalsouth:
**Describe the bug**
Octopoes' integration fail approximately 5% of the time.

**To Reproduce**
From inside a freshly installed clone -- as of this writing -- we find that running the (zsh) command:
```
F=0; for i in {1..100}; do make -C octopoes itest || F=$((F+1)); done; echo "$F/100"
```
yields `5/100` meaning approximately 5% of the time Octopoes' integration tests fail.

<details>

<summary>Typical error look like this:</summary>

```

================================================================================ test session starts =================================================================================
platform linux -- Python 3.11.6, pytest-7.4.0, pluggy-1.2.0
rootdir: /app/octopoes
plugins: env-0.8.2, anyio-3.7.1, timeout-2.1.0, cov-4.1.0, mock-3.11.1, requests-mock-1.11.0
timeout: 300.0s
timeout method: signal
timeout func_only: False
collected 10 items

tests/integration/test_api_connector.py ..                                                                                                                                     [ 20%]
tests/integration/test_ooi_repository.py .                                                                                                                                     [ 30%]
tests/integration/test_xtdb_client.py ......F                                                                                                                                  [100%]

====================================================================================== FAILURES ======================================================================================
____________________________________________________________________________ test_query_for_system_report ____________________________________________________________________________

self = <urllib3.connectionpool.HTTPConnectionPool object at 0x7fd340c9d210>, method = 'POST', url = '/test-69bca82f-5ec9-45b1-9ae0-c5b0a9a2f097/bits/recalculate', body = None
headers = {'User-Agent': 'python-requests/2.31.0', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive', 'Content-Length': '0'}
retries = Retry(total=3, connect=None, read=None, redirect=None, status=None), redirect = False, assert_same_host = False, timeout = Timeout(connect=None, read=None, total=None)
pool_timeout = None, release_conn = False, chunked = False, body_pos = None, preload_content = False, decode_content = False, response_kw = {}
parsed_url = Url(scheme=None, auth=None, host=None, port=None, path='/test-69bca82f-5ec9-45b1-9ae0-c5b0a9a2f097/bits/recalculate', query=None, fragment=None)
destination_scheme = None, conn = None, release_this_conn = True, http_tunnel_required = False, err = None, clean_exit = False

    def urlopen(  # type: ignore[override]
        self,
[...]

PR #2441, opened 2024-02-05
Make rtest fail less than 1 in 20

originalsouth:
### Changes
Patches for Celery and Robot files.

### Issue link
Address issue #2336 

Closes ...

### Demo
N/A (see pipeline, hopefully)

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


originalsouth:
Thanks for the feedback. I agree. `Wait Until Keyword Succeed` would be a better - indeed. I will experiment with it later  today/this week. 

The underlying problem here, also effecting #2083, is however that (re)calculating the bits takes a relatively rather large and inconsistent number of seconds while also being asynchronous. Incidentally, this is any issue which I am currently investigating.

dekkers:
I first thought there was a race condition in the tests, but when working to change the tests I saw that it actually does wait for all celery tasks to execute. How did you conclude that the underlying problem is the number of seconds recalculation takes?

Did you see that there were still tasks being executed or waiting to be executed after "Await Sync" (which calls "Await Tasks Done")?  If so we should fix "Await Tasks Done" instead of adding sleep statements.

The problem might also not be in the tests.. One strange thing I saw is that even after I added sleep statements of 10 seconds the number of tasks that were executed as reported by robotframework/celerymonitor logging wasn't stable in the test I was looking at even after I disabled celery beat. For example one time it was 92, then 95, then 93, etc. There might be race conditions in how the bits and scan profile propagation are executed in Octopoes itself instead of in the tests.

originalsouth:
> How did you conclude that the underlying problem is the number of seconds recalculation takes?

Similarly to #2083 the error occurs after a recalculation, adding a sleep "resolves" the problem indicating a timing issue in recalculation. I think on of the problems is that bits can trigger new bits recursively and we do not block till this process is finished. With the event manager mock in #2310 we dig deeper into the this issue (although high resolution time stamps might still be required).

octopoes/tests/robot/05_bits.robot
```diff
@@ -12,8 +14,8 @@ Bit With Scan Level 1
     Object List Should Not Contain    Finding|Hostname|internet|example.com|KAT-NXDOMAIN
     Declare Scan Profile    Hostname|internet|example.com    1
     Await Sync
-    Object List Should Contain    KATFindingType|KAT-NXDOMAIN
-    Object List Should Contain    Finding|Hostname|internet|example.com|KAT-NXDOMAIN
+    Wait Until Keyword Succeeds    60s    1s    Object List Should Contain    KATFindingType|KAT-NXDOMAIN
+    Wait Until Keyword Succeeds    60s    1s    Object List Should Contain    Finding|Hostname|internet|example.com|KAT-NXDOMAIN
```
dekkers, line 18: Does this still fail the test if it isn't there after 60 seconds? The robot framework documentation isn't very clear but I think they mean to say that it doesn't.

octopoes/tests/robot/CeleryMonitor.py
```diff
@@ -11,13 +11,17 @@
...
+        "worker_concurrency": 1,
+        "worker_prefetch_multiplier": 0,
         "result_backend": f"rpc://{queue_uri}",
         "task_serializer": "json",
         "result_serializer": "json",
         "event_serializer": "json",
         "accept_content": ["application/json", "application/x-python-serialize"],
         "result_accept_content": ["application/json", "application/x-python-serialize"],
         "task_queues": ("octopoes",),
+        "task_alway_eager": True,
```
dekkers, line 24: This is the config for the celerymonitor / the part running inside the robot tests, so won't do anything as far as I understand it. The tasks are created and executed inside octopoes api and api worker and not the robot tests.

I also think it is better if we don't configure celery in the integration tests differently from how it is normally run unless it is really necessary.

octopoes/tests/robot/05_bits.robot
originalsouth: Yes, I believe that is the gist of it; see [here](http://robotframework.org/robotframework/2.6.1/libraries/BuiltIn.html).

octopoes/tests/robot/CeleryMonitor.py
originalsouth: Addressed in 5dfb5afad4312791bc4b3dc6c6409884247f1425.

dekkers (approved):

PR #2451, opened 2024-02-07
Add task duration intervals to the scheduler

### LINKED ISSUE #1956: Task events for scheduler

jpbruinsslot:
### Changes

* Add functionality to track duration intervals for tasks.
* Columns `pending_at`, `queued_at`, `dispatched_at`, `running_at` are added, these signify the timestamp that a task was in that particular state.
* Method on `Task` class added `update_status` to calculate the duration of a task transition when the status updates.
* Added `meta` jsonb column to task. Added meta column for runners to relay and persist information about a task, e.g. cpu, mem load, run times etc.
* Added status update patch request to scheduler api from runners to update task status to `running`

Closes #1956


boefjes/boefjes/local.py
```diff
@@ -40,10 +42,14 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
     def __init__(self, local_repository: LocalPluginRepository):
         self.local_repository = local_repository
+        self.scheduler_client = SchedulerAPIClient(str(settings.scheduler_api))
 
     def run(self, boefje_meta: BoefjeMeta, environment: Dict[str, str]) -> List[Tuple[set, Union[bytes, str]]]:
         logger.info("Running local boefje plugin")
 
+        task_id = str(boefje_meta.id)
+        self.scheduler_client.patch_task(task_id, TaskStatus.RUNNING)
+
```
jpbruinsslot, line 52: I don't know if adding this is advisable or desirable. I took note from the docker boefje runner. @Donnype

boefjes/boefjes/local.py
```diff
@@ -58,10 +64,14 @@ def run(self, boefje_meta: BoefjeMeta, environment: Dict[str, str]) -> List[Tupl
 class LocalNormalizerJobRunner(NormalizerJobRunner):
     def __init__(self, local_repository: LocalPluginRepository):
         self.local_repository = local_repository
+        self.scheduler_client = SchedulerAPIClient(str(settings.scheduler_api))
 
     def run(self, normalizer_meta, raw) -> NormalizerOutput:
         logger.info("Running local normalizer plugin")
 
+        task_id = str(normalizer_meta.id)
```
jpbruinsslot, line 72: I don't know if adding this is advisable or desirable. I took note from the docker boefje runner. @Donnype

boefjes/boefjes/local.py
```diff
@@ -58,10 +64,14 @@ def run(self, boefje_meta: BoefjeMeta, environment: Dict[str, str]) -> List[Tupl
...
+        self.scheduler_client.patch_task(task_id, TaskStatus.RUNNING)
+
```
jpbruinsslot, line 74: I don't know if adding this is advisable or desirable. I took note from the docker boefje runner. @Donnype

boefjes/boefjes/local.py
jpbruinsslot, line 52: I don't know if adding this is advisable or desirable. I took note from the docker boefje runner. @Donnype

mula/scheduler/models/tasks.py
```diff
@@ -82,6 +132,13 @@ class TaskDB(Base):
         default=TaskStatus.PENDING,
     )
 
+    pending = Column(Interval)
+    queued = Column(Interval)
+    dispatched = Column(Interval)
+    running = Column(Interval)
+
+    meta = Column(JSONB)
```
jpbruinsslot, line 140: Added for runners to relay and persist information about a task, e.g. cpu, mem load, run times etc. If this info is already available through bytes we can opt to reference that. However, it might be more difficult/time-consuming to retrieve that information when we want to make rankings based on that information since we don't have it readily available in the scheduler.

mula/scheduler/schedulers/scheduler.py
```diff
@@ -195,6 +183,21 @@ def push_item_to_queue(self, p_item: models.PrioritizedItem) -> None:
...
+            id=p_item.id,
+            scheduler_id=self.scheduler_id,
+            type=self.queue.item_type.type,
+            p_item=p_item,
+            status=models.TaskStatus.PENDING,
+            created_at=datetime.now(timezone.utc),
+            modified_at=datetime.now(timezone.utc),
+        )
+        self.ctx.datastores.task_store.create_task(task)
+
```
jpbruinsslot, line 200: Refactored to make the sequence pending > queued > dispatched work correctly within the scheduler

mula/scheduler/server/server.py
```diff
@@ -470,6 +474,16 @@ def patch_task(self, task_id: str, item: Dict) -> Any:
...
+        # Retrieve updated task
+        try:
+            updated_task = self.ctx.datastores.task_store.get_task_by_id(task_id)
+        except Exception as exc:
+            self.logger.error(exc)
+            raise fastapi.HTTPException(
+                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                detail="failed to get updated task",
+            ) from exc
+
```
jpbruinsslot, line 486: We want to make sure that when the runner updates the status of the task. That the duration in a `Task` is calculated and updated correctly

mula/scheduler/models/tasks.py
```diff
@@ -57,10 +59,58 @@ class Task(BaseModel):
 
     status: TaskStatus
 
+    # Durations
+    pending: Optional[timedelta] = None
+    queued: Optional[timedelta] = None
+    dispatched: Optional[timedelta] = None
+    running: Optional[timedelta] = None
```
jpbruinsslot, line 66: Added new duration columns for setting specifically or by calculation

boefjes/boefjes/local.py
jpbruinsslot: The main issue was that the boefje and normalizer runner are not updating the task when they are processing the job. This change will make sure that it happens and that the duration calculation is done for a task

boefjes/boefjes/local.py
underdarknl: I think this 'self calculation' based on events is less precise than having the boefjesrunner send in its own metrics, but since we might not have those metrics for all runners we need to allow for both paths.

jpbruinsslot: Yes, and from the runner perspective, they are still able to update the value, or we can parse it from the received meta data if so desired.

mula/scheduler/models/tasks.py
Donnype: Perhaps it makes more sense to save datetimes of when the state was (last) changed? Then these fields could be called `[...]_at` so the naming would be clearer.

Also, when a task gets dispatched but re-queued because of an error or shutdown, this route would potentially bring a lot of complex calculations with it to get the right numbers. I am btw in general always a bit hesitant about working with interval-like data structures vs. static data points.

boefjes/boefjes/local.py
Donnype: @jpbruinsslot For the Docker setup there were indeed some limitation that meant that status handling had to happen in the runner. The initial setup that uses the local runner handles the status in the `app.py`, which I'd say for now is a better place to add this I think for the local setup.

jpbruinsslot: @Donnype updated, and added it to `app.py`. Would you please confirm that it is in the correct place?

mula/scheduler/models/tasks.py
jpbruinsslot: Updated, made it timestamps. Now if we want to make queries: "What tasks were queued longer than 5 minutes in the past 6 days" we need to write specific queries for it that calculate the difference between those timestamps

jpbruinsslot:
A rebase borked the PR a bit, probably going to put the changes into a separate PR. Unless there is a way to fix the history

mula/scheduler/models/tasks.py
jpbruinsslot: We can opt for removing this field for now since it depends on changes or implementation in runners

boefjes/tests/test_app.py
```diff
@@ -22,10 +22,13 @@ def test_one_process(manager: SchedulerWorkerManager, item_handler: MockHandler)
 
     patched_tasks = manager.scheduler_client.get_all_patched_tasks()
 
-    assert len(patched_tasks) == 3
-    assert patched_tasks[0] == ("70da7d4f-f41f-4940-901b-d98a92e9014b", "completed")
+    assert len(patched_tasks) == 6
```
jpbruinsslot, line 25: @Donnype would you mind taking a look if I'm making the correct assumptions within these tests.

boefjes/Makefile
```diff
@@ -62,6 +62,13 @@ ci-docker-compose := docker compose -f base.yml  -f .ci/docker-compose.yml
 
 test: itest ## Run all tests.
 
+utest: ## Run the unit tests.
```
Donnype, line 65: cool :+1:

boefjes/boefjes/app.py
```diff
@@ -240,6 +240,8 @@ def _start_working(
 
     while True:
         p_item = task_queue.get()
+        scheduler_client.patch_task(p_item.id, TaskStatus.RUNNING)  # Note: implicitly, we have p_item.id == task_id
```
Donnype, line 243: @jpbruinsslot :ok_hand:

boefjes/tests/test_app.py
Donnype: @jpbruinsslot These patches seem logical right? All of them go from running to either completed or failed :+1:

boefjes/boefjes/local.py
Donnype: Looks good @jpbruinsslot !

dekkers (commented):
I am missing a functional specification about what we are trying to achieve. If we want to store an accurate execution time of boefjes and normalizers (which is the primary reason of this change as far as I understood it), we would first need to define what we mean with with that/what we want to store. We will also be using new docker / kubernetes runners soon and I think it would also be good to take those in account.

These are just some ideas, I think we need to brainstorm/discuss this more:

- With the docker / kubernetes runner, if the container executes one task the execution time might be the time the container has run.
- If it would be a long-running container that will executes multiple tasks, we can't use the container execution time. One way would be to have the container report the execution time back when submitting the boefje results. We would have to trust the container to do this correctly, but I don't think this would be a big problem in practice.
- Another option would be to calculate the time between the input being fetched in the boefjes api and the output being submitted. This might be less accurate if there is latency between boefjes api and the container or the boefjes api is overloaded. If we want to add batching in the boefjes API this would also not work.
- For the current boefje/normalizer runner, we might take the execution time for the `run` function.

I think doing status updates queries from the runner to the schedule and then storing the timestamps won't give the most accurate execution time because it will include the latency of those requests. For very simple boefjes and most normalizer the overhead might even be more than the execution time. The latency might also be higher when if we are going to have runners that will be inside different internal networks in the future. If the scheduler and/or database is overloaded it would also give more inaccurate information.

The way it is currently implemented we also include the time it takes to fetch OOIs or raw files and save raw files or OOIs which I'm not sure we should include in the task duration.

I think that the best solution would be for the runner to be responsible for calculating the execution time and providing that to the scheduler to store it.

Database wise I think this can just be 32-bit integer that will store the execution time in miliseconds. That will provide a resolution of 1 ms and maximum of 24 days which should be enough. That way we would only need 4 bytes instead of 16 bytes for the PostgreSQL interval type that has very small resolution and high range.

With regards to the other timestamps being added, do we really need them all? It's not going to be the biggest bottleneck, but if we want to scale to millions of tasks, I think it is important to make the rows larger than nessary. The existing table design would also need improvements to achieve this, but that shouldn't stop us from doing new things in the most efficient way.

Pending_at and created_at will always be the same, so I don't think it is useful to have a pending_at? I'm also not sure how useful it is store queued_at and dispatched_at for every task. If I think from a user perspective, the interesting data are:
- When was the task created (created_at timestamp)
- When was the task executed (running_at timestamp, maybe executed_at would be a better name?)
- How long did task execution take (duration in ms).

While the other timestamp might be useful for debugging what is going on inside the scheduler, I'm not sure that is reason enough to store them for every task in production systems. For debugging we also already have debug logging with which we can see when a task is queued, fetched by the runner, etc.

dekkers:
As discussed (sorry for writing this up so late):

The duration that the user wants to see is the execution runtime of the boefje/normalizer job. The only way to do this is to have the runner provide the information.

The duration needs to be sortable with large number of tasks, because a user might want to sort the tasks by duration to see which tasks took a long time to run. So we should store the data in such a way that is easy for PostgreSQL to sort. The suration can be stored most effeciently as miliseconds using a 32-bit integer. This would be have a maximum of 2147483648 miliseconds or 24.8 days which should be more than enough.

jpbruinsslot:
Considerations:

* Do we want to and/or expect that we also want to track and query the durations of other statuses than `RUNNING`. If not, then we can add a single column to the task column with this duration. If so, then we need to consider if we want to have those status durations in the task column or to normalize in order to not clutter the task table.
* CPU/mem usage metrics seem specific towards a boefje runner and might not necessarily relate to other task runners. Additionally, we're not considering other metrics like disk usage and network usage which might not be relevant to other task runners. Considering the agnostic nature of the scheduler we then we need to evaluate whether to normalize those metrics fields as well. I suggest we need to discuss this in another ticket since it also begs questions about ux: how is this going to be used, queried and visualized, and if other tools that track these metrics are better suited than to integrate this into the scheduler.

mula/scheduler/models/tasks.py
```diff
@@ -57,10 +57,39 @@ class Task(BaseModel):
...
     status: TaskStatus
 
-    created_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
+    meta: dict | None = None
+
+    # Status transition timestamps
+    pending_at: datetime | None = Field(default_factory=lambda: datetime.now(timezone.utc))
+    queued_at: datetime | None = None
+    dispatched_at: datetime | None = None
+    running_at: datetime | None = None
```
Donnype, line 66: `started_at`?

mula/scheduler/models/tasks.py
```diff
@@ -57,10 +57,39 @@ class Task(BaseModel):
...
+    completed_at: datetime | None = None
 
+    created_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
     modified_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
```
Donnype, line 70: Is this needed if the only state changes are for the status that you use the ..._at fields for?

mula/scheduler/models/tasks.py
```diff
@@ -57,10 +57,39 @@ class Task(BaseModel):
 
     status: TaskStatus
 
-    created_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
+    meta: dict | None = None
+
+    # Status transition timestamps
+    pending_at: datetime | None = Field(default_factory=lambda: datetime.now(timezone.utc))
```
Donnype, line 63: This is just the same as created_at? Could we reduce this to:

`created_at`
`queued_at`
`started_at`
`ended_at`

The latter two being set by the runner?

Donnype: This way you know how long:
- it takes for tasks to complete from queueing them (ended_at - queued_at)
- When the task actually started (started_at)
- The runtime (ended_at - started_at)
- The complete lifecycle (ended_at - created_at) for throughput analysis.
- Queueing time (started_at - queued_at (?)) 

Also, I'd imagine for longrunning tasks that you'd be interested in when they started, not when they were queued. That's I'd say a valid use-case for the started_at/ended_at approach vs. the runtimes? Unless you want to infer that through modified_at. Or the end times in general would be a relevant column to analyse

Donnype: What I just thought of: I guess the default sorting is going to be the `modified_at` in the setup where we only save runtimes?

mula/scheduler/models/tasks.py
underdarknl: I'd rather collect this info in the scheduler for fast-access, as its something the user wants to filter on, see on the tasklist. Having those details only in Bytes would mean we end up reading those jobmeta's for the whole list / or worse for all jobs when rendering a filtered task list.

underdarknl: The question where and how we want to to store these meta values in the schedulers database is still very valid though.

PR #2469, opened 2024-02-08
Do not log an error on token refresh in bytes client

dekkers:
### Changes
In the bytes client in boefjes we always logged when the request returned an error. This was also done when a 401 was returned because the token needs to be refreshed. This resulted in a confusing error message about the 401, because the code in `retry_with_login` would just refresh the token and retry the request and nothing went wrong.

It also fixes the wrong re-raising of the exception when the status isn't 401.

---

---
## Checklist for QA:


boefjes/boefjes/clients/bytes_client.py
```diff
@@ -67,8 +67,9 @@ def login(self) -> None:
     def _verify_response(response: requests.Response) -> None:
         try:
             response.raise_for_status()
-        except HTTPError:
-            logger.error(response.text)
+        except HTTPError as error:
+            if error.response.status_code != 401:
+                logger.error(response.text)
```
underdarknl, line 72: Could we add a logger.debug() for 401 status_codes?

underdarknl, line 72:
```suggestion
                logger.error(response.text)
            else:
                logger.debug(response.text)
```

underdarknl (approved):

PR #2513, opened 2024-02-15
Mail Report unit tests

madelondohmen:
### Changes
This PR adds unit tests for Mail Report.

### Issue link
Closes #2472 

---

### Code Checklist
- [x] All the commits in this PR are properly PGP-signed and verified.
- [x] This PR only contains functionality relevant to the issue; tickets have been created for newly discovered issues.
- [x] I have written unit tests for the changes or fixes I made.
- [ ] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [x] I have performed a self-review of my code and refactored it to the best of my abilities.

### Communication
- [x] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [x] I have made corresponding changes to the documentation, if necessary.
- [x] I have included comments in the code to elaborate on what is not self-evident from the code itself, including references to issues and discussions online, or implicit behavior of an interface.

---
## Checklist for code reviewers:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_code.md) into your comment.

---
## Checklist for QA:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_qa.md) into your comment.

PR #2516, opened 2024-02-15
Feature/efficient reporting

### LINKED ISSUE #2446: Reduce queries in aggregate reports
We can reduce queries done in a report by allowing lists of OOIs as parameters in the gerenate_report methods and gathering all queries before running them. 

Also, we could add lru's to octopoes api for faster querying.

Donnype:
### Changes
This PR adds several changes that should be backward compatible with current reporting, but opens up the possibility to make significant performance improvements:
- The parent BaseReport class now has a `collect_data` method that takes in multiple oois. By default this generates the data through the already existing `generate_data` method. It returns the data grouped by the input_ooi for easy compatibility.
- The views have been rewritten to use the `collect_data` method, moving iteration to the report classes in a backward compatible way.
- Octopoes now has a `/query-many` endpoint that you also give a path query, but takes in several `sources` (references), and returns the same result as running that path on all sources separately.
- Octopoes and the query module have been updated to be able to handle this kind of query to make sure XTDB does the heavy lifting by using `or`-clauses. Still for compatibility we had to connect this back to the source through some datalog magic.
- The Mail report has been rewritten to override the `collect_data` method to make use of this feature. When the other reports do this as well, the amount of queries to octopoes will remain constant in stead of scale with the number of input objects. This is a TODO for a follow up PR that I will create, including benchmark results.

Note: the logic around the Query object in this endpoint gets a bit complicated at this point. We should reserve some time to refactor this at some point, but this would give quite some changes perhaps.

### Issue link

Closes #2446 (partly, follow up PR incoming)

### Demo
See tests.

### Notes for QA
Follow the regular report generation flow, but now expect a performance boost generating (aggregate) reports. 

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


dekkers (commented):
In general I am very happy to see this and it looks good, I have just a few small remarks. I think it is a nice big step towards more efficient report generation.

octopoes/octopoes/connector/octopoes.py
```diff
@@ -283,3 +283,23 @@ def query(
             TypeAdapter(OOIType).validate_python(ooi)
             for ooi in self.session.get(f"/{self.client}/query", params=params).json()
         ]
+
+    def query_many(
+        self,
+        path: str,
+        valid_time: datetime,
+        sources: list[OOI | Reference | str] = None,
```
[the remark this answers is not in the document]
Donnype: Indeed good spot!

madelondohmen:
In the meantime, some naming changes have been made to the Mail Report (#2513). I think your code will overwrite this, so it would be good to compare the two and make sure these changes won't be overwritten.

Donnype:
> In the meantime, some naming changes have been made to the Mail Report (#2513). I think your code will overwrite this, so it would be good to compare the two and make sure these changes won't be overwritten.

Yes I noticed those, should be fixed I think :+1:

octopoes/octopoes/xtdb/query.py
```diff
@@ -146,17 +167,19 @@ def _where_field_is(self, ref: Ref, field_name: str, value: Ref | str | set[str]
...
+            value = value.replace('"', r"\"")
+
         if ooi_type in abstract_types:
             if isinstance(value, str):
-                value = value.replace('"', r"\"")
-                self._add_or_statement(ref, field_name, f'"{value}"')
+                self._add_or_statement_for_abstract_types(ref, field_name, f'"{value}"')
                 return
 
             if not isinstance(value, type):
```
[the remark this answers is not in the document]
Donnype: I added a test that does some checking, but it took a while to realize that this check is needed because `issubclass` fails if the first argument is not a class:

We need the subclass check to verify if the type is an OOI Type:
```python
>>> value = Network
>>> isinstance(value, OOIType)
False
>>> isinstance(value, OOI)
False
>>> isinstance(ooi_type, type)
True
>>> issubclass(ooi_type, OOI)
True
>>> issubclass(3, OOI)
  [...]
TypeError: issubclass() arg 1 must be a class
```

The test also spotted a small bug :relieved:

rocky/reports/report_types/ipv6_report/report.py
```diff
@@ -50,12 +50,31 @@ def generate_data(self, input_ooi: str, valid_time: datetime) -> dict[str, Any]:
...
+                "Hostname.<hostname [is ResolvedHostname].address", valid_time, hostname.reference
+            )
 
             results = {
                 hostname.name: {"enabled": any(ip.reference.class_type == IPAddressV6 for ip in ips)}
                 for hostname in hostnames
             }
 
         return results
+
```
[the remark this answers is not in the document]
Donnype: Yeah I wasn't sure partly because I knew others were also working on these methods. I think this avoids some merge conflicts, but not sure if I there are any other usages.

rocky/reports/report_types/name_server_report/report.py
```diff
@@ -125,3 +126,46 @@ def generate_data(self, input_ooi: str, valid_time: datetime) -> dict[str, Any]:
...
+        result = {}
+
+        for input_ooi, hostname_references in hostnames_by_input_ooi.items():
+            name_server_checks = NameServerChecks()
+            finding_types = {}
+
+            for hostname in hostname_references:
+                check = NameServerCheck()
+                check.no_uncommon_ports = not any(port_finding_types.get(hostname, []))
+                check.has_dnssec = "KAT-NO-DNSSEC" not in [x.id for x in hostname_finding_types.get(hostname, [])]
```
ammar92, line 154: Use a generator instead, this will improve performance in this case
```suggestion
                check.has_dnssec = "KAT-NO-DNSSEC" not in (x.id for x in hostname_finding_types.get(hostname, []))
```

rocky/reports/report_types/name_server_report/report.py
```diff
@@ -125,3 +126,46 @@ def generate_data(self, input_ooi: str, valid_time: datetime) -> dict[str, Any]:
...
+                check.has_valid_dnssec = check.has_dnssec and "KAT-INVALID-DNSSEC" not in [
```
ammar92, line 155: Same here

rocky/reports/report_types/name_server_report/report.py
```diff
@@ -125,3 +126,46 @@ def generate_data(self, input_ooi: str, valid_time: datetime) -> dict[str, Any]:
...
+
+        query = "Hostname.<ooi[is Finding].finding_type"
+        hostname_finding_types = self.group_by_source(
+            self.octopoes_api_connector.query_many(query, valid_time, all_hostnames),
+            lambda ooi: ooi.id in ["KAT-NO-DNSSEC", "KAT-INVALID-DNSSEC"],
+        )
+        query = "Hostname.<hostname[is ResolvedHostname].address.<address[is IPPort].<ooi[is Finding].finding_type"
+        port_finding_types = self.group_by_source(
+            self.octopoes_api_connector.query_many(query, valid_time, all_hostnames),
+            lambda ooi: ooi.id in ["KAT-UNCOMMON-OPEN-PORT", "KAT-OPEN-SYSADMIN-PORT", "KAT-OPEN-DATABASE-PORT"],
```
ammar92, line 142: Since you're using the `check` argument quite often, this leads to a lot of repetitive code and lambda noise. I want to suggest to introduce a (partial) function that accepts a list of finding type names, something like:
```python3
def finding_type_checker(finding_types: list[str]):
    def _check(ooi, types):
        return ooi.id in types
     return functors.partial(_check, types=finding_types)

...
self.group_by_source(..., finding_types_checker(["KAT-UNCOMMON-OPEN-PORT", "...", ...]))
...
```

ammar92 (commented):
It looks good overal. I just have a few suggestions that might improve it even more

rocky/reports/report_types/aggregate_organisation_report/report.py
```diff
@@ -444,27 +444,39 @@ def aggregate_reports(
...
-                    if (
-                        Reference.from_str(ooi).class_type in report_type.input_ooi_types
-                        and report_type.id in selected_report_types
-                    ):
-                        report = report_type(connector)
-                        data = report.generate_data(ooi.primary_key, valid_time=valid_time)
-                        report_data[ooi.primary_key][report_type.id] = data
+            results = report_type(connector).collect_data(oois, valid_time)
         except ObjectNotFoundException:
-            logger.error("Object not found: %s", ooi.primary_key)
```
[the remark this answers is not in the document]
Donnype: Not sure here, I'll add it

octopoes/octopoes/xtdb/query.py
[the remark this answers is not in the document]
ammar92: `isinstance` only works for instances, so you'd have to give it `Network()` rather than type `Network`. `isinstance(value_type, type)` will always return `True` since everything derives from class `type`. And `issubclass` works for class types and not instances

rocky/reports/report_types/aggregate_organisation_report/report.py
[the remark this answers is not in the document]
Donnype: Ah now I see: we don't loop over OOIs so the ooi variable does not exist actually.

rocky/reports/report_types/name_server_report/report.py
Donnype: Is this really worth the effort? If I only change it here this probably will only introduce inconsistencies as most of our comprehensions are just list comprehensions and the underlying data structure is also just a plain list (so we do not save memory)

rocky/reports/report_types/name_server_report/report.py
Donnype: See comment above

rocky/reports/report_types/name_server_report/report.py
ammar92: Yep it's absolutely worth the effort. Because this optimization not only saves memory, it also enhances performance by minimizing the number of iterations. There's no need to create a new list based on another list and iterate over each list element if there's a greater chance that the condition (not in) is met earlier and just to throw the list away after it. But even if it's at the end of the list (because it's still an O(n) search), it will still be beneficial over creating a new list and disposing it.

Of course this should be applied on more places for consistency and overall readability and performance optimization, but this was just an idea and obviously I can't pin point every other example where this is applicable.

ammar92 (dismissed):

rocky/reports/report_types/mail_report/report.py
```diff
@@ -51,21 +51,21 @@ def generate_data(self, input_ooi: str, valid_time: datetime) -> dict[str, Any]:
 
             number_of_spf -= (
                 1
-                if any(
+                if list(
```
ammar92, line 54: What have you done or searched to fix this Mypy error? A quick lookup reveals that in this context it indeed requires a type guarded/ type narrowing function. An even simpler approach would be to rewrite the filter with lambda to a generator expression (`any(( expr for item in finding_types[...]))`)

Donnype: I don't think these kinds of comments really put a good tone on PR reviews. With these words you are:
1. Suggesting I didn't try anything to resolve it (by asking what I tried)
2. Saying that it was actually "quick" and easy to look the issue up and resolve it.

But if you must know I looked at:
1. Narrowing the return type of the `group_finding_types_by_source()` method, of which I expected the error to move towards perhaps a mismatch with the return type of the `group_by_source()` method, which it did not
2. Type hinting the variables, without any change
3. How it could be possible in general that the type error doesn't occur when we call `list()` but does occur when we call `any()`

At which point I figured this is perhaps not worth putting too much time into since I think this logic could be refactored in other ways and this PR focuses around other major API changes. 

P.S.
Isn't it strange that mypy doesn't complain about the id field now that it's wrapped in the plain `any()`?

rocky/reports/report_types/ipv6_report/report.py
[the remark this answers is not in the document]
dekkers: It avoids conflicts in git, but we still have a conflict when the generate_data is changed but the collect_data is not updated to reflet to changes, so I'm not really sure it is better. We can also remove the old code and check that the new code does the same as the old code in a PR that immediately follows this one, but I don't think it is a good idea to keep maintaining two implementations of the reports.

rocky/reports/report_types/mail_report/report.py
dekkers: Yes, I also don't really understand it from a quick look, maybe it is because mypy is still lacking in some areas. We also don't have all strict options turned on and the models are still excluded from mypy here (my octopoes typing PR will fix that) which might also cause this, not sure about that.

While in general we should use `any` and generators because they are better performance wise, I think that

1. The current code already does it in the suboptimal way, so I don't see this as a blocker for merging this PR.
2. I think the way we use filters, generators, count findings, etc. in the reports are symptoms that we aren't using the right datastructures and we should first focus on those things. For example use a single query to fetch all relevant findings of hostnames for multiple reports in one go and have it return a set of findingtype ids we can use `not in` on without needing a generator or filter/lambda at all. Or if we only need to know the total number of each findingtype we might be able to just let the database count them.

ammar92: > 3. How it could be possible in general that the type error doesn't occur when we call `list()` but does occur when we call `any()`

It's a bit strange indeed, it seems as if Mypy is picking the wrong overload of filter's initializer. It has been reported a while ago but hasn't been fixed. So far I've only seen people bypassing this by using a generator expression in the `any` instead.

> At which point I figured this is perhaps not worth putting too much time into since I think this logic could be refactored in other ways and this PR focuses around other major API changes.

Agree

> 2\. I think the way we use filters, generators, count findings, etc. in the reports are symptoms that we aren't using the right datastructures and we should first focus on those things. For example use a single query to fetch all relevant findings of hostnames for multiple reports in one go and have it return a set of findingtype ids we can use `not in` on without needing a generator or filter/lambda at all. Or if we only need to know the total number of each findingtype we might be able to just let the database count them.

Yep, that's some food for thoughts

rocky/reports/report_types/ipv6_report/report.py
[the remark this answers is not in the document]
Donnype: Done

stephanie0x00:
## Checklist for QA:

### What works:
Generated an aggregate report successfully for multiple domains. Tasks are created and seem to complete as expected. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

dekkers (approved):

PR #2536, opened 2024-02-20
Select all oois triggers toggle all

Rieven:
### Changes
Select all oois button triggered toggle all, which is not necessary.

- When pressing the button "Continue with all OOIs", it will not continue, saying that there are no oois selected from the list of oois.
- This is because the button is in the same form as the oois selection  (checkboxes)
- I move the button outside the form to seperate 2 submits.
- Test by clicking the button again and check whether it will continue to the next page (with no errors, see screenshot).

# The problem
Pressing the button "Continue with all OOIs" cannot go to the next page, because it triggers the toggle all functionality in JS which requires at least 1 checkbox to be selected from the list of objects(OOIs). No selection is needed with "Continue with all OOIs".

"Continue with all OOIs" should go to the next page with query parameter : `ooi="all"` and should not be interrupted by the toggle all functionality.

Under what circumstances is this problem occurring? 
screenshot:
image

## Steps to reproduce
1. Go to Reports in main menu
2. Go to "Aggregate reports"
3. Press "Continue with all OOIs" and see error at toggle all checkbox (see screenshot)

# Possible solutions
- Describe possible ways to solve the problem. 
Remove the button outside the ooi selection form and having its own form  

What other options have you considered? 
Put a `href` attribute, but then it also needs to continue with all the selection from query parameters, so need to refactor that as well. 

# Solution 
Remove the button outside the ooi selection form and having its own form

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


underdarknl (dismissed):

rocky/tests/conftest.py
```diff
@@ -820,3 +820,21 @@ def list_origins(
...
+        Hostname(network=network.reference, name="b.example.com"),
+        Hostname(network=network.reference, name="c.example.com"),
+        Hostname(network=network.reference, name="d.example.com"),
+        Hostname(network=network.reference, name="e.example.com"),
+        Hostname(network=network.reference, name="f.example.com"),
+    ]
+
+
+@pytest.fixture
+def now_formatted(valid_time):
```
dekkers, line 839: This isn't really a good name. I also don't think it very useful use of a fixture, I think it is better to just put `valid_time.strftime("%Y-%m-%d")` in the test.

PR #2543, opened 2024-02-21
Make valid time required parameter in the octopoes API

### LINKED ISSUE #1961: Make valid time required parameter in octopoes API
Currently the valid time in the octopoes API isn't a required parameter in most of its API endpoints, but is required for a few. If the valid time isn't given we fall back to the current time. I think we should always require the valid time to be given. 

As far as I know rocky will already provide the valid time, because a call to `get_observed_at` will return the current time if it isn't specified. Other services that call octopoes might not do this, I haven't checked it yet, but I think it would be better if they would also be forced to explicitly specify the valid time, because then you have to consider what the valid time should be. Forgetting it won't result in silently doing the wrong thing. If something would do multiple calls to octopoes, being forced to specify the valid time will also prevent race conditions if the same valid time is specified.

dekkers:
### Changes
This PR makes valid_time a required parameter. It also adds mypy checking to octopoes/connector.

I also refactored the rocky code a bit, some code that was only used in `SingleOOITreeMixin` but defined in one of the superclasses have been moved to that class.

### Issue link
Closes #1961 

---

---
## Checklist for QA:


ammar92 (approved):
No remarks

underdarknl (approved):

PR #2544, opened 2024-02-21
Disable ruff split-on-trailing-comma and update ruff

dekkers:
### Changes
See https://docs.astral.sh/ruff/settings/#lint_isort_split-on-trailing-comma

I see this cause unnecessary noise in PRs. Probably it is because if you add an import while coding, the line can be split because it gets too long, and if later the import is removed again it will not be folded back.

It also updates ruff and move the linting config to ruff.lint (this has been changed in ruff 0.2 because they also added a formatter)

---

---
## Checklist for QA:


ammar92 (approved):
That's unfortunate, because I did actually like the newlines in imports as they're easier to read and more tidy. However, I like consistency more and too bad it doesn't reformat it properly again if the import line is shortened.

underdarknl (approved):

Issue #2116, opened 2023-12-11
Use mypy in boefjes, octopoes and rocky


dekkers:
Currently mypy is excluded in boefjes, octopoes and rocky. We should also do type checking there because it can find bugs such as #2115. There are a lot of errors, but most look easy to fix. If some typing problems are hard to fix we can also use Any or ask mypy to ignore things because that would be better than not running mypy at all.

PR #2555, opened 2024-02-23
Fix octopoes typing

dekkers:
### Changes
This fixes all typing in Octopoes.

### Issue link
#2116 (Almost done, only boefjes plugins directory still needs to be done)

---

---
## Checklist for QA:


octopoes/.ci/docker-compose.yml
```diff
@@ -20,7 +20,7 @@ services:
       args:
         ENVIRONMENT: dev
       context: .
-    command: pytest -x tests/integration --timeout=300
+    command: pytest tests/integration --timeout=300
```
dekkers, line 23: I think it's nicer to see whether just one tests fails or multiple because that gives a better indication of what goes wrong. If you want to stop pytest after a failing test you can always press ctrl-c.

octopoes/bits/spf_discovery/spf_discovery.py
```diff
@@ -86,7 +86,7 @@ def parse_a_mx_qualifiers(mechanism: str, input_ooi: DNSTXTRecord, spf_record: D
             spf_record=spf_record.reference, hostname=hostname.reference, mechanism=mechanism_type
         )
     if mechanism.startswith("a/") or mechanism.startswith("mx/"):
-        mechanism_type, domain = mechanism.split("/", 1)[1]
+        mechanism_type, domain = mechanism.split("/", 1)
```
dekkers, line 89: This is actually a real bug that was catched by mypy.

octopoes/octopoes/api/models.py
```diff
@@ -59,18 +59,21 @@ class ScanProfileDeclaration(BaseModel):
     valid_time: datetime
 
 
+ValidatedOOIType = Annotated[ConcreteOOIType, Field(discriminator="object_type")]
```
dekkers, line 62: It took me quite a long time to figure out what was going wrong when I changed the `object_type` from the base OOI form `Literal["OOI"]` to `str` to fix typing conflicts with all the subclasses. What happened was that it created an IPAddress object instead of an IPAddressv4 object because the IPAddress suddenly also matched because `object_type` wasn't `Literal["OOI"]` anymore.

Using [discriminated unions](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions) is actually a better way to do this matching. Pydantic will only try to validate against the type that matches the object_type instead of all the types. This is more efficient and any validation error will only about that type and not all every other type it would otherwise try. This doesn't work with abstract types that don't define `object_type` so I had to create the `ConcreteOOIType` type that remove those.

octopoes/octopoes/events/events.py
```diff
@@ -33,7 +33,9 @@ class OOIDBEvent(DBEvent):
 
     @property
     def primary_key(self) -> str:
-        return self.new_data.primary_key if self.new_data else self.old_data.primary_key
+        # There doesn't seem to be an easy way to tell mypy that if new_data is
+        # None then old_data is never None.
+        return self.new_data.primary_key if self.new_data else self.old_data.primary_key  # type: ignore[union-attr]
```
dekkers, line 38: Definining the event types better and not have an old_data and new_data which are `None` depending on the operation_type would fix this. I think we also want to change the way this works, so I decided not to refactor everything and just ignore the type errors in a lot of places.

octopoes/octopoes/core/service.py
```diff
@@ -154,7 +165,7 @@ def save_origin(self, origin: Origin, oois: list[OOI], valid_time: datetime) ->
             try:
                 self.ooi_repository.get(origin.source, valid_time)
             except ObjectNotFoundException:
-                return
+                raise ValueError("Origin source of observation does not exist")
```
dekkers, line 168: I spotted this while debugging what went wrong in saving. I think we should never silently not save something when the data is not correct, but always return an error...

octopoes/octopoes/xtdb/query.py
```diff
@@ -43,7 +43,6 @@ class Aliased:
 
 
 Ref = type[OOI] | Aliased
-A = Aliased
```
dekkers, line 46: This was not a typing issue, but I think we shouldn't create one-level classes but always use proper names...

octopoes/.ci/docker-compose.yml
ammar92: Agree; and we possibly save a lot of CI runtime as well

octopoes/octopoes/api/models.py
ammar92: This is indeed a huge change since v2. Discriminated unions weren't available before, hence the big union of concrete types. In v2 unions are also handled differently now.

It's something I've always wanted to do as soon as possible after the migration, but found it a bit risky. I'm not sure about the name yet, but I'm happy you picked this up. (Perhaps we could rename the old *OOIType* to something like *OOIBaseType* or *OOIGeneralType*, and make your suggested discriminator type *OOIType*)

octopoes/octopoes/models/path.py
```diff
@@ -61,7 +61,10 @@ def reverse(self) -> Segment:
             self.source_type,
         )
 
-    def __eq__(self, other: Segment) -> bool:
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Segment):
+            return NotImplemented
```
ammar92, line 66: You probably meant `raise` instead of `return`?

octopoes/octopoes/models/path.py
```diff
@@ -100,7 +103,10 @@ def __str__(self) -> str:
         segments = ".".join(map(str, self.segments))
         return f"{start_type}.{segments}"
 
-    def __eq__(self, other: Path):
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Path | str):
+            return NotImplemented
```
ammar92, line 108: See earlier comment

octopoes/octopoes/models/path.py
```diff
@@ -100,7 +103,10 @@ def __str__(self) -> str:
         segments = ".".join(map(str, self.segments))
         return f"{start_type}.{segments}"
 
-    def __eq__(self, other: Path):
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Path | str):
```
ammar92, line 107: You probably want a tuple of `Path` and `str`, otherwise you're checking against an instance of a `Union[Path, str]`
```suggestion
        if not isinstance(other, (Path, str)):
```

rocky/reports/report_types/name_server_report/report.py
```diff
@@ -71,9 +71,9 @@ def generate_data(self, input_ooi: str, valid_time: datetime) -> dict[str, Any]:
 
         try:
             ooi = self.octopoes_api_connector.get(Reference.from_str(input_ooi), valid_time)
-        except ObjectNotFoundException as e:
-            logger.error("No data found for OOI '%s' on date %s.", str(e), str(valid_time))
-            raise ObjectNotFoundException(e)
+        except ObjectNotFoundException:
+            logger.error("No data found for OOI '%s' on date %s.", str(ooi), str(valid_time))
```
ammar92, line 75: The formatter when using `%s` already implicitly casts/ transforms the argument to a string right?

ammar92 (commented):
Very nice stuff. Just a few remarks/ questions, but consider it approved

octopoes/octopoes/models/path.py
dekkers: No, see https://docs.python.org/3/reference/datamodel.html#object.__eq__

octopoes/octopoes/models/path.py
dekkers: `isinstance` supports Union as argument, see https://docs.python.org/3/library/functions.html#isinstance

ammar92: Ah you're right, this changed in version 3.10 👍

octopoes/octopoes/api/models.py
dekkers: Discriminated unions were already available in 1.9 that was released in december 2021: https://docs.pydantic.dev/latest/changelog/#v190-2021-12-31

ammar92: Yep you're right. I meant to say that the way we handled these types was implemented even before 1.9, and after that we didn't implement discriminated unions yet.

rocky/reports/report_types/name_server_report/report.py
dekkers: Yes you are right, I just kept that from the original without paying attention to it. I've fixed it now.

octopoes/.ci/docker-compose.yml
Donnype: Sounds good, I vaguely remember seeing this suite freeze from time to time back then, but that was the race condition you fixed I think 👍

stephanie0x00:
## Checklist for QA:

### What works:
Nicely done. Things seem to work as expected: objects can be added and are created, boefjes can be enabled and run accordingly, tasks complete and reports seem to generate too. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

underdarknl (dismissed):

underdarknl (approved):

Issue #1913, opened 2023-10-17
Testssl boefje does not terminate


tobiasBDO:
_Please add `bug`, the name of any relevant modules (e.g. `rocky`), and any other relevant labels to your issue._

**Describe the bug**
While running the [testssl boefje](https://github.com/minvws/nl-kat-coordination/tree/main/boefjes/boefjes/plugins/kat_testssl_sh_ciphers) I run into an issue where it gets "stuck" on some jobs, without output and it can't terminate itself. It is stuck for days on end and I've had this happen multiple times, any clue as to what might be causing this?

**To Reproduce**
1. Add IP addresses to kat
2. Add corresponding port 443
3. Run boefje
4. Wait

**Expected behavior**
I expect the boefje to either terminate by itself, or give output and stop gracefully. Neither of which happens right now.

**OpenKAT version**
KAT `V1.11.0`

**Additional context**
With docker inspect we can see that it's been running since the 15th, and right now its up status is  `Up 2 days`.

```
        "State": {
            "Status": "running",
            "Running": true,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 3282281,
            "ExitCode": 0,
            "Error": "",
            "StartedAt": "2023-10-15T02:47:13.201095881Z",
            "FinishedAt": "0001-01-01T00:00:00Z"
```

ammar92:
Can you please (partially) share the logs of the container?

tobiasBDO:
> Can you please (partially) share the logs of the container?

This is the log actually:

```sh
###########################################################
    testssl.sh       3.2rc2 from https://testssl.sh/dev/

      This program is free software. Distribution and
             modification under GPLv2 permitted.
      USAGE w/o ANY WARRANTY. USE IT AT YOUR OWN RISK!

       Please file bugs @ https://testssl.sh/bugs/

###########################################################

 Using "OpenSSL 1.0.2-bad (1.0.2k-dev)" [~183 ciphers]
 on 29afe5648a1c:/home/testssl/bin/openssl.Linux.x86_64
 (built: "Sep  1 14:03:44 2022", platform: "linux-x86_64")


 Start 2023-09-15 08:20:00                -->> 145.131.16.58:80 (145.131.16.58) <<--

 rDNS (145.131.16.58):   arg-pldacl13.amsiohosting.net.
```

ammar92:
Can you also post the entire inspect result?

[...]

PR #2560, opened 2024-02-26
Set a timeout on hanging test ssl container

noamblitz:
### Changes
Set a timeout on test ssl container of 5 minutes. Also containers are now correctly cleaned up.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/1913

### Demo

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


boefjes/boefjes/plugins/kat_testssl_sh_ciphers/main.py
```diff
@@ -26,13 +27,14 @@ def run(boefje_meta: BoefjeMeta) -> list[tuple[set, bytes | str]]:
...
     )
 
-    container.wait()
-
-    output = get_file_from_container(container, "tmp/output.json")
-
     try:
+        container.wait(timeout=300)
+        output = get_file_from_container(container, "tmp/output.json")
+    except (docker.errors.DockerException, RequestException) as e:
```
ammar92, line 33: It's better to stick to the [API's expected exceptions](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.wait): `requests.exceptions.ReadTimeout` and `docker.errors.APIError`.

boefjes/boefjes/plugins/kat_testssl_sh_ciphers/main.py
```diff
@@ -26,13 +27,14 @@ def run(boefje_meta: BoefjeMeta) -> list[tuple[set, bytes | str]]:
...
+        logging.warning("DockerException occurred: %s", e)
+        container.stop()
+        output = ""
```
ammar92, line 36: @underdarknl Wouldn't we still want some kind of boefje failure (and possibly reschedule)? I believe testssl would still produce some JSON output even if nothing interesting comes out. I'd rather see something than silently ignoring this because this will still show as a successful boefje run.

ammar92 (commented):
I can't confidently say that this is the final solution as we don't exactly know why this container freezes, but it's definitely some improvement. Just had a few tiny remarks.

boefjes/boefjes/plugins/kat_testssl_sh_ciphers/main.py
noamblitz: Yeah its weird, because I have tried those and it doesnt catch them correctly... Will try again tomorrow

boefjes/boefjes/plugins/kat_testssl_sh_ciphers/main.py
noamblitz: Testssl doesn't produce an output unfortunately. It hangs before doing anything, see: https://github.com/minvws/nl-kat-coordination/issues/1913#issuecomment-1765991030

noamblitz: We can choose to let the boefje fail! That would make sense indeed. Would also mean that I can remove the changes on the normalizer

underdarknl: We'd need to differentiate between a timeout in reaching the target (which would mean we need to reschedule), and a hanging process because testssl is unhappy about something else (like trying ssl tests on a non ssl port). The later could just be shortcutted by having the boefje return a succes state but not do any work like we do for other boefjes that do their own work filtering.

boefjes/boefjes/plugins/kat_testssl_sh_ciphers/main.py
ammar92: For this, you need to upgrade the `docker` package to the latest version (7.0.0)

boefjes/boefjes/plugins/kat_testssl_sh_ciphers/schema.json
```diff
@@ -0,0 +1,12 @@
+{
+  "title": "Arguments",
+  "type": "object",
+  "properties": {
+    "TIMEOUT": {
+      "title": "TIMEOUT",
+      "type": "integer",
+      "description": "Timeout for requests in the test ssl script",
+      "default": 30
```
ammar92, line 9: You need to include a `minimum` (of 0) property, otherwise this allows negative numbers

boefjes/boefjes/plugins/kat_testssl_sh_ciphers/main.py
noamblitz: We use docker V6, not sure if we should upgrade for the sake of this PR

noamblitz:
Exceptions seem to not be caught correctly with the new changes.

noamblitz:
With the latest commit it works again. Can be reviewed, docker should be backward compatible.

noamblitz:
Can be qa'd by:

Scan a specific ip that hangs (ping me). Boefje should fail with output: this ipservice does not have tls. Set timeout in katalogus setting to 5000. Now boefje should fail.

boefjes/boefjes/plugins/kat_testssl_sh_ciphers/main.py
```diff
@@ -19,20 +22,29 @@ def run(boefje_meta: BoefjeMeta) -> list[tuple[set, bytes | str]]:
...
-
-    output = get_file_from_container(container, "tmp/output.json")
-
     try:
+        container.wait(timeout=300)
+        output = get_file_from_container(container, "tmp/output.json")
+    except (APIError, RequestException) as e:
+        logging.warning("DockerException occurred: %s", e)
+        container.stop()
+        raise TimeoutError("Timeout occurred while running testssl.sh")
```
dekkers, line 46: I don't think timeout is the only error that can happen, so maybe raise a more general error?

boefjes/boefjes/plugins/kat_testssl_sh_ciphers/normalize.py
```diff
@@ -44,7 +44,8 @@ def run(normalizer_meta: NormalizerMeta, raw: bytes | str) -> Iterable[OOI]:
     boefje_meta = normalizer_meta.raw_data.boefje_meta
     input_ooi = Reference.from_str(boefje_meta.input_ooi)
     output = json.loads(raw)
-
+    if not output:
+        return
```
dekkers, line 48: I think this can be removed now that we raise an error?

ammar92:
> Exceptions seem to not be caught correctly with the new changes.

Just out of curiosity, did you get any other exceptions? According to the docs the `ReadTimeout` would be the expected exception to catch. But since `ReadTimeout` also extends `RequestException` and the latter is the broader exception, I think the suggested implementation is appropriate here. Also because the package depends on `requests` and other exceptions than read timeouts may occur while talking to the Docker API.

ammar92:
> > Exceptions seem to not be caught correctly with the new changes.
> 
> Just out of curiosity, did you get any other exceptions? According to the docs the `ReadTimeout` would be the expected exception to catch. But since `ReadTimeout` also extends `RequestException` and the latter is the broader exception, I think the suggested implementation is appropriate here. Also because the package depends on `requests` and other exceptions than read timeouts may occur while talking to the Docker API.

Just for documentation and knowledge sharing: after examining the logs with @noamblitz we found out that `docker` doesn't actually yield a `requests.exceptions.ReadTimeout`, but rather in this case a `requests.exceptions.ConnectionError` (that was caused by a `urllib3.exceptions.ReadTimeoutError`, which was caused by a socket's `TimeoutError`).

ammar92 (dismissed):

boefjes/boefjes/plugins/kat_testssl_sh_ciphers/main.py
```diff
@@ -19,20 +22,29 @@ def run(boefje_meta: BoefjeMeta) -> list[tuple[set, bytes | str]]:
...
-
-    output = get_file_from_container(container, "tmp/output.json")
-
     try:
+        container.wait(timeout=300)
+        output = get_file_from_container(container, "tmp/output.json")
+    except (APIError, RequestException) as e:
+        logging.warning("DockerException occurred: %s", e)
+        container.stop()
+        raise Exception("Timeout occurred while running testssl.sh")
```
dekkers, line 46:
```suggestion
        raise Exception("Error occurred (possibly a timeout) while running testssl.sh")
```

The reason I am nitpicking about this is that in my experience software that gives hard to understand or incorrect error message makes it hard to figure out what goes wrong and can cause a lot of frustation when you use such software.

underdarknl (approved):

Issue #2526, opened 2024-02-19
Use new DNS report generation and remove Keiko DNS report from onboarding


Rieven:
In the onboarding we still are generating a DNS report using Keiko DNS Report.

We have now a new flow for DNS report generation.

Issue:
Remove Keiko DNS Report --> DNS report from "generate report" in Reports flow.

Issue #1854, opened 2023-10-02
Boefjes tiles shows (partially) old design in onboarding


Rieven:
The design of required plugins at onboarding showing old design:
![image](https://github.com/minvws/nl-kat-coordination/assets/30832687/2aa3bef6-2554-4853-8e94-14fe65c49116)

Should see the new design style:
![image](https://github.com/minvws/nl-kat-coordination/assets/30832687/add27b7a-aad4-422c-90df-87b05ff498fe)

Issue #2139, opened 2023-12-15
Hide optional select options when no options are available in ooi-forms


underdarknl:
**Is your feature request related to a problem? Please describe.**
If (for example during onboarding) an OOI has an optional reference, but there are no choices available. we can hide the whole select.

**Describe the solution you'd like**
Hide the selection box for references where there is no choice available.

dekkers:
I think for the onboarding it wouldn't be a problem to hide it, but I am not sure whether hiding it in other places is the right thing to do UX wise, or we should still show it and disable it?

I think part of the problem is that we also expose all the internal data structures to the user and some of those fields aren't supposed to be specified by the user at all, whether there are choices for it or not. For example when adding an URL you get asked the web_url field which will show all the WebURL objects (so is only empty when there is no data in KAT yet) but it doesn't really make any sense to ask the user for that.

Also if a field  is required but there is only one option (for example what we have with Network usually) we can just select that option. I think in that case we should always display the field, but can disable it if there is only one option.

PR #2561, opened 2024-02-26
Refactor onboarding wizard to use Generate Report flow to create a DNS report

Rieven:
### Changes
- We have now only DNS report option for onboarding. Make it in a way future proof to catch also other reports in views. 
- Add URL form is simplified using a single URL field with validation (Django URLField). 
- `web_url` field is removed from the onboarding URL form as for the onboarding this is not necessary. There is no `web_url` yet because there is no URL yet created when onboarding at the moment of creation. 
- Clearance levels warnings are repeated throughout rocky. Move this code to `OrganizationView` so we can call it directly to check if member can set clearance levels. We need to later think about a better way to handle warnings, we want to reuse same warnings for the same functionality. (this case raising clearance levels)
- Boefje tiles are refactored which was using an old design. `SelectBoefjes` is removed as this class was a bit complex. It is fixed using the same partial as plugin tiles adding extra possibilities to insert checkboxes, and hidden input types. 
- Transform URL object into a Hostname object which is required to generate a DNS report. We need to find a way to easily get a Hostname from a URL object. Would be perfect using queries but it does not work for now. New ticket to fix this you can find here: https://github.com/minvws/nl-kat-coordination/issues/2579
- Make url queryparameter ready to send to Generate report url. Also `observed_at` is included at moment of creation.
- DNS report is generated using url queryparameter to get all the data to generate a DNS report. Same as if you would do it from the general generate report flow.

### Issue link
https://github.com/minvws/nl-kat-coordination/issues/2526
https://github.com/minvws/nl-kat-coordination/issues/1854

Closes https://github.com/minvws/nl-kat-coordination/issues/2526
Closes https://github.com/minvws/nl-kat-coordination/issues/1854

### Demo
You can skip 3 minutes waiting time of report from 1:05 - 4:05

https://github.com/minvws/nl-kat-coordination/assets/30832687/6bc0a158-355a-4bcb-aa60-8d0c09e28bcb

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


rocky/onboarding/forms.py
```diff
@@ -36,3 +36,17 @@ class OnboardingSetClearanceLevelForm(forms.Form):
             },
         ),
     )
+
```
underdarknl, line 39: I'd rather see a solution where all the OOI-add forms do not display their optional selection fields when there are no options for such a field. There is a separate issue for that available here:
https://github.com/minvws/nl-kat-coordination/issues/2139 

following that route would also negate adding more code for this specific onboarding form.

rocky/onboarding/templates/step_2b_choose_report_type.html
```diff
@@ -29,19 +29,22 @@ <h2>{% translate "Choose a report - Type" %}</h2>
...
-                    {% endif %}
-                </div>
+                <form method="get"
+                      action="{% url "step_setup_scan_ooi_info" organization.code %}"
+                      class="inline">
+                    <div class="button-container">
+                        {% if organization %}
+                            <button name="report_type" value="dns-report" type="submit" class="button">{% translate "DNS report" %}</button>
+                        {% endif %}
+                        <a href="" class="button disabled">Pen test</a>
```
underdarknl, line 39: Could we list the available reports instead of making them up? All but the DNS can still be disabled but this way we can use the same code in the onboarding as we will on the report flow.

rocky/onboarding/templates/step_3b_setup_scan_ooi_add.html
```diff
@@ -48,14 +48,16 @@ <h3>{% translate "Dependencies" %}</h3>
...
-                    <button type="submit">{% translate "Create object" %}</button>
-                </form>
-                {% if organization %}
                     <div class="button-container">
-                        <a href="{% url "complete_onboarding" organization.code %}"
-                           class="button ghost">{% translate "Skip onboarding" %}</a>
-                    </div>
-                {% endif %}
+                        <button type="submit">{% translate "Create object" %}</button>
+                        {% if organization %}
```
underdarknl, line 53: Is it possible to get here without having an organization?

rocky/onboarding/templates/step_3c_setup_scan_ooi_detail.html
```diff
@@ -48,29 +49,21 @@ <h3>{% translate "Creating an object" %}</h3>
...
-                            <a href="{% url "complete_onboarding" organization.code %}"
-                               class="button ghost">{% translate "Skip onboarding" %}</a>
-                        {% endif %}
-                    </div>
-                </form>
+            </div>
+            <div class="button-container">
+                <a href="{% url "step_report" organization.code %}?{{ request.GET.urlencode }}"
+                   class="button">{% translate "Start scanning" %}</a>
+                {% if organization %}
```
underdarknl, line 63: Same here.

rocky/onboarding/templates/step_3d_clearance_level_introduction.html
```diff
@@ -57,7 +57,7 @@ <h3>{% translate "How to know required clearance level" %}</h3>
                     </div>
                 </div>
                 <div class="button-container">
-                    <a href="{% url "step_acknowledge_clearance_level" organization.code %}?ooi_id={{ ooi }}"
+                    <a href="{% url "step_acknowledge_clearance_level" organization.code %}?{{ request.GET.urlencode }}"
```
underdarknl, line 60: The url already has an argument attached, adding an extra ? is not valid.

rocky/onboarding/forms.py
Rieven: Yes, I will resolve this in another PR, it will be then applicable in the Create object form. I think for the onboarding a simple URL field is enough. As you can see that I need to reformat the form too much using init, while I can simple create a new object.

rocky/onboarding/templates/step_2b_choose_report_type.html
Rieven: Yes I was thinking of this indeed. Pass all the Reports we want to show in onboarding through from views to template

rocky/onboarding/templates/step_3b_setup_scan_ooi_add.html
underdarknl: Then we can remove the check on if organization to clean things up further.

rocky/onboarding/views.py
```diff
@@ -100,237 +92,74 @@ class OnboardingChooseReportTypeView(
...
-                        "OpenKAT added the following required object to your object list to complete your request: {}"
-                    ).format(str(ooi)),
-                )
-        return hidden_fields
-
-    def get_form_kwargs(self):
-        kwargs = super().get_form_kwargs()
-        hidden_fields = self.get_hidden_form_fields()
-        kwargs.update({"hidden_fields": hidden_fields, "initial": hidden_fields})
+        self.report_type = self.request.GET.get("report_type", "")
```
dekkers, line 136: I thought the onboarding always used the DNS report?

Rieven: Yes, the idea here is to use more report types in the future. But that will follow in another PR if necessary.

rocky/onboarding/templates/step_2b_choose_report_type.html
Rieven: We need to do this in another PR, there is more complexity to it. Choosing a different report type, means that we need to populate a different form to create the desired OOI that applies to that report type. DNS report would need a Hostname fetched from URL, but other report types needs different forms. Also the forms are nested, you need to create this first to create the second and so on ....

rocky/onboarding/forms.py
underdarknl: I agree, and that should be fixed in the report flow first.

Adding an OOI itself should also be made simpler by not showing optional fields that have no options. that's where the other issue (#2139) comes in.

rocky/onboarding/templates/step_2b_choose_report_type.html
underdarknl: Allowing *just* the dns report i the onboarding is fine. But lets keep this PR small, and create a new issue for that.

rocky/rocky/views/scan_profile.py
```diff
@@ -40,56 +35,9 @@ def post(self, request, *args, **kwargs):
...
-                    )
-                    % (
-                        self.ooi.reference.human_readable,
-                        level,
-                        self.organization_member.acknowledged_clearance_level,
-                    ),
-                )
-                return redirect(reverse("account_detail", kwargs={"organization_code": self.organization.code}))
-
+            return redirect(reverse("account_detail", kwargs={"organization_code": self.organization.code}))
```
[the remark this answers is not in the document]
Rieven: Good that you saw this, I need to find out what the desired flow is.

ammar92 (commented):
Don't have much to add to what has already been said. Seems like the rework is already done and most points have been addressed. I suppose we move this to QA and test/ verify it once more.

stephanie0x00:
## Checklist for QA:

### What works:
The onboarding flow is now more user friendly. 
- When adding the URL there is no optional 'web_url' part anymore. This section was never used as it could not be generated or added as a first time user. 
- The 'DNSSEC' boefje is now enabled by default. Instead of manually having to select this boefje every time. 
- The HTML version with the summary severity overview of the findings is now replaced by the PDF version of the DNS report. 
- Downloading the DNS report

### What doesn't work:
Nothing identified at this time. 

### Bug or feature?:
n/a.

underdarknl (approved):

rocky/rocky/views/ooi_detail.py
```diff
@@ -59,6 +60,12 @@ def post(self, request, *args, **kwargs):
 
     def handle_page_action(self, action: str) -> bool:
         try:
+            if action == PageActions.CHANGE_CLEARANCE_LEVEL.value:
+                clearance_level = int(self.request.POST.get("level"))
+                if not self.can_raise_clearance_level(self.ooi, clearance_level):
+                    return redirect("account_detail", organization_code=self.organization.code)
+                return self.get(self.request, *self.args, **self.kwargs)
```
dekkers, line 67: 1) The problem I described in https://github.com/minvws/nl-kat-coordination/pull/2561#discussion_r1514420064 has not been fixed. The flow was:

- No indemnification present or not high enough clearance level -> error message
- Did not acknowledge clearance level -> redirect to the account page.

The flow will always redirect to the account detail now, but that does not make any sense, because the user can't give themself a higher clearance level and also can't set an organization indemnification on that page.

2) This code is for ScanProfileDetailView, so should be in ScanProfileDetailView, not in OOIDetailView. Code specific to a certain view should live in that view class and not be moved to superclasses.

Issue #1456, opened 2023-07-25
Make sure boefje worker only accepts 'reasonable' mime-types from boefjes


underdarknl:
**Is your feature request related to a problem? Please describe.**
We need to make sure that boefjes label their raw files with reasonable mime-types. Or at least not accept single letter mine-types as is an easy mistake to make when returning a set of 1 items.

**Describe the solution you'd like**
check for the length of each mime-type, and check for the presence of a / char. If not, throw a warning.

**Describe alternatives you've considered**
None come to mind

**Additional context**
When returning a raw file with a single mimetype, its too easy to fall in this trap:
`set('image/png')`  or `set(('image/png'))`   would result in `set('m', '/', 'e', 'p', 'n', 'i', 'a', 'g')` , where  `set(('image/png',)) ` would result in the correct mimetype.

PR #2563, opened 2024-02-27
Filter invalid mimetypes in boefjes output.

underdarknl:
### Changes
Filters out mimetypes that are too short, or don't contain a slash

### Issue link
closes https://github.com/minvws/nl-kat-coordination/issues/1456

### Code Checklist

### Communication

---

---
## Checklist for QA:


dekkers:
Wouldn't it be better to have the boefje fail instead of ignoring the mimetype?

underdarknl:
> Wouldn't it be better to have the boefje fail instead of ignoring the mimetype?

Not all returned mimetypes might be invalid. And not all mime-types might be hard coded in the boefje (think http-headers being used to fill mime-types). Throwing a warning for invalid mime-types seems reasonable to me, as it might still allow other normalizers to ingest the data based on for example the boefje/$boefjename mimetype. the mime-types for each raw file in a job are visible in the task-list. but maybe we should find a way to also show any warnings there.

dekkers:
> Not all returned mimetypes might be invalid. And not all mime-types might be hard coded in the boefje (think http-headers being used to fill mime-types). Throwing a warning for invalid mime-types seems reasonable to me, as it might still allow other normalizers to ingest the data based on for example the boefje/$boefjename mimetype. the mime-types for each raw file in a job are visible in the task-list. but maybe we should find a way to also show any warnings there.

All mimetypes are currently hardcoded/whitelisted as far as I know. I also don't think that creating an opportunity for attackers to inject raw files with arbitrary mimetypes is a good idea. I actually just realised that we fixed the `produces` field of boefje definition a few months ago, so we can simplify this by checking if the mimetype is listed in the `produces` of the boefje definition and give an error if that is not the case.

underdarknl:
> > Not all returned mimetypes might be invalid. And not all mime-types might be hard coded in the boefje (think http-headers being used to fill mime-types). Throwing a warning for invalid mime-types seems reasonable to me, as it might still allow other normalizers to ingest the data based on for example the boefje/$boefjename mimetype. the mime-types for each raw file in a job are visible in the task-list. but maybe we should find a way to also show any warnings there.
> 
> All mimetypes are currently hardcoded/whitelisted as far as I know. I also don't think that creating an opportunity for attackers to inject raw files with arbitrary mimetypes is a good idea. I actually just realised that we fixed the `produces` field of boefje definition a few months ago, so we can simplify this by checking if the mimetype is listed in the `produces` of the boefje definition and give an error if that is not the case.

The produces list was never intended to be a complete list of all mime-types possible for a boefje. The whitelisting is valid though, and this is why we have/had a whitelist of mimetypes in for example the download boefje which makes sure a remote webserver cannot inject any 'special' mimetypes into openkat. (boefje/$boefje, or info/* error/* being the special ones).

We can start seeing the produces list as each boefje's whitelist. It would make things bit more strict, which is good for security (enforcing this would move to the runner, not the boefje itself), but we would still need to handle the special mime-types for errors and info message etc inside the boefje.

boefjes/boefjes/job_handler.py
```diff
@@ -23,6 +23,8 @@
 from octopoes.models.exception import ObjectNotFoundException
 from octopoes.models.types import OOIType
 
+MIMETYPE_MIN_LENGTH = 5  # two chars before, and 2 chars after the slash ought to be reasonable
```
ammar92, line 26: Although the suggested implementation works and seems as a practical approach, I was just curious if this is the right approach to validate MIME types. Why not just validate against the format `type/subtype;parameter=value`? A MIME type always has both a type and a subtype, never just one or the other, and an optional parameter can be added to provide additional details.

In my view a MIME type in the format `<two characters> / <two characters>` is not a more valid or reasonable MIME type than a non-existing or defined (with valid format) MIME type or a bogus one (e.g. `ab/cd`)

dekkers (approved):

PR #2620, opened 2024-03-07
Fix architecture check for Wappalyzer boefje

noamblitz:
### Changes
Fixed the check which architecture you use for the Wappalyzer check.


Donnype (approved):
PC (with & without container):
`x86_64`

Macbook pro M2:
`arm64`

Macbook pro M2 in container:
`aarch64`

So seems to work!

madelondohmen:
Seems to work on Windows 11 with WSL2. Output is "x86_64".

dekkers (approved):

Issue #2571, opened 2024-02-28
bit, warn on configurable list of world-writable domains in csp headers


underdarknl:
**Is your feature request related to a problem? Please describe.**
We should create a bit that does the following:
for all domains allow-listed in CSP headers:
  check if domain is listed in a list of configured dangerous domains.

This list could be a config_ooi listing domains like:
* www.jsdelivr.com
* githubpages.com
* cdn.skypack.dev

Those domains can host any JS or files from anyone, which immediately negates the whole point of using CSP to allow only secure domains.

**Describe alternatives you've considered**
Dont use a config, just hard-code the list in the bit.

underdarknl:
https://github.com/mayakyler/link-shorteners is a list of url shorteners, which by definition are also world-writable

PR #2624, opened 2024-03-08
Feature/disallowed domains in csp

noamblitz:
### Changes
Add bit that checks for disallowed domains in csp header
Add bit that asks the relevant question

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/2571

TO DO: unit test

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


boefjes/boefjes/plugins/kat_answer_parser/normalize.py
```diff
@@ -18,3 +18,13 @@ def run(normalizer_meta: NormalizerMeta, raw: bytes | str) -> Iterable[OOI]:
             bit_id="port-classification-ip",
             config=json.loads(raw),
         )
+
+    if "/bit/disallowed-csp-hostnames" in mime_types:
```
underdarknl, line 22: That leading / is weird here. I think we should fix that for all question outputs.

noamblitz: Will do! Creating questions is also a bit of a pain I noticed. Took me a while. We should revisit that when the time comes.

octopoes/bits/ask_disallowed_domains/question_schema.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "$id": "/bit/disallowed-csp-hostnames",
+  "type": "object",
+  "default": {},
+  "Port Configuration": "Root Schema",
+  "required": [],
+  "properties": {
+    "disallowed_hostnames": {
```
underdarknl, line 9: Maybe we should make sure we also allow the user to optionally deny url-shorteners instead of doing this by definition.

noamblitz: Good suggestion. Will do that

boefjes/boefjes/plugins/kat_answer_parser/normalize.py
underdarknl: When you have the Json you want (as in, this json contains all the data We need), you can use something like https://www.jsonschema.net/ to create a version of the jsonschema/question that would produce such a json.

noamblitz: Yeah but I mean that you have to add a bit that creates the question and then add it to the answer parser just feels a bit redundant. Maybe we could have one bit that adds all the questions on network for example.

noamblitz: Checked code and leading / has several uses. Cant remove for now.

octopoes/bits/ask_disallowed_domains/question_schema.json
```diff
@@ -0,0 +1,22 @@
...
+      "description": "Comma separated list of disallowed hostnames in CSP headers",
+      "type": "string",
+      "pattern": "^(\\s*(,*)[^,]+,?\\s*)*$",
+      "default": "github.com,google.com"
+    },
+    "disallow_url_shorteners": {
+      "description": "Do you want to disallow url shorteners in csp headers?",
+      "type": "string",
```
underdarknl, line 17: JsonSchema has a Boolean type. please use that.

octopoes/bits/disallowed_csp_hostnames/disallowed_csp_hostnames.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def run(input_ooi: HTTPHeaderHostname, additional_oois: list, config: dict[str, str]) -> Iterator[OOI]:
+    header_hostname = input_ooi
+    header = header_hostname.header
+
+    if header.tokenized.key.lower() != "content-security-policy":
+        return
+
+    disallow_url_shorteners = config.get("disallow_url_shorteners", "True").lower() == "true" if config else True
```
underdarknl, line 29:
```suggestion
    disallow_url_shorteners = config.get("disallow_url_shorteners", True) if config else True
```

octopoes/tests/test_disallowed_csp_hostnames.py
```diff
@@ -0,0 +1,56 @@
...
+
+def test_disallowed_csp_headers_allow_url_shortener():
+    http_header_hostname = HTTPHeaderHostname(
+        hostname=Reference.from_str("Hostname|internet|bit.ly"),
+        header=Reference.from_str(
+            "HTTPHeader|internet|1.1.1.1|tcp|443|https|internet|example.com|https|internet|example.com|443||Content-Security-Policy"
+        ),
+    )
+
+    results = list(run(http_header_hostname, [], {"disallow_url_shorteners": "false"}))
```
underdarknl, line 41:
```suggestion
    results = list(run(http_header_hostname, [], {"disallow_url_shorteners": False}))
```

noamblitz:
Unfortunately, something goes wrong in rocky which causes an error on submitting the form with booleans instead of strings

octopoes/bits/disallowed_csp_hostnames/disallowed_csp_hostnames.py
```diff
@@ -0,0 +1,44 @@
...
+LINK_SHORTENERS = link_shorteners_list()
+
+logger = logging.getLogger(__name__)
+
+
+def get_disallowed_hostnames_from_config(config, config_key, default):
+    disallowed_hostnames = config.get(config_key, None)
+    if disallowed_hostnames is None:
+        return default
+    return list(disallowed_hostnames.strip(" ").split(",")) if disallowed_hostnames else []
```
ammar92, line 19:
```suggestion
    return list(disallowed_hostnames.strip().split(",")) if disallowed_hostnames else []
```

octopoes/tests/test_disallowed_csp_hostnames.py
ammar92, line 1: Other than empty lists, the checks on the lengths of the results don't prove much. You should verify the returned output against real values (or partially at least)

ammar92 (commented):
Nice one, almost there! If you implement te test like this:
```
assert results = [ KATFindingType, Finding ]
```

This way you also verify the length and order for free 👍

ammar92 (dismissed):

octopoes/octopoes/models/ooi/config.py
```diff
@@ -9,7 +9,7 @@ class Config(OOI):
 
     ooi: Reference = ReferenceField(OOI)
     bit_id: str
-    config: dict[str, str]
+    config: dict
```
dekkers, line 12: It is possible to type this and the bit function arguments better if we want that. I don't know if we want config to be arbitrary JSON or we want to be more strict, but for arbitrary json the following type can be used:

`JSON: TypeAlias = dict[str, "JSON"] | list["JSON"] | str | int | float | bool | None`

See also https://github.com/python/typing/issues/182#issuecomment-1320974824 for more information

If we want config to always be a key-value pair where the value can be JSON we could use `dict[str, JSON]` as type. Although that might mean that we have to use casts or asserts in the bit code because mypy won't know about the json schema. If that gives too much problems another option might be `dict[str, Any]`.

@ammar92 any thoughts about what would be best?

ammar92: @dekkers I think this indeed is the best alternative, with the appropriate types and naming. I hope this will eventually end up as a builtin type in Python.

In many use cases they indeed use `dict[str, Any]`, but if we really must enforce JSON (without serialization) then the `JSON` type is the way to go. Although for me personally, the `dict[str, Any]` would also work, given that it has a good variable name, clear context or documented properly.

ammar92:
I fixed the merge conflict, should be testable now

octopoes/octopoes/models/ooi/config.py
underdarknl: Bit configs can be arbitrary json, where the schema is defined in the JsonSchema supplied in the related QuestionOOI.

stephanie0x00:
## Checklist for QA:

### What works:
- Manually adding a CSP header for an existing site/host with a bad domain in it triggers the vulnerability as shown below. Couldn't verify with just an IP-address as it's currently not possible to generate a HTTPResource without a hostname. The finding is shown below:  

image

### What doesn't work:
n/a

### Bug or feature?:
- Do we want this feature added to the CSP compliance checklist in the Web report? I don't think it is added right now. 

image

underdarknl:
I would expect the report to say 'CSP header present 1/1` but then not list the CSP as secure in the following line so '0/1'.

noamblitz:
Lets create a ticket to add this to the report. Since it was not part of the original ticket i think its not a good idea to let this PR hang on it.

octopoes/octopoes/models/ooi/config.py
noamblitz: What is the outcome of this?

noamblitz: @ammar92 @dekkers @underdarknl Everyone okay if I refactor to str, Any?

dekkers: Fine with me, we can always change this later if we want to.

underdarknl (approved):

dekkers (approved):

PR #2673, opened 2024-03-19
Add why container stdout/stderr can't be used for boefjes output

dekkers:
### Changes
This adds information why container stdout/stderr can't be used for boefjes output to the boefjes runner design document so it is clear for everybody.

---

---
## Checklist for QA:


ammar92 (approved):
Clear

underdarknl:
Would'nt the http upload integration be part of the outer layer of the boefje, Eg Boefjes now have a run() method, which returns the raw files. The http uploading would need to be done *in* the container, and not in the runner, since there is no way of getting this data cleanly out of the boefje and into that uploader.

stephanie0x00:
The only thing not clear for me is the following part. Why shouldn't this be done with OpenKAT? 

> This is something that can be done with a cluster component like filebeat, but must not be done with an application like OpenKAT.

dekkers:
> The only thing not clear for me is the following part. Why shouldn't this be done with OpenKAT?
> 
> > This is something that can be done with a cluster component like filebeat, but must not be done with an application like OpenKAT.

I added a bit more detail, so I hope it is clear now.

stephanie0x00:
Yes thanks :+1:

underdarknl (approved):

PR #2692, opened 2024-03-19
Add docker definition for load script in scheduler

jpbruinsslot:
### Changes

Add docker definition for load script in scheduler. Note this script is only used for development purposes.


ammar92:
I was curious and had some thoughts:
- I miss some context, as in why is this needed? I think it might be a bit overkill to have another Python based image for just this purpose
  - See also `cveapi` module and `boefjes/tools` package, they're relatively simple, and based on the project file and documentation it's quite clear what environment is needed to get them running
  - I'm worried that some alignment is missing project wide and introduces inconsistency, e.g. what to containerize and what not
- This seems to target the load module only, whilst there is also a benchmark module
  - I believe this could've been just a Python image with Python as the entry point such that you can target both modules
- Why not use the Mula image instead? (assuming dev environment) It contains the relevant environment and It's already built, because you need it to load data/ benchmark against in the first place

jpbruinsslot:
I'll close the PR was mainly used for development purposes, not necessary to take it into the code base.

Issue #2443, opened 2024-02-06
Build CI/CD for building versioned boefje image containers


underdarknl:
**Is your feature request related to a problem? Please describe.**
Currently our boefjes are mostly locally stored folders with python code. This needs to be transformed into a Container or archive file based system where each Boefje is self-contained with a well defined entrypoint. These artifacts should be distributable as an archive and a oci-image and contain al needed requirements for the internals of this boefje to function.

**Describe the solution you'd like**

* Describe the requirements per boefje in their own local requirements file.
* Create a build script for each boefje (or a general one if possible)
* Build ci/cd to create versioned artifacts (archive + ico-image) of released boefjes. possibly build for various architectures.
* Setup distribution channel for artifacts
* Connect katalogus to distribution channel
* Have boefje runner collect required artifact (based on local arch, runner type) from distribution channel.

Donnype:
Related:
- https://github.com/minvws/nl-kat-coordination/issues/1628
- https://github.com/minvws/nl-kat-coordination/pull/1620/files

Donnype:
Also in the future it would be great if we could provide easy tooling to build the images outside of kat, but the current scripts at least depend on some imports outside of its scope

dekkers:
### Building images

I don't think we can build every boefje that currently already uses an existing container image using a single builder Dockerfile. So we would need some logic to use a Dockerfile in the boefje directory if it exists and fallback to using the general builder Dockerfile if it doesn't. But maybe it easier to just put a Dockerfile in the boefje directory even if that would cause a bit of duplication.

I think it would be good to not increase the scope too much and tag the boefjes container the same way we do the current containers, so with the OpenKAT version when we add a git tag. This also leaves a bit of room to change the boefjes HTTP API, because we haven't really used it yet and it would be nice to be able to make changes it if necessary.

I think it shouldn't be a problem to just build all boefje images with a single build command in a development setup from source. It takes a bit of time the first time, but after that it would just use the default docker image caching. Those images will then be available in the local docker registry for KAT to use. So I don't think it is necessary to have logic to decide to pull images or build them.

For github actions I think we can use the path filter that used is in the Debian action to only build images when there are changes to the boefje in a PR. I think for releases and maybe also main we always want to build all the images, because you want to make sure that you have the latest base image / security updates.

[...]

PR #2709, opened 2024-03-22
Feature/boefjes to oci images

Donnype:
### Changes
This adds Docker (OCI) image builder setups for both python-native boefjes and boefjes that were using docker containers already. As an example for this first version, only ~`dns-records` and~ `dns-sec` were changed.

To get it working for the dnssec image, I had to change the boefje_meta argument to be a dict since else we'd be adding pydantic there and a job_models file. This can be a point of discussion.

I also added a step in the `make build` that we run during installations. We now also do `make -C boefjes images` that builds the two images. There the target also provides examples of how to manually run the docker build commands with the right arguments and tags.

~One issue here is that some alpine images only have python 3.5 available which messes a bit with some of the APIs in the standard lib. Hence the ignoring the dnssec main.py for pre-commit.~ We'll try to migrate to new python3.11-slim images where possible. In this case, it was possible.

I used `boefje.Dockerfile` per the Dockerfile naming convention and the fact that potentially normalizer Dockerfiles could perhaps also live in the same directory with the name `normalizer.Dockerfile` for instance.

~Follow up: do we still care about the group and user ids? We do not pass them anyways currently and the main reason for implementing them was specific requirements for the apps themselves, if I recall correctly.~ No, removed.

### Issue link
Closes https://github.com/minvws/nl-kat-coordination/issues/2443

### Demo & QA notes
This feature should not change the current flow or setup, but please verify that both the `dns-records` and `dns-sec` boefje work properly on a clean install (i.e. after `make kat` or `make reset`, which should build the images). Perhaps a rough notion of the new performance would also be nice as the risk here is that the new setup will induce significant overhead to smaller boefjes.

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


boefjes/Makefile
```diff
@@ -15,6 +15,7 @@ export m		# Message for alembic migration
 export revid	# Revision id to generate raw sql for
 export rev1		# Previous revision id for generating migrations
 export rev2		# New revision id for the new migration file
+export DOCKER_DEFAULT_PLATFORM=$(shell docker system info --format '{{.OSType}}/{{.Architecture}}')
```
dekkers, line 18: As far as I understand this is necessary for selecting to build linux/amd64 images on OS X? Is it always the case that docker system info return linux/amd64 on OS X or is ot also possible to run an arm64 docker?

It would also be nice to add this as a comment because it wasn't clear for me why this was set.

boefjes/boefjes/docker_boefjes_runner.py
```diff
@@ -53,16 +53,22 @@ def run(self) -> None:
 
             # save container log (stderr) to bytes
             self.bytes_api_client.login()
-            self.bytes_api_client.save_raw(task_id, container_logs, stderr_mime_types)
+            task = self.scheduler_client.get_task(task_id)
+
+            if task.status == TaskStatus.RUNNING:
+                self.boefje_meta.ended_at = datetime.now(timezone.utc)
+                self.bytes_api_client.save_boefje_meta(self.boefje_meta)  # The task didn't create a boefje_meta object
```
dekkers, line 60: I think it would be good to add that the status means the container was finished running without sending data back.

boefjes/boefjes/docker_boefjes_runner.py
```diff
@@ -53,16 +53,22 @@ def run(self) -> None:
...
+
+            self.bytes_api_client.save_raw(task_id, container_logs, stderr_mime_types.union({"stderr/boefje"}))
```
dekkers, line 62: It was already in the old code, but this saves the container log as raw file using the default boefje raw file types, which I think is not what we want.

boefjes/boefjes/plugins/kat_dnssec/boefje.Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM noamblitz/drill:latest
+
+RUN apk add --update --no-cache python3 curl && ln -sf python3 /usr/bin/python && python3 -m ensurepip
```
dekkers, line 3: I think it would be better to use python:3.11 as base image and do `apt-get install ldnsutils` to install `drill` than continue using this ancient container with outdated and probably unpatched software.

That would also mean it isn't necessary to add the exceptions to pre-commit and pyproject.toml.

boefjes/boefjes/plugins/kat_dns/main.py
dekkers, line 1: I am not really sure we should already merge this, because this would result in having to start a container for just doing a few simple dns lookups and would mean a performance downgrade.

boefjes/images/base.Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+ENV PYTHONPATH=/app:$BOEFJE_PATH
+
+# requirements.txt file is optional this way
+COPY $BOEFJE_PATH/requirements.txt* .
+
+RUN --mount=type=cache,target=/root/.cache pip install --upgrade pip  && (pip install -r requirements.txt || true)
+
+COPY ./images/docker_adapter.py .
+COPY $BOEFJE_PATH $BOEFJE_PATH
```
dekkers, line 23: There should also be `USER nonroot` at the end to make sure we don't run as root in the container.

boefjes/images/boefje_entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+curl -s "$1" > input.json
+OUTPUT_URL=$(python3 -c "import sys, json; print(json.loads(sys.stdin.read())['output_url'])" < input.json)
+
+# Pass the task to the boefje entrypoint and save the output, or send back that the task has failed
+python -m docker_adapter < input.json > out.json || curl -s -X POST -d '{"status": "FAILED"}' "$OUTPUT_URL"
+curl -s -X POST -f -d '@out.json' -H "Content-type: application/json" "$OUTPUT_URL"
```
dekkers, line 10: I think it would be more efficient to do everything in python than to start python twice and parse the json twice.

boefjes/boefjes/plugins/kat_dnssec/boefje.Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+ARG BOEFJE_PATH
+ENV PYTHONPATH=/app:$BOEFJE_PATH
+
+ENTRYPOINT ["/app/boefje_entrypoint.sh"]
+
+COPY ./images/boefje_entrypoint.sh /app/boefje_entrypoint.sh
+COPY ./images/docker_adapter.py .
+COPY $BOEFJE_PATH $BOEFJE_PATH
```
dekkers, line 12: We also need to make sure we don't run as root in the container.

boefjes/Makefile
Donnype: This is mainly for _building_ the local images with the `images` target. When I build the image locally on my macbook, the default platform still targets `linux/amd64` which is then not usable on the same macbook unfortunately.

Donnype: Comment added 👍

boefjes/images/base.Dockerfile
Donnype: Good spot!

Donnype: Follow up: do we still care about the group and user ids? We do not pass them anyways currently and the main reason for implementing them was specific requirements for the apps themselves, if I recall correctly.

boefjes/boefjes/plugins/kat_dns/main.py
Donnype: Perhaps we could plan/create for a proper benchmark after this ticket. Then I'll just remove the `oci_image` field from the current json file.

boefjes/boefjes/plugins/kat_dnssec/boefje.Dockerfile
Donnype: I guess this is not a big change and `drill` is just a widely used "standard" library? Then agreed. I also considered this image as a challenge to see that we could at least easily support the alpine 3.4.0 image, but where we can upgrade the base images we can definitely upgrade them in this process.

boefjes/images/boefje_entrypoint.sh
Donnype: The main reason for starting out this way (first versions I must say) was for one part seeing if we could leverage more generic shell tooling for this part that could also live between other adapters for other languages. But with the decision to use the `docker_adapter.py` as an adapter for other images as well now this is now a bit redundant. The other reason was not introducing requests as a dependency in the image, but now that we're installing python that's not a big extra step and otherwise `urllib` is not a scary package for just a small script doing 2 requests.

boefjes/boefjes/docker_boefjes_runner.py
Donnype: I also think that's the case. I'll remove it

Donnype: Or do you mean that we should use other mime types?

Donnype: Update: the scheduler needs the mime-types to start with `error/`, so this will not work in general. If we only save the error logs when the task is still running we could just use the `error/boefje` mime type here now. Follow up ticket: https://github.com/minvws/nl-kat-coordination/issues/2728

boefjes/images/base.Dockerfile
Donnype: Update: removed

boefjes/images/boefje_entrypoint.sh
Donnype: Update: installed httpx and removed the entrypoint

stephanie0x00:
## Checklist for QA:

### What works:
DNSSec boefje works and appear no error messages in the logs.  After Review is done it can be moved to Ready to merge. 

I followed the following approach while timing: I let the DNSrecords and DNSzone boefjes run and finish, then enable the dnssec boefje and start the stopwatch. Data is against mispo.es with L1 and timed until the KAT-NO-DNSSEC Finding Type is completed under the Normalizer tab: 

- Current main it takes the times below to finish. It takes about ~30s for the first dnssec task to show up. 
     - 1st time:-  ~40 seconds
     - 2nd time: ~23 seconds
     - 3rd time: ~45 seconds
     - 4th time: ~47 seconds

- This branch to run. It takes about ~40seconds for the first dnssec related task to show up in the list:
    - 1st time: 1m15seconds
    - 2nd time: ~40 seconds
    - 3rd time: ~55 seconds
    - 4th time: 1m05s.

### What doesn't work:
n/a

### Bug or feature?:
n/a

boefjes/boefjes/plugins/kat_dnssec/main.py
```diff
@@ -1,22 +1,13 @@
...
-    input_ = boefje_meta.arguments["input"]
+def run(boefje_meta):
+    input_ = boefje_meta["arguments"]["input"]
     domain = input_["name"]
 
-    client = docker.from_env()
-
     # check for string pollution in domain. This check will fail if anything other characters than a-zA-Z0-9_.- are
     # present in the hostname
     if not re.search(r"^[\w.]+[\w\-.]+$", domain.lower()):
```
underdarknl, line 13: Have we checked punicode domainname compatibility here?

underdarknl (approved):

PR #2727, opened 2024-03-26
Replace Wappalyzer

### LINKED ISSUE #2533: Replace Wappalyzer
## User Story
The current Wappalyzer Boefje works on HTTPHostnameURL OOIs which not ideal (as for instance the respective A-records are not checked causing issues like described in #1963). 

Additionally, the current code is outdated leaving place for improvements, the current up to date Wappalyzer code is however restricted.

## Acceptance Criteria:
- Replace Wappalyzer with something open source, locally run-able and up to date like for instance https://github.com/rverton/webanalyze
- Have Wappalyzer replacement work on (resolved) IP's.

ammar92:
### Changes
This implements [python-Wappalyzer](https://github.com/chorsley/python-Wappalyzer) that replaces the Wappalyzer boefje and normalizer. Unfortunately, the most recent version of that package has not yet been released on pypi, hence the git dependency in the pyproject file.

### Issue link
Closes #2533 

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


boefjes/boefjes/plugins/kat_webpage_analysis/main.py
```diff
@@ -66,13 +67,36 @@ def run(boefje_meta: BoefjeMeta) -> list[tuple[set, bytes | str]]:
...
 
+    # in case of a full response object, we hexdump to avoid issues with binary data or different encoding
+    response_dump = json.dumps(create_response_object(response))
+    content = binascii.hexlify(response.content).decode()
+
     return [
-        ({"openkat-http/full"}, f"{response.headers}\n\n{response.content}"),
-        ({"openkat-http/headers"}, json.dumps(dict(response.headers))),
+        ({"openkat-http/response"}, f"{response_dump}\n\n{content}"),
+        ({"openkat-http/headers"}, response_dump),
```
dekkers, line 76: This changes the format of openkat-http/headers raw files which will result in that the changed headers normalizer won't be able to parse the old raw files. Also if other people have created their own normalizers this will also break. I don't think we should just change the format of raw files this way.

boefjes/boefjes/plugins/kat_webpage_analysis/analyze_page/normalizer.json
```diff
@@ -1,7 +1,7 @@
 {
-  "id": "kat_website_software_normalize",
+  "id": "kat_webpage_analysis_wappalyzer_normalize",
```
dekkers, line 2: I think it would be better if the directory of the boefje/normalizer matches with the id of the boefje/normalizer like we usually do.

I also think that normalizer would be better placed in the boefjes/plugins directory and not in boefjes/plugins/kat_webpage_analysis directory. It makes it easier to find, but there is also only a loose connection between boefje and normalizer. In the future there might be another boefje that also produces `openkat-http/response` and then it also wouldn't make a lot of sense that the normalizer inside the directory of another boefje. It also would make it easier to create a container image with the just the boefjes.

ammar92: The reason that it was placed under `kat_webpage_analysis` was because they share a common function, namely analyzing webpages. However, you're point about its high coupling is a good one, hence I moved the directory to the plugins base folder.

originalsouth:
Seems to work as advertised.

boefjes/boefjes/plugins/kat_webpage_analysis/main.py
```diff
@@ -66,13 +67,36 @@ def run(boefje_meta: BoefjeMeta) -> list[tuple[set, bytes | str]]:
         if content_type[0] in ALLOWED_CONTENT_TYPES:
             body_mimetypes.add(content_type[0])
 
+    # in case of a full response object, we hexdump to avoid issues with binary data or different encoding
+    response_dump = json.dumps(create_response_object(response))
+    content = binascii.hexlify(response.content).decode()
```
dekkers, line 72: I missed this the first time, but doing a hexdump will increase the storage requirements. Bytes support binary data, so if we can avoid this it would be better.

ammar92: I've thought about this, but the problem I had was differentiating between the encoding formats for the response meta and response content (which could be different). I agree on the size though and it's probably worth it to spend some little time to find come up with a better solution.

originalsouth: Base64?

ammar92: > Base64?

Base64 and other more efficient encodings could be interesting, but I'm currently trying something out that doesn't change the content

dekkers:
Looks good to me, moved to QA

stephanie0x00:
## Checklist for QA:

### What works:
The new Wappalyzer works as expected, when scanned against mispo.es both 36 findings are shown on the main branch compared to this branch (DNS, nmap TCP, wappalyzer and webpage analysis boefjes are enabled). 


### What doesn't work:
For another added domain the number of findings are different 61 (this branch) vs 69 (main) findings when scanning mispo.es and an additional domain. The difference appears to be in the identification of some jQuery findings for the additional domain. My guess is that the javascript files are not detected based on the extension for the additional domain, as the javascript files have a .gjs extension. 

### Bug or feature?:
See above.

ammar92:
> For another added domain the number of findings are different 61 (this branch) vs 69 (main) findings when scanning mispo.es and an additional domain. The difference appears to be in the identification of some jQuery findings for the additional domain. My guess is that the javascript files are not detected based on the extension for the additional domain, as the javascript files have a .gjs extension.

Good observation and thanks for the review. The new implementation is not a fully compatible replacement for the current implementation, so (minor) differences are expected.

underdarknl (approved):

Issue #2140, opened 2023-12-15
Incorrect Findings (with snyk.io), Because the Version Number Cannot Be Found (by Wappalyzer)


jordy-kennisnet:
**Incorrect Findings (with snyk.io), Because the Version Number Cannot Be Found (by Wappalyzer)**

When software is identified on a website without an associated version, all potential vulnerabilities are displayed. However, this is often unjustified because the software component used is, in fact, up-to-date.

**To Reproduce**
1.	Scan a website with Wappalyzer (boefje).
2.	Scan a software object without a version number with Snyk.io (boefje).
3.	You will receive all vulnerabilities because the version number is 'null.'

**Expected Behavior**
When Wappalyzer identifies software without a version number, I don't want to see all the potential vulnerabilities. I expect a finding that indicates the inability to retrieve the version of a software component. This finding can then be corrected by manually entering the version number or ignoring it. If there is a policy against displaying version numbers, and you don't want findings related to version numbers not being found, a solution could be implemented with a setting per organization.

**Screenshots**
None

**OpenKAT version**
1.12

**Additional context**
In the RAW file of Wappalyzer you can see that angular is found with 100% confidence, but with version null.
{"slug":"angular","name":"Angular","confidence":100,"version":null,"icon":"Angular.svg","website": "https://angular.io","cpe":"cpe:/a:angularjs:angular.js"

underdarknl:
Maybe we need to ask the user for the version via a BIT, or maybe we make a config and BIT that allows the user to select the desired findings.

PR #2730, opened 2024-03-27
Dont report vulnerabilites without version info of the software for snyk

noamblitz:
### Changes
Before, when there was no version info available, the snyk boefje would report all vulns, now it will not do that anymore.

Reproduce by adding Handlebars software and softwareinstance without version info and enable snyk boefje, no vulns will appear.

### Closes
Closes https://github.com/minvws/nl-kat-coordination/issues/2140


boefjes/boefjes/plugins/kat_snyk/check_version.py
```diff
@@ -170,4 +170,4 @@ def check_version_in(version: str, versions: str):
     in_range = False
     while not in_range and all_versions:
         in_range, all_versions = check_version_agains_versionlist(version, all_versions)
-    return in_range
+    return in_range if version else False
```
underdarknl, line 173:
```
if not version:
  return False
```
 can be the first line of this method. Or better yet, we could never call this method at all of there is no version.

noamblitz: hmm that makes sense yes, not the smartest solution this...

underdarknl (approved):

Issue #2535, opened 2024-02-20
RPKI boefje doesn't return the expected "KAT-INVALID-RPKI" finding


dekkers:
We have a bug in the RPKI boefje. Is it possible to use the deliberate known 'bad' IP space in our unit tests? The following subnet is marked as 'invalid': 209.24.0.0/24 as shown on:  https://bgp.he.net/net/209.24.0.0/24. 

Other IPs that can be used are shown on this site: https://nlnog.net/tools/rpki-mapper/. The prefixes listed are:
- 209.24.0.0/24: ROA Signed and Invalid;  IRR Match - Parent entry found
- 209.24.1.0/24: ROA Signed and Valid; IRR Match - Parent Entry Found
- 194.32.71.0/24: ROA Signed and Invalid; IRR Valid

When checking against a known 'invalid' IP address OpenKAT doesn't return the expected "KAT-INVALID-RPKI" finding. Below are the snippets from the raw file for this IP. 

```
{"vrps_records": [{"prefix": "209.24.0.0/24", "expires": "2024-02-20T21:00", "ta": "arin"}], "notexpired": true, "exists": true}
```

```
[...]

Issue #2331, opened 2024-01-15
RPKI boefje should perform signing checks


stephanie0x00:
As mentioned in #2255:

> Current RPKI boefje doesn't perform any signing checks, it only checks if a key is present and if this isn't expired.

> A separate boefje would also be needed to verify if once's upstream provider checks for RKPI validity and blocks any invalid or expired routes.

PR #2759, opened 2024-04-02
RPKI Improvements

noamblitz:
### Changes
Added a validity check of RPKI through bgp tools. Also removed expiry checks because there are no expired record in the vprs.json.

### Issue link
Closes https://github.com/minvws/nl-kat-coordination/issues/2535
Closes https://github.com/minvws/nl-kat-coordination/issues/2331

### Demo

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


boefjes/boefjes/plugins/kat_rpki/main.py
```diff
@@ -13,39 +13,62 @@
...
-            rpki_meta = json.load(json_meta_file)
+        rpki_json = load_json(RPKI_PATH)
+        rpki_meta = load_json(RPKI_META_PATH)
+
+    if not BGP_PATH.exists() or cache_out_of_date(BGP_META_PATH):
+        bgp_data, bgp_meta = refresh_cache(
+            BGP_SOURCE_URL, BGP_PATH, BGP_META_PATH, hash_algorithm, file_extension="jsonl"
+        )
+    else:
+        bgp_data = load_jsonl(BGP_PATH)
```
ammar92, line 49: Not sure if it's used anywhere, but  perhaps you forgot to load `bgp_meta` here as well?

boefjes/boefjes/plugins/kat_rpki/main.py
```diff
@@ -13,39 +13,62 @@
...
     for roa in rpki_json["roas"]:
         if IPAddress(ip) in IPNetwork(roa["prefix"]):
-            exists = True
             expires = datetime.fromtimestamp(roa["expires"])
-            roas.append({"prefix": roa["prefix"], "expires": expires.strftime("%Y-%m-%dT%H:%M"), "ta": roa["ta"]})
-            if expires > now:
-                notexpired = True
+            asn = roa["asn"]
+            roas.append(
+                {"prefix": roa["prefix"], "expires": expires.strftime("%Y-%m-%dT%H:%M"), "ta": roa["ta"], "asn": asn}
```
ammar92, line 60: I'm not sure if the datetime format and timezone is specified, but you might want to include a timezone in this ISO format?

ammar92 (commented):
Good stuff 👍 Just a few suggestions that might improve it even more.

Also in general; some of the environment variables, such as user agent and paths, are almost guaranteed to be used on every run. Extracting them to constants, or apply dependency injection by passing them as arguments instead, will make the code less coupled and better (unit) testable.

ammar92 (approved):

synnack:
Haven't tested yet, from review it looks like it's checking the right sources, but I'm not sure all relevant cases are handled.

The code currently checks if a ROA exists and if the ROA matches a record what's seen in the global routing table.
That's good.

Then we have the following cases:
1. No ROA exists: Unprotected by RPKI (Should be a finding probably if no shorter prefix has coverage)
2. The route does not exist in the global routing table (BGP offline? Maybe a shorter prefix has coverage?, This is *not* invalid, just absent)
3. Routes exists, but no route matches the ROA (RPKI invalid, but likely there would be an incorrect ROA, severe reachability problems)
4. A route exists that matches, but another one that does not match, the incorrect route would be RPKI invalid, indicative of a hijack attempt or misconfiguration in BGP by somebody on the internet

The last two cases have RPKI-INVALID routes, though the most serious one from a security standpoint is probably case 4. Currently the code would *not* catch case 4, but *would* catch case 2 as invalid.

synnack:
The other case that is not checked, which probably should be in a separate boefje is mentioned by Jan in case A in #2535. 

> A, is my upstream connections blocking traffic if a signing key is invalid? https://rpkitest.nlnetlabs.net/ A simple ping to 77.162.48.82 and 2a02:a445:fb2f:1:222d:342d:9ffe:9d7f should tell you if this is the case.

This tests if INVALID routes can be pinged by the ISP that connects the openkat instance to the internet. Could be useful when openkat remote agents are supported.

cat agent

synnack:
Looking into seeing which cases can be covered with small code changes:
Case 1 NO-RPKI-ROA check can be added by checking if the roas list is empty and yielding that FINDING.

Oversimplified code,because yielding findings are abstracted in different function :)
```
if len(roas_list) == 0 and len(bgp_entries) > 0 and IPAddress(ip) not in rfc1918_private_ipspace:
   yield FINDING_NO_ROA
``` 

For case 2-4:
```
            invalid_exists = False
            for entry in bgp_data:
                if IPAddress(ip) in IPNetwork(entry["CIDR"]):
                    bgp_entries.append(entry)
                    if entry["ASN"] != asn:
                        invalid_exists = True

            if invalid_exists:
               yield FINDING_RPKI_INVALID_EXISTS
```

Not sure if you want to later distinguish between cases 2-4 for diagnostic purposes, but for the findings this should yield the correct results.

You can test with:
- 209.24.0.0/24: ROA Signed and Invalid; IRR Match - Parent entry found
- 209.24.1.0/24: ROA Signed and Valid; IRR Match - Parent Entry Found
- 194.32.71.0/24: ROA Signed and Invalid; IRR Valid

noamblitz:
Thanks for the feedback @synnack it really helps a lot. 

So basically for case 2-4 we are going to invert the check and not call it "ok" when there is a match, but call it "ok" when there is not a mismatch. 

That makes sense! 

In this check: 

```
if len(roas_list) == 0 and len(bgp_entries) > 0 and IPAddress(ip) not in rfc1918_private_ipspace:
   yield FINDING_NO_ROA
```
   
Why would we check on the length of bgp_entries? Isnt it true that the ROA doesn't exist even when there are also no bgp entries? Currently, we have this check
   
```
       for roa in rpki_json["roas"]:
        if IPAddress(ip) in IPNetwork(roa["prefix"]):
            exists = True
```
In the normalizer we create a finding if not exists. We can easily add the bgp check as well but I am having difficulty finding the reason.

stephanie0x00:
Mostly if there is no BGP route covering that route, then we don't need a ROA for it. It may catch things that the rfc1918 check does not, like multicast routes or other IP space that is not globally reachable.

It is not necessary if the private space list is comprehensive enough to include all IPs we don't want to create a finding for if no ROA exists.

boefjes/boefjes/plugins/kat_kat_finding_types/kat_finding_types.json
```diff
@@ -461,6 +461,12 @@
     "impact": "Without RPKI validation, your servers might be more vulnerable to unintended or malicious routing configuration errors, potentially leading to inaccessibility of your servers or interception of internet traffic directed to them.",
     "recommendation": "Make sure that the Route Origin Authorizations (ROAs) that specify which Autonomous Systems (AS) are authorized to announce your IP addresses are valid and not expired."
   },
+  "KAT-INVALID-RPKI": {
+    "description": "The route announcement that is matched by the published Route Policy and Authorization (RPKI) is invalid",
```
synnack, line 465:
```suggestion
    "description": "A route announcement that is matched by the published Route Policy and Authorization (RPKI) is invalid",
```

boefjes/boefjes/plugins/kat_rpki/normalize.py
```diff
@@ -10,14 +11,18 @@ def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
...
-        yield f
+    address = IPv4Address(ooi.tokenized.address)
 
-    if not results.get("valid") and not results.get("notexpired"):
-        ft = KATFindingType(id="KAT-EXPIRED-RPKI")
-        f = Finding(finding_type=ft.reference, ooi=ooi)
-        yield ft
-        yield f
+    # if the address is private, we do not need a ROA
+    if not address.is_private:
```
synnack, line 17:
```suggestion
    if address.is_global:
```
not is_private fails for 100.64.0.0/10, which is not global but also not technically private.

It's also sementically more correct, there is only RPKI on global IP space :)

Both cases would also fail for globally addressable IP space that is not publically routed, like 7.0.0.0/8 (US Department of Defense).  But that's probably not significant among users of openkat :)

boefjes/boefjes/plugins/kat_rpki/normalize.py
```diff
@@ -10,14 +11,18 @@ def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
     results = json.loads(raw)
     ooi = Reference.from_str(input_ooi["primary_key"])
 
-    if not results["exists"]:
-        ft = KATFindingType(id="KAT-NO-RPKI")
-        f = Finding(finding_type=ft.reference, ooi=ooi)
-        yield ft
-        yield f
+    address = IPv4Address(ooi.tokenized.address)
```
synnack, line 14: Does not seem to support IPv6? ipaddress.ip_address is the dual stack supporting function for this.

synnack (commented):
Mostly looks good, but:
- IPv4-only for the normalizer as far as I can tell can be fixed with ip_address() instead of IPv4Address
- a bug on 100.64.0.0/10 IP space may yield false warnings about no RPKI existing (impossible for that space), can be fixed with is_global() instead of not is_private()
- Minor language tweak

boefjes/boefjes/plugins/kat_rpki/normalize.py
noamblitz: Thanks! Stupid.... :)

noamblitz:
Thanks @synnack for the awesome feedback again!!!!!

boefjes/boefjes/plugins/kat_rpki/main.py
```diff
@@ -13,78 +13,149 @@
...
 from boefjes.job_models import BoefjeMeta
 
+# Paths and URLs for RPKI
 BASE_PATH = Path(getenv("OPENKAT_CACHE_PATH", Path(__file__).parent))
 RPKI_PATH = BASE_PATH / "rpki.json"
 RPKI_META_PATH = BASE_PATH / "rpki-meta.json"
 RPKI_SOURCE_URL = "https://console.rpki-client.org/vrps.json"
+
+# Paths and URLs for BGP
+BGP_PATH = BASE_PATH / "bgp.jsonl"
```
underdarknl, line 23: Did not know about jsonl, nice!

boefjes/boefjes/plugins/kat_rpki/normalize.py
```diff
@@ -10,14 +11,18 @@ def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
...
-        yield f
+    address = ip_address(ooi.tokenized.address)
 
-    if not results.get("valid") and not results.get("notexpired"):
-        ft = KATFindingType(id="KAT-EXPIRED-RPKI")
-        f = Finding(finding_type=ft.reference, ooi=ooi)
-        yield ft
-        yield f
+    # if the address is private, we do not need a ROA
+    if address.is_global:
```
underdarknl, line 17: Should we also avoid running the lookup in the boefje itself when the address is not global? seems like a quick and easy optimization.

noamblitz: We absolutely could, just felt this was a more natural split between gathering data in the boefje and drawing conclusions in the normalizer, but I'll move it tomorrow!

noamblitz: Optimized the lookup!

ammar92 (dismissed):
Looks good to me. I was worried that the latest addition (ignoring private addresses) would yield in a normalizer error since it's not the same format as the normalizer expects (a JSON dict/ object instead of a string), but it seems unlikely that this will happen.

boefjes/boefjes/plugins/kat_rpki/main.py
```diff
@@ -13,78 +14,155 @@
...
-def create_hash(data: bytes, algo: str) -> str:
-    hashfunc = getattr(hashlib, algo)
-    return hashfunc(data).hexdigest()
-
-
-def cache_out_of_date() -> bool:
-    """Returns True if the file is older than the allowed cache_timout"""
+def cache_out_of_date(meta_path: Path) -> bool:
+    """Returns True if the cache file is older than the allowed cache_timeout"""
     now = datetime.utcnow()
```
underdarknl, line 112: https://sonarcloud.io/project/issues?resolved=false&id=minvws_nl-kat-coordination&open=AY_dVgxOGgO8mB8X8BEZ
datetime.utcnow does not yield a timezone aware datetime object, changing this means we'd also need to compare against a timezone aware datetime later on, but it is cleaner to do so.

dekkers: I think the easiest way to do that is to use `isoformat` and `fromisoformat` methods that will also write the timezone information:

```
In [2]: now = datetime.now(timezone.utc).isoformat()

In [3]: now
Out[3]: '2024-07-04T09:14:56.348539+00:00'

In [4]: datetime.fromisoformat(now)
Out[4]: datetime.datetime(2024, 7, 4, 9, 14, 56, 348539, tzinfo=datetime.timezone.utc)
```

boefjes/boefjes/plugins/kat_rpki/main.py
```diff
@@ -13,78 +14,155 @@
...
+# Paths and URLs for BGP
+BGP_PATH = BASE_PATH / "bgp.jsonl"
+BGP_META_PATH = BASE_PATH / "bgp-meta.json"
+BGP_SOURCE_URL = "https://bgp.tools/table.jsonl"
+
+# Cache timeout and default hash function
 RPKI_CACHE_TIMEOUT = 1800  # in seconds
-HASHFUNC = "sha256"
+HASHFUNC = getenv("HASHFUNC", "sha256")
+USER_AGENT = getenv("USERAGENT", default="OpenKAT")
```
dekkers, line 31: Moving the call to `getenv` outside `run` to the module scope means `getenv` is called when the boefje is imported instead of when the boefje is run. As far as I can see this will mean that any boefje settings in the katalogus will not be used.

boefjes/boefjes/plugins/kat_rpki/main.py
noamblitz: Since we just use `now` to check against max-age `return (now - cached_file_timestamp).total_seconds() > maxage`, would this make any difference?

boefjes/boefjes/plugins/kat_rpki/main.py
noamblitz: Thanks for this. I moved everything in `run`!

boefjes/boefjes/plugins/kat_rpki/main.py
underdarknl: Technically, no. As both are as timezone unaware as each other, but for cleanness it would be better to specify which timezone we are dealing with, especially if we end up comparing a timestamp from an external source against any of these at some point.

noamblitz: I actually agree with you haha. I have added timezone awareness with backward compatibility.

boefjes/boefjes/plugins/kat_rpki/main.py
underdarknl: Oeh, that is a tricky little issue I did not realize, we should check the other boefjes for this behaviour.

stephanie0x00:
## Checklist for QA:

### What works:
The "RPKI-Invalid" finding is created for the IPv4 addresses 194.32.71.0 and 209.24.0.0 as expected. The IP 209.24.1.0 doesn't result in a finding (as expected). 

### What doesn't work:
See below. 

### Bug or feature?:
Generating the RPKI report for the 3 IPv4 addresses shows that all the checks are passed. However this is incorrect for 194.32.71.0 and 209.24.0.0, as these are supposed to show an incompliance. The text in the report should also be updated from "RPKI not expired" to "RPKI invalid". This can be picked up in a different ticket if preferred. 

image

dekkers (approved):

PR #2765, opened 2024-04-02
Remove sigrid workflows, no longer needed

sigio:
Data is uploaded to sigrid from a central location, once per day
These workflows are no longer needed


dekkers:
How does this provide us with feedback about the proposed changes in a PR?

underdarknl:
For now we are sticking to our own, the general push is do be disabled.

Issue #204, opened 2023-01-02
Flexible scheduling


jpbruinsslot:
## Context

The scheduler now references the `/random` endpoint from octopoes to get ooi's for rescheduling of tasks  (ooi * available boefjes = tasks). To keep track of recurring tasks within the scheduler we allow for:

* a more deterministic approach to scheduling and rescheduling of tasks, so we don't rely on randomness from objects returned by octopoes.
* more fine-grained control of the scheduling of specific tasks, potential for scheduling job independent of signals from octopoes, katalogus, and octopoes random endpoint.
* allows for scheduling other tasks such as reports
* scheduling of tasks without an associated ooi, i.e. independent run boefjes
* scheduling of normalizer tasks, e.g. run a specific normalizer task at an interval regardless of a signal from bytes.
* give better insight into what tasks are scheduled and at what time, and how many times the specific task has been executed
* leads into new ranking possibilities

#### Change 1: `Schedule` model

Create a new model `Schedule` which contains the necessary information to create a new `Task`. A `Schedule` has a 1-to-many relationship with an instanced `Task`. The updated entity relationship diagram for the scheduler:

```mermaid
erDiagram
task {
    id uuid PK
    scheduler_id str
    schedule_id uuid FK
    hash str
    priority int
    status taskstatus
    data jsonb
    created_at timestamp
    modified_at timestamp
}

schedule {
    id uuid PK
    scheduler_id str
    hash str
    data jsonb
    enabled bool
    schedule str
    tasks list[task]
    deadline_at timestamp
    created_at timestamp
    modified_at timestamp
}

task }o--|| schedule: ""
```

With both models looking like follows:

https://github.com/minvws/nl-kat-coordination/blob/394202903d434cbb9f950a89b403421feb44ed63/mula/scheduler/models/schema.py#L22-L41

> [!NOTE] This model is in the process to be rename to `Schedule`.

https://github.com/minvws/nl-kat-coordination/blob/394202903d434cbb9f950a89b403421feb44ed63/mula/scheduler/models/task.py#L47-L66

The `data` field contains the actual data that is needed for a task runner to execute its task, in case of a `BoefjeScheduler` this is a `BoefjeTask`:

```javascript
{
  "id": "",
  "boefje": {},
  "input_ooi":  "",
  "organization": ""
}
```

#### Change 2: Removal of the `PrioritizedItem` table and model

[...]

PR #2786, opened 2024-04-04
Flexible scheduling

jpbruinsslot:
Please refer to https://github.com/minvws/nl-kat-coordination/issues/204 for more background on this PR

## Changes

Since this is a big PR with substantial changes, I've added the following for more guidance on review and qa. It is helpful to read the above-linked issue to get more context. And whenever you need me I can give you this as well in a call.

The main changes are:

* Addition of a new `Schedule` model to support a more deterministic approach to (re)scheduling of tasks
* Removal of the `PrioritizedItem` model and its database table. The queue is not a separate table anymore, a 'view' of the database filtered on status `queued` is the current state of the queued items of a scheduler.

Specific code changes of interest:

* Model changes
  * `Schedule` model added to `models/schedule.py`
    * migrations added ( take specific note of the `p_item` change to `data` in the task table)
    * store added to `storage/schedule_store.py`
  * Removed `PrioritizedItem`
     * migration added for removal of `items` table
  
 * `BoefjeScheduler` changes
   * added a new process that adds new tasks to be picked up by checking the schedule of prior tasks: `push_tasks_for_rescheduling()`

* `NormalizerScheduler` changes
  * minor changes made to support the new setup

* Abstract base class `Scheduler` changes - some methods have been moved around so below are the main points of interest:
  * minor changes made to support the new setup
  * The `post_push()` method creates a `Schedule` for when a task has been added to the queue.

* API server changes
  * Added endpoints for `Schedule`
  * Updated `Task` endpoints to support the new setup

* Boefje Runner changes to support the new setup

* Rocky changes to support the new setup

## QA notes

* Since this pr is now a replacement for the rescheduling of tasks within the scheduler the functionality of KAT should remain the same. So the main issues that could arise is from starting, (automatic) rescheduling of scans. All other actions of scanning that would touch the scheduler.
  * To check if `schedule` objects have been created for the boefje tasks,  you can run through the onboarding, do some scans  and check the schedules `curl http://localhost:8004/schedules`
* Migrations, since there are substantial data migrations with this pr, care should be given on checking whether migrations from data prior to this pr will applied correctly with this pr.
  * Check out `main`
  * `make reset`
  * Walk through the onboarding, do some scans
  * Check out `feature/mula/refactor-queue`
  * `docker compose kill` shutdown all the containers
  * `make kat` (we don't want to reset we want the migration to be applied on the initial run)

---

### Code Checklist

---

---

## Checklist for QA:


mula/scheduler/models/queue.py
```diff
@@ -1,65 +1,24 @@
...
-        server_default=func.now(),
-    )
-
-    modified_at = Column(
-        DateTime(timezone=True),
-        nullable=False,
-        server_default=func.now(),
-        onupdate=func.now(),
-    )
+class PrioritizedItem(Task):
```
jpbruinsslot, line 20: To make it easier for development I've opted (for now) to subclass a `PrioritizedItem` from a `Task`. This can be refactored eventually to fully remove the `PrioritizedItem` class from the scheduler

mula/scheduler/models/schema.py
```diff
@@ -0,0 +1,90 @@
...
+
+    hash = Column(String(32), nullable=True)  # TODO: unique=True
+
+    data = Column(JSONB, nullable=False)
+
+    enabled = Column(Boolean, nullable=False, default=True)
+
+    schedule = Column(String, nullable=True)
+
+    # TODO: cascade
```
jpbruinsslot, line 71: Check and write tests when deletion of either `TaskSchema` or `Task`

mula/scheduler/models/task.py
```diff
@@ -47,34 +47,40 @@ class TaskStatus(str, enum.Enum):
 class Task(BaseModel):
```
jpbruinsslot, line 47: `Task` now get the full functionality of what a `PrioritizedItem` had.

mula/scheduler/queues/boefje.py
```diff
@@ -21,4 +21,7 @@ def create_hash(self, p_item: models.PrioritizedItem) -> str:
         input_ooi = dict_utils.deep_get(p_item.model_dump(), ["data", "input_ooi"])
         organization = dict_utils.deep_get(p_item.model_dump(), ["data", "organization"])
 
-        return mmh3.hash_bytes(f"{input_ooi}-{boefje_id}-{organization}").hex()
+        if input_ooi:
```
jpbruinsslot, line 24: This change will allow for the execution of boefjes that don't have a related OOI

mula/scheduler/schedulers/boefje.py
```diff
@@ -282,59 +283,118 @@ def push_tasks_for_random_objects(self) -> None:
...
+                        filters.Filter(
+                            column="deadline_at",
+                            operator="lt",
+                            value=datetime.now(timezone.utc),
+                        ),
+                        # TODO: ad enabled filter
+                    ]
+                )
             )
-        except httpx.HTTPError:
```
jpbruinsslot, line 290: Check why, and how this got deleted and if it is intentional

jpbruinsslot: Or is it an artifact of the random method

mula/scheduler/schedulers/scheduler.py
```diff
@@ -245,119 +248,140 @@ def push_item_to_queue(self, p_item: models.PrioritizedItem) -> None:
...
+                    schema.id,
+                    schema_id=schema.id,
+                    task_hash=task.p_item.hash,
+                    scheduler_id=self.scheduler_id,
+                )
+                schema.enabled = False
+
+            self.ctx.datastores.schema_store.update_schema(schema)
+
+        # TODO: explain why we're using executor here
```
jpbruinsslot, line 365: Calculation of deadlines could potentially take a long time. Since this is called from the server API, we want to have it be calculated in the background since we want to return a response without keeping the connection open.

mula/scheduler/alembic/versions/0008_restructure.py
```diff
@@ -0,0 +1,96 @@
...
+    op.drop_table("items")
+    op.add_column("tasks", sa.Column("schema_id", scheduler.utils.datastore.GUID(), nullable=True))
+    op.add_column("tasks", sa.Column("hash", sa.String(length=32), nullable=True))
+    op.add_column("tasks", sa.Column("priority", sa.Integer(), nullable=True))
+    op.add_column("tasks", sa.Column("data", postgresql.JSONB(astext_type=sa.Text()), nullable=False))
+    op.alter_column("tasks", "scheduler_id", existing_type=sa.VARCHAR(), nullable=False)
+    op.drop_index("ix_tasks_p_item_hash", table_name="tasks")
+    op.create_index(op.f("ix_tasks_hash"), "tasks", ["hash"], unique=False)
+    op.create_foreign_key(None, "tasks", "schemas", ["schema_id"], ["id"], ondelete="SET NULL")
+    op.drop_column("tasks", "p_item")
```
Donnype, line 47: Shouldn't we be more careful and actually migrate this data to the new data field? Or does that contain a new kind of data structure?

mula/scheduler/utils/cron.py
```diff
@@ -0,0 +1,8 @@
+from datetime import datetime, timezone
+
+from croniter import croniter
+
+
+def next_run(expression: str, start_time: datetime = datetime.now(timezone.utc)):
```
Donnype, line 6: Perhaps `calculate_from` or `after` or ... not sure

mula/scheduler/storage/task_store.py
```diff
@@ -73,7 +73,7 @@ def get_tasks_by_hash(self, task_hash: str) -> list[models.Task] | None:
         with self.dbconn.session.begin() as session:
             tasks_orm = (
                 session.query(models.TaskDB)
-                .filter(models.TaskDB.p_item["hash"].as_string() == task_hash)
+                .filter(models.TaskDB.hash == task_hash)
```
Donnype, line 76: 👍

mula/scheduler/models/queue.py
Donnype: So there are no issues here also when trying to query PrioritizedItems?

mula/scheduler/schedulers/normalizer.py
```diff
@@ -221,22 +221,21 @@ def push_task(self, normalizer: Plugin, raw_data: RawData, caller: str = "") ->
...
-        p_item = PrioritizedItem(
-            id=task.id,
+        # TODO: check the correct attributes, schema
+        task = Task(
             scheduler_id=self.scheduler_id,
             priority=score,
-            data=task.model_dump(),
-            hash=task.hash,
+            hash=normalizer_task.hash,
+            data=normalizer_task.model_dump(),
```
Donnype, line 233: Perhaps it makes sense to rename either `data` or `normalizer_task` so that it makes sense that `task.data == normalizer_task.model_dump()`? Either `task.arguments` and `normalizer_arguments` or `task.meta` and `normalizer_meta` or just `normalizer_task_data`?

mula/scheduler/rankers/deadline.py
```diff
@@ -0,0 +1,41 @@
...
+from datetime import datetime, timedelta, timezone
+from typing import Any
+
+from scheduler.utils import cron
+
+from .ranker import Ranker
+
+
+class DefaultDeadlineRanker(Ranker):
+    def rank(self, obj: Any) -> int:
```
Donnype, line 11: So semantically the rank will now be a number that represents a timestamp equalling the "deadline" right? I believe that's what we discussed also. Meaning we will can infer that a boefje should run now because the current timestamp is really close to or even greater than the deadline?

mula/tests/integration/test_ranker.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def test_calculate_deadline_cron(self):
+        schedule = models.Schedule(
+            scheduler_id="test",
+            p_item=models.PrioritizedItem(hash="test", priority=1),
+            cron_expression="0 12 * * 1",  # every Monday at noon
+        )
+
+        deadline = self.ranker.rank(schedule)
+        self.assertIsNotNone(deadline)
```
Donnype, line 45: I suppose we can also assert that it is at least greater than the current timestamp? (With or without mocking)

mula/tests/integration/test_scheduler.py
```diff
@@ -204,3 +208,73 @@ def test_enable_scheduler(self):
 
         # Stop the scheduler
         self.scheduler.stop()
+
+    # FIXME: deadline calculation is done in thread, so it's hard to test
+    def test_signal_handler_task(self):
```
Donnype, line 213: I remember us having some discussion about having deadlines and how/when to set them. But I remember an argument from my side where I mentioned that tasks can depend on a schedule, but if we update schedule deadlines through a signal when the tasks finish, we also create a dependency of the schedule on the running tasks, which sounded a bit tricky to me. Perhaps we can infer more at query time by looking at the most recent task?

mula/scheduler/models/queue.py
jpbruinsslot: There were some un pushed changes. It should now be removed

mula/scheduler/alembic/versions/0008_restructure.py
jpbruinsslot: Yeah it going to change to data, the prioritized wrapper is going to be removed. However, you make a good point the `data` in the `p_item` should be migrated to the new `data` field

mula/scheduler/utils/cron.py
jpbruinsslot: perhaps use the croniter naming? `get_next`?

mula/scheduler/schedulers/normalizer.py
jpbruinsslot: Yeah will do. Should've prefaced that the normalizer scheduler needed some work.

mula/scheduler/rankers/deadline.py
jpbruinsslot: Correct. However, I can see that this can be confusing. In the case of deadline calculation we're not ranking but we're scheduling a job in the future. When the deadline has passed the job will be evaluated, ranked and then be pushed on to the queue.

Would you also agree to refactor/restructure this outside of the rankers module?

mula/tests/integration/test_scheduler.py
jpbruinsslot: Yeah that would be an elegant solution, I'll look into it

mula/scheduler/alembic/versions/0008_restructure.py
jpbruinsslot: @Donnype mentioned he has done this work in bytes as well

mula/tests/integration/test_scheduler.py
Donnype: Yes so it boils down dropping the `deadline_at` field by calculating whether we've passed the deadline when we are going to reschedule jobs. I believe we could do that by calculating it the same way: using the `schedule` field and the `ended_at` of the last task

Donnype: Perhaps another argument:

If we schedule dns-scan to run on Mondays and nmap to run on Tuesdays, we might (?) see that nmap will already run on Monday due to it finding the ip adress. Then we run nmap anyways on Tuesdays using a `deadline_at`, because a lot can happend between now and the deadline that perhaps makes running the task at the deadline not necessary. The query-time approach of calculating whether we should run the task would handle this (if we make sure to match the schedule and tasks on task-hashes in that case?).

Donnype (commented):
Good stuff @jpbruinsslot! Do you have a todo list or timeline on when we can merge this and make the switch?

mula/Makefile
```diff
@@ -67,6 +67,10 @@ cov: ## Generate a test coverage report
 		--cov-report term-missing:skip-covered \
 		--cov=scheduler tests/
 
+shell: ## Start a shell in the scheduler container.
```
ammar92, line 70: The comment is a bit misleading, since this actually starts a *new* container with a Python shell as the entry point, and not exec into a running container (at least this is how I read it the first time)

mula/scheduler/context/context.py
```diff
@@ -99,7 +99,7 @@ def __init__(self) -> None:
                     structlog.dev.set_exc_info,
                     structlog.stdlib.PositionalArgumentsFormatter(),
                     structlog.processors.TimeStamper("iso", utc=False),
-                    structlog.dev.ConsoleRenderer(),
+                    structlog.dev.ConsoleRenderer(colors=False),
```
ammar92, line 102: Any particular reason why this is disabled? In the other modules, when the `text` log format is selected we do colorize the log records

mula/scheduler/schedulers/normalizer.py
```diff
@@ -318,7 +306,7 @@ def is_task_allowed_to_run(self, normalizer: Plugin) -> bool:
         Returns:
             True if the task is allowed to run, False otherwise.
         """
-        if not normalizer.enabled:
+        if normalizer.enabled is False:
```
ammar92, line 309: The previous implementation is the more Pythonic way of dealing with this. The current suggested implementation is generally unnecessary unless you need to differentiate between `False` and other falsy values (e.g., `None`, `0`, `''`, etc.).

mula/scheduler/server/serializers/schedule.py
```diff
@@ -0,0 +1,17 @@
...
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+class Schedule(BaseModel):
+    model_config = ConfigDict(from_attributes=True)
+
+    hash: str | None = Field(None, max_length=32)
+
+    data: dict | None = {}
```
ammar92, line 11: I think using a `Field` with a default factory for dict is safer. Also, if the field is not supposed to be nullable, but does have a default value, then the `None` type should be removed

mula/scheduler/server/serializers/task.py
```diff
@@ -37,8 +36,18 @@ class Task(BaseModel):
 
     scheduler_id: str | None = None
 
+    schedule_id: uuid.UUID | None = None
+
+    priority: int | None = 0
```
ammar92, line 41: The `None` doesn't make sense if it's not nullable (I inferred this from the `0` assignment)

mula/scheduler/utils/cron.py
ammar92: You can't assign `datetime.now(timezone.utc)` in the signature, otherwise it will always have the same value (based on the module import time). The Pythonic way is to assign a `None` as default, and assign a new datetime in the function body if the passed value is indeed `None`.

ammar92 (commented):
Looks good. Just a few remarks. We've already discussed the PR in a high level overview offline

rocky/tests/conftest.py
```diff
@@ -1692,19 +1684,50 @@ def onboarding_collect_data():
...
                 },
             ],
-            "security": {"spf": False, "dkim": False, "dmarc": False, "dnssec": False, "caa": False},
+            "security": {
+                "spf": False,
+                "dkim": False,
+                "dmarc": False,
+                "dnssec": False,
+                "caa": False,
+            },
```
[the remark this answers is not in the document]
jpbruinsslot: I'll tidy it up for ya :)

mula/scheduler/connectors/services/octopoes.py
```diff
@@ -88,7 +88,7 @@ def get_random_objects(self, organisation_id: str, n: int, scan_level: list[int]
     @exception_handler
     def get_object(self, organisation_id: str, reference: str) -> OOI:
         """Get an ooi from octopoes"""
-        url = f"{self.host}/{organisation_id}"
+        url = f"{self.host}/{organisation_id}/object"
```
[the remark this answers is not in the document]
jpbruinsslot: yeah wasn't used before

boefjes/boefjes/clients/bytes_client.py
```diff
@@ -85,7 +85,11 @@ def get_boefje_meta(self, boefje_meta_id: str) -> BoefjeMeta:
 
     @retry_with_login
     def save_normalizer_meta(self, normalizer_meta: NormalizerMeta) -> None:
-        response = self._session.post("/bytes/normalizer_meta", content=normalizer_meta.json(), headers=self.headers)
+        response = self._session.post(
```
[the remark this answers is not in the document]
jpbruinsslot: line length again, ill fix it

mula/scheduler/server/handlers/queues.py
```diff
@@ -108,22 +108,13 @@ def push(self, queue_id: str, item: serializers.PrioritizedItem) -> Any:
 
         try:
             # Load default values
-            p_item = models.PrioritizedItem()
+            default_item = models.Task()
+            patch_data = item_in.model_dump(exclude_unset=True)
+            new_item = default_item.model_copy(update=patch_data)
```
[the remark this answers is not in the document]
jpbruinsslot: I'll have a look

mula/scheduler/alembic/versions/0008_add_task_schedule.py
```diff
@@ -0,0 +1,182 @@
...
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("hash"),
+    )
+
+    op.drop_index("ix_items_hash", table_name="items")
+    op.drop_table("items")
```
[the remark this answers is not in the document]
jpbruinsslot: Yeah exactly

mula/scheduler/alembic/versions/0008_add_task_schedule.py
```diff
@@ -0,0 +1,182 @@
...
+import scheduler
+
+# revision identifiers, used by Alembic.
+revision = "0008"
+down_revision = "0007"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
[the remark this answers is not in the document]
jpbruinsslot: Yeah, tested it with a minimal set. But would be good to fully test it

mula/Makefile
jpbruinsslot: Yeah I'll remove it since, I'm not using it at the moment. And indeed the comment is a bit misleading

jpbruinsslot:
> Good stuff @jpbruinsslot! Do you have a todo list or timeline on when we can merge this and make the switch?

Thanks! And thank you for your suggestions. I wanted to get the reviews out first. For the next steps, I'll implement the suggestions, get QA to look at it and focus on the migration for a bit to make sure that works out fine. When merged I'd like to have it deployed in order to test to uncover edge cases, and how it will fare with more throughput/load.

mula/scheduler/server/handlers/queues.py
```diff
@@ -96,9 +96,9 @@ def pop(self, queue_id: str, filters: storage.filters.FilterRequest | None = Non
                 detail="could not pop item from queue, check your filters",
             )
 
-        return models.PrioritizedItem(**p_item.model_dump())
+        return models.Task(**p_item.model_dump())
 
-    def push(self, queue_id: str, item: serializers.PrioritizedItem) -> Any:
+    def push(self, queue_id: str, item_in: serializers.Task) -> Any:
```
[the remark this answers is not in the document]
jpbruinsslot: Was a bit torn on the naming. Since prioritized item is not there anymore there is still the notion of when a task is put on the queue it is considered an item within the `PriorityQueue` object. Meaning renaming every instance of item to task I'd considered to create a lot of noise in the PR. I agree that it would be more descriptive and less confusing to rename all instances of item to task. So I propose to pick up renaming these in a separate PR.

mula/scheduler/storage/schedule_store.py
```diff
@@ -0,0 +1,120 @@
...
+        enabled: bool | None = True,
+        min_deadline_at: datetime | None = None,
+        max_deadline_at: datetime | None = None,
+        min_created_at: datetime | None = None,
+        max_created_at: datetime | None = None,
+        offset: int = 0,
+        limit: int = 100,
+        filters: FilterRequest | None = None,
+    ) -> tuple[list[models.Schedule], int]:
+        with self.dbconn.session.begin() as session:
```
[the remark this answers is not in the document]
jpbruinsslot: Let's discuss this, and move it to a new issue

mula/scheduler/server/serializers/task.py
jpbruinsslot: This has to do with these models being added specifically to allow partial updates from the API patch endpoints. However, it is one of the issues that I want to tackle in the fastapi ticket we have for the scheduler. For now I'll set the assignment to `None`

mula/scheduler/schedulers/scheduler.py
```diff
@@ -235,129 +269,181 @@ def push_item_to_queue(self, p_item: models.PrioritizedItem) -> None:
...
 
-        return True
+        # Check if the adjusted time is earlier than the minimum, and
+        # ensure that the adjusted time is not earlier than the deadline
+        adjusted_time = deadline + jitter_offset
+        adjusted_time = max(adjusted_time, deadline)
 
-    def is_item_on_queue_by_hash(self, item_hash: str) -> bool:
-        return self.queue.is_item_on_queue_by_hash(item_hash)
+        return adjusted_time
```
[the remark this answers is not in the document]
jpbruinsslot: @Donnype I've updated it, would you mind checking?

mula/scheduler/server/handlers/queues.py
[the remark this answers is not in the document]
jpbruinsslot: Been able to make it a one-liner

stephanie0x00:
## Checklist for QA:

### What works:
General scheduling works on a 'make reset', tasks can be rescheduled, links in the UI to/from the task list seem to work. Findings are created. 

### What doesn't work:
When migrating (followed the steps as described in the QA notes), the task list history disappears and an error message is shown in the UI. The task statistics table does show that multiple tasks are present. 

### Bug or feature?:
See above.

jpbruinsslot:
Updated and amended qa instructions, for additional testing

underdarknl (approved):

PR #2760, opened 2024-04-02
Improve error handling by better exception aggregation

originalsouth:
### Changes
Improve error handling by better exception aggregation

### Issue link
#2721 

### Demo
image

---

### Code Checklist
- [x] All the commits in this PR are properly PGP-signed and verified.
- [x] This PR only contains functionality relevant to the issue; tickets have been created for newly discovered issues.
- [ ] I have written unit tests for the changes or fixes I made.
- [x] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [x] I have performed a self-review of my code and refactored it to the best of my abilities.

### Communication
- [x] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [x] I have made corresponding changes to the documentation, if necessary.
- [x] I have included comments in the code to elaborate on what is not self-evident from the code itself, including references to issues and discussions online, or implicit behavior of an interface.

---
## Checklist for code reviewers:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_code.md) into your comment.

---
## Checklist for QA:
Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_qa.md) into your comment.

PR #2795, opened 2024-04-08
Fix/2721 improve error handling by better exception aggregation v2

originalsouth:
See #2760 but now with a cooler branch name.


ammar92:
Can we please rename the branch instead of duplicating PRs? Duplicating PRs can lead to confusion and loss of context. Also, since we do squash merges for main, the branch name may not appear in the main tree anyways.

originalsouth:
> Can we please rename the branch instead of duplicating PRs? Duplicating PRs can lead to confusion and loss of context. Also, since we do squash merges for main, the branch name may not appear in the main tree anyways.

I agree, but https://stackoverflow.com/questions/20007578/renaming-a-branch-while-on-pull-request

ammar92:
> > Can we please rename the branch instead of duplicating PRs? Duplicating PRs can lead to confusion and loss of context. Also, since we do squash merges for main, the branch name may not appear in the main tree anyways.
> 
> I agree, but https://stackoverflow.com/questions/20007578/renaming-a-branch-while-on-pull-request

That's unfortunate, I've been going through the GitHub docs for renaming branches and indeed, if the renamed branch is the head branch of an open pull request, this pull request is closed.

If it's merely about the branch name, then I suggest to close this PR. Sure a nicer branch name would've been good from the start, but renaming it at this stage doesn't add anything anymore imho.

originalsouth:
@ammar92 @stephanie0x00 both this PR #2795 and #2760 contain the same patches. I really have no opinion on which one should be merged and which one should be closed. As far as I am aware we have no policy or guidelines on branch name naming and as mentioned after squashing there should be no difference. Both remain open for now, and I am happy with either one being merged or rejected.

PR #2802, opened 2024-04-08
Fix and improve running boefjes/normalizer by hand

dekkers:
### Changes
- This fixes run_normalizer that was broken because the import was added before adding the path.
- Add force to the log level configuration because else it doesn't get picked up (probably something has changed in how logging is configured)
- Set the log level to info instead of debug, because debug gives too much detail
- Log "Saved raw file" at info level instead of the "Done with boefje". Logging the raw file id makes it easy to fetch the raw file of a normalizer, also when not using the run_boefje tool.
- Log that there are no results if we don't have any results instead of "Done with boefje"
- Log "Running local boefje/normalizer plugin" at debug level because that doesn't give any useful information not already in the log.
- Rename `BOEFJE_DOCKER_NETWORK` to `BOEFJES_DOCKER_NETWORK` in docker compose to get rid of the warning about deprecated name.

---

---
## Checklist for QA:


ammar92 (approved):
Looks good to me. About the `force` parameter suggestion; I think it could also be caused by calls to `logging.basicConfig` in one of the imported modules?

underdarknl (approved):

Issue #2828, opened 2024-04-11
I want to duplicate/create boefjes with custom arguments


Donnype:
Also see the related epic: https://github.com/minvws/nl-kat-coordination/issues/2736.

## To Be Discussed:

- Can users with level 2 create boefjes that have level 4? If so, they should be disabled by default.
- What _role_ is suitable for the "copy-boefjes" functionality? Admins, redteamers etc.
- Should new boefjes be added to all orgs for the user or only for the current? Perhaps add a global "copy-from org X to org Y" for the newly created boefjes?
- Should we group "copied" boefjes by their base-image (nested)?
- What would it look like when we also have external boefjes/images that are "added" to the local install from other repositories?

### Notes from discussion:
- New boefjes always have to be approved by a superuser/admin, unless it is created by an admin
- New boefjes are available for all organizations but only enabled for the current organization.

### Feature request
If I enable `nmap` in the Katalogus, I get a fixed set of nmap arguments/flags I can pick (through e.g. `nmap-tcp`, `nmap-udp` and `nmap-ports`). But there are thousands of combinations to make here, so unless a thousand boefjes are added by default, I would like to be able to add my own flags and give that setup a name myself.

#### Inspiration from Airbyte for Adding an Image with some configs

![image](https://github.com/minvws/nl-kat-coordination/assets/46660228/74b15cb2-8ebf-4cb5-9c84-9f44c70c12ae)
![image](https://github.com/minvws/nl-kat-coordination/assets/46660228/7486d3e1-9a5b-4591-ae07-4ea1ba4f07cf)

### Technical Notes
Note that this functionality is in progress, so in the current branch I am working on I have an implementation for **only** nmap.

### Personal input from frontend perpective (@Donnype)
To throw out some ideas, I imagine as a first step having more information on the boefje detail page (for boefjes that support this) with respect to the fields `oci_image` and `oci_arguments`. For `nmap-tcp` these are currently set to the following:
```json
{
  "id": "nmap",
  "name": "Nmap TCP",
  "description": "Defaults to top 250 TCP ports. Includes service detection.",
  "consumes": [
    "IPAddressV4",
    "IPAddressV6"
  ],
  "environment_keys": [
    "TOP_PORTS"
  ],
  "scan_level": 2,
  "oci_image": "openkat/nmap",
  "oci_arguments": [
    "--open",
    "-T4",
    "-Pn",
    "-r",
    "-v10",
    "-sV",
    "-sS"
  ]
}
[...]

PR #2832, opened 2024-04-11
Build nmap OCI image that handles both nmap-tcp and nmap-udp using a new `oci_arguments` field

### LINKED ISSUE #2785: Implement nmap container that accepts custom nmap args (POC)
Epic: #2736

As per our discussion on 02/04/2024, we decided to take the nmap boefje as a first example of a boefje image that can take arguments. Fundamentally this means that with the new usage of the `oci_image` field, we will allow multiple boefje manifests (i.e. `boefje.json`) to point to the same OCI image tag, but provide a new field in this definition that specifies the specific arguments for nmap. (This still means that we need to have different boefje-ids for the different argument sets.)

The steps to reach this are:
- [x] Create a new optional POC field in the boefje meta with the oci_arguments
- [x] Create a new-boefje-style docker image (Dockerfile)
- [x] Read out the new `oci_arguments` field in the nmap run() function to pass
- [ ] Update the other nmap boefjes to only a boefje.json file with their custom oci_arguments (and a cover image)
 

### Discussion points from implementing

There is a correlation between the `oci_arguments` and the `environment_keys`. For instance, nmap-udp has a custom TOP_PORTS_UDP (although these would not clash if they had the same name..). Changing this can be tricky due to the settings table, but the alternative would be if-elsing on the boefje-id to get the right environment variable key. Then you end up not needing the `oci_arguments` field after all, because why not if-else on the command line arguments?...

But I feel like in general we'll get this issue in the place where we now use environment variables as a workaround for arguments. Hence my proposal is to introduce the following conventions:

**Environment variables** can be set throught the settings table/api from the katalogus and can be defined dynamically through it _per boefje image_. Multiple boefje variants (multiple boefje_id) using the same boefje image have to use the same set of environment variables. They should act as a pre-requisite or necessary configuration to make the boefje succeed, and not impact the output format or cardinality.

**OCI arguments** are static and cannot be defined dynamically. They are defined _per boefje variant_ (per boefje_id) and can in fact do all of the above. They act as the Dockerfile's `CMD` and may impact how the input_ooi is scanned and the output.

As a result, we do not need to worry that per boefje_id the environment could also dictate possible output formats. It also means that we cannot impact deletion propagation behavior dynamically, which is probably not what we want. For instance: if I set `TOP_PORTS` to 10 now dynamically, we will delete all open ports we previously found outside this top 10. This would require `TOP_PORTS` to become an `oci_argument`, which means it will not be configurable until we implement such a flow.

Donnype:
### Changes
This creates a new build command and implementation that merges the functionality from `nmap-tcp` with `nmap-udp` into one oci_image. It adds the required commands to the definition file of the boefje.

Other changes include:
- Rename `docker_adapter.py` to `oci_adapter.py` since using Docker is technically an implementation detail of how we build our oci images.
- Add a sane default for BOEFJE_PATH in the custom Dockerfiles, making the build command shorter and hence easier to read.
- Actually set the environment variables received from the API in the adapter.
- Do not require a `main.py` in the directory when looking for boefje resources, but throw an error when trying to run it in "local" mode.
 
### Issue link
Closes #2785

### Demo
:thinking:

### Notes on QA
This change is backward compatible, but we need to make sure `nmap-udp` and `nmap-tcp` still work after running either `make kat`, or `make -C boefjes images`.

---

### Code Checklist

### Communication

---

---
## Checklist for QA:


boefjes/boefjes/plugins/kat_nmap_tcp/main.py
```diff
@@ -1,42 +1,21 @@
...
 TOP_PORTS_DEFAULT = 250
-TOP_PORTS_MIN = 1
 
 
-def run_nmap(args: list[str]) -> str:
-    """Run Nmap in Docker."""
-    client = docker.from_env()
-    return client.containers.run(NMAP_IMAGE, args, remove=True).decode()
+def run(boefje_meta: dict):
+    top_ports_key = "TOP_PORTS"
```
underdarknl, line 9: Ideally we'd use just the same env-var, and specify the env vars per boefje in the katalogus instead of trying the boefje-name inside the boefje itself.

Donnype: Indeed, but also see the considerations in the issue with respect to backward compatibility and the fact that we do not have a UI to configure the arguments. We discussed last week that we can move forward with this code/approach and now start a more proper Design refinement flow as this ticket is supposed to kick off: #2828

boefjes/images/base.Dockerfile
```diff
@@ -11,8 +11,8 @@ COPY $BOEFJE_PATH/requirements.txt* .
 RUN --mount=type=cache,target=/root/.cache pip install --upgrade pip && pip install httpx
 RUN if test -f requirements.txt; then pip install -r requirements.txt; fi
 
-COPY ./images/docker_adapter.py .
+COPY ./images/oci_adapter.py .
 COPY $BOEFJE_PATH $BOEFJE_PATH
 
-ENTRYPOINT ["/usr/local/bin/python", "-m", "docker_adapter"]
+ENTRYPOINT ["/usr/local/bin/python", "-m", "oci_adapter"]
```
ammar92, line 17: Just an idea: Imho `docker_adapter` (or just `adapter` or `plugin_adapter`) as a module name is more suitable, since in the near future we could also have adapters from other technologies such as kata, gVisor and Firecracker

boefjes/boefjes/local.py
```diff
@@ -50,6 +50,9 @@ def run(self, boefje_meta: BoefjeMeta, environment: dict[str, str]) -> list[tupl
         boefjes = self.local_repository.resolve_boefjes()
         boefje_resource = boefjes[boefje_meta.boefje.id]
 
+        if not boefje_resource.module and boefje_resource.boefje.oci_image:
```
ammar92, line 53: Nullable fields should probably be checked explicitly using `is None`. For `oci_image` it may be different, if the intention is to also validate against an empty string.

ammar92 (commented):
Nice stuff 👍 Just a few questions/ remarks, but looks good otherwise. Mostly about the term *oci* being too much present imho

stephanie0x00:
## Checklist for QA:

### What works:
I can start the TCP scan and it'll finish for the hostname. It'll also create the expected finding. The UDP scan also works, but takes a long time to complete. The docker container is also cleaned up after it's done. 

### What doesn't work:
Nothing found during QA. 

### Bug or feature?:
n/a

boefjes/images/base.Dockerfile
Donnype: Yes, but it is not evident that we would need to use this same script, so for now this is a better name I would say.

PR #2878, opened 2024-04-23
Add some additional exception handling to the scheduler

jpbruinsslot:


ammar92 (approved):
Looks good as discussed 👍 

Just a few remarks/ thoughts:
- The except clause on `HTTPStatusError` could still be useful, because you probably cannot guarantee that this exception is caught and always expect `ExternalServiceHTTPStatusError` to be raised
  - You could also catch both (`except (HTTPStatusError, ExternalServiceHTTPStatusError) as exc`)
- Now that I read it again, I think `ExternalServiceResponseError` would be a more suitable and generic name in this context, and imho the `HTTPStatusError` postfix makes this a bit more coupled to the httpx API
- The `httpx.Response` object already contains a `httpx.Request` object accessible via the `request` property
  - I would suggest to remove the `request` parameter out of the `ExternalServiceResponseError` class, even if it's just in case for accidentally passing non-related request objects for the current response

ammar92 (approved):

underdarknl (approved):

PR #2900, opened 2024-05-01
Specify bit in the answer raw file instead of as mimetype

### LINKED ISSUE #2808: kat_answer_normalizer uses mimetypes as input
The kat_answer_normalizer uses the mime type as input: https://github.com/minvws/nl-kat-coordination/blob/5273eaa897206e18775f939d2f426189fc383e21/boefjes/boefjes/plugins/kat_answer_parser/normalize.py#L12

This won't work if the normalizer only gets the input OOI dict as proposed in #2564. This is the only normalizer that uses the mimetype as input for what to do.  I see the following possible solutions:

1. Add the mimetypes to the normalizer function signature
2. Change the answer raw files so that they include the bit instead of using the metadata of the raw file for this information.

My preference would be to do second. I think it would be better if we use the mimetype only to decide which normalizer to run and that raw files include all necessary information for the normalizer instead of using the raw file metadata for this.

Original ticket that implemented this also mentioned this but doesn't clearly state why it was implemented this way: https://github.com/minvws/nl-kat-coordination/issues/288

dekkers:
### Changes
This changes the answer that is generated by rocky and parsed by the normalizer to include the bit id the answer is for.

### Issue link
Closes #2808 

---

---
## Checklist for QA:


ammar92 (approved):

underdarknl (approved):

stephanie0x00:
## Checklist for QA:

### What works:
Tested by manually adding a 'Network' object. On the Objects page a Question object is created, the questions on this form have to be submitted in order to trigger the normalizer. Which results in yielded objects and the raw files that can be downloaded and contain the content. Looks good. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

PR #2904, opened 2024-05-03
Saving reports

### LINKED ISSUE #2669: Create report history page frontend
_This ticket is part of epic #2626_

## Description
The report flow is going to change a bit. The first step is to save the generated reports and show them in a history page. 
This ticket will only create the frontend of the history page. Saving the generated reports is part of issue #2720

On this page, the following sections are included:
* Generate report button
* Tabs
* Filtering and searching
* Table with all the reports (which is foldable an can show more details)
* Pagination
* Options to rerun, share, save and delete selected reports from the table

### Design
https://www.figma.com/design/GJCDuJYinANHYQ6Ystfzgn/%E2%9C%8D%F0%9F%8F%BC-KAT-|-For-Review-|-Reports-Page?node-id=1325-874

## To do
- [x] Save and show "Selected plugins" @dekkers 
- [x] Generate a report -> Refer to new URL with report ID @dekkers 
- [x] Fix schaling/styling of table 
- [x] Bug: "Selected objects()" is empty when opening a saved report @madelondohmen 
- [x] Change "Objects" column name to "Input Objects" @madelondohmen 
- [x] "Input Objects" column: One object = show object, multiple objects = show a number @Rieven 
- [x] "Report types" column: More than 2 rapport types -> add "+..." @Rieven 
- [x] Show maximum of 20 reports @Rieven 
- [x] Get maximum of 6 children for each report when loading the history page @Rieven 
- [x] Expanded row: report types tabel @Rieven 
- [x] Expanded row: subreports (< 6 reports = show, >5 reports = "View all subreports" button) @Rieven 
- [x] New subreports page template with all subreports @madelondohmen 
- [x] Get one report by parent_id to show on subreport page @madelondohmen 
- [x] Rename report_history_list.html to report_history.html @madelondohmen 
- [x] Write unit tests

### LINKED ISSUE #2720: Saving reports for history page
_This ticket is part of epic #2626_

## Description
All the generated reports need to be saved to XTDB, so they can be retrieved and showed in the history page.

### Design
Design of the history page:
https://www.figma.com/file/8r4OSFIJt1PF4iDXf9MbqM/%E2%9C%85-KAT-%7C-Approved-for-implementation?type=design&node-id=1183-3471&mode=design

## To do
- [x] Add reference date, object-count and name to Octopoes ReportData
- [x] Save all the reports to XTDB

noamblitz:
Closes #2720 
Closes #2669

This PR adds the following features:

- A Report OOI which has all the required information about a report, also features an attribute that points to the report data stored in bytes
- Some improvements to the XTDB connector enable it to query parent and child reports in one query
- A history table showing all previously created reports
- Some template tags that make the transition from Python objects to JSON easier (mostly temporary)

Currently, creating large reports can take some time because each report has to be individually saved. This can be improved with scheduled reporting and/or letting normalizers handle the creation of Report OOIs.


rocky/reports/views/generate_report.py
```diff
@@ -217,18 +222,44 @@ def generate_reports_for_oois(self) -> dict[str, dict[str, dict[str, Any]]]:
...
-            if report_type.name not in report_data:
-                report_data[report_type.name] = {}
-
             for ooi, data in results.items():
                 ooi_human_readable = Reference.from_str(ooi).human_readable
+                if report_type.id not in report_data:
+                    report_data[report_type.id] = {}
 
-                report_data[report_type.name][ooi] = {
+                report_data[report_type.id][ooi] = {
```
noamblitz, line 230: This is better, but it looks ugly in UI. Need to fix it.

rocky/tools/templatetags/ooi_extra.py
```diff
@@ -81,3 +81,8 @@ def index(indexable, i):
 @register.filter
 def pretty_json(obj: dict):
     return json.dumps(obj, default=str, indent=4)
+
+
+@register.filter
+def human_readable(reference_string: str):
+    return Reference.from_str(reference_string).human_readable
```
underdarknl, line 88: Is building a reference from str really needed (and not overly expensive) when we already have the referenced object in scope when rendering the form?

underdarknl: I love the idea of a helper function, but I'd rather have a that function just try and return the OOI classes method if thats possible instead of recreating the class from the ref.

noamblitz: Good point, its prefered that the creator of the report gives the whole object as context but sometimes its not as easy (for example when using the PK as key in a dict. Its nice to have this kind of helper in that case. Fortunately, creating a Reference from string is basically a fancy string split and shouldn't be too slow. But I agree with you, it should usually be avoided! 

image

octopoes/octopoes/xtdb/query.py
```diff
@@ -146,10 +147,10 @@ def from_path(cls, path: Path) -> "Query":
 
         return query
 
-    def pull(self, ooi_type: Ref) -> "Query":
+    def pull(self, ooi_type: Ref, *, fields: str = "[*]") -> "Query":
         """By default, we pull the target type. But when using find, count, etc., you have to pull explicitly."""
 
-        self._find_clauses.append(f"(pull {self._get_object_alias(ooi_type)} [*])")
+        self._find_clauses.append(f"(pull {self._get_object_alias(ooi_type)} {fields})")
```
underdarknl, line 153: Is there any hardening / cleaning of this input to avoid query injection?

rocky/rocky/views/mixins.py
```diff
@@ -314,18 +314,18 @@ def __getitem__(self, key: int | slice) -> list[HydratedReport | Report]:
         raise NotImplementedError("ReportList only supports slicing")
 
     def get_subreports(self, report_id: str) -> list[Report]:
```
noamblitz, line 316: Hi, this whole method is not needed as the get_reports already returned tuples with child and parent. It causes huge huge performance drawbacks doing these queries in Rocky instead of in Octopoes (where they should be done). Please see the return types of list_reports in Octopoes where you'll find that all reports are always returned together with their children.

dekkers:
I just pushed some changes:
- The save report view use a post now and redirect to the view that fetches and shows the report
- Saving child reports is an option for aggregate reports. I am not sure whether we should keep this as an option however, because I don't see how this is going to scale to large number of OOIs and report types.
- The `input_oois` on the report object is a list of strings that will contain the primary key of the input OOIs, so this is available for the aggregate report without child reports

Some changes that still need to be done:
- Report object should not be an OOI. It is not an object of interest that we scan. Generating scan profile for reports does not make sense. It should be just an object in XTDB similar as for example scan profiles aren't OOIs.
- `ReportList` should not do queries in its `__init__` method.
- The generate report / concatenated reports still needs some fixing up. But the question is also what the usecase for this is because the current implementation of concatenated reports also doesn't scale to 10.000.
- I think it should be possible to simply view/download a single report of a single OOI without having to save the report.
- Tests still need to be fixed

octopoes/octopoes/xtdb/query.py
dekkers: This shouldn't be a problem as long as we don't put any user controlled string in the fields. The current code only uses strings that are statically defined in the code.

dekkers (approved):

PR #2944, opened 2024-05-13
Fix pydantic serialization of enum on PATCH of tasks

jpbruinsslot:
### Changes

* Added the `use_enum_value=True` for the `Task` pydantic models to ensure the use of the value of enums instead of raw enum values
* Added a `Task` serializer for the server API to serialize correctly, especially when partial updates on `PATCH` endpoints are considered
* For sqlalchemy task update statement added the `task.model_dump(mode="json")` parameter to make sure that the correct values are passed in the `update()` method

### Issue

Closes https://github.com/minvws/nl-kat-coordination/issues/2945

### QA

See the initial issue (https://github.com/minvws/nl-kat-coordination/issues/2945) for the specific exception that was raised. This occurred with normal task creation, so to check if the exception is still raised, is to go through onboarding and create a scan task.


ammar92:
Any progress so far? As discussed offline; I think the `class TaskStatus(str, enum.Enum)` might be causing problems. A string enum should just inherit from the `Enum`class. Let me know if that helps

jpbruinsslot:
Thank you for the reminder! I'll make the implementations and report back to you!

jpbruinsslot:
Removing it results in: `psycopg2.errors.InvalidTextRepresentation) invalid input value for enum` trying to debug this specific error

dekkers:
There is also `StrEnum` that is added in Python 3.11. Because we still support Python 3.10, the `strenum` package needs to be added if we want to use this.

mula/scheduler/models/tasks.py
```diff
@@ -61,6 +61,12 @@ class Task(BaseModel):
 
     modified_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
 
+    @validator("status", pre=True)
+    def parse_status(cls, v):
+        if isinstance(v, str):
+            return TaskStatus(v)
+        return v
+
```
jpbruinsslot, line 69: This might interfere with the other solutions

Donnype:
@jpbruinsslot Was this one ready to be reviewed, or put back on draft perhaps?

jpbruinsslot:
> @jpbruinsslot Was this one ready to be reviewed, or put back on draft perhaps?

Think I didn't attach it to an issue. But should be on draft.

mula/scheduler/models/tasks.py
jpbruinsslot: It does. Needs to be removed

mula/scheduler/server/server.py
```diff
@@ -468,7 +466,21 @@ def patch_task(self, task_id: uuid.UUID, item: dict) -> Any:
...
+        patch_data = item.model_dump(exclude_unset=True)
+        if len(patch_data) == 0:
+            raise fastapi.HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail="no data to patch",
+            )
+
+        try:
+            updated_task = task_db.model_copy(update=patch_data)
+        except Exception as exc:
```
jpbruinsslot, line 478: Probably to broad and not a 400

jpbruinsslot:
Suggestions from preliminary review (@ammar92):
* implement extra models to reduce code duplication: https://fastapi.tiangolo.com/tutorial/extra-models/
* implement python 3.11 `StrEnum` for the enums

Can also be picked up in new tickets if we want to merge these changes

jpbruinsslot:
Unfortunately the extra models, afaics don't solve the problem of creating an extra model that allows for optional fields for partial updates by subclassing as is proposed in (https://fastapi.tiangolo.com/tutorial/extra-models/).

When using internal models I want to make sure that specific fields are required, e.g.:

```python
class TaskBase(BaseModel):
    model_config = ConfigDict(from_attributes=True, use_enum_values=True)

    id: uuid.UUID

    scheduler_id: str

    type: str

    p_item: PrioritizedItem

    status: TaskStatus

```

When specifying a model for a `PATCH` endpoint I want to ensure that fields aren't required, e.g:

```python
class Task(BaseModel):
    model_config = ConfigDict(from_attributes=True, use_enum_values=True)

    id: uuid.UUID | None = None

    scheduler_id: str | None = None

    type: str | None = None

    p_item: PrioritizedItem | None = None

    status: TaskStatus | None = None
```

jpbruinsslot:
> Suggestions from preliminary review (@ammar92):
> 
> * implement extra models to reduce code duplication: https://fastapi.tiangolo.com/tutorial/extra-models/
> * implement python 3.11 `StrEnum` for the enums
> 
> Can also be picked up in new tickets if we want to merge these changes

I would recommend to merge this pull request and pick up the python 3.11 `strenum` at a later moment. Implementing the "extra models" will not code duplication for internal models and serializer that are used for the rest api.

mula/scheduler/server/serializers/task.py
```diff
@@ -0,0 +1,44 @@
...
+    COMPLETED = "completed"
+
+    # Task has failed
+    FAILED = "failed"
+
+    # Task has been cancelled
+    CANCELLED = "cancelled"
+
+
+class Task(BaseModel):
```
Donnype, line 33: So these are mainly used as BaseModels that allow for patching and hence the default values? I also looked into this recently for the boefjes, where @ammar92 pointed to [the suggested naming convention](https://fastapi.tiangolo.com/tutorial/extra-models/) of, in this case `TaskIn`.

Donnype (approved):

mula/scheduler/server/serializers/task.py
jpbruinsslot: https://github.com/minvws/nl-kat-coordination/pull/2944#issuecomment-2185777061

stephanie0x00:
## Checklist for QA:

### What works:
Looks good. The serialization error is not shown anymore when finishing the onboarding flow. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

underdarknl (approved):

Issue #1392, opened 2023-07-13
Add domain information to IP findings if available


zcrt:
**Is your feature request related to a problem? Please describe.**
Currently the following flow is available in KAT:
1. Add a hostname
2. Use the DNS-boefje to find the corresponding IP
3. Use Shodan to get information on the IP

If Shodan finds information on this IP, one has to use the clearance level inheritance function to find out the domain that the IP came from. This information is not included in a report.

**Describe the solution you'd like**
When a finding is found on an IP for which a hostname is known, it would be nice to add the hostname to a report in KAT. This could be implemented for individual findings, or if that is difficult, a table with IP-Hostnames could be added to the report (optionally filtered on IPs/hostnames for which findings do exist in the report.)

**Additional context**
A person with only access to the exported `.pdf` would have difficulty with finding out the source of a finding on an (initially) unknown IP address.

underdarknl:
Hm, this is not easy to do in a non-specific way. For ip-addresses you would need to somehow specify that you'd want to include te dnsA/DnsAAAARecords and their respective hostnames. but that would also branch out to cnameRecords i fyou want to have a full view of where this IP is used. From where you would probably want to know what websites or other services are running on those ip's.
A better route would be to include a new report in what is being send out to those recipients, these extra reports could contain these specific queries, and answer those questions.

zcrt:
Manually writing extra queries for specific systems to create specific reports would not scale to many organisations with many systems. I would then opt to start with using the hostnames from dnsA/dnsAAArecords, and see from there how large the issue is that remains after (cnames, websites, services)

IshangoNL:
From an end-user perspective, acting on the report, this is a major issue. It is not possible to act on a finding, based on the IP address alone. Especially if it originates from a cloud servce that uses changing (or multiple) IP addresses. The goal of OpenKAT should be to fix issues, not just pinpointing them. And in order to fix an issue I have to know who I have to set to work. It is not user friendly if I have to look up each IP address of each finding in a separate report.

PR #2952, opened 2024-05-14
feat: :memo: add human readable hostnames to vulnerability report

zcrt:
### Changes

Allows hostnames to be included with the vulnerability report. To this end also adds an option to group input_oois in dictionary form with a human readable hostname string as value. This option can be used to extend the same functionality to other reports.

If no hostnames are available only the IP will be shown. 
If one hostname is available it will be shown in brackets behind the IP.
If more than one hostname is available, the shortest one will be shown and an indication will be added that there are more hostnames available for this IP.

### Issue link

Related: https://github.com/minvws/nl-kat-coordination/issues/1392

### Demo
image

---

### Code Checklist

### Communication

---

---

## Checklist for QA:


rocky/reports/report_types/definitions.py
```diff
@@ -107,6 +107,34 @@ def to_hostnames(self, input_oois: Iterable[str], valid_time: datetime) -> dict[
...
 
+    def to_human_readable_hostnames(
+        self,
+        input_oois: Iterable[str],
+        valid_time: datetime,
+        extra_indicator: str = ", ...",
+        open_indicator: str = "(",
+        close_indicator: str = ")",
+        hostnames_by_input_ooi: dict | None = None,
+    ) -> dict:
```
ammar92, line 118:
```suggestion
    ) -> dict[str, str]:
```

rocky/reports/report_types/definitions.py
```diff
@@ -107,6 +107,34 @@ def to_hostnames(self, input_oois: Iterable[str], valid_time: datetime) -> dict[
...
+        """Converts input_oois to human readable hostname strings.
+
+        Turns a list of either Hostname and IPAddress references into a string
+        of shortest related hostname and indication on more hostnames present,
+        grouped by input ooi.
+        """
+        if hostnames_by_input_ooi is None:
+            hostnames_by_input_ooi = self.to_hostnames(input_oois=input_oois, valid_time=valid_time)
```
ammar92, line 126: `to_hostnames` method accepts positional arguments, so there's no need to pass them as keyword arguments

```suggestion
            hostnames_by_input_ooi = self.to_hostnames(input_oois, valid_time)
```

ammar92 (dismissed):
Looks good to me 👍 Just a few very minor remarks to improve it even more

rocky/reports/report_types/definitions.py
zcrt: Yeah, I followed the "explicit is better than implicit" doctrine here. There may be a [special syntax](https://peps.python.org/pep-0736/) someday available for this. However, if you prefer the positional arguments I can remove the keywords

ammar92: I would still prefer the positional arguments at least for consistency. I would agree to use keyword arguments for more complex signatures, but this one is relatively simple and by using them as positional arguments, it would minimise verbosity and visual noise.

rocky/reports/report_types/definitions.py
```diff
@@ -107,6 +107,34 @@ def to_hostnames(self, input_oois: Iterable[str], valid_time: datetime) -> dict[
...
         return hostnames_by_input_ooi
 
+    def to_human_readable_hostnames(
+        self,
+        input_oois: Iterable[str],
+        valid_time: datetime,
+        extra_indicator: str = ", ...",
+        open_indicator: str = "(",
+        close_indicator: str = ")",
+        hostnames_by_input_ooi: dict | None = None,
```
dekkers, line 117: These optional arguments aren't used by any caller. It makes the code more complex and harder to understand than is necessary. In my opinion it is always good to follow the YAGNI principle and not write code that doesn't have a direct use but might be used in the future.

```suggestion
```

rocky/reports/report_types/definitions.py
```diff
@@ -107,6 +107,34 @@ def to_hostnames(self, input_oois: Iterable[str], valid_time: datetime) -> dict[
...
+    ) -> dict[str, str]:
+        """Converts input_oois to human readable hostname strings.
+
+        Turns a list of either Hostname and IPAddress references into a string
+        of shortest related hostname and indication on more hostnames present,
+        grouped by input ooi.
+        """
+        if hostnames_by_input_ooi is None:
+            hostnames_by_input_ooi = self.to_hostnames(input_oois, valid_time)
```
dekkers, line 126:
```suggestion
        hostnames_by_input_ooi = self.to_hostnames(input_oois, valid_time)
```

rocky/reports/report_types/definitions.py
```diff
@@ -107,6 +107,34 @@ def to_hostnames(self, input_oois: Iterable[str], valid_time: datetime) -> dict[
...
+        human_readable = {}
+        for input_ooi, hostnames in hostnames_by_input_ooi.items():
+            hostname_str = ""
+            if hostnames:
+                hostnames = [h.human_readable for h in hostnames]
+                add_str = extra_indicator if len(hostnames) > 1 else ""
+                hostname_str = open_indicator + min(hostnames, key=len) + add_str + close_indicator
+            human_readable[input_ooi] = hostname_str
```
dekkers, line 134: I think writing it like this makes it easier to understand. I didn't test the code.

```suggestion
            if hostnames:
                hostname = min([h.human_readable for h in hostnames], key=len)
                if len(hostnames) > 1:
                    human_readable[input_ooi] = f"({hostname}, ...)"
                else:
                    human_readable[input_ooi] = f"({hostname})"
            else:
                human_readable[input_ooi] = ""
```

rocky/reports/report_types/definitions.py
ammar92: While I agree with the principles of YAGNI, I believe this is intended to be used as a helper/ utility function. I've seen codebases that provide such utilities, but don't actually use them theirselves or only reference them in their tests. For me it's a gray area so either way is fine, but leaving it out is probably more preferable. I'll leave it up to you.

Imho I'm more concerned about the `hostnames_by_input_ooi` being an optional argument, as it results in a database query as a side effect when the argument is not provided. It would be better to instantiate this variable before passing it to the function, following the principle of dependency injection. This approach would make the function’s behavior more predictable, easier to understand and test.

zcrt: > I believe this is intended to be used as a helper/ utility function.

This is indeed the case here.

> It would be better to instantiate this variable before passing it to the function, following the principle of dependency injection.

I considered this, It would result in extra function calls outside the helper and thus a less easy to use helper. Logically this functionality felt at place next to `to_hostnames` and `to_ips`. However, if the `to_hostnames` is already called close to this proposed function it would result in unnecessary database queries, which I why I ended up with this hybrid solution. Without the call to `to_hostnames` it would become a static method of the `Report` class. Is that the intended solution, or would it have to move to `rocky/reports/report_types/helpers.py` then? (I tried that first, but it felt illogical in practice.) I can move it around either way, let me know what you prefer

ammar92: > I considered this, It would result in extra function calls outside the helper and thus a less easy to use helper. Logically this functionality felt at place next to `to_hostnames` and `to_ips`. However, if the `to_hostnames` is already called close to this proposed function it would result in unnecessary database queries, which I why I ended up with this hybrid solution. Without the call to `to_hostnames` it would become a static method of the `Report` class. Is that the intended solution, or would it have to move to `rocky/reports/report_types/helpers.py` then? (I tried that first, but it felt illogical in practice.) I can move it around either way, let me know what you prefer

Thanks for your reaction. I think it's preferable if we move it to `helpers` module. (Imho the `to_hostnames` and `to_ips` should've also been helper functions outside the `Report` class, since otherwise the class tends to become a god class violating the single responsibility principle; but that's out of scope/ can be refactored soon).

rocky/reports/report_types/definitions.py
ammar92: > I actually prefer a single assignment to the `human_readable` dictionary, as that keeps the code more maintainable in my experience. However, line 131 is indeed redundant as the length does not need to be determined from the human readable strings.
> 
> Or maybe the `add_str` and `hostname_str` could be combined in a single f-string expression

Your suggestion looks fine. I'm not sure if an f-string would make it more readable in this case; but I'll leave that up to you.

dekkers: > I actually prefer a single assignment to the `human_readable` dictionary, as that keeps the code more maintainable in my experience. However, line 131 is indeed redundant as the length does not need to be determined from the human readable strings.

I find the way you wrote the method is a lot harder to understand than is necessary and thus harder to maintain. They way I changed it is very clear that there are three different cases and what the string in those three different cases would be. With your code there are 6 more variables you need to keep track of in your head to understand what this does.

rocky/reports/report_types/definitions.py
zcrt: It looks like it is not easily possible to move it to the `helpers` module, as that will result in circular imports. I made it a static method for now, it can then be refactored together with the other methods of the `Report` class.

rocky/reports/report_types/definitions.py
zcrt: I have now rewritten the whole method

zcrt:
I rewrote some methods, I noticed that `to_hostnames` will not return `Hostname`s that have the same resolved address for a specific hostname, is that the intended behavior?

(For example if you have a.nl -> A -> 1.2.3.4 and b.nl -> A -> 1.2.3.4, and ask to_hostnames with `a.nl`, it will return something like  `{'a.nl': [ResolvedHostname("a.nl")]` and not `b.nl` in that list as well, so `b.nl` does not end up in the vulnerability report.)

rocky/reports/report_types/vulnerability_report/report.html
```diff
@@ -4,7 +4,7 @@
     {% for ip, vulnerability_data in data.items %}
         {% if vulnerability_data.summary.total_findings > 0 %}
             {% if show_heading %}
-                <h3>{% translate "Host:" %} {{ ip.human_readable }}</h3>
+                <h3>{% translate "Host:" %} {{ ip.human_readable }} {{ vulnerability_data.hostnames }}</h3>
```
underdarknl, line 7: should we limit these to a maximum of maybe three?

ammar92 (approved):
Looks good to me now; I don't have any more remarks

stephanie0x00:
## Checklist for QA:

### What works:
Seems to work nicely. It now shows the IP address and/or the hostname, if applicable. 

### What doesn't work:
- The docs on [docs.openkat.nl](https://docs.openkat.nl/manual/reports.html#reports) should be updated, as the vulnerability report now isn't only attached to the Hostname, but also to the IP address. 
- The fix is only applicable to the normal report, not the aggregate report. Is that intended? See below. 

### Bug or feature?:
The aggregate report only shows the IP address, as shown in the screenshot below. It doesn't matter if the IP address and hostname were selected or only the Hostname. In both cases only the IP address is shown. 

image

zcrt:
> The fix is only applicable to the normal report, not the aggregate report. Is that intended? See below.

This is not intended, during testing it has been available in the aggregate report, if I remember correctly

zcrt:
> This is not intended, during testing it has been available in the aggregate report, if I remember correctly

Just tested the branch again, and it seems to work for aggregate report. For the 'normal' vulnerability report I only get domain names, but that might be intended? (This PR add hostnames to IPs, not the other way around)

stephanie0x00:
> > This is not intended, during testing it has been available in the aggregate report, if I remember correctly
> 
> Just tested the branch again, and it seems to work for aggregate report. For the 'normal' vulnerability report I only get domain names, but that might be intended? (This PR add hostnames to IPs, not the other way around)

I'm slightly confused as I now observe this behaviour the other way around (while my initial screenshot shows it was definitely the aggregate report that showed this behaviour). The aggregate report now shows the IP address with hostname in brackets (as expected), but the normal report doesn't. Instead it shows the IP address and hostname separate (screenshots below). I've added my testing steps below for reproducing, just in case. 

My guess is that this is related to the asset overview? In the aggregate report asset overview there is a link between the IP and Hostname, while the asset overview for the normal report shows this as separate hosts? See the screenshots below. Either case, this feature is an improvement thus we can already merge it, even if we have different view in the normal and aggregate report (though it's nicer if they match). 

Current observation for the normal report:

image

Asset overview in aggregate report:

image

Asset overview in normal report: 

image


> 
> > The docs on [docs.openkat.nl](https://docs.openkat.nl/manual/reports.html#reports) should be updated, as the vulnerability report now isn't only attached to the Hostname, but also to the IP address.
> 
> There has been no change as far as I can see. The aggregate vulnerability report always reported IPs.

I'll double check this with the team. 

### Testing steps

I've added mispo.es as object with L2 clearance, walked through the onboarding, and enabled the boefjes nmap TCP and webpage analysis. After completing these steps I've taken the following steps to create the both reports: 

- Reports > Change the default filter to include L1 through L4 and both Declared and Inherited objects.
- Click the 'Select all objects' button. 
- Select the vulnerability report. 
- Continue the wizard
- Repeat these steps for the aggregate report.

stephanie0x00:
As discussed with @madelondohmen we will push this change for the aggregate report and create a new ticket (#3110) for this fix in the normal report as it might take a bit more time.

underdarknl (approved):

PR #2975, opened 2024-05-21
Enable ruff format skip-magic-trailing-comma

dekkers:
### Changes

This enables ruff format [skip-magic-trailing-comma](https://docs.astral.sh/ruff/settings/#format_skip-magic-trailing-comma). This prevents formatting only changes in PRs when an argument is added and later removed.

The downside is that in some cases it might be more readable to have each argument on a single line, but in my opinion it more important to have consistent formatting and prevent formatting only changes in PRs.

---

---

## Checklist for QA:


Donnype (dismissed):
At first I thought this PR did the exact opposite, but now I realize I am 200% for this change. I want to read the code naturally and see a lot of context on my screen, and not
keep
having
to
scroll
down
after
every
three
function
calls.

dekkers:
We currently don't have a lot of big PRs open, so merging this now shouldn't generate a lot of merge conflicts.

I have looked into alternatives, but enabling this is the only way to make sure that there is one way to format the same code instead of two. So while in some cases the code might be a little bit more readable when not collapsed into a single line, I think having only one way to format code and less noise in PRs is more important.

underdarknl:
I agree, and I was waiting for a moment like this to pull this PR back to the front. Lets see if this is the moment.

Donnype (approved):

Donnype (approved):

underdarknl (approved):

PR #3001, opened 2024-05-28
Implement SonarCloud integrations

ammar92:
### Changes

(draft) First steps to Sonar Cloud integration

### Issue link

Closes ...

### Demo

---

### Code Checklist

### Communication

---

---

## Checklist for QA:


.github/workflows/sonar-cloud.yml
```diff
@@ -0,0 +1,27 @@
+name: SonarCloud
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
+    tags: [v*]
```
dekkers, line 7: I can't find anything about support for tags in the sonarcloud documenation. It does support branch analysis (https://docs.sonarsource.com/sonarcloud/enriching/branch-analysis/) so I think it makes more sense to run this on the `release-*` branches.

.github/workflows/sonar-cloud.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Download artifacts
+        uses: actions/download-artifact@v4
+
+      - name: SonarCloud
+        if: always()
```
dekkers, line 24: Why the if? It doesn't seem like this does anything and this isn't the sonarcloud documentation.

.pre-commit-config.yaml
```diff
@@ -79,9 +79,16 @@ repos:
         args: [--target-version, "5.0"]
 
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.9.0
+    rev: v1.10.0
     hooks:
       - id: mypy
+        args:
+          [
+            "--ignore-missing-imports",
```
dekkers, line 87: Mypy ran successfully before without this ignore, so why do we need this?

Also mypy is configured in pyproject.toml, I think we should keep the configuration there and not configure mypy in two different places.

.pre-commit-config.yaml
```diff
@@ -79,9 +79,16 @@ repos:
...
+            "--scripts-are-modules",
+            "--linecoverage-report",
+            "mypy/",
```
dekkers, line 90: This will create a `mypy` directory in everyone's git repository. I think it would be better to call it `.mypy-report` to hide it by default and it should be added to .gitignore so it doesn't get committed.

.github/workflows/sonar-cloud.yml
```diff
@@ -0,0 +1,93 @@
+name: SonarCloud
+
+on:
+  workflow_dispatch:
+  # workflow_run:
+  #   workflows: ["Pre-commit checks"]
+  #   # branches: [main]
+  #   types: [completed]
```
dekkers, line 8: This looks like leftover things that should be removed?

.github/workflows/sonar-cloud.yml
```diff
@@ -0,0 +1,93 @@
...
+
+  # push:
```
dekkers, line 10: Shouldn't we also run this on the main branch? Then we need to run it for pushes to main branch like we do it in other workflows.

.github/workflows/sonar-cloud.yml
```diff
@@ -0,0 +1,93 @@
...
+  octopoes-tests:
+    uses: minvws/nl-kat-coordination/.github/workflows/octopoes-tests.yml@feature/sonar-cloud
+  bytes-tests:
+    uses: minvws/nl-kat-coordination/.github/workflows/bytes-tests.yml@feature/sonar-cloud
+  mula-tests:
+    uses: minvws/nl-kat-coordination/.github/workflows/mula-tests.yml@feature/sonar-cloud
+  rocky-tests:
+    uses: minvws/nl-kat-coordination/.github/workflows/rocky-tests.yml@feature/sonar-cloud
+
+  # todo: boefjes
```
dekkers, line 23: Boefjes are still missing?

dekkers (commented):
This adds extra workflows to run the tests which means we run the same tests multiple times. This is not wrong, but also not optimal and might mean the CI takes longer, so we should probably look into improving that in the future.

I actually think that we should run with coverage by default, but only in the CI. In rocky we show coverage by default at the moment, but in my experience when working on tests the coverage information always hides the errors that you do want to see. But this is also not very important and we can change this later.

I also think this PR shows one more reason why it would be better to have one python namespace for all OpenKAT code instead of it being split over multiple namespaces like we have now. This is also not something we have to chane now, but we should look into this later.

dekkers (approved):

Issue #2645, opened 2024-03-13
Task list view for super users for all organisations


zcrt:
**Is your feature request related to a problem? Please describe.**
Debugging tasks in a running OpenKAT installation with multiple organisations can be hard. It would be easier with a single view where all tasks within OpenKAT are listed. Such a view could for example be used to more quickly determine what tasks are erroring, if a specific boefje/normalizer/bit always fails for organisations or if a specific task uses all workers accross organisations.

**Describe the solution you'd like**
When no organization is selected, and the user is a super-user, a menu item "Tasks" showing the tasks page as for a single organisation, but with data for all organizations, including a column naming the organisation associated with that task.

underdarknl:
Optionally, we could make this function also work for regular users who have multiple organizations. If so, we'd need to send the organizations that user is part of to the scheduler's api  as the scheduler has no notion of users or their rights.

jpbruinsslot:
From the scheduler point of the view, the following can be posted to the `/tasks` endpoint to get all the tasks for specific schedulers:

```javascript
POST /tasks

{
  "filters": [
    {
      "column": "scheduler_id",
      "operator": "in",
      "value": ["boefje-org-1", "boefje-org-2"]
    }
  ]
}
```

PR #3005, opened 2024-05-29
feat: :rocket: generic task view

zcrt:
### Changes

This PR introduces a task view for all organizations simultaneously. To this end:
- Add new taskviews
- Enhances scheduler endpoints
- Updates templates to use organisation information from tasks
- Extends the main menu on the generic page

What needs to be done is to make sure the new views are only accessible for staff users. I tried this with the commented out django code, but the user object seems to be missing. I left some loggingdebugstatements that need to be removed after that is fixed.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/2645

### Demo

image

image

image

---

### Code Checklist

### Communication

---

---

## Checklist for QA:


underdarknl:
nice work!
This needs checks for user-rights, so the user can only see 'all' tasks if he has those rights, Or we should filter the task-lists based on the list of organisations the user has access to if we want to offer this to everyone.

zcrt:
Indeed, how would I go about checking for user-rights in OpenKAT?

rocky/rocky/views/tasks.py
```diff
@@ -115,3 +118,88 @@ def get_context_data(self, **kwargs):
...
+
+class AllTaskListView(ListView):
+    paginate_by = 100
+    paginator_class = RockyPaginator
+
+    # @staff_member_required
+    def get_queryset(self):
+        task_type = self.request.GET.get("type", self.plugin_type)
+        status = self.request.GET.get("scan_history_status") if self.request.GET.get("scan_history_status") else None
+        input_ooi = self.request.GET.get("scan_history_search") if self.request.GET.get("scan_history_search") else None
```
ammar92, line 131:
```suggestion
        status = self.request.GET.get("scan_history_status", None)
        input_ooi = self.request.GET.get("scan_history_search", None)
```
or if you're doing a string emptiness check and a None check at the same time:
```suggestion
        status = self.request.GET.get("scan_history_status", "") or None
        input_ooi = self.request.GET.get("scan_history_search", "") or None
```

rocky/rocky/views/tasks.py
```diff
@@ -115,3 +118,88 @@ def get_context_data(self, **kwargs):
...
+
+        if self.request.GET.get("scan_history_from"):
```
ammar92, line 133:
```suggestion
        if "scan_history_from" in self.request.GET:
```

rocky/rocky/views/tasks.py
```diff
@@ -115,3 +118,88 @@ def get_context_data(self, **kwargs):
...
+            min_created_at = datetime.strptime(self.request.GET.get("scan_history_from"), "%Y-%m-%d")
+        else:
+            min_created_at = None
+        if self.request.GET.get("scan_history_to"):
```
ammar92, line 137:
```suggestion
        if "scan_history_to" in self.request.GET:
```

ammar92 (commented):
Nice feature and work overall. I do have a few remarks to improve it even more.

Regarding user rights, you might be able to combine the `OrganizationMember` and the current user to create a list of accessible schedulers

underdarknl:
nice work!
This needs checks for user-rights, so the user can only see 'all' tasks if he has those rights, Or we should filter the task-lists based on the list of organisations the user has access to if we want to offer this to everyone.

> Nice feature and work overall. I do have a few remarks to improve it even more.
> 
> Regarding user rights, you might be able to combine the `OrganizationMember` and the current user to create a list of accessible schedulers

That could work on the rocky end, but the scheduler api does not allow for filtering on many organizations I think?

ammar92:
> That could work on the rocky end, but the scheduler api does not allow for filtering on many organizations I think?

If the scheduler API doesn't support this yet, you'd still have to retrieve the user's organizations to know which task lists the user can see or retrieve

jpbruinsslot:
From the scheduler side I'd recommend to retrieve the task for all organizations from the tasks endpoint, this makes sure that filtering is done in db-time instead of python-time. Filtering on multiple scheduler ids and retrieving tasks is possible from this endpoint.

rocky/rocky/views/tasks.py
zcrt: Isn't there a difference between these checks? I think the first one also checks the value of the key, while your suggestion only checks existence of the key. In any case; I copied it from your `TaskListView` ;)

rocky/rocky/views/tasks.py
zcrt: See https://github.com/minvws/nl-kat-coordination/pull/3005#discussion_r1652841568

zcrt:
I updated the PR to retrieve the user accessible schedulers and show all related tasks and stats

rocky/rocky/scheduler.py
```diff
@@ -196,30 +198,43 @@ def list_tasks(
         **kwargs,
     ) -> PaginatedTasksResponse:
         kwargs = {k: v for k, v in kwargs.items() if v is not None}  # filter Nones from kwargs
-        res = self._client.get("/tasks", params=kwargs)
+        endpoint = "/tasks"
+        logger.debug("Making request to %s with args %s", endpoint, kwargs)
```
ammar92, line 202: The debug statement seems unnecessary since requests are already logged by the HTTP client. If needed, we can make this logging configurable. This would enable consistent and fine-tuned logging to meet the user's needs

rocky/rocky/scheduler.py
```diff
@@ -196,30 +198,43 @@ def list_tasks(
...
+        # Endpoint expects a str.
+        res = self._client.post(endpoint, params=kwargs, data=json.dumps(kwargs.get("filters", None)))  # type: ignore
```
ammar92, line 204: `filters` seems to be already passed in `params`, but it's also sent within the request body too?

ammar92: `data` should be a mapping, not a `str` or `bytes` input

For binary request data, such as JSON, use `content` instead of `data` (which is intended for form-encoded data) and set the correct content type

rocky/rocky/scheduler.py
```diff
@@ -196,30 +198,43 @@ def list_tasks(
...
+        if res.status_code != httpx.codes.OK:
+            logger.warning(
+                "Unexpected response %s from endpoint %s with args %s which returns code %s.",
+                res.content,
+                endpoint,
+                kwargs,
+                res.status_code,
+            )
```
ammar92, line 212: You might be better of using `res.raise_for_status()`, and if necessary, catch it and produce a nice user error or throw a `SchedulerError`, or a specific scheduler error (e.g. implement `class TaskFilterError(SchedulerError): ...`)

The logged warning will not be seen by the user, while the error could've definitely be a user (input) error

ammar92 (commented):
Thanks for the rework and updates. Just a few tiny remarks on the scheduler client. Should be ready to be functionally tested (can you take a look at this, @stephanie0x00?)

mula/scheduler/server/server.py
```diff
@@ -105,6 +105,15 @@ def __init__(
...
 
+        self.api.add_api_route(
+            path="/schedulers/by_type/{item_type}",
+            endpoint=self.get_schedulers_by_type,
+            methods=["GET"],
+            response_model=list[models.Scheduler],
+            status_code=status.HTTP_200_OK,
+            description="List all schedulers of a specific type.",
+        )
+
```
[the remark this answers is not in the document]
jpbruinsslot: I see that you've opted for using filtering on the `/tasks` endpoint, in that case I would recommend removing the changes made here in the scheduler.

mula/scheduler/server/server.py
```diff
@@ -236,6 +245,9 @@ def metrics(self) -> Any:
     def get_schedulers(self) -> Any:
         return [models.Scheduler(**s.dict()) for s in self.schedulers.values()]
 
+    def get_schedulers_by_type(self, item_type: str) -> Any:
+        return [models.Scheduler(**s.dict()) for s in self.schedulers.values() if s.queue.item_type.type == item_type]
+
```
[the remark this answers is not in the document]
jpbruinsslot: I see that you've opted for using filtering on the `/tasks` endpoint, in that case I would recommend removing the changes made here in the scheduler.

rocky/rocky/scheduler.py
```diff
@@ -262,14 +277,41 @@ def health(self) -> ServiceHealth:
...
+
+    def get_all_task_stats(self, task_type: str, scheduler_ids: list | None = None) -> dict:
+        """Return all task stats for specific task type (e.g. 'boefje')."""
+        return SchedulerClient._merge_stat_dicts(
+            dicts=[
+                self._get_task_stats(scheduler_id=scheduler.get("id"))
+                for scheduler in self._get(f"/schedulers/by_type/{task_type}")
+                if (scheduler_ids is None or scheduler["id"] in scheduler_ids)
+            ]
+        )
```
jpbruinsslot, line 314: This would be a nice addition to implement this in the scheduler to be able to get stats for multiple scheduler id's. That way we could off-load rocky doing the filtering, by letting the scheduler filter this in db-time.

rocky/rocky/scheduler.py
zcrt: Ah, I did not see the HTTP client logging during debugging, how do I make this visible within the KAT environment?

zcrt: Found it

rocky/rocky/scheduler.py
underdarknl: I agree filtering these jobs in Rocky should not be done, they should be filtered in The scheduler.
the complexity is that we have no easy API on the scheduler to filter on tasks for a list of schedulers, correct?

rocky/rocky/urls.py
```diff
@@ -72,6 +78,9 @@
         PrivacyStatementView.as_view(),
         name="privacy_statement",
     ),
+    path("tasks/", AllBoefjesTaskListView.as_view(), name="all_task_list"),
```
underdarknl, line 81: Drop the trailing / for consistency with other routes

mula/scheduler/server/server.py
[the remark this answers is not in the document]
zcrt: I removed the type filtering

mula/scheduler/server/server.py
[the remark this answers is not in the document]
zcrt: I removed the type filtering

rocky/rocky/scheduler.py
zcrt: I simplified this function. I would say that extending the scheduler to create stats for multiple schedulers is a nice feature request but out of scope for a MVP (unless it turns out later that python in rocky is too slow to merge a couple of dictionaries)

rocky/rocky/urls.py
zcrt: Consistency with which routes? There does not seem to be consistency.

underdarknl: I mistakenly assumed we did not have routes with trailing /'es.

rocky/rocky/views/tasks.py
```diff
@@ -115,3 +118,88 @@ def get_context_data(self, **kwargs):
...
+            return []
+
+    def post(self, request, *args, **kwargs):
+        self.handle_page_action(request.POST["action"])
+        return redirect(request.path)
+
+    def handle_page_action(self, action: str) -> None:
+        if action == PageActions.RESCHEDULE_TASK.value:
+            task_id = self.request.POST.get("task_id")
+            organization_code = self.request.POST.get("organization_code")
```
underdarknl, line 165: We need to validate that the user can access this organization. There is some code further down the line that checks if the given organization ID matches the Job's organization ID, but checking for access rights needs to be done here.

underdarknl: Another route could be to just link to the organization's taskview page action, which because it loads with the organization in the url, will aready do this check for us. That would mean we just need to point the rerun form to that url in the same way the organiaton specific view does it. In that case this whole post handler and handle_page_action ca be removed.

zcrt: You mean something like
```python
organization_code in [o.code for o in self.request.user.organizations]
```
?

underdarknl:
merged as another PR after fixing the conflicts.

Issue #2997, opened 2024-05-28
Introduce the `Boefje` and `Normalizer` models in the KATalogus


Donnype:
Part of this epic: https://github.com/minvws/nl-kat-coordination/issues/2881

### Todo
- [x] Introduce the database models and migrations
- [x] Add plugin repository for boefjes and normalizers
- [x] Add router endpoints: post, get, patch, delete
- [x] Merge plugins with the local plugins in the PluginsService 
- [x] Expose to the API which plugins can be "deleted", since the local directories cannot
- [x] Handle id conflicts: override a set of fields that make sense to edit

PR #3011, opened 2024-05-30
Add Boefje and Normalizer models in the KATalogus

Donnype:
### Changes

Also see https://github.com/minvws/nl-kat-coordination/issues/2997 and the epic https://github.com/orgs/minvws/projects/7/views/33?pane=issue&itemId=60723200.

This adds the models, APIs and logic behind the following part of the new database schema:

```mermaid
classDiagram

class boefje {
    id
    plugin_id
    scan_level
    consumes
    produces
    version
    created
    description
    name
    environment_keys
    oci_image
    oci_arguments
}

class normalizer {
    id
    plugin_id
    scan_level
    consumes
    produces
    version
    created
    description
    name
    environment_keys
}
```

Some particular choices I faced where I decided myself how this should probably work, but are up for discussion:
- For local plugins, we can also add entries with the same id to overwrite these fields:
  - `name`
  - `description`
  - `scan_level`
  - `oci_image`
  - `oci_arguments`
- I removed unused fields such as `options` and `authors` for the time being.
- We now expose a `static` field to be able to differentiate between plugins that can be hard-deleted and the local plugins. For now it doesn't make sense to be able to be able to delete these local plugins, perhaps never. This field is not saved on a database level though since it can simply be configured when fetching the data. The editable-fields logic above makes sure that also for local plugins we overwrite this works well.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/2997

### Demo

Although next PRs will showcase more functionality and the power in the frontend, one funny thing you could to to see that this works is go to the [katalogus api ](http://localhost:8003/docs#/plugins/add_plugin_v1_organisations__organisation_id__plugins_post) and add a random plugin using the post endpoint:

image

To suddenly find a new entry in your KATalogus:
image

This entry is useless unless you configure a real `oci_image` pointing to a boefje image though.

---

### Code Checklist

### Communication

---

---

## Checklist for QA:


underdarknl:
Would it make sense to also include a path to a normalizer image? zip file / oci image or otherwise? And would it make sense to track the author? (eg, if this plugin was created by a local user, or (if null) comes from upstream?

boefjes/boefjes/katalogus/tests/integration/test_api.py
ammar92, line 1: I noticed that your `test_add_*` tests might be testing more than necessary for a single unit test. Because not only an add operation is tested, but also a list, filter and delete operation in the same unit. Ideally, unit tests should focus on one specific functionality to keep them simple and maintainable. In this case a simple idempotency tests would suffice (just one add, and one get)

ammar92 (commented):
Nice work. I do have a few remarks to improve it even more. Let me know what you think

Donnype:
> Would it make sense to also include a path to a normalizer image? zip file / oci image or otherwise? And would it make sense to track the author? (eg, if this plugin was created by a local user, or (if null) comes from upstream?
> 
> I'd also guess that one of the things a user Might want to overwrite (without changing the actual oci_image) is the jsonschema that contains the settings their variant of the plugin consumes/reqiures

Eventually, yes custom images make sense probably. Authors could make sense, although perhaps the first iterations could live without it, especially if authors could be inferred from the oci_image somehow. I also thought about the schema, since actually now there is not support for it for these kinds of boefjes. Perhaps we need to have a json schema field as well. Could be implemented once we tie the frontend and backend together and (probably) find out we need this, and then consider if we want to change the API with respect to the schema's?

boefjes/boefjes/katalogus/models.py
```diff
@@ -46,8 +44,8 @@ class Normalizer(Plugin):
 class Bit(Plugin):
     type: Literal["bit"] = "bit"
     consumes: str
-    produces: list[str]
-    parameters: list[str]  # ooi.relation-name
+    produces: list[str] = Field(default_factory=list)
+    parameters: list[str] = Field(default_factory=list)  # ooi.relation-name
```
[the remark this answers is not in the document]
Donnype: 🤔 Not sure but perhaps the goal was consistency here, but I'll remove these again

boefjes/boefjes/katalogus/api/plugins.py
```diff
@@ -93,35 +104,88 @@ def update_plugin_state(
...
-    try:
-        with plugin_service as p:
-            p.update_by_id(plugin_id, organisation_id, enabled)
-    except HTTPStatusError as ex:
-        raise HTTPException(ex.response.status_code)
+    with plugin_service as p:
+        p.set_enabled_by_id(plugin_id, organisation_id, enabled)
+
+
+class PatchBoefje(BaseModel):
```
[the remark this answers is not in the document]
ammar92: I meant the convention as used in [here](https://fastapi.tiangolo.com/tutorial/extra-models/)

boefjes/boefjes/katalogus/tests/integration/test_api.py
Donnype: Let's not forget that there's value in the mix as well though (note also that these are integration tests, not unit tests). One bug in the past was the fact that a `PUT` was not idempotent as we crashed on the second call, things we have a higher chance of finding with a bit of entropy in the tests created by following a regular access patterns (save, get, filter, remove). To have both we'd either be writing _many_ tests per endpoint, or invest in proper input parametrization/hypothesis testing because combinations explode, especially as you want to combine steps:

_If I add a {boefje/normalizer} and then add another {boefje/normalizer/already existing plugin} then filter on plugin type {boefje/normalizer} {with/without} a limit and offset and then delete a {boefje/normalizer} again, does the other plugin stay unchanged?_

This could yield `2*3*2*2*2 = 48` tests, and being integration tests (dropping/building the database, making API calls) these will start slow down rather soon.

I think in practice there's often a sweet spot: you know that pagination can be tested in isolation since you know the implementation. And you also know that deleting a model requires adding it , so you get two for one there, etc. You might argue that then you don't know if the `add` or `delete` failed. But in my experience, I prefer not having to fix 2 tests when 1 thing is broken.

Donnype:
@underdarknl Follow up ticket: https://github.com/minvws/nl-kat-coordination/issues/3052

boefjes/boefjes/katalogus/tests/integration/test_api.py
ammar92: Thanks for the explanation. My reasoning indeed came from a unit test perspective, but from the perspective of an integration test, this pattern could work as long as it covers the test and doesn't do more than necessary. For example, it should focus on critical workflows rather than exhaustive input combinations. Looking at the tests again I agree this approach makes sense.

> You might argue that then you don't know if the `add` or `delete` failed. But in my experience, I prefer not having to fix 2 tests when 1 thing is broken.

I don't fully agree. Because it's not often one thing that is broken, but could have multiple reasons for its failure, such as a bad implementation of the function or improper usage. Combining multiple operations in a test could obscure the root cause of a failure, which usually leads to more debugging time and unclear test results.

But anyways, it's a gray area sometimes and I think as long as we're fully aware of what we're doing and focus on write maintainable tests, we're doing a good thing.

ammar92 (approved):
Looks good to me 👍 I don't have any remarks anymore

boefjes/boefjes/katalogus/tests/integration/test_api.py
Donnype: Definitely gray area, and absolutely agreed on focusing on writing maintainable tests 👍

stephanie0x00:
## Checklist for QA:

### What works:
Nothing obvious is broken. Performed 

### What doesn't work:
n/a

### Bug or feature?:
n/a

Issue #3015, opened 2024-06-03
Investigate bug empty boefje consumes in scheduler


jpbruinsslot:
During development it is possible to set the plugin field `consumes` to an empty list. When a boefje gets enabled the scheduler will check whether this boefje can run on ooi's that allow this boefje to run on. To do that the scheduler checks what types a boefje can consume. The scheduler will ask octopoes to return those ooi's. When the consumes field of a boefje is empty it will ask octopoes to return ooi's and it will do so when an empty list is provided. Resulting in tasks being created.

PR #3017, opened 2024-06-03
Fix empty consumes of boefjes will trigger tasks in scheduler

jpbruinsslot:
### Changes

When a boefje gets enabled the scheduler will check whether this boefje can run on ooi's that allow this boefje to run on. To do that the scheduler checks what types a boefje can consume. The scheduler will ask octopoes to return those ooi's. When the `consumes` field of a boefje is empty it will ask octopoes to return ooi's and it will do so when an empty list is provided.

This PR adds a check before the scheduler will ask octopoes to return ooi's, it will check if the `consumes` field is empty or not and will not request octopoes for ooi's when it is.

### Issue link

Closes [#3015](https://github.com/minvws/nl-kat-coordination/issues/3015)

### QA steps

From what I can see it is not possible to set the consumes of a boefje to an empty list from the webapp. In order to make sure that everything should be working as intended, is to enable a boefje and make sure that the types the ooi's consumes, tasks are being created for ooi's of that type.

### DIFF
--- a/mula/scheduler/schedulers/boefje.py
+++ b/mula/scheduler/schedulers/boefje.py
@@ -215,9 +215,7 @@ class BoefjeScheduler(Scheduler):
         boefjes can run on, and create tasks for it."""
         new_boefjes = None
         try:
-            new_boefjes = self.ctx.services.katalogus.get_new_boefjes_by_org_id(
-                self.organisation.id
-            )
+            new_boefjes = self.ctx.services.katalogus.get_new_boefjes_by_org_id(self.organisation.id)
         except ExternalServiceError:
             self.logger.error(
                 "Failed to get new boefjes for organisation: %s from katalogus",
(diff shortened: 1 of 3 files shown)


mula/scheduler/schedulers/boefje.py
```diff
@@ -242,6 +242,16 @@ def push_tasks_for_new_boefjes(self) -> None:
         )
 
         for boefje in new_boefjes:
+            if not boefje.consumes:
```
underdarknl, line 245: Would it make sense to drop any boefje that has no inputs from the list of boefjes before we add it to the local 'boefjes' cache? Thath way we could clean the new_boefjes list before it even gets here.

stephanie0x00:
## Checklist for QA:

### What works:
Seems to work as expected, nothing identified that isn't working. 

### What doesn't work:

n/a

### Bug or feature?:

n/a

underdarknl (approved):

PR #3092, opened 2024-06-17
Fix rabbtimq exception handling in scheduler

### LINKED ISSUE #2948: Investigate test server results that impact scheduler
Since the test server has run quite a while, it would be helpful to inspect the results and see what impact the scheduler on performance.

jpbruinsslot:
### Changes

The following scheduler errors were encountered:

```
root-scheduler-1  | pika.exceptions.AMQPHeartbeatTimeout: No activity or too many missed heartbeats in the last 60 seconds 
root-scheduler-1  | pika.exceptions.ConnectionWrongStateError: BlockingConnection.add_callback_threadsafe() called on closed or closing connection.
root-scheduler-1  | RuntimeError: cannot schedule new futures after interpreter shutdown
```

We were expected that the scheduler would clean up after an unhandled exception. However, when trying to shutdown and clean up the listeners the scheduler throws another exception because the rabbitmq channel is closed. Which in turn is another unhandled exception triggering yet another shutdown. This results in a deadlock where the clean up of the whole scheduler application is stalled and other running threads keep running.

Within this PR we made sure we try-catch the `basic_consume` (start of consuming rabbitmq messaging), since we saw from the stack trace that the `AMQPHeartbeatTimeout:` was thrown there. Prior to this PR we didn't catch all exception that could've been throw by `basic_consume`, which resulted in unhandled exception we couldn't recover from.

Additionally, we made sure that when we shutdown a listener we also check whether we have an active connection and channel that we can close. This was the cause of the second unhandled exception.

### Issue link

Closes #2948 

### QA Instructions

Since the initial error was a lost heartbeat timeout, which could be hard to trigger, we can try to mimic this behaviour to start openkat, run a few jobs and terminate the rabbitmq instance to make sure the scheduler shut downs correctly.

```
$ docker-compose kill rabbitmq
```


mula/scheduler/connectors/listeners/listeners.py
```diff
@@ -193,14 +220,29 @@ def stop(self) -> None:
...
+        if self.connection.is_closed:
+            self.logger.debug("Connection already closed")
+            return
+
+        # When there is still an open connection, close the connection
+        # and the channel. Using the `add_callback_threadsafe` method
+        # ensures that the close can be done from any thread.
+        try:
             self.connection.add_callback_threadsafe(self._close_callback)
+        except Exception as exc:
```
jpbruinsslot, line 236: split this out in expected pika exceptions and unexpected

mula/scheduler/connectors/listeners/listeners.py
```diff
@@ -141,10 +164,11 @@ def basic_consume(self, queue: str, durable: bool, prefetch_count: int) -> None:
         try:
             self.channel.basic_qos(prefetch_count=prefetch_count)
             self.channel.basic_consume(queue, on_message_callback=self.callback)
+
+            # Blocking call that processes the messages
             self.channel.start_consuming()
-        except pika.exceptions.AMQPChannelError as exc:
-            # Do not recover on channel errors
-            self.logger.error("AMQPChannelError: %s", exc)
+        except Exception as exc:
```
jpbruinsslot, line 170: split this out in expected pika exceptions and unexpected

mula/scheduler/connectors/listeners/listeners.py
```diff
@@ -158,20 +185,23 @@ def callback(
...
+            message=body,
         )
 
         # Submit the message to the thread pool executor
         self.executor.submit(self.dispatch, channel, method.delivery_tag, body)
 
     def dispatch(self, channel, delivery_tag, body: bytes) -> None:
         # Check if we still have a connection
-        if not self.connection:
+        if not self.connection or self.connection.is_closed:
```
ammar92, line 199:
```suggestion
        if self.connection is None or self.connection.is_closed:
```

mula/scheduler/connectors/listeners/listeners.py
```diff
@@ -158,20 +185,23 @@ def callback(
...
             self.logger.debug("No connection available, cannot dispatch message!")
             return
 
         # Check if we still have a channel
-        if not self.channel:
+        if not self.channel or self.channel.is_closed:
```
ammar92, line 204:
```suggestion
        if self.channel is None or self.channel.is_closed:
```

mula/scheduler/connectors/listeners/listeners.py
```diff
@@ -193,14 +223,29 @@ def stop(self) -> None:
 
         self.executor.shutdown(wait=True)
 
-        if self.connection:
+        if not self.connection:
```
ammar92, line 226:
```suggestion
        if self.connection is None:
```

mula/tests/integration/test_listeners.py
```diff
@@ -0,0 +1,216 @@
...
+
+import pika
+from scheduler import connectors, utils
+
+from tests.mocks import listener
+
+
+class RabbitMQTestCase(unittest.TestCase):
+    def setUp(self):
+        self.dsn = "amqp://guest:guest@ci_rabbitmq:5672/%2Fkat"
```
ammar92, line 14: This could be a constant out of the class, or a class constant, as it's not a class instance variable that needs to be instantiated for each test

mula/tests/integration/test_listeners.py
ammar92, line 1: Please make sure to use more common names, e.g.:
- `t` -> `thread`
- `lst` -> `listener`

They do improve readability as they're more self-explanatory and less ambiguous

ammar92 (commented):
Looks good 👍 Just a few remarks to improve it even more

Generally I think this is also the best approach to deal with these kind of exceptions

mula/tests/integration/test_listeners.py
jpbruinsslot: for now I leave them as is, since there is a conflict with `from tests.mocks import listener`, and I can change it with  using `as` but I reckon that `lst` should be descriptive enough in this context

ammar92: It's still ambiguous and non conventional imho. I often read it as `list`. There are a few options:
- Rename variable to `listener` (see next points for the name conflict)
- Rename variable to `listener_`, adding an underscore as a postfix is often a solution to deal with name conflicts (with commonly defined keywords, built-in classes, functions, or other objects, but also for named objects within the same module)
- Revise import; e.g. `import tests.mocks.listener` or `from tests.mocks.listener import MockListener`
- Rename `test.mocks.listener` module to `test.mocks.listeners` which is in line with the source codebase (`scheduler.connectors.listeners`)

ammar92 (dismissed):
No remarks 👍

stephanie0x00:
## Checklist for QA:

### What works:
Enabled a lot of boefjes to create a large task list, then killed rabbitmq as mentioned in the QA notes. The scheduler logs created some error messages. I verified with @jpbruinsslot if this was/wasn't expected behaviour. He confirmed that it was, for the current PR, and will create new tickets for the errors in the log. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

underdarknl (approved):

PR #3174, opened 2024-07-02
Change report flow to POST requests

Rieven:
### Changes
- This PR is a part of a larger PR. 
- It implements POST request for OOIs selection (instead of GET)
- OOI selection will do a post request to report type selection (here you can view which OOIs you've selected)
- At report type selection page, you can go back to change selection (This is also done with a post request back to the OOI selection page.
- The other PR part 2 will implement the rest of the flow from report type selection posting to Setup scan/configuration page and saving report for viewing. Other pr: https://github.com/minvws/nl-kat-coordination/pull/3182

### Issue link
https://github.com/minvws/nl-kat-coordination/issues/3199

Closes https://github.com/minvws/nl-kat-coordination/issues/3199

### QA notes

Will only work for select oois (also works for select all) and next you will go to the report type selection page.
This PR stops there for post request. This is only step 1 which willl havdle only ooi selection and showing report type selection part.

---

### Code Checklist

---

---

## Checklist for QA:


rocky/reports/templates/partials/report_ooi_list.html
```diff
@@ -78,10 +81,10 @@ <h2>
                         <th>{% translate "Clearance type" %}</th>
                     </thead>
                     <tbody>
-                        {% for object in object_list %}
+                        {% for object in ooi_list %}
                             <tr>
                                 <td>
-                                    <input type="checkbox" id="submit-form" name="ooi" value={{ object.primary_key }} {% if object.primary_key in selected_oois or "all" in request.GET.ooi %}checked{% endif %} {% if "all" in request.GET.ooi %}disabled{% endif %}>
+                                    <input type="checkbox" name="ooi" value={{ object.primary_key }} {% if object.primary_key in selected_oois or "all" in request.GET.ooi %}checked{% endif %} {% if "all" in request.GET.ooi %}disabled{% endif %}>
```
ammar92, line 87: The value attribute should be enclosed with double quotes (and preferable encoded?). It's unsafe otherwise, since the primary key is basically user input.

rocky/reports/views/base.py
```diff
@@ -161,23 +142,34 @@ def get_ooi_filter_forms(self, ooi_types: Iterable[type[OOI]]) -> dict[str, Form
             )
         }
 
+    def ooi_selection_is_valid(self) -> bool:
+        is_valid = True
+        if not self.selected_oois:
+            is_valid = False
+            messages.error(self.request, _("Select at least one OOI to proceed."))
+        return is_valid
```
ammar92, line 150: This is a getter method with a side effect? The callee of the function should decide wether or not to throw the error based on the returned boolean

rocky/reports/views/base.py
```diff
@@ -202,8 +194,15 @@ def get_report_types(reports: set[type[BaseReportType]]) -> list[dict[str, str]]
             for report_type in reports
         ]
 
+    def report_type_selection_is_valid(self):
+        is_valid = True
+        if not self.selected_report_types:
+            is_valid = False
+            messages.error(self.request, _("Select at least one report type to proceed."))
+        return is_valid
```
ammar92, line 202: Also a getter with a side effect. See previous point about how to use such functions

rocky/reports/templates/partials/report_ooi_list.html
Rieven: I forgot to put the double quotes.

rocky/reports/views/base.py
Rieven: Yes, the thing I want to accomplish here are 2 things. 
1. I want to keep 1 message central to use for all other views
2. I want to check the validity just to redirect at certain places. In some cases I just show the message, at other times I use. If it is not valid redirect to another url with the messaging. 

I cannot really find a good way to have both functionality in one, because just throwing the message will return a None. I cannot check for None values since that anything can be None even when the message is not thrown.

Rieven: or another possibility is to create a dictionary with error messaging to keep it central and use it where needed.

Rieven: Should I rename this function so it doesn't look like a plain getter function?

Rieven: @ammar92 

What do you think of this?

```
from enum import StrEnum

class OOISelectionError(StrEnum):
    none_selection = _("Select at least one OOI to proceed.")
```

We can add as many error messages so we an pull it where needed?

ammar92: You can use a constant that holds the message if it's shared between several functions or modules. A dictionary or enums of possible strings would make it worse

Rieven: fixed: https://github.com/minvws/nl-kat-coordination/blob/ea12f0edbbe3de17bc97ad95eb3dcc1416b39436/rocky/reports/views/aggregate_report.py#L83

ammar92: Such constants are usually put outside of the class and somewhere at the top in the module to make them easier to find. This way will improve code readability, because configuration values are separated from implementation logic.

TwistMeister:
I believe the work done in this particular PR works as intended as I've been able to verify that the mentioned actions are indeed POST requests. What doesn't work is going from the report selection to the next step. That results in an error. After consulting @Rieven I learned that this is indeed intended since that part will be fixed in Part 2 he mentioned above.

So as long as this doesn't get merged without part 2, QA seems all good.

rocky/reports/templates/partials/report_ooi_list.html
```diff
@@ -103,7 +110,7 @@ <h2>
         {% endif %}
     </div>
 </section>
-<form id="select_all_objects_form" class="inline" method="get">
+<form novalidate id="select_all_objects_form" class="inline" method="get">
```
underdarknl, line 113: Is this 'novalidate' needed?

Rieven: We put here `novalidate` to show form error wrapped in Manon styling, like for example:
image

underdarknl (approved):

PR #3194, opened 2024-07-05
cve-2024-6387 from RickGeex

noamblitz:
### Changes

Servicebanner download boefje with a normalizer for cve 2024 6387.

Normalizer from https://github.com/RickGeex/CVE-2024-6387-Checker/blob/main/CVE-2024-6387.py


boefjes/boefjes/plugins/kat_cve_2024_6387/main.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def run(boefje_meta: BoefjeMeta) -> list[tuple[set, str | bytes]]:
+    input_ = boefje_meta.arguments["input"]  # input is IPPort
+    port = input_["port"]
+    ip = input_["address"]["address"]
+
+    status, banner = SSHChecker(ip, port, TIMEOUT).check_vulnerability()
+
+    return [(set(), json.dumps({"status": status, "banner": banner}))]
```
underdarknl, line 92: Lets return the banner in a separate raw file.

underdarknl:
```suggestion
    return [(set('openkat/servicebanner',), banner)]
```

boefjes/boefjes/plugins/kat_cve_2024_6387/normalizer.json
```diff
@@ -0,0 +1,10 @@
+{
+  "id": "kat_cve_2024_6387_normalize",
+  "consumes": [
+    "boefje/CVE-2024-6387"
```
underdarknl, line 4:
```suggestion
    "openkat/servicebanner"
```

boefjes/boefjes/plugins/kat_cve_2024_6387/normalize.py
```diff
@@ -0,0 +1,22 @@
...
+from octopoes.models import Reference
+from octopoes.models.ooi.findings import CVEFindingType, Finding
+
+
+def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
+    ooi = Reference.from_str(input_ooi["primary_key"])
+
+    results = json.loads(raw)
+
+    if results["status"] == "vulnerable":
```
underdarknl, line 14:
```suggestion
    if "SSH-2.0-OpenSSH" in raw:
        if (
            any(version in raw for version in SSHChecker.vulnerable_versions)
            and raw not in SSHChecker.excluded_versions
        ):
```

boefjes/boefjes/plugins/kat_servicebanner/main.py
```diff
@@ -0,0 +1,40 @@
...
+    except Exception as e:
+        return str(e)
+
+
+def run(boefje_meta: BoefjeMeta) -> list[tuple[set, str | bytes]]:
+    input_ = boefje_meta.arguments["input"]  # input is IPPort
+    port = input_["port"]
+    ip = input_["address"]["address"]
+
+    sshsock = get_ssh_sock(ip, port, TIMEOUT)
```
underdarknl, line 36: Handle the case where None is returned.

boefjes/boefjes/plugins/kat_servicebanner/main.py
```diff
@@ -0,0 +1,40 @@
...
+def get_ssh_banner(sock):
+    try:
+        banner = sock.recv(1024)
+        try:
+            banner = banner.decode().strip()
+        except UnicodeDecodeError:
+            banner = banner.decode("latin1").strip()
+        sock.close()
+        return banner
+    except Exception as e:
```
underdarknl, line 27: ReRaise this error, or dont catch it here, catch it around line 38 and return a different output-set if we find an error instead of a banner.

boefjes/boefjes/plugins/kat_servicebanner/main.py
```diff
@@ -0,0 +1,40 @@
+import socket
+
+from boefjes.job_models import BoefjeMeta
+
+TIMEOUT = 1.0
+
+
+def get_ssh_sock(ip, port, timeout):
```
underdarknl, line 8: this captures *any* banner, not just ssh right? rename the function

boefjes/boefjes/plugins/kat_servicebanner/main.py
```diff
@@ -0,0 +1,40 @@
...
+    try:
+        sock.connect((ip, port))
+        return sock
+    except Exception:
+        return None
+
+
+def get_ssh_banner(sock):
+    try:
+        banner = sock.recv(1024)
```
underdarknl, line 20: Maybe add another timeout call for this read? I'm not sure the initial timeout you set also applies to slow reads, since it was set before the connect call.

ammar92 (commented):
Cool stuff! Looks good overall. I had only one suggestion regarding naming convention. It would be more readable to separate 'servicebanner' into two words, since they are two words. So `openkat/servicebanner` (MIME type) would become `openkat/service-banner` and `kat_servicebanner` (plugin ID) would become `kat_service_banner`. Also it would be nice if the cover image could be compressed as it is relatively big in size.

boefjes/boefjes/plugins/kat_cve_2024_6387/normalize.py
```diff
@@ -0,0 +1,56 @@
...
+
+def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
+    ooi = Reference.from_str(input_ooi["primary_key"])
+
+    banner = raw.decode()
+
+    if (
+        "SSH-2.0-OpenSSH" in banner
+        and any(version in banner for version in VULNERABLE_VERSIONS)
+        and banner not in EXCLUDED_VERSIONS
```
dekkers, line 47: This will start to give only false positives when Debian or Ubuntu will do another update, because then the version will be `SSH-2.0-OpenSSH_9.2p1 Debian-2+deb12u4` for example and that will not be in `EXCLUDED_VERSIONS`.

ammar92 (approved):

underdarknl (approved):

PR #3231, opened 2024-07-11
Feat/human readable dates

### LINKED ISSUE #3221: Human readable time on Object table page
**Describe the feature**
PR #3186 introduces the 'When' column in the Last observed by table to show when an object was last identified. The current timestamp is not human readable, but is already an improvement. A better human readable time stamp would be nicer to the user.  For example `2024-10-31 13:37:37` or another time format. I suppose it would also be nice if the user can choose his/her preference for the timestamp formatting. 

**To Reproduce**

1. Make sure you have some data in the interface. 
2. Click on Objects > Open any object.
3. Scroll down to the 'Last observed by' section and view the timestamp

**Expected behavior**
A human readable timestamp for the user. 

**Screenshots**
image

**OpenKAT version**
main

underdarknl:
### Changes

Adds human readable timetamps to the origin table. 

### Issue link

closes #3221 

### QA notes

Human readable dates can now be seen on the origin table on the OOI-Detail page. Boefjes that have ran too long ago and have produced stale data will be highlighted with a warning.

### Code Checklist

---

---

## Checklist for QA:


rocky/rocky/views/mixins.py
```diff
@@ -46,6 +48,19 @@ class OriginData(BaseModel):
     boefje: Boefje | None = None
     params: dict[str, str] | None = None
 
+    @property
+    def is_old(self, maxage: Optional[int] = None) -> bool
+        if not maxage:
```
ammar92, line 53: See suggestion above. But even if it was nullable, it would make more sense to explicitly check for `None` instead. Otherwise, it's ambiguous; are you checking for `None`, a value of `0`, or both?

Additionally, consider implementing a range check. For example, check if the value is greater than 0 and raise a `ValueError` otherwise

rocky/rocky/views/mixins.py
```diff
@@ -46,6 +48,19 @@ class OriginData(BaseModel):
...
+            maxage = ORIGIN_MAX_AGE
+        if not self.normalizer:
```
ammar92, line 55:
```suggestion
        if self.normalizer is None:
```
or leave as is, if you're specifically validating against both a `None` and an empty dict

rocky/rocky/views/mixins.py
```diff
@@ -46,6 +48,19 @@ class OriginData(BaseModel):
...
+            return False
+            
+        observationdate = self.normalizer.get("raw_data", {}).get("boefje_meta", {}).get("ended_at", False)
```
ammar92, line 58: Two things:
- `observartion_date` is a more clear and Pythonic name
- The latest `False` which is used as a default value doesn't make sense. By default it returns `None`, which is a better way of indicating non-existence. Also, try not to mix too much types: Ideally you want at most a `datetime | None`, and not a `datetime | bool | None`.

```suggestion
        observation_date = self.normalizer.get("raw_data", {}).get("boefje_meta", {}).get("ended_at")
```

rocky/rocky/views/mixins.py
```diff
@@ -46,6 +48,19 @@ class OriginData(BaseModel):
...
+        old = datetime.datetime.utcnow() - datetime.timedelta(seconds=maxage)
+        if observationdate and observationdate < old:
+            return True
+        return False
```
ammar92, line 62: See above and this suggestion:
```suggestion
        return observation_date and observation_date < old:
```

rocky/rocky/views/mixins.py
```diff
@@ -46,6 +48,19 @@ class OriginData(BaseModel):
     boefje: Boefje | None = None
     params: dict[str, str] | None = None
 
+    @property
+    def is_old(self, maxage: Optional[int] = None) -> bool
```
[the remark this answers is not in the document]
ammar92: Also, in our code base we use new style hinting; so `Optional[X]` would be `X | None`

rocky/rocky/views/mixins.py
```diff
@@ -32,6 +33,7 @@
 
 logger = structlog.get_logger(__name__)
 
+ORIGIN_MAX_AGE = 60*60*24 * 2 # 2 days in seconds
```
ammar92, line 36: I suggest using a `timedelta` directly. This way you don't have to create a `timedelta` later on and you get documentation for free.

```suggestion
ORIGIN_MAX_AGE = datetime.timedelta(days=2)
```

rocky/rocky/views/mixins.py
```diff
@@ -3,6 +3,7 @@
 from datetime import datetime, timezone
 from functools import cached_property
 from operator import attrgetter
+from typing import Optional
```
ammar92, line 6: Not necessary to use anymore, see comment below

rocky/rocky/views/mixins.py
```diff
@@ -46,6 +48,19 @@ class OriginData(BaseModel):
     boefje: Boefje | None = None
     params: dict[str, str] | None = None
 
+    @property
```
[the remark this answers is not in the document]
ammar92: Another solution, assuming a property would be convenient to use within the templates, is to create a `is_older_than` method, and still have this property (without arguments) calling the `is_older_than` method with the default value.

But since I haven't seen any uses of `max_age`, I suggest leaving the `max_age` value out and just have
```
@property
def is_old() -> bool:
    ... # check against `ORIGIN_MAX_AGE`
```

rocky/rocky/templates/oois/ooi_detail_origins_observations.html
```diff
@@ -35,7 +37,12 @@ <h2>{% translate "Last observed by" %}</h2>
                                     {% translate "No Raw file could be found, this might point to an error in OpenKAT" %}
                                 {% endif %}
                             </td>
-                            <td>{{ observation.normalizer.raw_data.boefje_meta.ended_at }}</td>
+                            <td {% if observation.is_old %} aria-label="{% translate "Warning" %}" {% endif %}>
+                                {% if observation.is_old %}
+                                    <span>{% translate "Warning" %}:</span>
```
ammar92, line 42:
```suggestion
                                    <span>{% translate "Warning:" %}</span>
```

ammar92 (commented):
Nice feature. I just have a few remarks to improve the implementation

TwistMeister:
## Checklist for QA:

### What works:
- Could not verify because the page won't load

### What doesn't work:
- The OOI detail page wouldn't load because of a template error:
Screenshot 2024-07-19 at 12 08 47

rocky/rocky/views/mixins.py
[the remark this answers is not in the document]
underdarknl: Creating a property (for easy access with a default) makes sense, and creating a is_older_than with an argument makes sense.
Especially when we can replace the default based on the freshness expectations we have of each OOI with a data-driven timeout later on.

rocky/rocky/views/mixins.py
underdarknl: Since the job might not have a normalizer (due to various errors), I was trying to do some defensive coding.

ammar92:
> ### What works:
> * Could not verify because the page won't load

Thanks for the review. My bad, I forgot to check it after the rework. I've fixed the bug and verified it. It should be working now. 
image

underdarknl:
I expected More from the naturaltime template function. this looks like its not doing much?

ammar92:
> I expected More from the naturaltime template function. this looks like its not doing much?

I don't know, I didn't touch the template code.

I'll look into it

ammar92:
> > I expected More from the naturaltime template function. this looks like its not doing much?
> 
> I don't know, I didn't touch the template code.
> 
> I'll look into it

Seems as if the time passed into the template context wasn't a datetime object but rather a string. Should be fixed now
image

underdarknl:
We need a nice and upstreamed 'human readable' and accessible form of this in Manon.

PR #3292, opened 2024-07-24
fix open redirects in plugin_enable_disable.py

underdarknl:
This fixes two open redirects by binding them onto our own domain.

### Changes

These changes add the requested url, after the current hostname making it an absolute redirect limited to our own webserver/hostname instead of anywhere.
Allowing redirects to anywhere is a potential security issue.

### Issue link

https://github.com/minvws/nl-kat-coordination/security/code-scanning/1
https://github.com/minvws/nl-kat-coordination/security/code-scanning/2

### QA notes

Enabling / Disabling plugins should still work.

### Code Checklist

---

---

## Checklist for QA:

### DIFF
--- a/rocky/katalogus/views/plugin_enable_disable.py
+++ b/rocky/katalogus/views/plugin_enable_disable.py
@@ -24,7 +24,8 @@ class PluginEnableDisableView(SinglePluginView):
                 messages.WARNING,
                 _("{} '{}' disabled.").format(self.plugin.type.title(), self.plugin.name),
             )
-            return HttpResponseRedirect(request.POST.get("current_url"))
+            url = request.build_absolute_uri(request.POST.get("current_url"))
+            return HttpResponseRedirect(url)
 
         try:
             plugin_settings = self.katalogus_client.get_plugin_settings(self.plugin.id)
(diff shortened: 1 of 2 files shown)


ammar92:
Not sure why the tests are failing, the tests seems to work locally

rocky/katalogus/views/plugin_enable_disable.py
```diff
@@ -91,5 +92,5 @@ def post(self, request, *args, **kwargs):
                     self.plugin.scan_level.value,
                 ),
             )
-
-        return HttpResponseRedirect(request.POST.get("current_url"))
+            url = request.build_absolute_uri(request.POST.get("current_url"))
+            return HttpResponseRedirect(url)
```
dekkers, line 96: The indentation changed here which results in the unit test errors.

The change also doesn't fix anything, because if you give an absolute uri to `request.build_absolute_uri` it will return the same uri.

dekkers (commented):
I don't see the open redirect vulnerability here. We have CSRF protection, so I don't see how `current_url` can be provided by an attacker. Am I missing something or is this a false positive?

rocky/katalogus/views/plugin_enable_disable.py
ammar92: Nice catch.

What would be the best solution in this case? Check against an allowed domain list?

underdarknl: You are right, its a False positive in combination with our use of CSRF. There's no way to influence the current_url param and still make it trough the CSRF without the user being involved. This PR can be closed.
I also did not notice the indentation change. Weird that it happend in the first place , but probably because of my copying of the line from the earlier method.

Issue #3287, opened 2024-07-23
Flexible scheduling: new boefjes not working correctly


jpbruinsslot:
See description of https://github.com/minvws/nl-kat-coordination/pull/3297

PR #3297, opened 2024-07-24
Fix new boefjes issue for scheduler

jpbruinsslot:
#### Wrong cache gets flushed during flushing of another cache

While working on https://github.com/minvws/nl-kat-coordination/pull/2786 I encountered an issue with how newly enabled boefjes were being handled by the scheduler. By taking note of the logs it seemed that the scheduler identified all current enabled boefjes to be new and would check if for those boefjes ooi's were available to run on.

This was mainly because of the following line in the code:

https://github.com/minvws/nl-kat-coordination/blob/7e333a3d66f064eb4a4f1d12c0945037e6e36066/mula/scheduler/connectors/services/katalogus.py#L65

It would clear the `new_boefjes` cache every time the `plugin_cache` was being flushed. These caches are unrelated and flushing one should not clear the other one.

However the plugin cache in main isn't used and it's cache is also not being flushed by either `get_plugin_by_id_and_org_id()` or `get_plugins_by_org_id()` methods in the katalogus service. So we don't encounter this issue on main.

#### Thread locking mechanism for `new_boefjes` cache added

I noticed the `new_boefjes` cache did not implement a locking mechanism. Since multiple threads are accessing and writing to the shared cache I've added a thread lock for this cache as well.

#### Implemented separate locks for individual katalogus caches

Instead of one thread lock for all the katalogus caches, I've implemented individual locks for the caches. This ensures that calls to one cache will not block calls from other threads to the other caches.

#### Added thread safety tests

To ensure everything would be working accordingly and safeguard any future regression this PR specifically adds thread safety tests.

#### General refactoring of variable names, and code clean-ups

The long variable cache names were unnecessary and did not help with making the variable names more descriptive. Additionally, this PR also added several closures to some of the methods that access the caches and need to lock the resource, this to reduce some code duplication and enhance readability.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/3287

### QA notes

This is issue is not seen on main but I can show its workings on the flexible scheduling feature branch. The main focus for QA would be checking if enabling of plugins will continue working correctly. E.g. when enabling a boefje that would start and execute scan because ooi's are the correct scan level.

---

---

## Checklist for QA:


mula/scheduler/connectors/services/katalogus.py
```diff
@@ -161,74 +164,98 @@ def get_organisations(self) -> list[Organisation]:
...
+        with self.new_boefjes_cache_lock:
+            # Get the enabled boefjes for the organisation from katalogus
+            plugins = self.get_plugins_by_organisation(organisation_id)
+            enabled_boefjes = {
+                plugin.id: plugin
+                for plugin in plugins
+                if plugin.enabled is True and plugin.type == "boefje" and plugin.consumes
+            }
+
+            self.logger.info(
```
jpbruinsslot, line 230: Remove. Were used during development

mula/scheduler/connectors/services/katalogus.py
```diff
@@ -161,74 +164,98 @@ def get_organisations(self) -> list[Organisation]:
...
+
+            # Check if there are new boefjes
+            new_boefjes = []
+            for boefje_id, boefje in enabled_boefjes.items():
+                if boefje_id in self.new_boefjes_cache.get(organisation_id, {}):
+                    continue
+
+                new_boefjes.append(boefje)
+
+            self.logger.info(
```
jpbruinsslot, line 244: Remove. Were used during development

mula/scheduler/connectors/services/katalogus.py
```diff
@@ -22,69 +22,71 @@ def __init__(
...
-        self.organisations_plugin_cache: dict_utils.ExpiringDict = dict_utils.ExpiringDict(lifetime=cache_ttl)
+        self.plugin_cache_lock = threading.Lock()
+        self.plugin_cache: dict_utils.ExpiringDict = dict_utils.ExpiringDict(lifetime=cache_ttl)
 
         # For every organisation we cache on which type of object (consumes)
         # the boefjes consume, it references the object type (consumes)
         # as the key and a dict of boefjes as value.
-        self.organisations_boefje_type_cache: dict_utils.ExpiringDict = dict_utils.ExpiringDict(lifetime=cache_ttl)
+        self.boefje_cache_lock = threading.Lock()
+        self.boefje_cache: dict_utils.ExpiringDict = dict_utils.ExpiringDict(lifetime=cache_ttl)
```
ammar92, line 34: The type hints for the `ExpiringDict` and other `dict`s below are unnecessary

mula/scheduler/connectors/services/katalogus.py
```diff
@@ -22,69 +22,71 @@ def __init__(
...
+            self.plugin_cache.expiration_enabled = False
+            self.plugin_cache.reset()
 
             orgs = self.get_organisations()
             for org in orgs:
-                if org.id not in self.organisations_plugin_cache:
-                    self.organisations_plugin_cache[org.id] = {}
-                    self.organisations_new_boefjes_cache[org.id] = {}
+                if org.id not in self.plugin_cache:
+                    self.plugin_cache[org.id] = {}
```
ammar92, line 66: You can use `dict.setdefault` for this purpose

```suggestion
                plugin_cache.setdefault(org.id, {})
```

mula/scheduler/connectors/services/katalogus.py
```diff
@@ -22,69 +22,71 @@ def __init__(
...
+            self.boefje_cache.reset()
 
             orgs = self.get_organisations()
             for org in orgs:
-                self.organisations_boefje_type_cache[org.id] = {}
+                self.boefje_cache[org.id] = {}
 
-                for plugin in self.get_plugins_by_organisation(org.id):
+                plugins = self.get_plugins_by_organisation(org.id)
+                for plugin in plugins:
```
ammar92, line 89: Since you're not using the `plugins` variable, you might as well leave it out (like the previous implementation)

mula/scheduler/connectors/services/katalogus.py
```diff
@@ -97,30 +99,31 @@ def flush_organisations_boefje_type_cache(self) -> None:
...
+            self.normalizer_cache.reset()
 
             orgs = self.get_organisations()
             for org in orgs:
-                self.organisations_normalizer_type_cache[org.id] = {}
+                self.normalizer_cache[org.id] = {}
 
-                for plugin in self.get_plugins_by_organisation(org.id):
+                plugins = self.get_plugins_by_organisation(org.id)
+                for plugin in plugins:
```
ammar92, line 126: See previous point

mula/scheduler/connectors/services/katalogus.py
```diff
@@ -161,74 +164,86 @@ def get_organisations(self) -> list[Organisation]:
...
+                if plugin.enabled is True and plugin.type == "boefje" and plugin.consumes
+            }
+
+            # Check if there are new boefjes
+            new_boefjes = []
+            for boefje_id, boefje in enabled_boefjes.items():
+                if boefje_id in self.new_boefjes_cache.get(organisation_id, {}):
+                    continue
+
+                new_boefjes.append(boefje)
```
ammar92, line 236: If you use flip the condition by using a `not in` operator, you'll lose the extra else-clause, which makes this imho a bit more maintainable and cleaner

ammar92 (commented):
Looks good! I just have a few remarks/ suggestions to improve it even more. The provided PR context is really good and helpful, very much appreciated it. Keep up the good work

mula/scheduler/connectors/services/katalogus.py
jpbruinsslot: yes love it

mula/scheduler/connectors/services/katalogus.py
jpbruinsslot: Unfortunately `self.plugin_cache` is of type `ExpiringDict` that doesn't implement the `setdefault()` method. I've added the method to the `ExpiringDict` to make this possible as well.

jpbruinsslot:
> Looks good! I just have a few remarks/ suggestions to improve it even more. The provided PR context is really good and helpful, very much appreciated it. Keep up the good work

Thanks! Made changes according your suggestions

mula/scheduler/connectors/services/katalogus.py
ammar92: Cool. Maybe a bit late but something to think about (for another PR); if `ExpiringDict` is supposed to act like a dict, maybe it could extend a `dict` or `UserDict` instead

ammar92 (dismissed):

stephanie0x00:
## Checklist for QA:

### What works:
Seems to look good. Have enabled multiple boefjes, rescheduled some tasks (boefjes and normalizers). Findings are created and tasks complete. 

### What doesn't work:
Nothing identified. 

### Bug or feature?:
n/a

underdarknl (approved):

Issue #3310, opened 2024-07-31
Scan profile declaration from External Database not working


zcrt:
**Describe the bug**
https://github.com/minvws/nl-kat-coordination/blob/eefc519c9b40e7b5ce826d91e24242a8c8c5e608/boefjes/boefjes/plugins/kat_external_db/normalize.py#L53 and https://github.com/minvws/nl-kat-coordination/blob/eefc519c9b40e7b5ce826d91e24242a8c8c5e608/boefjes/boefjes/plugins/kat_external_db/normalize.py#L71 do not seem to be working

**To Reproduce**

1. Connect an external DB
2. Notice clearance level stays 0.

**Expected behavior**
Using the hardcoded clearance `level=3`.

**Screenshots**
![image](https://github.com/user-attachments/assets/00296257-c23c-4f81-a6fe-49abcc9ad5ce)

**OpenKAT version**
main (31 july 2024)

originalsouth:
Can replicate bug with https://gist.github.com/originalsouth/b922ecef262ef4dac6802c0410f2a058

originalsouth:
Desired result can be achieved by adding: `BOEFJES_SCAN_PROFILE_WHITELIST='{"kat_external_db_normalize": 3}'` to the `.env` file.

zcrt:
Would this solution still work with https://github.com/minvws/nl-kat-coordination/pull/3311? Because this looks like a hardcoded solution for a specific declaration.

originalsouth:
> Would this solution still work with #3311? Because this looks like a hardcoded solution for a specific declaration.

@dekkers

PR #3311, opened 2024-07-31
feat: :hammer: Add indemnification level from external DB

zcrt:
### Changes

Adds the option to the external DB normalizer to add indemnification levels from external databases.

### Issue link

Does not close, but needs a solve of https://github.com/minvws/nl-kat-coordination/issues/3310 in order to work properly.

### Demo
image

### QA notes

The `.json` from the boefje with an item should also contain `indemnification_level` with a valid integer.

---

### Code Checklist

---

---

## Checklist for QA:


boefjes/boefjes/plugins/kat_external_db/normalize.py
```diff
@@ -29,6 +31,18 @@ def follow_path_in_dict(path, path_dict):
     return path_dict
 
 
+def get_indemnification_level(path_dict, path=INDEMNIFICATION_ITEM_PATH, default=DEFAULT_INDEMNIFICATION_LEVEL):
```
underdarknl, line 34: I'm not sure using tuples as default values is a good idea here.

dekkers: The code also doesn't use these arguments, so the arguments should be removed and the default can be used directory in the code.

boefjes/boefjes/plugins/kat_external_db/normalize.py
```diff
@@ -17,6 +17,8 @@
 IP_ADDRESS_ITEM_PATH = ["address"]
 DOMAIN_LIST_PATH = ["domains"]
 DOMAIN_ITEM_PATH = ["name"]
+INDEMNIFICATION_ITEM_PATH = ["indemnification_level"]
+DEFAULT_INDEMNIFICATION_LEVEL = 3
```
dekkers, line 21: I am not sure we should have a default indemnification level of 3. Wouldn't 2 be more appropriate?

boefjes/boefjes/plugins/kat_external_db/normalize.py
zcrt: I can change it, although using variables was per-sorting for when normalizer settings are available.

boefjes/boefjes/plugins/kat_external_db/normalize.py
zcrt: It used to be 3 already, this PR only adds it from an external database

boefjes/boefjes/plugins/kat_external_db/normalize.py
zcrt: This way it is easier to change the settings for a different DB because they are grouped together in the file.

dekkers: I am fine with the constants defined in the top of the file.

What I meant we can just use the constants in the code instead of defining function arguments that are never used by callers. Adding function arguments implies that the function is called with those arguments, making it in my opinion confusing because that doesn't happen. It's just a small thing here and on itself it is not that really a big problem, but what I am concerned with is that if we do this everywhere we get a more complex and harder to understand codebase in general.

To prevent this I am fan of the YAGNI (You Aren't Gonna Need It) principle: only create the code that is needed now and create the code that is needed later when it needed. In my opinion the principle applies both to whole features and smaller code parts such as here.

originalsouth:
Confirm the code works as advertised; only feedback is that `clearance` as a keyword instead of `indemnification_level`., as also used in the CSV importer, would be more simple.

See also:
https://github.com/minvws/nl-kat-coordination/blob/eba91950513211442aa0ace66c20f113341201cd/rocky/rocky/views/upload_csv.py#L40-L41

originalsouth:
Code can easily be tested with
```
#!/usr/bin/env python
import json
from http.server import BaseHTTPRequestHandler, HTTPServer


class PotatoServer(BaseHTTPRequestHandler):
    def do_GET(self):
        self.send_response(200)
        self.send_header("Content-type", "application/json")
        self.end_headers()
        self.wfile.write(json.dumps({
            "ip_addresses": [{"address": f"10.0.0.{i}", "indemnification_level": 2} for i in range(0xFF)],
            "domains": [{"name": f"potato{i}.com", "indemnification_level": 1} for i in range(0xFF)]
        }).encode())


if __name__ == "__main__":
    HTTPServer(("", 8080), PotatoServer).serve_forever()
```

originalsouth (approved):

stephanie0x00:
I'm very confused by this PR, but when doing the following:

> `Desired result can be achieved by adding: BOEFJES_SCAN_PROFILE_WHITELIST='{"kat_external_db_normalize": 3}' to the .env file.`

While having the pototoserver running and serving the file, I get a lot of potato hosts which end up with a L3 clearance. So I think it works. 

I'd rather see the clearance level setting available in the Katalogus instead of the .env file, because regular users will not be able to find this, and thus use this boefje.

originalsouth:
> I'm very confused by this PR, but when doing the following:
> 
> > `Desired result can be achieved by adding: BOEFJES_SCAN_PROFILE_WHITELIST='{"kat_external_db_normalize": 3}' to the .env file.`
> 
> While having the pototoserver running and serving the file, I get a lot of potato hosts which end up with a L3 clearance. So I think it works.
> 
> I'd rather see the clearance level setting available in the Katalogus instead of the .env file, because regular users will not be able to find this, and thus use this boefje.

Shouldn't the potato hosts have L2?

underdarknl:
> I'd rather see the clearance level setting available in the Katalogus instead of the .env file, because regular users will not be able to find this, and thus use this boefje.

It's by design that not every normalizer can set indemnifications. Because we also don't want every user to have the power to enable this, we opted to make this an env-var that is controllable by the system admins.

underdarknl:
> > I'm very confused by this PR, but when doing the following:
> > > `Desired result can be achieved by adding: BOEFJES_SCAN_PROFILE_WHITELIST='{"kat_external_db_normalize": 3}' to the .env file.`
> > 
> > 
> > While having the pototoserver running and serving the file, I get a lot of potato hosts which end up with a L3 clearance. So I think it works.
> > I'd rather see the clearance level setting available in the Katalogus instead of the .env file, because regular users will not be able to find this, and thus use this boefje.
> 
> Shouldn't the potato hosts have L1?

Yes, they should.

underdarknl (approved):

PR #3314, opened 2024-08-01
Add audit logging to CRUD actions using Django signals

dekkers:
### Changes

This adds audit logging for CRUD actions in Django using signals.

The check for token auth so that tokens can be used for non-DRF views has been moved to its own middleware. The auth required middleware should come after the structlog middleware, but the new auth token middleware is put before the structlog middleware so that middleware can use the user.

### Issue link

Closes #3178

### Demo

image

### QA notes

Check that adding, changing and removing rocky objects (users, organizations, members, indemnifications, etc.) still work and log the right message.

---

## Checklist for code reviewers:

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_qa.md) into your comment.

PR #3317, opened 2024-08-05
Basic audit trails via logging

### LINKED ISSUE #3043: We need to investigate which events we want to add the our Audit Log.
We need to investigate which events we want to add the our Audit Log.

_Originally posted by @underdarknl in https://github.com/minvws/nl-kat-coordination/issues/3038#issuecomment-2152755534_
            
I currently see the following:
* Adding OOI's [1]
* Editing OOI's [1]
* Deleting OOI's [1]
* Setting Declared indemnifications on OOIs [1]
* Deleting Declared indemnifications on OOIs [1]
* Enabling Plugins [2]
* Disabling plugins [2]
* Editing plugin settings [2]
* Creating Plugin jobs [3]
* Downloading Raw files [4] 
* Login [4]
* Logout[4]
* Password reset [4]
* Account creation [4]
* Account deletion [4]
* Member block/unblock [4]
* Member organization membership changes [4]
* Organization creation [4]
* Organization deletion [4]
* Setting Max indemnificatoin levels for users [4]
* Accepting Indemnification levels for yourself [2]

1: Can be stored on Octopoes on actual object model by adding specific fields
2: Needs a new table design to track changes over time (possibly move to XTDB?)
3: Store in Scheduler Job?
4: Newly created Audit Log?

All of the above should also be send out as structured logs to a remote aggregator.

ammar92:
### Changes

This PR implements some more audit trails via logging:
- Adding, creating and deleting objects
- Plugins
  - Change plugin state
  - Update plugin settings
- Raw data has been uploaded
- Login, logout and failed logins
- Organization creation and deletion

Most of the other logging related to models have been implemented in #3314 and merged into this branch.

Also implemented basic event codes for specific ORM record trails. This is a very basic implementation to get things started. I also considered using a tuple for the event codes (an element for each CRUDE action), but since there's probably still room for improvement and development around the audit trails I used a dict for now for flexibility.

Ideally we'd use more descriptive codes (using alphanumerics) when possible. In both cases, whether we use alphanumeric or numeric codes, we should work towards automatically labeling them as much as possible.

### Issue link

Partially closes #3043 

### QA notes

You should see more logging on certain actions, such as creating an object or enabling a plugin. On certain model changes (such as an organization or organization tags), you should also see an `event_code` variable in the log record.

---

### Code Checklist

---

---

## Checklist for QA:


rocky/katalogus/client.py
```diff
@@ -84,6 +84,7 @@ def __init__(self, base_uri: str, organization: str):
         self.session = httpx.Client(base_url=base_uri)
         self.organization = organization
         self.organization_uri = f"/v1/organisations/{organization}"
+        self.logger = logger.bind(organization_code=organization)
```
dekkers, line 87: I think it would better to add the organization code to the context in OrganizationView once instead of having to add it to a lot of different places. We can also use the module level logger here.

octopoes/octopoes/connector/octopoes.py
```diff
@@ -38,8 +39,8 @@ class OctopoesAPIConnector:
     def __init__(self, base_uri: str, client: str):
         self.base_uri = base_uri
         self.client = client
-
         self.session = httpx.Client(base_url=base_uri, timeout=30, event_hooks={"response": [self._verify_response]})
+        self.logger = structlog.get_logger("octopoes-connector", client=client)
```
dekkers, line 43: The client object doesn't have a string representation, so this doesn't seem useful? The organisation code can be added in the OrganisationView. We can also use a module level logger here in that case.

rocky/tools/models.py
```diff
@@ -266,13 +270,17 @@ class Indemnification(models.Model):
...
+    EVENT_CODES = {"created": 900211, "updated": 900212, "deleted": 900213}
+
 
 class OOIInformation(models.Model):
     id = models.CharField(max_length=256, primary_key=True)
     last_updated = models.DateTimeField(auto_now=True)
     data = models.JSONField(null=True)
     consult_api = models.BooleanField(default=False)
 
+    EVENT_CODES = {"created": 900221, "updated": 900222, "deleted": 900223}
```
dekkers, line 282: The event codes of the above 3 models are the same?

ammar92: Nice catch, fixed it

octopoes/octopoes/connector/octopoes.py
ammar92: The `client` here is a string representation (of a node).

> The organisation code can be added in the OrganisationView

This would work within the context of Rocky, but this client is also used elsewhere, e.g. by the Boefje workers

stephanie0x00:
## Checklist for QA:

### What works:
Looks good. I see various events in the logs. Not observing anything weird, like error messages. Onboarding works and boefjes can be started and complete.

### What doesn't work:
n/a

### Bug or feature?:
n/a

octopoes/octopoes/connector/octopoes.py
dekkers: Ok, but then we should call it `organization_code` and not use different keys for the same thing, because that makes it difficult/impossible to filter/search log messages.

```suggestion
        self.logger = structlog.get_logger("octopoes-connector", organisation_code=client)
```

ammar92: You're absolutely right, will commit the suggestion

rocky/katalogus/client.py
ammar92: Should be fixed now

underdarknl (approved):

PR #3338, opened 2024-08-09
Implement subreport rename form in table and remove it from dialog

### LINKED ISSUE #3286: Update implementation of 'give my report a name' with in-table input fields and 'undo' option
This is an update to the implementation of 'give my report a name' [#2775](https://github.com/minvws/nl-kat-coordination/issues/3250). After discussion meeting and a design review [#3250](https://github.com/minvws/nl-kat-coordination/issues/3250). 

Implementation Figma file link will follow after the design review.

### LINKED ISSUE #3339: Add the ability to use date placeholders in report names in the config flow

TwistMeister:
### Changes

Updates the implementation of 'give my report a name' with in-table input fields and 'undo' option, instead of a dialog per report row.

Closes #3286 
Closes #3339 

### Demo

### QA notes

In step 4 of a multi report you can now edit the names in the table directly. When the input differs from the default generated name a "reset arrow" button should be shown besides the input to reset its contents to the original generated name.

---

### Code Checklist

---

---

## Checklist for QA:


ammar92 (commented):
Nice work. Just a few tiny remarks, but can QA'd tested already

rocky/assets/js/renameReports.js
```diff
@@ -1,35 +1,72 @@
...
+//   let new_name = target.querySelector(".new-report-name").value;
+
+//   if (new_name) {
+//     if (reference_date) {
+//       new_name += " (" + reference_date + ")";
+//     }
+//     update_target_input.setAttribute("value", new_name);
+//     update_target_text.textContent = new_name;
+//   }
+// }
```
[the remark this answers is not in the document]
TwistMeister: Whoops, thanks for pointing this out. Completely forgot about this part

stephanie0x00:
## Checklist for QA:

### What works:
Seems to work as described. Couldn't test the Multi Report section as the file upload for report data appears to be broken. For normal reports it does work. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

underdarknl (dismissed):

rocky/reports/views/mixins.py
```diff
@@ -78,9 +80,6 @@ def save_report(self, report_names: list) -> ReportOOI:
                         if ooi_name in default_name and report_type_name in default_name:
```
underdarknl, line 80: When this if does not trigger. name stays untouched, but is still used later on at line 93 leading to a unassigned var

rocky/assets/js/renameReports.js
```diff
@@ -1,35 +1,38 @@
...
-  let confirm_buttons = root.querySelectorAll(".confirm-button");
+function initResetButtons() {
+  const resetButtons = document.querySelectorAll(".reset-button");
 
-  confirm_buttons.forEach((button) => initClickHandlers(button));
-}
+  resetButtons.forEach((button) => {
+    button.classList.add("hidden");
+
+    let input = button.closest("tr").querySelector(".name-input");
```
ppvg, line 11: You could consider hardening against the tr or input not existing;
```suggestion
    let input = button.closest("tr")?.querySelector(".name-input");
    if (!input) return;
```

rocky/assets/js/renameReports.js
```diff
@@ -1,35 +1,38 @@
...
+    input.defaultvalue = input.value;
 
-    editReportName(target);
+    button.addEventListener("click", function (event) {
+      input.value = input.defaultvalue;
+      button.classList.add("hidden");
+    });
   });
+
+  watchInputChanges();
```
ppvg, line 21: You could consider using event delegation instead of adding event listeners to each individual button and input. E.g.

```js
function initRenameReports() {
  const table = document.querySelector(".reset-button").closest("table");
  if (!table) return;

  table.addEventListener("click", function (event) {
    if (!event.target.matches(".reset-button")) return;
    event.stopPropagation();
    const button = event.target;
    const input = button.closest("tr")?.querySelector(".name-input");
    if (!input) return;
    input.value = input.defaultvalue;
    button.classList.add("hidden");
  });

  table.addEventListener("change", function (event) {
    if (!event.target.matches(".name-input")) return;
    event.stopPropagation();
    const input = event.target
    const button = input.closest("tr")?.querySelector(".reset-button");
    if (!button) return;
    if (input.defaultValue === input.value) {
      button.classList.add("hidden");
    } else {
      button.classList.remove("hidden");
    }
  });
}
```

rocky/reports/templates/partials/export_report_settings.html
```diff
@@ -13,115 +12,65 @@ <h2>{% translate "Report name" %}</h2>
...
-                                               type="text"
-                                               value="{{ report_name }}"
-                                               required>
-                                    </div>
-                                    <br />
-                                    <div>
-                                        <label for="reference-date">{% translate "Add reference date" %}</label>
-                                        <select class="reference-date" name="reference-date">
-                                            <option disabled selected value="">-- {% translate "Select option" %} --</option>
+                                                class="icon ti-arrow-back-up action-button reset-button"></button>
```
ppvg, line 43: This button needs an accessible name, preferably using an `aria-label` to shadow the icon's pseudo content:

```suggestion
                                                class="icon ti-arrow-back-up action-button reset-button" aria-label="Reset"></button>
```

Alternatively, you could move the icon to a nested aria-hidden span and add a visually hidden accessible text alongside it, i.e.:

```html
<button class="reset-button">
  <span class="icon ti-arrow-back-up" aria-hidden="true">
  </span><span class="visually-hidden">Reset</span>
</button>
```

rocky/reports/templates/partials/export_report_settings.html
```diff
@@ -13,115 +12,65 @@ <h2>{% translate "Report name" %}</h2>
...
+                {% csrf_token %}
+                <button type="submit" form="generate_report">
+                    {% include "forms/report_form_fields.html" %}
 
-    {% translate "Generate report" %}<span class="icon ti-chevron-right"></span>
-</button>
-</form>
-{% else %}
-{% include "partials/return_button.html" with btn_text="Go back" %}
+                    {% translate "Generate report" %}<span class="icon ti-chevron-right"></span>
```
ppvg, line 63: This icon should get an `aria-label` or `aria-hidden="true"` - probably the latter, as there is already accessible text on the button.

rocky/reports/views/mixins.py
TwistMeister: Could you check the update code, if this still applies? I think I addressed it properly now.

rocky/reports/views/generate_report.py
```diff
@@ -170,15 +171,19 @@ class SaveGenerateReportView(SaveGenerateReportMixin, BreadcrumbsGenerateReportV
...
-            + "?"
-            + urlencode({"report_id": report_ooi.reference})
-        )
+        report_names = request.POST.getlist("report_name")
+        reference_dates = request.POST.getlist("reference_date")
+        final_report_names = list(zip(old_report_names, self.finalise_report_names(report_names, reference_dates)))
+        report_ooi = self.save_report(final_report_names)
+
+        if "" in report_names:
+            raise SuspiciousOperation(_("Empty name should not be possible."))
```
underdarknl, line 180: Would this mean we saved the report, eventhough we found the naming suspicious?

rocky/reports/views/mixins.py
underdarknl: I think so, as the unit-test does not fail anymore. But the problem leading up to this is that we end up trying to create a named report where we don't have enough info. Your current fix leaves the name an empty-string. which is still un-useful and unwanted.

rocky/assets/js/renameReports.js
```diff
@@ -1,35 +1,46 @@
...
+onDomReady(initRenameReports);
+
+function initRenameReports() {
+  initResetButtons();
+
+  const table = document.getElementById("report-name-table");
+  if (!table) return;
+
+  table.addEventListener("click", function (event) {
+    if (!event.target.matches(".reset-button")) return;
```
ppvg, line 12: The reset buttons are now handled by this delegated event handler, right? So `initResetButtons` is no longer needed?

ppvg: Since storing the initial `value` for later use doesn't necessarily involve the reset buttons, the name might be confusing. And I wonder if it would make sense to use the [`dataset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) to store the initial value instead of a non-standard property on the element instance. Maybe something like...

```js
Array.prototype.forEach.call(table.querySelectorAll('.name-input'), function (input) {
  input.dataset.defaultValue = input.value;
});
```

...and...

```diff
-input.value = input.defaultvalue;
+input.value = input.dataset.defaultvalue;
```

?

rocky/reports/views/mixins.py
TwistMeister: True, but imo not something created in this PR and it's better this way than it was before, so I'd like to see this merged asap :)

TwistMeister: On second thought: ooi_name and report_type_name are always present in default_name, since we create default_name programmatically using both ooi_name and report_type_name. We only use it here to check against to determine if we're at the right instance in the list, to update with "updated_name". When the user didn't actually put in a new name "update_name" will be the same as "default_name" and therefor "name_to_save" will never be empty in practise. Is it neat? Not really. Will it break? I don't believe it will.

rocky/assets/js/renameReports.js
TwistMeister: Thanks for your effort to make this as slick as it is now :)

PR #3342, opened 2024-08-12
Fix async code calling blocking sync code

dekkers:
### Changes

This fixes async functions calling blocking sync code. After spotting this in a code review, I checked and saw that we already did this incorrectly in a few other places. Async code should not call blocking sync code because that will block the whole async event loop.

### QA notes

It is hard to verify that it was not correct before, because the only effect is that the code would block in a place where it shouldn't.

---

---

## Checklist for QA:


octopoes/octopoes/api/router.py
```diff
@@ -568,18 +568,18 @@ def importer(data: bytes, xtdb_session_: XTDBSession, reset: bool = False) -> di
...
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail="Error receiving objects") from e
     return importer(data, xtdb_session_)
 
 
 @router.post("/io/import/new", tags=["io"])
-async def importer_new(request: Request, xtdb_session_: XTDBSession = Depends(xtdb_session)) -> dict[str, int]:
+def importer_new(request: Request, xtdb_session_: XTDBSession = Depends(xtdb_session)) -> dict[str, int]:
     try:
-        data = await request.body()
+        data = request.body()
```
Donnype, line 582: Here we do call async code though?

bytes/bytes/api/router.py
```diff
@@ -161,7 +161,7 @@ async def create_raw(
         if meta_repository.has_raw(meta, parsed_mime_types):
             return RawResponse(status="success", message="Raw data already present")
 
-        data = await request.body()
+        data = request.body()
```
Donnype, line 164: Same here?

boefjes/boefjes/api.py
```diff
@@ -102,7 +102,7 @@ async def boefje_input(
 
 
 @app.post("/api/v0/tasks/{task_id}")
-async def boefje_output(
+def boefje_output(
```
Donnype, line 105: :+1:

octopoes/octopoes/api/router.py
dekkers: Yes, so I need to fix it differently. I've added `sync_to_async` decorator to the importer so we can await the importer. Even if you call async code, you can't just mix sync and async code.

bytes/bytes/api/router.py
dekkers: I have wrapped this with `async_to_sync` because all the other code in the function also calls sync code.

stephanie0x00:
## Checklist for QA:

### What works:
Looks good. Onboarding works, scanning works, findings are created.

### What doesn't work:
n/a

### Bug or feature?:
n/a

underdarknl (approved):

PR #3353, opened 2024-08-13
Add create schedule functionality to scheduler api

jpbruinsslot:
### Context

Adds the functionality of creating a schedule for an arbitrary task through the scheduler API. This functionality is needed for the upcoming recurring report tasks. It is necessary to specify at what interval this task needs to be executed for an individual report task. A schedule for a task can be created as follows:

```javascript
POST /schedules
(
    "scheduler_id": "id-of-the-scheduler",
    "schedule": "cron-expression",
    "data": {"place": "your", "task": "here"}
}
```

### Changes

* `models/schedule.py` 
  * ~~computed field for deadline; we want to be able to calculate the deadline for the schedule when we create a schedule from the create endpoint. This allows the API client just to supply the cron expression and not do the deadline calculation itself.~~ Update: a post init to calculate the `deadline_at` when it isn't supplied and a schedule is present did the job better with less code.
* `server/`
  * `schedules.py` added create endpoint for schedules, adds validation of `data` by referencing the scheduler with `scheduler_id`, it also generates the `hash` of the contents of the `data` field so this doesn't have to be done by the client.
  * `serializers/schedule.py` adding `ScheduleCreate` model to correctly serialize/validate the client request. Take note that issues are already created to tackle the server API like: https://github.com/minvws/nl-kat-coordination/issues/3162

### QA notes

We're going to create a `Schedule` through the newly created `/schedules` endpoint:

* `make reset`
* Fire up openkat
*  Create an organisation called `dev`. Don't start any scans though!
* `curl -X POST -H "Content-Type: application/json" -d '{"scheduler_id": "boefje-dev", "schedule": "*/5 * * * *", "data": {"id": "12c3f698-5036-40ce-a1d5-1d59e50116d8", "boefje": { "id": "dns-records", "name": "DnsRecords", "version": null }, "input_ooi": "Hostname|internet|mispo.es", "dispatches": [], "organization": "dev" }}' http://localhost:8004/schedules
`
* `curl HTTP://localhost:8004/schedules` it should now list a schedule

---

### Code Checklist

---

---

## Checklist for QA:


mula/scheduler/models/schedule.py
```diff
@@ -30,14 +30,41 @@ class Schedule(BaseModel):
...
     created_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
     modified_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
 
+    def __init__(self, **data):
+        super().__init__(**data)
+
+        if "deadline_at" in data:
+            self.deadline_at = data["deadline_at"]
+        elif "deadline_at" not in data and self.schedule is not None:
+            self.deadline_at = cron.next_run(self.schedule)
```
ammar92, line 43: This seems more like a [model_post_init](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_post_init) functionality, since there's no validation here anymore and the object is already initialized. Consider using that, since you rarely need to use `__init__` here nowadays

mula/scheduler/models/schedule.py
```diff
@@ -47,6 +74,16 @@ def validate_schedule(cls, value: str) -> str:
         except Exception as exc:
             raise ValueError(f"Invalid cron expression: {value}") from exc
 
+    @classmethod
+    def model_validate(cls, data):
```
ammar92, line 78: See earlier remark about model_post_init, which could make this block unnecessary

mula/scheduler/server/handlers/schedules.py
```diff
@@ -5,26 +5,42 @@
...
+    def __init__(
+        self,
+        api: fastapi.FastAPI,
+        ctx: context.AppContext,
+        s: dict[str, schedulers.Scheduler],
+    ) -> None:
         self.logger: structlog.BoundLogger = structlog.getLogger(__name__)
         self.api = api
         self.ctx = ctx
+        self.schedulers: dict[str, schedulers.Scheduler] = s
```
ammar92, line 24:
```suggestion
        schedulers: dict[str, schedulers.Scheduler],
    ) -> None:
        self.logger: structlog.BoundLogger = structlog.get_logger(__name__)
        self.api = api
        self.ctx = ctx
        self.schedulers = schedulers
```

mula/scheduler/server/handlers/schedules.py
```diff
@@ -98,6 +114,75 @@ def list(
...
+        s = self.schedulers.get(new_schedule.scheduler_id)
+        if s is None:
+            raise fastapi.HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail="scheduler not found",
+            )
+
+        # Validate data with task type of the scheduler
+        try:
+            instance = s.ITEM_TYPE.parse_obj(new_schedule.data)
```
ammar92, line 140: Use `model_validate`

ammar92 (commented):
Nice work. Just a few remarks that might improve it even more

mula/scheduler/models/schedule.py
jpbruinsslot: Using your suggestion, was able to create a simpler implementation 🙏

underdarknl (approved):

Issue #3373, opened 2024-08-15
POST redirects when all required plugins are enabled


Rieven:
The flow currently jumps to the plugins selection of the report flow. This feels a little confusing as I cannot enable any plugins. Are there scenario's where you could actually enable plugins for the findings report? Otherwise I think it is more logical from the user experience view to go directly to the 'naming' page for the report. 
 
A screenshot below to show what it currently looks like: 

![image](https://github.com/user-attachments/assets/d5e1b4bc-4921-40de-b8ed-9ad97d3536fb)

_Originally posted by @stephanie0x00 in https://github.com/minvws/nl-kat-coordination/issues/3369#issuecomment-2291274154_

PR #3380, opened 2024-08-15
Redirect to desired view when all plugins are enabled.

Rieven:
### Changes
At the report step where the user enables plugins, when all plugins are enabled we want to redirect the POST to another view.
- An important point to address here is that when the user wants to change their selection and needing to go back to the plugins page, even here if plugins are enabled it needs to go further back to the report type selection page.

### Issue link
https://github.com/minvws/nl-kat-coordination/issues/3373
https://github.com/minvws/nl-kat-coordination/issues/3435

Closes https://github.com/minvws/nl-kat-coordination/issues/3373
Closes https://github.com/minvws/nl-kat-coordination/issues/3435

### Code Checklist

---

---

## Checklist for QA:


rocky/reports/views/aggregate_report.py
```diff
@@ -116,6 +116,9 @@ def setup(self, request, *args, **kwargs):
         )
 
     def post(self, request, *args, **kwargs):
+        if not hasattr(ReportTypesSelectionAggregateReportView, "selected_oois"):
+            self.setup(request, *args, **kwargs)
+
```
ammar92, line 121: This is becoming spaghetti here. `setup` is called by automatically by the framework and always before the `dispatch` method (which in turn calls `post`, `get` or other HTTP methods), so there shouldn't be a call to `setup` here. The `setup` method is responsible for initializing some attributes, including `self.args` and `self.kwargs`. By calling `setup` manually in the `post` method, you're potentially re-initializing or overwriting attributes that the framework has already set up, which could lead to unintended behavior. Since Django guarantees that setup is called before the post method, there's no need to call it again.

---

Also, for a next usage of `hasattr`, the correct way to use `hasattr` in this context is to use: `hasattr(self, '<attribute>')`. Otherwise, what you're doing is checking for a class-level attribute, which is not the intended use and will probably not work, since `selected_oois` hasn't been set in this class, or there were no assignments such as `ReportTypesSelectionAggregateReportView.selected_oois = ...`

rocky/reports/views/aggregate_report.py
```diff
@@ -167,6 +175,14 @@ class ExportSetupAggregateReportView(AggregateReportStepsMixin, BreadcrumbsAggre
     current_step = 4
 
     def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        # POST can come from a redirected POST which needs to initiate setup to fetch variables
+        if not hasattr(ExportSetupAggregateReportView, "oois_pk") and not hasattr(
+            ExportSetupAggregateReportView, "report_types"
+        ):
+            self.setup(request, *args, **kwargs)
```
ammar92, line 182: See earlier remark

rocky/reports/views/generate_report.py
```diff
@@ -105,6 +105,9 @@ class ReportTypesSelectionGenerateReportView(
     current_step = 2
 
     def post(self, request, *args, **kwargs):
+        if not hasattr(ReportTypesSelectionGenerateReportView, "selected_oois"):
+            self.setup(request, *args, **kwargs)
```
ammar92, line 109: See earlier remark

rocky/reports/views/generate_report.py
```diff
@@ -132,6 +135,12 @@ class SetupScanGenerateReportView(
     current_step = 3
 
     def post(self, request, *args, **kwargs):
+        if "return" in self.request.POST and self.plugins_enabled():
+            return ReportTypesSelectionGenerateReportView().post(request, *args, **kwargs)
```
ammar92, line 139: I've rarely seen views being initialized in another view. This also skips the view execution flow the framework (setup, dispatch, render). Is this the desired behavior?

rocky/reports/views/generate_report.py
```diff
@@ -149,6 +158,13 @@ class ExportSetupGenerateReportView(GenerateReportStepsMixin, BreadcrumbsGenerat
     reports: dict[str, str] = {}
 
     def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        if not hasattr(ExportSetupGenerateReportView, "oois_pk") and not hasattr(
+            ExportSetupGenerateReportView, "report_types"
+        ):
+            self.setup(request, *args, **kwargs)
```
ammar92, line 164: See earlier remark

stephanie0x00:
## Checklist for QA:

### What works:
The general flow seems to work. 

### What doesn't work:
Steps to reproduce:
- Go through the onboarding for mispo.es. 
- Select all reports in the report flow
- On the plugins page return to the previous page (report selection0. 
- Change from all reports to only 1 report. 

This results in the following behaviour for the normal + aggregate report: 
- The toggle needs to be pressed twice before the checkboxes disappear. 
- When I select 1 report, eg. DNS report, and continue to the plugins page, you will see the plugins page for all reports. The expected behaviour would be to only view the plugins for the selected DNS report and/or skip this page if the requirements are met. 

For the normal + aggregate report the following steps are required to reproduce: 
- Select the Hostname for mispoes. 
- Select a report in the report flow for which you do not have boefjes enabled, such as RPKI, Vulnerability report etc. 
- Enable the required plugins for this report 

This results in the following behaviour: 
- Observe that the 'Configuration' page now suddenly is empty. 
- When you click on the 'Go Back' button (there is not button to continue) you have lost your selection and have to start the flow again. 

### Bug or feature?:
See above.

ammar92 (commented):
Tiny remarks. Can be tested already

stephanie0x00:
The bugs in the earlier post are indeed on main, thus can be merged.

rocky/tools/view_helpers.py
```diff
@@ -163,3 +164,11 @@ def build_breadcrumbs(self):
...
             }
         ]
+
+
+class PostRedirect(HttpResponse):
+    status_code = 307
+
+    def __init__(self, redirect_to, *args, **kwargs):
+        super().__init__(content=b"", *args, **kwargs)
+        self["Location"] = iri_to_uri(redirect_to)
```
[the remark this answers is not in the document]
Rieven: Great this is even better!

dekkers (approved):

PR #3381, opened 2024-08-16
burpsuite fix

### LINKED ISSUE #3050: Burpsuite normalizer cannot parse xml file
**Describe the bug**
A community member ran into the issue that Burpsuite XML files are not parsed properly by the burpsuite-normalizer. 

Uploading a raw file with the mime-type `xml/burp-export` works, however when the raw file is being parsed by the scheduler an error is triggered. 

Reproducing this issue also showed that it's currently unclear how a user can create the XML file (standard export format is a binary .burp file). Thus an update on the documentation would also be required to make this more clear. 

**To Reproduce**

1. In BurpSuitePro create an XML file by right-click on one or more hosts (e.g. Targets page) and then click 'Issues > Report issues for this host'. Choose the XML export and continue with the wizard. 
2. In Openkat click on Objects. 
3. Select Upload raw file and use the mime-type 'xml/burp-export'.
4. Observe the error in the logs from the normalizer. 

**Expected behavior**
Under Object you upload a raw file with the `xml/burp-export` mime-type. The raw file is uploaded and should automatically be parsed by the scheduler. 

**Screenshots**
The error message: 

image

**OpenKAT version**
commit 8b29e48f229d872174dd56b7b9e4e007f490e393 (HEAD -> main, origin/main, origin/HEAD)

underdarknl:
### Changes

Fixes the normalizer for burpsuite, it crashed on the non decoded raw file earlier.

### Issue link

closes #3050

### QA notes

Importing should (in combination with the other fixes) now work

### Code Checklist

---

---

## Checklist for QA:


ammar92 (dismissed):
No remarks, but some additional info: Looking at the bug and implementation, it wasn't actually an encoding error, but the code failed due wrong usage of `parse`. `parse` expects a file-like object or path, so if `raw` was wrapped in a `BytesIO` object for example, it would have worked. `parseString` basically does the same, but wraps the string in a `StringIO` object and passes it to `parse`.

stephanie0x00:
This PR depends on the fixing of two bugs. Thus moving this ticket to 'Blocked' for now. 
- When you create an ExternalScan object, you have to wait a minute before you can select this object in the 'Raw file upload' interface, since the ExternalScan object is otherwise not recognised. Waiting a minute will let you pass this bug, but it should be fixed as it is confusing to the end user. 
- The second bug is that the raw file upload times out on bytes when the timestamp is actually valid. This prevents you from uploading a file altogether. 

I've already discussed both bugs with @noamblitz. 

image

noamblitz:
It works now when I solve bytes problem in this issue

https://github.com/minvws/nl-kat-coordination/issues/3488

boefjes/boefjes/plugins/kat_burpsuite/normalize.py
```diff
@@ -16,15 +15,9 @@
...
-    # assume that input ooi is none or a HostnameHTTPURL
-    if input_ooi:
-        ooi = Reference.from_str(input_ooi["primary_key"])
-        network = Network(name=ooi.tokenized.netloc.network.name)
-    else:
-        network = Network(name="internet")
-        yield network
+    parser = minidom.parseString(raw.decode("UTF-8"))
+
+    network = Network(name="internet")
```
underdarknl, line 20: Burpsuite seems to be one of those tools that would be used on the internal networks a lot.

noamblitz: I can make support for either scan oois -> internet or hostnames -> hostname.network

stephanie0x00:
Whooo!! It now works again \o/ we still have the magical 1 minute wait bug, but that is an easy workaround. There is now a bug in the normalizer: #3537

noamblitz:
For QA: now any OOI that has a network somewhere in its tree can be used as input and that network will be used.

ammar92 (approved):

stephanie0x00:
> For QA: now any OOI that has a network somewhere in its tree can be used as input and that network will be used.

Seems to work. Was able to directly upload it under the mispo.es Hostname and the objects get added to the Objects list. :+1:

noamblitz:
> > For QA: now any OOI that has a network somewhere in its tree can be used as input and that network will be used.
> 
> Seems to work. Was able to directly upload it under the mispo.es Hostname and the objects get added to the Objects list. 👍

Could you try it with another network then internet?

dekkers (approved):

Issue #3356, opened 2024-08-13
Improve env/settings/allowlist configuration logic for the boefjes


Donnype:
TODO:
- [x] Make adhering to the schema to be enabled optional in the API
- [x] Use the schema keys instead of environment keys in the runner, override with katalogus entries instead of vice versa
- [x] Phase out environment keys in the `boefje.json` files in favor of the `schema.json` keys, as these are now redundant.
~- [ ] Update schema/api checks in Rocky if needed~

part of https://github.com/minvws/nl-kat-coordination/issues/3052.

underdarknl:
> Make adhering to the schema to be enabled optional in the API

N.b. this is needed in situations where the runner injects global env-vars, and as such the fields in the boefje config themselves can be left empty. Instead of bringing the global vars to the validator, we have two options:
Ignore the validator, or ignore the required fields. We opted for the first for now

Issue #1906, opened 2023-10-12
Rename `BOEFJE_` environment prefix to `SYSTEM_`


Darwinkel:
Requested by @ammar92. 

It's currently a bit confusing to use `BOEFJE_` as a prefix to pass variables on to Boefjes, considering:

- We now also have `BOEFJE_API` (https://github.com/minvws/nl-kat-coordination/pull/1811, https://github.com/minvws/nl-kat-coordination/pull/1868) which should not actually be passed on to boefjes;
- We may also want to pass on variables like this to normalizers.

It would probably be better to rename it to `SYSTEM_` (`HOST_` and `GLOBAL_` could conflict with existing conventions). In addition, [the documentation](https://docs.openkat.nl/developer_documentation/boefjes.html#environment-variables) could be made more explicit regarding how to pass a host system variable to a boefje and the exact transformations that happen. For example, that variable interpolation is required to get that to work.

(Original PR for context: https://github.com/minvws/nl-kat-coordination/pull/1083)

@ammar92 also suggests that we could/should separate plugin-scoped and global-scoped settings within the Boefje runner.

underdarknl:
this is a fair point. We should consider making sure we select a clean namespace for our 'safe' env vars.

PR #3384, opened 2024-08-20
Feature: improve settings and environment logic and phase out redundant environment keys

Donnype:
### Changes
This improves the environment logic to pass all BOEFJES_* environment variables to the boefjes and allows you to override them with the KATalogus settings. This makes the flow more simple and intuitive.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/3356

### Demo

-

### QA notes

Please verify the Katalogus and boefjes such as `dns-records` still work properly, creating settings and review the documentation update.

---

### Code Checklist

---

---

## Checklist for QA:


boefjes/boefjes/dependencies/plugins.py
```diff
@@ -150,12 +150,7 @@ def _put_normalizer(self, normalizer_id: str) -> None:
     def delete_settings(self, organisation_id: str, plugin_id: str):
         self.config_storage.delete(organisation_id, plugin_id)
 
-        try:
-            self._assert_settings_match_schema({}, organisation_id, plugin_id)
-        except SettingsNotConformingToSchema:
-            logger.warning("Making sure %s is disabled for %s because settings are deleted", plugin_id, organisation_id)
-
-            self.set_enabled_by_id(plugin_id, organisation_id, False)
+        # We don't check the schema anymore because we can provide entries through the global environment as well
```
underdarknl, line 153: I think It would  be better if we 'optionally' skip this validation. That way we can do validation by default, but offer the user to skip validation if they want to do so by re-posting the form, with an extra argument.

Donnype: Shouldn't we just handle this in Rocky instead of requiring a call towards the KATalogus while we already know that the schema doesn't match?

boefjes/boefjes/job_handler.py
```diff
@@ -35,30 +38,40 @@ def get_octopoes_api_connector(org_code: str) -> OctopoesAPIConnector:
...
-                if katalogus_key in environment_keys and katalogus_key not in environment:
-                    environment[katalogus_key] = value
-
-        return {k: str(v) for k, v in environment.items() if k in environment_keys}
     except HTTPError:
         logger.exception("Error getting environment settings")
         raise
 
+    settings_from_katalogus = response.json()
+    new_env = {}
```
ammar92, line 54: You could also copy `settings_from_katalogus` here like `new_env = settings_from_katalogus.copy()`, unless the order of the loops below is of importance

ammar92 (dismissed):
Looks good to me. Btw; this is very much related to https://github.com/minvws/nl-kat-coordination/issues/1906, any possibility we could discuss this one soon?

boefjes/boefjes/dependencies/plugins.py
underdarknl: Oh, yes. I did not notice this was the kat-alogus.

boefjes/boefjes/job_handler.py
```diff
@@ -35,30 +38,40 @@ def get_octopoes_api_connector(org_code: str) -> OctopoesAPIConnector:
...
+
+    for key, value in os.environ.items():
+        if key.startswith("BOEFJE_"):
```
underdarknl, line 57: I think it would be better if we limit the passed on env vars to only those listed in the settings json. As boefje A does not need Boefje B's api-keys.

boefjes/boefjes/job_handler.py
Donnype: The order indeed matters in terms of which config overwrites the other, but I added a dict comprehension

ammar92 (approved):
Looks good to me.

For the next iteration, I suggest we start using pytest markers to distinguish between different test types (e.g., unit, integration, and performance tests). This approach will help us avoid using code/environment guards and streamline the CI process

stephanie0x00:
## Checklist for QA:

### What works:
Functionally this PR seems to work as expected. I enabled some general boefjes (onboarding + nmap ports) and they run. I also enabled LeaxIX with an API key and this runs as well, including the normalizers. 

### What doesn't work:
The docs have some room for improvement on how a user can work with the environment keys. Based on the initial ticket I would say you have to do the following: 

Add `BOEFJES_MYNEWBOEFJE=yourkey` to the .env file, then in the schema.json file for MYNEWBOEFJE you would add the BOEFJES_MYNEWBOEFJE text and it should work? I think it would be very useful to have an example + description like this added, such that it is initially clear what/how it works. 

### Bug or feature?:
n/a

Donnype:
@stephanie0x00 Done :)

underdarknl (approved):

PR #3389, opened 2024-08-20
Generic tasks view refactor

zcrt:
Rewrite of https://github.com/minvws/nl-kat-coordination/pull/3005


rocky/rocky/settings.py
```diff
@@ -78,7 +78,7 @@
     "loggers": {
         "root": {
             "handlers": ["console"],
-            "level": "INFO",
+            "level": env("LOG_LEVEL", default="WARNING").upper(),
```
underdarknl, line 81: This increases the log level default from info to warning, for the entire Rocky app.

underdarknl, line 81:
```suggestion
            "level": env("LOG_LEVEL", default="INFO").upper(),
```

rocky/rocky/templates/tasks/boefjes.html
```diff
@@ -56,7 +66,7 @@ <h1>{% translate "Boefjes" %}</h1>
                                         </td>
                                     </tr>
                                     <tr class="expando-row">
-                                        <td colspan="5">
+                                        <td colspan="7">
```
underdarknl, line 69:
```suggestion
                                        <td colspan="{% if not organization.code %}7{% else %}5{% endif %}">
```

rocky/assets/js/renderNormalizerOutputOOIs.js
```diff
@@ -11,11 +11,14 @@ buttons.forEach((button) => {
     .closest("tr")
     .getAttribute("data-task-id")
     .replace(/-/g, "");
+  const organization =
+    organization_code ||
+    button.closest("tr").getAttribute("data-organization-code");
```
ppvg, line 16: The "closest tr" is selected for `raw_task_is` and `task_id` as well, so you could consider reusing it. You could also consider hardening against there not being a `<tr>`. E.g.:

```js
const tr = button.closest("tr");
if (!tr) return;
const raw_task_id = tr.closest("tr").getAttribute("data-task-id");
const task_id = raw_task_id.replace(/-/g, "");
const organization = organization_code || tr.getAttribute("data-organization-code");
```

rocky/rocky/templates/tasks/boefjes.html
```diff
@@ -56,7 +66,7 @@ <h1>{% translate "Boefjes" %}</h1>
                                         </td>
                                     </tr>
                                     <tr class="expando-row">
-                                        <td colspan="5">
+                                        <td colspan="{% if not organization.code %}7{% else %}5{% endif %}">
```
ppvg, line 69: The "modified at" column is new and not conditional, right? So this should be 6 instead of 5?
```suggestion
                                        <td colspan="{% if not organization.code %}7{% else %}6{% endif %}">
```

rocky/rocky/templates/tasks/partials/task_actions.html
```diff
@@ -13,18 +13,25 @@ <h2>{% translate "Yielded objects" %}</h2>
         {% if task.status.value in "completed,failed" %}
             {% if task.type == "normalizer" %}
                 <a class="button"
-                   href="{% url 'bytes_raw' organization_code=organization.code boefje_meta_id=task.data.raw_data.boefje_meta.id %}"><span class="icon ti-download"></span>{% translate "Download meta and raw data" %}</a>
-                {% include "partials/single_action_form.html" with btn_text=_("Reschedule") btn_class="ghost" btn_icon="icon ti-refresh" action="reschedule_task" key="task_id" value=task.id %}
+                   href="{% url 'bytes_raw' organization_code=task.data.raw_data.boefje_meta.organization boefje_meta_id=task.data.raw_data.boefje_meta.id %}"><span class="icon ti-download"></span>{% translate "Download meta and raw data" %}</a>
```
ppvg, line 16: Not new in this PR, but do these icons use an icon font and pseudo-content? If so, they should either get an `aria-label` or `aria-hidden="true"` - most likely the latter since the links already have accessible text.

zcrt: As this is not new in the PR, I would like to suggest raising a separate issue for this. I also think this may be out of scope for a minimum viable product

rocky/assets/js/renderNormalizerOutputOOIs.js
zcrt: @underdarknl can you comment on this? It does not seem a blocking issue to me, especially since this issue was already present

underdarknl: @ppvg true, we will fix this in a new ticket, as @zcrt mentions, it's not a new regression, but they are useful points to improve on.
@TwistMeister could you pick this up?

rocky/rocky/scheduler.py
```diff
@@ -223,31 +225,28 @@ def list_tasks(
...
             raise SchedulerValidationError(extra_message=_("Task list: "))
         except ConnectError:
             raise SchedulerConnectError(extra_message=_("Task list: "))
 
     def get_task_details(self, task_id: str) -> Task:
-        try:
-            res = self._client.get(f"/tasks/{task_id}")
-            res.raise_for_status()
-            task_details = Task.model_validate_json(res.content)
+        task_details = Task.model_validate_json(self._get(f"/tasks/{task_id}", "content"))
```
underdarknl, line 239: Seeing that we will check *if* the organization of the returned job is the same as the now optionally set self.organization we should probably just don't do anything when there is not self.organization defined.

rocky/rocky/scheduler.py
```diff
@@ -223,31 +225,28 @@ def list_tasks(
...
+            params = {k: v for k, v in kwargs.items() if v is not None if k != filter_key}  # filter Nones from kwargs
+            endpoint = "/tasks"
+            res = self._client.post(endpoint, params=params, json=kwargs.get(filter_key, None))
             return PaginatedTasksResponse.model_validate_json(res.content)
         except ValidationError:
             raise SchedulerValidationError(extra_message=_("Task list: "))
         except ConnectError:
             raise SchedulerConnectError(extra_message=_("Task list: "))
 
     def get_task_details(self, task_id: str) -> Task:
```
underdarknl, line 238:
```suggestion
    def get_task_details(self, task_id: str) -> Task:
    if not self.organization_code:
        raise SchedulerTaskNotFound("No organization defined in client.")
```

zcrt:
```suggestion
    def get_task_details(self, task_id: str) -> Task:
        if self.organization_code is None:
            raise SchedulerTaskNotFound("No organization defined in client.")
```

rocky/rocky/templates/tasks/partials/task_actions.html
HeleenSG: Ticket: https://github.com/minvws/nl-kat-coordination/issues/3404

underdarknl (approved):

PR #3391, opened 2024-08-21
Introduce a-sync recalculate bits with status polling

originalsouth:
### Changes

When recalculate bits takes longer than a timeout in the the current setup, a timeout is achieved and the result is seen in the front-end.

This patch fixes that by launching recalculate_bits a-sync (with a session_id) and then providing a a-sync retrieve result function that indicates either if the jobs exists, is running or is done.

This way the front-end can await the recalculation without timing out and retrieving the results.

### Issue link

N/A

### Demo

Create a large system where calculations take sufficiently long (and special thanks to @Rieven).

Launch a recalculation:
```
curl -s -X POST "http://localhost:8001/_rieven/bits/recalculate"

{"session_id":"e197a7f2-c72e-4265-ae57-30dbdb1e4d8c"}
```
Check the sessions with:
```
curl -s -X POST "http://localhost:8001/_rieven/bits/recalculate/status"

{"sessions":["e197a7f2-c72e-4265-ae57-30dbdb1e4d8c"]}
```
Check session `"e197a7f2-c72e-4265-ae57-30dbdb1e4d8c"` with:
```
curl -s -X POST "http://localhost:8001/_rieven/bits/recalculate/status" -d "e197a7f2-c72e-4265-ae57-30dbdb1e4d8c"

{"status":"awaiting"}
```

Some time later `"e197a7f2-c72e-4265-ae57-30dbdb1e4d8c"` with:
```
curl -s -X POST "http://localhost:8001/_rieven/bits/recalculate/status" -d "e197a7f2-c72e-4265-ae57-30dbdb1e4d8c"

{"status":5240}
```

Sessions are empty:
```
 curl -s -X POST "http://localhost:8001/_rieven/bits/recalculate/status"
 
{"sessions":[]}
```

And hence `"e197a7f2-c72e-4265-ae57-30dbdb1e4d8c"`:
```
curl -s -X POST "http://localhost:8001/_rieven/bits/recalculate/status" -d "e197a7f2-c72e-4265-ae57-30dbdb1e4d8c"

{"status":"unavailable"}
```

### QA notes

Run the demo!
The current recalculate button in the front-end is broken (because that is not done in this patch).

---

### Code Checklist

---

---

## Checklist for QA:

### DIFF
--- a/octopoes/octopoes/api/router.py
+++ b/octopoes/octopoes/api/router.py
@@ -527,6 +527,8 @@ async def recalculate_bits_launcher(octopoes: OctopoesService, session_id: uuid.
 
     octopoes.commit()
 
+    await asyncio.sleep(60)
+
     RECALCULATE_BITS_SESSIONS[session_id] = inference_count
 
 
@@ -547,15 +549,18 @@ async def recalculate_bits_status(request: Request) -> Any:
     except Exception as e:
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail="Error receiving objects") from e
     session_id = uuid.UUID(data.decode())
-    if session_id in RECALCULATE_BITS_SESSIONS:
-        retval = RECALCULATE_BITS_SESSIONS[session_id]
-        if retval > 0:
-            RECALCULATE_BITS_SESSIONS.pop(session_id)
-            return {"status": retval}
+    if session_id:
+        if session_id in RECALCULATE_BITS_SESSIONS:
+            retval = RECALCULATE_BITS_SESSIONS[session_id]
+            if retval > 0:
+                RECALCULATE_BITS_SESSIONS.pop(session_id)
+                return {"status": retval}
+            else:
+                return {"status": "awaiting"}
         else:
-            return {"status": "awaiting"}
+            return {"status": "unavailable"}
     else:
-        return {"status": "unavailable"}
+        return RECALCULATE_BITS_SESSIONS
 
 
 @router.get("/io/export", tags=["io"])
--- a/octopoes/octopoes/api/router.py
+++ b/octopoes/octopoes/api/router.py
@@ -527,8 +527,6 @@ async def recalculate_bits_launcher(octopoes: OctopoesService, session_id: uuid.
 
     octopoes.commit()
 
-    await asyncio.sleep(60)
-
     RECALCULATE_BITS_SESSIONS[session_id] = inference_count
 
 
--- a/octopoes/octopoes/api/router.py
+++ b/octopoes/octopoes/api/router.py
@@ -546,8 +546,8 @@ async def recalculate_bits_status(request: Request) -> Any:
         data = await request.body()
     except Exception as e:
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail="Error receiving objects") from e
-    session_id = uuid.UUID(data.decode())
-    if session_id:
+    if data:
+        session_id = uuid.UUID(data.decode())
         if session_id in RECALCULATE_BITS_SESSIONS:
             retval = RECALCULATE_BITS_SESSIONS[session_id]
             if retval > 0:
@@ -558,7 +558,7 @@ async def recalculate_bits_status(request: Request) -> Any:
         else:
             return {"status": "unavailable"}
     else:
-        return RECALCULATE_BITS_SESSIONS
+        return {"sessions": list(RECALCULATE_BITS_SESSIONS.keys())}
 
 
 @router.get("/io/export", tags=["io"])
--- a/octopoes/octopoes/api/router.py
+++ b/octopoes/octopoes/api/router.py
@@ -10,6 +10,7 @@ from typing import Any
 
 from asgiref.sync import sync_to_async
 from fastapi import APIRouter, Body, Depends, HTTPException, Path, Query, Request, status
+from fastapi.responses import JSONResponse
 from httpx import HTTPError
 from pydantic import AwareDatetime
 
@@ -556,7 +557,7 @@ async def recalculate_bits_status(request: Request) -> Any:
             else:
                 return {"status": "awaiting"}
         else:
-            return {"status": "unavailable"}
+            return JSONResponse(status_code=status.HTTP_404_NOT_FOUND, content={"status": "session not found"})
     else:
         return {"sessions": list(RECALCULATE_BITS_SESSIONS.keys())}
 
--- a/octopoes/octopoes/api/router.py
+++ b/octopoes/octopoes/api/router.py
@@ -1,3 +1,4 @@
+import asyncio
 import json
 import uuid
 from collections import Counter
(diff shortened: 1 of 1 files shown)


octopoes/octopoes/api/router.py
```diff
@@ -524,7 +527,38 @@ def recalculate_bits(octopoes: OctopoesService = Depends(octopoes_service)) -> i
...
+async def recalculate_bits(octopoes: OctopoesService = Depends(octopoes_service)) -> Any:
+    session_id = uuid.uuid4()
+    RECALCULATE_BITS_SESSIONS[session_id] = -1
+
+    asyncio.create_task(recalculate_bits_launcher(octopoes, session_id))
+
+    return {"session_id": session_id}
+
+
+@router.post("/bits/recalculate/status", tags=["Bits"])
```
Donnype, line 543: Perhaps a GET request makes more sense?

octopoes/octopoes/api/router.py
```diff
@@ -524,7 +527,38 @@ def recalculate_bits(octopoes: OctopoesService = Depends(octopoes_service)) -> i
...
+async def recalculate_bits_status(request: Request) -> Any:
+    try:
+        data = await request.body()
+    except Exception as e:
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail="Error receiving objects") from e
+    if data:
+        session_id = uuid.UUID(data.decode())
```
Donnype, line 550: Would be nice in a url parameter if we use a GET so that we can just add it to the arguments with a proper UUID type and let FastAPI handle the rest.

octopoes/octopoes/api/router.py
```diff
@@ -524,7 +527,38 @@ def recalculate_bits(octopoes: OctopoesService = Depends(octopoes_service)) -> i
...
+        if session_id in RECALCULATE_BITS_SESSIONS:
+            retval = RECALCULATE_BITS_SESSIONS[session_id]
+            if retval > 0:
+                RECALCULATE_BITS_SESSIONS.pop(session_id)
+                return {"status": retval}
+            else:
+                return {"status": "awaiting"}
+        else:
+            return {"status": "unavailable"}
```
Donnype, line 559: Some suggestions: 
```suggestion
                return {"status": "running"}
        else:
            return JSONResponse(status_code=status.HTTP_404_NOT_FOUND, content={"status": "session not found"})
```

octopoes/octopoes/api/router.py
```diff
@@ -524,7 +527,38 @@ def recalculate_bits(octopoes: OctopoesService = Depends(octopoes_service)) -> i
...
+    else:
+        return {"sessions": list(RECALCULATE_BITS_SESSIONS.keys())}
```
Donnype, line 561: This makes this endpoint both, in "Django-rest-framework-terms", a list() and get() at the same time. I'd say we should separate this into two endpoints (if we need actually need a list) and add the session_id perhaps as a url path parameter even for the "get" version.

octopoes/octopoes/api/router.py
```diff
@@ -514,8 +515,10 @@ def delete_node(xtdb_session_: XTDBSession = Depends(xtdb_session)) -> None:
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail="Deleting node failed") from e
 
 
-@router.post("/bits/recalculate", tags=["Bits"])
-def recalculate_bits(octopoes: OctopoesService = Depends(octopoes_service)) -> int:
+RECALCULATE_BITS_SESSIONS: dict[uuid.UUID, int] = {}
```
Donnype, line 518: Although simple, there is an issue with this approach: what if clients configure `uvicorn` to run with more than 1 worker, each worker has its own state. Hence we wouldn't have read-your-writes consistency, and you randomly see or don't see the sessions you created, depending on which Uvicorn worker handles the request.

Donnype (commented):
Sounds like a good fix for this endpoint being slow. But there are some caveats with the global session dict approach, see my comment! :)

octopoes/octopoes/api/router.py
originalsouth: Right the other approach would be to dump the data in XTDB, @underdarknl and I talked about that scenario. 

1. Do clients do this?
1. Does a user "switch" worker on every request?

octopoes/octopoes/api/router.py
originalsouth: I opted for a POST for two reasons:
1. To be consistent with the current recalculate bits implementation
1. So that we can pass parameters in the body, not in the URL

I'm cool either way, but this implementation works and I do not seem  the harm in it.

octopoes/octopoes/api/router.py
originalsouth: See comment above.

octopoes/octopoes/api/router.py
originalsouth: See 3fddf135fe209c63d2989737e2740733417e7b55

octopoes/octopoes/api/router.py
Donnype: I did a quick check using this code:

```python3
import os
import uvicorn
from contextlib import asynccontextmanager

from fastapi import FastAPI

SESSIONS = []

@asynccontextmanager
async def lifespan(app: FastAPI):
    SESSIONS.append(os.getpid())
    yield

app = FastAPI(lifespan=lifespan)

@app.get("/")
async def index():
    return {"worker_id": os.getpid(), "sessions": SESSIONS}

if __name__ == "__main__":
    uvicorn.run("main:app", host="127.0.0.1", port=8009, workers=2)
```

And get different responses quite randomly:
image

So I'd answer "yes" on the second question. Is your first question: do clients use multiple workers?

dekkers: I think they would have to if they want to scale OpenKAT, because I doubt a single octopoes process is going to be enough for a big installation. And if you would run OpenKAT on Kubernetes, you would normally also create a Deployment with multiple replicas. In both cases there is no guarantee that a subsequent request goes to the same process.

The problem is that we are trying to run a task inside a http request for which it is always better to use task runner. We already have celery, so in my opinion the right approach would be to create a celery task for recalculating bits. I can help with that if needed, I have a lot of experience with celery.

octopoes/octopoes/api/router.py
originalsouth: Right but by the same argument, we are also changing the internal state of the dictionary (and later XTDB) on success... anyway... I will yield to the GET.

octopoes/octopoes/api/router.py
originalsouth: Thanks @Donnype that proves you point quite elegantly. If @dekkers and @underdarknl agree I will rewrite this for the using XTDB as central storage -- with the all important GET. Thanks for the feedback!

originalsouth:
I think @dekkers has better ideas with regards to fixings #3392

PR #3383, opened 2024-08-19
Generic Finding normalizer

noamblitz:
### Changes

Create generic CVE normalizer that consumes a cve/id and creates the CVEFindingype and Finding OOIs

### Issue link

_You have to create an issue to link to this PR. If this really is not possible, write a very detailed description here and add this PR to the project board directly._

_Please add the link to the issue after "Closes"._

Closes ...

### Demo

_Please add some proof in the form of screenshots or screen recordings to show (off) new functionality, if there are interesting new features for end-users._

### QA notes

Upload raw file with either one findintype or a comma separated list of finding types with mime type openkat/finding

---

### Code Checklist

- [x] All the commits in this PR are properly PGP-signed and verified.
- [x] This PR only contains functionality relevant to the issue.
- [x] I have written unit tests for the changes or fixes I made.
- [x] I have checked the documentation and made changes where necessary.
- [x] I have performed a self-review of my code and refactored it to the best of my abilities.

- [ ] Tickets have been created for newly discovered issues.
- [ ] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [ ] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [ ] I have included comments in the code to elaborate on what is not self-evident from the code itself, including references to issues and discussions online, or implicit behavior of an interface.

---

## Checklist for code reviewers:

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_qa.md) into your comment.

PR #3409, opened 2024-08-27
Add Nikto boefje that scans for outdated software

Souf149:
### Changes

Added a Nikto boefje written in javascript and a normalizer that scans/reports webservers for outdated software.

This boefje also makes use of https://github.com/minvws/nl-kat-coordination/pull/3383 to create a finding without having to create a new normalizer. (which could be removed since a normalizer also gets added with this pr)

### Demo

image
image

### QA notes

After enabling the Nikto boefje and normalizer, create an `IPAddress` or `Hostname` of a website and give it a clearance level of 3 or higher.
After doing this a new task should be created which will scan for outdated software.
(When adding a hostname with the name `cynalytics.nl` you should get a finding of an outdated software.)
---

### Code Checklist

---

---

## Checklist for QA:

### What works:

### What doesn't work:

### Bug or feature?:


boefjes/boefjes/plugins/kat_nikto/main.js
```diff
@@ -0,0 +1,54 @@
...
+  let ooi = "";
+  if (["IPAddressV4", "IPAddressV6"].includes(object_type))
+    ooi = boefje_meta.arguments.input.address;
+  else if (object_type == "Hostname") ooi = boefje_meta.arguments.input.name;
+  else throw new Error("Unexpected boefje_meta");
+
+  // Running nikto and outputting to a file
+  try {
+    execSync(`./nikto/program/nikto.pl -h ${ooi} -o ./output.json`, {
+      stdio: "inherit",
```
underdarknl, line 20: We might want to add:
UPDATES=no
To make sure Nikto does not report back to home about unrecognized webservers. (or make this an config argument for the boefje)

Also useful to avoid hanging boefjes:
PROMPTS=NO 

Also useful to be configurable from the boefje settings:
PROXYHOST=
PROXYPORT=
PROXYUSER=
PROXYPASS=
USERAGENT=

boefjes/boefjes/plugins/kat_nikto/boefje.json
```diff
@@ -0,0 +1,14 @@
+{
+  "id": "nikto",
+  "name": "Nikto",
+  "description": "Uses Nikto",
+  "consumes": [
+    "IPAddressV4",
+    "IPAddressV6",
+    "Hostname"
```
underdarknl, line 8: I think it makes more sense to start Nikto on Ipports that we know are talking http/https.
This way we avoid nikto on services that are not webservers, or even better:

It looks like there's also a way to differentiate between different hostnames on the same IP, so scanning on websites could also make sense as that would provide the best data-set including any differences between ipv4 and ipv6 addresses or the same host. Use the cli argument -host+ for this.

boefjes/boefjes/plugins/kat_nikto/main.js
underdarknl: -404code=301,302,307,308 can be used to make nikto not follow redirects to different hosts.

boefjes/boefjes/plugins/kat_nikto/boefje.json
Souf149: I have looked more into it and think using the OOI `HostnameHTTPURL` would be perfect. It has a hostname, port, and scheme which has everything. I was not able to find information from `IpPort` whether it was a webserver or not (unless you meant looking at the port's number). 

`-h+` is supposed to be an alias for `-host+` although this is not specified in [nikto's wiki](https://github.com/sullo/nikto/wiki/Annotated-Option-List). I will make sure to use `-nolookup`.

Souf149: The boefje now makes use of `HostnameHTTPURL` which tells the boefje whether its HTTP or HTTPS and contains a hostname as well. 

In theory `IPAddressHTTPURL` could also be used, I plan to add this once the boefje is satisfactory.

boefjes/boefjes/plugins/kat_nikto/main.js
Souf149: These have been added

boefjes/boefjes/plugins/kat_nikto/schema.json
```diff
@@ -0,0 +1,32 @@
...
+    "PROXYPASS": {
+      "title": "PROXYPASS",
+      "maxLength": 256,
+      "type": "string"
+    },
+    "USERAGENT": {
+      "title": "USERAGENT",
+      "maxLength": 256,
+      "type": "string"
+    }
```
ammar92, line 29: Perhaps you could use the `HTTP_PROXY` (and optionally `HTTPS_PROXY`) for this? This is a common proxy URI configuration. This will save a few variables and make the form less complex (given that it has a good example, e.g. `http://user:pass@1.2.3.4:3128/`

boefjes/boefjes/plugins/kat_nikto/boefje.json
```diff
@@ -0,0 +1,18 @@
...
+  "name": "Nikto",
+  "description": "Uses Nikto",
+  "consumes": [
+    "HostnameHTTPURL"
+  ],
+  "environment_keys": [
+    "PROXYHOST",
+    "PROXYPORT",
+    "PROXYUSER",
+    "PROXYPASS",
```
ammar92, line 12: See earlier remark

ammar92 (commented):
Looks good in general. Just a suggestion about the variables

boefjes/boefjes/plugins/kat_nikto/schema.json
Souf149: These are config variables required for Nikto. If we would make the boefje's schema less complex, then it would move the complexity to the code of the boefje. 

I believe since the user chooses to activate the Nikto boefje, they would know that Nikto requires these variables which would make them self-explaining.

ammar92: It's not complex code. You can use `urlparse` from `urllib.parse` and simply pass on the parsed components (hostname, port, user and password) to Nikto.

The reason I'm bringing this up is to reduce overall complexity and work towards a standard approach. A lot of Boefjes we make usually contain some kind of (HTTP) web request and usually use tools or dependencies that work with these `HTTP_PROXY` and `HTTPS_PROXY` variables convention. If we'd have to specify these multiple variables just to specify proxy options for every other proxy-enabled Boefje, that would be quite cumbersome.

What do you think @underdarknl?

Souf149: My thought when developing this boefje was that users that have experience with Nikto would use this boefje. But I realize now that is not the case. It is OpenKAT users that (might never have used Nikto) will use this boefje.

I will implement it in the way you have suggested, thank you 😅

underdarknl: We should still put the globally available HTTP_PROXY var in the boefje's schema for it to bleed trough into the boefjes runtime env. Beyond that we could easily transpose whatever value is in there to the vars that NIKTO requires.

Souf149: Is HTTP_PROXY something that OpenKAT contains? I have not heard of that before. I can also not find anything about it in the repo

Souf149: I understand the comment correctly now, I assumed `HTTP_PROXY` was something provided by OpenKAT.

e6cf370 fixes this.

Souf149:
Changed test because 1 more normalizer has been added.

Nikto is ready to be merged

ammar92 (approved):

stephanie0x00:
## Checklist for QA:

### What works:
Seems to work for the tested domains. 

### What doesn't work:
Scanning against IPv6 hosts. The latest version of Nikto support IPv6 hosts by adding the `-ipv6` parameter. Would be nice if IPv6 support is also available as an addition to this PR. 

### Bug or feature?:
n/a

underdarknl (approved):

PR #3412, opened 2024-08-28
Move variables from utils.js to renderNormalizerOutputOOIs.js

dekkers:
### Changes

The normalizer output OOIs weren't visible when DEBUG is set to False because django-compressor doesn't support the scripts that have module type. Moving the variables from utils.js to renderNormalizerOutputOOIs.js fixes this.

The other quick solution would be to not use compress for renderNormalizerOutputOOIs.js, but having to fetch an extra javascript file for just two variables also didn't seem very useful to me. renderNormalizerOutputOOIs.js was also the only file that included utils.js

Long-term we probably need to look into how we build/load javascript, but that is not for now.

### QA notes

The problem is only visible with DEBUG=false. I've deployed the fix on kattest03. This can be compared with kattest04 that doesn't have the fix.

---

---

## Checklist for QA:


underdarknl:
We also don't need the module argument on the loading script tags anymore.

https://github.com/minvws/nl-kat-coordination/blob/7a51134590709b645010a426e12f72e548e3691d/rocky/rocky/templates/tasks/normalizers.html#L96C17-L96C30

stephanie0x00:
## Checklist for QA:

### What works:
Solves the issue. Clicking the links on the Tasks Normalizer page also works. 

### What doesn't work:
n/a

### Bug or feature?:
The tasks list on the Normalizer page doesn't seem to work. Though this appears to be also on main with DEBUG=True, thus unrelated to this PR. 

Also unrelated to this PR, but the test server shows a lot of failed tasks after running for a few days. Haven't looked into this yet, but that should be checked out.

underdarknl (approved):

Issue #3434, opened 2024-09-02
Report types selection does not override previous selection


Rieven:
Steps to reproduce:
- Go through the onboarding for mispo.es. 
- Select all reports in the report flow
- On the plugins page return to the previous page (report selection0. 
- Change from all reports to only 1 report. 

This results in the following behaviour for the normal + aggregate report: 
- The toggle needs to be pressed twice before the checkboxes disappear. 
- When I select 1 report, eg. DNS report, and continue to the plugins page, you will see the plugins page for all reports. The expected behaviour would be to only view the plugins for the selected DNS report and/or skip this page if the requirements are met. 

_Originally posted by @stephanie0x00 in https://github.com/minvws/nl-kat-coordination/issues/3380#issuecomment-2301315511_

PR #3436, opened 2024-09-02
Fix report types selection not being overriden

Rieven:
### Changes

- At report types selection page, remove the previous selection from the report form fields, so that the new selection for report types is overriden.
- When all checkboxes are checked, the toggle-on class must be set other wise you have to click twice to un check all checked checkboxes.

### Issue link
https://github.com/minvws/nl-kat-coordination/issues/3434

Closes https://github.com/minvws/nl-kat-coordination/issues/3434

### QA notes

See issue link for for reproduction.

---

### Code Checklist

---

---

## Checklist for QA:


rocky/assets/js/checkboxToggler.js
```diff
@@ -1,4 +1,31 @@
...
+      ) {
+        if (checkboxes[i].checked) {
+          count_checked++;
+        }
+      }
+    }
+
+    if (count_checked == checkboxes.length) {
+      checkbox_target.classList.toggle("toggle-on");
+    }
```
ppvg, line 24: The `toggle()` method will also remove the class if it's already present. Is that desired behaviour?

If not, you could consider using [`toggle()`'s `force` param](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle#force), either to make sure it's only ever added:

```suggestion
    if (count_checked == checkboxes.length) {
      checkbox_target.classList.toggle("toggle-on", true);
    }
```

...or by using `count_checked == checkboxes.length` directly:

```suggestion
    checkbox_target.classList.toggle("toggle-on", count_checked == checkboxes.length);
```

Rieven: Oeehhh nice parameters I can use indeed to get the desired results. thanks!

ammar92 (approved):

rocky/assets/js/checkboxToggler.js
```diff
@@ -1,4 +1,32 @@
...
+      }
+    }
+
+    checkbox_target.classList.toggle(
+      "toggle-on",
+      count_checked == checkboxes.length,
+    );
+  }
+}
+SetToggleONWhenAllChecked();
```
underdarknl, line 28: Execute this on dom load, not before.

rocky/assets/js/checkboxToggler.js
```diff
@@ -1,4 +1,32 @@
...
+  checkbox_target = document.querySelector(".toggle-all[data-toggle-target]");
+  if (checkbox_target) {
+    var checkboxes = document.getElementsByName(
+      checkbox_target.dataset.toggleTarget,
+    );
+
+    for (var i = 0; i < checkboxes.length; i++) {
+      if (
+        checkboxes[i].tagName == "INPUT" &&
+        checkboxes[i].type == "checkbox"
```
underdarknl, line 14: Can't these be handled in a direct dom query on line 7?

underdarknl:
> * When all checkboxes are checked, the toggle-on class must be set other wise you have to click twice to un check all checked checkboxes.

This feels like a bug that we need to fix, instead of adding more code? When 'toggle all' is clicked, we should simply check if all available checkboxes have in fact been checked. We can do that by simple counting all checkboxes, and all checked checkboxes. Thats two simple queries against the dom, and avoids adding an extra class.

Rieven:
> > * When all checkboxes are checked, the toggle-on class must be set other wise you have to click twice to un check all checked checkboxes.
> 
> This feels like a bug that we need to fix, instead of adding more code? When 'toggle all' is clicked, we should simply check if all available checkboxes have in fact been checked. We can do that by simple counting all checkboxes, and all checked checkboxes. Thats two simple queries against the dom, and avoids adding an extra class.

This is not a bug, the toggle is working as it supposed to when you are at the same page. But when all is checked and you go to the next page and you go back, then the toggle does not know on load that it needs to add `toggle-on` as default. It is only triggered when clicked.

stephanie0x00:
## Checklist for QA:

### What works:
The flow now works, the 'double select' is gone, and when you move forward and backward between the pages the selection holds and the plugins toggle along with it. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

underdarknl:
> > > * When all checkboxes are checked, the toggle-on class must be set other wise you have to click twice to un check all checked checkboxes.
> > 
> > 
> > This feels like a bug that we need to fix, instead of adding more code? When 'toggle all' is clicked, we should simply check if all available checkboxes have in fact been checked. We can do that by simple counting all checkboxes, and all checked checkboxes. Thats two simple queries against the dom, and avoids adding an extra class.
> 
> This is not a bug, the toggle is working as it supposed to when you are at the same page. But when all is checked and you go to the next page and you go back, then the toggle does not know on load that it needs to add `toggle-on` as suppose that the next click would uncheck all checkboxes. Because the toggle on is not set, it would first add it on the first click (checking all checkboxes that is already checked) and then the second click would uncheck all.

Sure, but adding this new code simply to add a  state that we could already ascertain during onLoad seems wasteful.
In OnLoad we can just check If all the available checkboxes are checked, and flip our internal state / set the class. Or am i missing something?

underdarknl (approved):

PR #3505, opened 2024-09-10
Add s3 functionality in Bytes

### LINKED ISSUE #3191: s3 functionality in bytes
## About this feature
When hosting OpenKAT on k8s it's a good idea (at least scalability) to use the available s3 storage, or setup your own.
Bytes is a prime candidate as its storage reruirements fit neatly onto S3's capabilities.

First untested try available at:

https://github.com/minvws/nl-kat-coordination/tree/fix/bytes-s3

Souf149:
### Changes

Made it so that if the user adds `S3_BUCKET_PREFIX` or `S3_BUCKET` to their environment variables when building OpenKAT, that OpenKAT will use S3-buckets for saving the data from boefje's raw files instead of saving them on the local disk.

### Issue link

Closes #3191 

### Demo

Example files made inside s3
image

Example of one of the files
image

### QA notes

To test this feature you will need access to an existing s3 server. And inside of you `.env` file you will need to provide the following variables which are required for [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables):

- AWS_ACCESS_KEY_ID
- AWS_SECRET_ACCESS_KEY
- AWS_ENDPOINT_URL

Outside of these you also have to specify either of the following variables to let OpenKAT know you want to use s3 for the storage of bytes:

- S3_BUCKET_PREFIX
- S3_BUCKET

Make sure that these are compliant with the [bucket naming rules of AWS](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html).

These are the variables I used for testing:
```env
S3_BUCKET_PREFIX=souf-prefix
S3_BUCKET=souf-bucket

# Required env variables for `boto3`
AWS_ACCESS_KEY_ID=GKbf4e89d08d6c5250e7f1de60
AWS_SECRET_ACCESS_KEY=406fef311a69f02fa1c3b58d73a6299651c7890dc0ef5448104f1762e29887b7
AWS_ENDPOINT_URL=http://garage:3900
```

I made use of the s3 service [garage](https://garagehq.deuxfleurs.fr/)
[How to setup garage](https://garagehq.deuxfleurs.fr/documentation/quick-start/)
[How to work with garage in a container](https://garagehq.deuxfleurs.fr/documentation/cookbook/real-world/)

I built a container for garage with the following docker compose file on the same host machine as `bytes`
```yml
services:
  garage:
    image: dxflrs/garage:v1.0.0
    restart: unless-stopped
    ports:
      - "127.0.0.1:3900:3900"
    volumes:
      - /etc/garage.toml:/etc/garage.toml
      - /var/lib/garage/meta:/var/lib/garage/meta
      - /var/lib/garage/data:/var/lib/garage/data
    networks:
      - network1

networks:
  network1:
    name: nl-kat-coordination_default
    external: true
```

---

### Code Checklist

---

---

## Checklist for QA:

### What works:

### What doesn't work:

### Bug or feature?:


bytes/bytes/raw/file_raw_repository.py
```diff
@@ -62,3 +72,68 @@ def _raw_file_path(self, raw_id: UUID, boefje_meta: BoefjeMeta) -> Path:
...
         return str(raw_id)[: self.UUID_INDEX]
+
+
+class S3RawRepository(RawRepository):
+    def __init__(
+        self,
+        file_middleware: FileMiddleware,
+        bucket_per_org: bool,
+        s3_bucket_prefix: str | None = "OpenKAT-",
+        s3_bucket_name: str | None = "OpenKAT",
```
ammar92, line 83: You probably want a default (even empty) value, but not a `None`? This could lead to weird formats below if someone passes a `None` explicitly

bytes/bytes/raw/file_raw_repository.py
```diff
@@ -62,3 +72,68 @@ def _raw_file_path(self, raw_id: UUID, boefje_meta: BoefjeMeta) -> Path:
...
+    ) -> None:
+        self.file_middleware = file_middleware
+        self.bucket_per_org = bucket_per_org
+        self.s3_bucket_prefix = s3_bucket_prefix
+        self.s3_bucket_name = s3_bucket_name
+
+        self._session: Session = None
+        self._s3resource = None
+
+    @property
```
ammar92, line 93: I suggest using `@functools.cached_property`, this will cache the created S3 resource. Or leave it as is

bytes/bytes/raw/file_raw_repository.py
```diff
@@ -62,3 +72,68 @@ def _raw_file_path(self, raw_id: UUID, boefje_meta: BoefjeMeta) -> Path:
...
+    def s3resource(self):
+        if self._s3resource:
+            return self._s3resource
+        self._session = BotoSession()
+        self._s3resource = self._session.resource("s3")
+        return self._s3resource
+
+    def get_or_create_bucket(self, organization: str):
```
ammar92, line 101: Missing return type

bytes/bytes/raw/file_raw_repository.py
```diff
@@ -62,3 +72,68 @@ def _raw_file_path(self, raw_id: UUID, boefje_meta: BoefjeMeta) -> Path:
...
+
+class S3RawRepository(RawRepository):
+    def __init__(
+        self,
+        file_middleware: FileMiddleware,
+        bucket_per_org: bool,
+        s3_bucket_prefix: str | None = "OpenKAT-",
+        s3_bucket_name: str | None = "OpenKAT",
+    ) -> None:
+        self.file_middleware = file_middleware
```
ammar92, line 85: You probably don't want to expose this attribute; otherwise leave as is

```suggestion
        self._file_middleware = file_middleware
```

bytes/bytes/raw/file_raw_repository.py
```diff
@@ -62,3 +72,68 @@ def _raw_file_path(self, raw_id: UUID, boefje_meta: BoefjeMeta) -> Path:
...
+        self.bucket_per_org = bucket_per_org
+        self.s3_bucket_prefix = s3_bucket_prefix
+        self.s3_bucket_name = s3_bucket_name
+
+        self._session: Session = None
```
ammar92, line 90: This object seems to be used only to create the s3resource object, you might consider removing this to simplify the code as it doesn't have any other value

ammar92 (commented):
Nice work 👍 Very few remarks to improve it even more

bytes/bytes/raw/file_raw_repository.py
Souf149: I meant to write this as "You do not have to give this argument, and if you do not, we will assign something to it". I will refactor it, thank you 👍

bytes/bytes/raw/file_raw_repository.py
Souf149: Is that needed? Since the method "caches" the property (`_s3resource`) already

bytes/bytes/raw/file_raw_repository.py
```diff
@@ -62,3 +72,68 @@ def _raw_file_path(self, raw_id: UUID, boefje_meta: BoefjeMeta) -> Path:
...
+        self._s3resource = None
```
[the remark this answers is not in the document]
Souf149: `boto3` has very limited type hinting. The only way I could find to get type hinting for `boto3` is by installing `boto3-stubs-full[boto3]` through pip. 
This makes us able to give `S3RawRepository._s3resource` a type by utilizing `from mypy_boto3_s3 import S3ServiceResource`. Is this an acceptable solution?

bytes/bytes/raw/file_raw_repository.py
Souf149: Or do you think it is better to let `@functools.cached_property` handle the caching?

bytes/bytes/raw/file_raw_repository.py
```diff
@@ -62,3 +72,68 @@ def _raw_file_path(self, raw_id: UUID, boefje_meta: BoefjeMeta) -> Path:
...
+        self._s3resource = self._session.resource("s3")
+        return self._s3resource
+
+    def get_or_create_bucket(self, organization: str):
+        # Create a bucket, and if it exists already return that instead
+        bucket_name = self.s3_bucket_name
+        if self.bucket_per_org:
+            bucket_name = f"{self.s3_bucket_prefix}{organization}"
+            try:
+                bucket = self.s3resource.create_bucket(Bucket=bucket_name)
```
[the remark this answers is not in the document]
Souf149: I think we should keep using the method in case `s3resource` does not exist yet

bytes/bytes/raw/file_raw_repository.py
ammar92: Ah I see. But then in that case you could simply remove the `None` type and give it a default value; which makes this then an optional keyword argument with a default

bytes/bytes/raw/file_raw_repository.py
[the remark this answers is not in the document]
ammar92: Thanks for clearing that up. I would advise against relying on `mypy_boto3` in the codebase. Instead, you can determine the type dynamically by inspecting what `resource()` returns (e.g. using `print(type(...resource())))`. This approach avoids external dependencies and keeps the codebase simpler, while still giving you the information needed for type hinting. But anyways, looking at the [API reference](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session.resource), it should return a `ServiceResource`, so you could also use just that

bytes/bytes/raw/file_raw_repository.py
ammar92: It's not really better in the sense of a better cache utility, but it's simpler. Although it's a common pattern, it wasn't immediately clear that the property is cached; whereas using the decorator would make it more obvious. But it's a gray area, so I'll leave it up to you. Both approaches are valid

bytes/bytes/raw/file_raw_repository.py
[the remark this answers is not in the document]
ammar92: While I understand why this is implemented as such, we should avoid using properties for instantiating such resources. Also, using this method (`get_or_create_bucket`) without a resource is a misuse of the method and should naturally yield in an error.

If you need this resource to be available when using the method, consider instantiating it in the `S3RawRepository` instantiation

bytes/bytes/raw/file_raw_repository.py
Souf149: since the attribute `_s3resource` never gets used outside of the property-method. I think it is indeed better to use the property-method exclusively. And make it a `@functools.cached_property`. And the original  `_s3resource` will be removed

bytes/bytes/raw/file_raw_repository.py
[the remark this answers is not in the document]
Souf149: You are totally correct. There is no need to have a method cache this property when it can just be a property that is assigned with the `S3RawRepository` instantiation

bytes/bytes/raw/file_raw_repository.py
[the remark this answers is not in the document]
Souf149: The `ServiceResource` provided inside `boto3.resources.base` does not provide an interface for the developer. 
image

I will check what `print(type(...resource()))` results in

bytes/bytes/raw/file_raw_repository.py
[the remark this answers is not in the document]
ammar92: Agreed. I believe instantiating the S3 connection shouldn't be expensive since it's just an HTTP API client. It wil usually only establish a connection when an actual request is made, so there shouldn't be much overhead in instantiation/ resources. Now that I think of this, it's probably even safe to instantiate it without needing to cache it

bytes/pyproject.toml
```diff
@@ -41,6 +41,7 @@ pyjwt = "^2.8.0"
 fastapi-slim = "^0.111.0"
 structlog = "^24.2.0"
 cryptography = "^43.0.1"
+boto3 = "^1.35.15"
```
Donnype, line 44: We could consider creating a separate poetry group for this to make it an optional dependency, so we don't always ship with `boto3` when it's not needed:

```toml
boto3 = { version = "^1.35.15", optional = true }
other_s3_requirement = { version = "^1.2", optional = true }

[tool.poetry.extras]
s3 = ["boto3", "other_s3_requirement"]
```
This can then be installed by `poetry install --extras s3`, for which we should add a build argument to the Dockerfile.

The downside however is that we would have to distribute separate Docker images as well in a release or require users of this functionality to build Bytes from source. A bit of extra work but makes the default installation leaner.

@underdarknl thoughts?

bytes/bytes/raw/file_raw_repository.py
```diff
@@ -62,3 +72,63 @@ def _raw_file_path(self, raw_id: UUID, boefje_meta: BoefjeMeta) -> Path:
 
     def _index(self, raw_id: UUID) -> str:
         return str(raw_id)[: self.UUID_INDEX]
+
+
+class S3RawRepository(RawRepository):
```
Donnype, line 77: Good to see that it was relatively easy to extend the interface to provide this functionality 👌

bytes/bytes/raw/file_raw_repository.py
```diff
@@ -62,3 +72,63 @@ def _raw_file_path(self, raw_id: UUID, boefje_meta: BoefjeMeta) -> Path:
...
+    def _s3resource(self):
+        session = BotoSession()
+        return session.resource("s3")
+
+    def get_or_create_bucket(self, organization: str):
+        # Create a bucket, and if it exists already return that instead
+        bucket_name = self.s3_bucket_name
+        if self.bucket_per_org:
+            bucket_name = f"{self.s3_bucket_prefix}{organization}"
+            try:
```
Donnype, line 100: If we expect that the common case is that the bucket exists (because we create it only once), would it be possible to flip the logic here and do something like this?

```python3
try:
  return self._s3resource.Bucket(name=bucket_name)
except BucketDoesNotExist:
  bucket = self._s3resource.create_bucket(Bucket=bucket_name)
  bucket.wait_until_exists()
  return bucket
```

Then at least we have 1 error and N happy flows in general instead of the other way around, but not sure if that Bucket() call fails if no bucket exists, meaning we make only 1 call instead of 2 calls always.

bytes/bytes/raw/file_raw_repository.py
```diff
@@ -62,3 +72,63 @@ def _raw_file_path(self, raw_id: UUID, boefje_meta: BoefjeMeta) -> Path:
...
+        bucket = self.get_or_create_bucket(raw.boefje_meta.organization)
+        bucket.Object(file_name).put(Body=contents)
+
+    def get_raw(self, raw_id: UUID, boefje_meta: BoefjeMeta) -> RawData:
+        file_name = self._raw_file_name(raw_id, boefje_meta)
+        bucket = self.get_or_create_bucket(boefje_meta.organization)
+
+        try:
+            contents = bucket.Object(file_name).get()["Body"].read()
+        except self._s3resource.meta.client.exceptions.NoSuchBucket as error:
```
Donnype, line 123: (Perhaps here we could catch whether the bucket for the organization wasn't created yet and create it in that case?)

bytes/bytes/raw/file_raw_repository.py
```diff
@@ -62,3 +72,63 @@ def _raw_file_path(self, raw_id: UUID, boefje_meta: BoefjeMeta) -> Path:
...
+                raise BytesFileNotFoundException(error)
+            logger.error("Could not get file from s3: %s/%s due to %s", bucket.name, file_name, error)
+            raise error
+
+        return RawData(value=self._file_middleware.decode(contents), boefje_meta=boefje_meta)
+
+    def _raw_file_name(self, raw_id: UUID, boefje_meta: BoefjeMeta) -> str:
+        if self.bucket_per_org:
+            return str(raw_id)
+        return f"{boefje_meta.organization}/{raw_id}"
```
Donnype, line 134: Does S3 scale properly when there are many files in a "directory"? On disk we needed an extra `self._index(raw_id)` namespace but I know object storage has different properties.

Donnype (commented):
Looking good! I have some high level considerations 😁👍

bytes/bytes/raw/file_raw_repository.py
Souf149: I don't have the experience of working with S3 that much. But reading off of multiple answers from stackoverflow:
["Even though your object's key might contain a '/', S3 treats the path as a plain string and puts all objects in a flat namespace."](https://stackoverflow.com/a/394505/9216830)

["The files are not really in folders, they are just strings as locations."](https://stackoverflow.com/a/30194799/9216830)

["My understanding is that S3 hashes object keys, so that lookup time for a given object is O(1)."](https://stackoverflow.com/a/48678474/9216830)

I assume that s3 storage would not benefit from using something like your `self._index(raw_id)`

bytes/bytes/raw/file_raw_repository.py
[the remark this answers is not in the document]
Souf149: I forgot to mention that it does tell me that it returns `ServiceResource` but this is not seemingly importable in a useful manner. Most of the classes from boto3 make use of adding methods and attributes at runtime.

bytes/pyproject.toml
ammar92: Although I like the idea of having contrib and extra package requirements, it does come with a few consequences. For example, we'd end up with more requirements files, and we'd have to modify build pipelines and scripts (e.g., `poetry-dependencies`) and the argument you mentioned. For now, I don't think it's worth the effort if it's only for this functionality (assuming this S3 storage will eventually be part of OpenKAT and not an optional extension). However it's something we could consider for the (near?) future or maybe implement it differently, e.g. storage and middleware as plugins?

bytes/bytes/raw/file_raw_repository.py
underdarknl: s3 is designed to handle "unlimited" amounts of files in one namespace. We dont need to worry about this.

bytes/bytes/raw/file_raw_repository.py
underdarknl: Usually, buckets cannot be created by the same user who accesses them.

bytes/pyproject.toml
underdarknl: for any k8s hosted openkat I'd expect them to run on the s3 storage backend, as that's usually a more performant and cheaper (and managed) solution available to these platforms.

bytes/bytes/raw/file_raw_repository.py
Souf149: So a bucket should not be created by the OpenKAT instance ever?

bytes/bytes/raw/file_raw_repository.py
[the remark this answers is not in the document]
Souf149: Using the vscode package [AWS boto3](https://marketplace.visualstudio.com/items?itemName=Boto3typed.boto3-ide) we can get full type coverage, but the code has to be unannotated. AWS has stated that typehinting

bytes/bytes/raw/file_raw_repository.py
Souf149: For now with [264733a](https://github.com/minvws/nl-kat-coordination/pull/3505/commits/264733a1d127df640eaabbcad9c9b1b76cfe5538) a bucket should always get created by `S3RawRepository.get_or_create_bucket()` and if it does not, it should throw an exception before this piece of code.

bytes/bytes/raw/file_raw_repository.py
Souf149: After some testing, I found out that no exception gets thrown when trying to create a bucket that already exists (according to the documentation it does throw an exception, but that part is outdated then I assume).

So now instead, we can create our bucket every time it is needed and it will give us the old bucket.

```python
try:
    bucket = self._s3resource.create_bucket(Bucket=bucket_name)
    bucket.wait_until_exists()
    return bucket
except bucket.meta.client.exceptions.ClientError as error:
    logger.error("Something went wrong)
    raise error 
```

An exception can still occur, but is not expected.
Implemented in [264733a](https://github.com/minvws/nl-kat-coordination/pull/3505/commits/264733a1d127df640eaabbcad9c9b1b76cfe5538)

Donnype: I'm not sure if the solution is reliable but one suggestion here is to use the creation_date field to check if we need to create the bucket still: https://stackoverflow.com/questions/26871884/how-can-i-easily-determine-if-a-boto-3-s3-bucket-resource-exists. One instance where we might still have significant overhead is when creating multiple raw files at once under higher workloads. It still feels a bit like the creating-a-connection-per-database-operation anti-pattern to create the bucket redundantly on every request, but perhaps most users just use the one-bucket setup

bytes/bytes/raw/file_raw_repository.py
Souf149: > If we expect users to not use this functionality and create buckets outside of OpenKAT, we could perhaps drop bucket creation?

I think giving the user the responsibility to create buckets is more than fair. *Unless* they have chosen to have `bucket_per_org` enabled. Because in that case we might want to give that responsibility to OpenKAT to make adding new organisations less troublesome. Or do you think it should still be the job of the user then?

Donnype: Perhaps we can also make it configurable whether OpenKAT should create buckets? It makes the PR a bit more tricky in terms of configuration, so I'd say that we do need to know if creating buckets in OpenKAT is a requirement, which might not be the case (yet)? If not, we just assume the buckets are present and raise a descriptive error if they are not.

stephanie0x00:
## Checklist for QA:

### What works:
We did a live demo of the functionality to check if it works. The code stores data to a local (Garage) S3 bucket and we are also able to retrieve the data stored in bytes. It doesn't seem to break anything on main. 

### What doesn't work:
n/a

### Bug or feature?:
We identified a few improvements and things that could probably be solved nicer. There are picked up in different tickets and will be created by @Souf149. These include documentation and whether or not both the PREXIX and Bucket env vars are both mandatory (by aws)  or whether these can be also used separately.

PR #3515, opened 2024-09-11
remove unneeded column from filtered plugin table view

### LINKED ISSUE #3474: Remove unnecessary repetition within katalogus
**Describe the issue**
We could remove the 'plugin type' column when we are looking at the boefjes or normalizers lists, as it will list the same type over and over.

**To Reproduce**

1. Go to KAT-alogus
2. Click on: Toggle view to see list view (instead of default card view)
3. See table: header "plugin type". When looking at either the boefjes or the normalizers view. This column will be filled with the same content for all. And can be removed. 

**Expected behavior**
Expect the table to have useful information for our users. Repeating the same information over and over while in a specific view isn't. Remove unnecessary information making it easier to see the useful information.

**Screenshots**
364079720-c9238c3e-5884-4dc9-8e19-7fb08b45eb73

underdarknl:
### Changes

Only displays the plugin type column for the type-unfiltered katalogus page.

### Issue link

closes #3474


stephanie0x00:
On the 'All' page the label for boefjes/normalizers is missing. 

image

dekkers (approved):

Issue #3143, opened 2024-06-25
Add Pattern to Nmap boefje to match a range of numbers, eg 0 to 65000 are not overly easy to do, but here's a try:


underdarknl:
> To revisit this. Patterns that match a range of numbers, eg 0 to 65000 are not overly easy to do, but here's a try:
> 
> `(([1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-4]),)*?(([1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-4])-)?([1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])`
>
> This will match the following:
> 
> - 1
> - 1,2
> - 1-65000
> - 1,2,100-200
> - 65534-65535
> 
> Rules:
> 
> - The number of single ports is unlimited, the number of range groups is unlimited, but expected to come after the single ports.
> - Range groups are intended to be separated by a dash
> - No spaces / other chars are allowed.
> - Port numbers (single or in range) are limited from 1 to 65535
> - A small number is allowed as the second argument in a range, this is not easily fixed in the regex, but we might be able to flip them round in the boefje code.

_Originally posted by @underdarknl in https://github.com/minvws/nl-kat-coordination/issues/2999#issuecomment-2177941702_

Donnype:
A different, perhaps more modular approach, including passing unit tests:
```python3

max_65535 = "(6553[0-5]|655[0-2]\d|65[0-4]\d{2}|6[0-4]\d{3}|[1-5]\d{0,4}|\d)"  # matches 1 to 65535
max_65535_or_range = f"({max_65535}|{max_65535}-{max_65535})"  # matches 1 to 65535 or a range of two
pattern = re.compile(f"{max_65535_or_range}|{max_65535_or_range}(,{max_65535_or_range})+")  # comma separated list of the latter

def test_single_port_pattern():
    for single_port in ["1", "2", "20", "200", "2000", "20000", "65535"]:
        assert pattern.fullmatch(single_port) is not None


def test_bad_single_port_pattern():
    for bad_single_port in ["-1", "-2000", "65536", "222222"]:
[...]

PR #3516, opened 2024-09-11
Add regex pattern check to PORTS setting of `nmap-ports`

Donnype:
### Changes

Adds regex pattern validation to the PORTS setting for `nmap-ports`. The only thing we cannot really check without exploding the pattern is if the left side of a `1-20` style range is smaller than the right side, which I think is an OK edge case not to handle if we treat is as an empty list.

See the unit test to see how the pattern is constructed:
```python3
def get_pattern():
    max_65535 = r"(6553[0-5]|655[0-2]\d|65[0-4]\d{2}|6[0-4]\d{3}|[1-5]\d{0,4}|\d)"
    max_65535_or_port_range = f"({max_65535}|{max_65535}-{max_65535})"
    one_or_comma_separated = f"^{max_65535_or_port_range}$|^{max_65535_or_port_range}(,{max_65535_or_port_range})+$"

    return re.compile(one_or_comma_separated)
```

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/3143

### Demo

See the unit tests for the behavior and how this pattern was constructed!

### QA notes

Please check that in the UI bad patterns are not allowed for this settings!

---

### Code Checklist

---

---

## Checklist for QA:


ammar92 (approved):
Approved because it's implemented as suggested in the issue ticket and includes well written tests. But in my opinion, we're overcomplicating it by using regex. The only advantage I see using regex is that we get free validation using `jsonschema` on the frontend side. This is a valid and reasonable advantage, but we might want to reconsider the complexity trade-off in future iterations.

(my two cents) My approach would be validating the argument in the boefje itself, for example:
```python3
def validate_nmap_ports(port_str: str) -> bool:
    port_ranges = port_str.split(',')

    for port_range in port_ranges:
        # handle single port
        if '-' not in port_range:
            try:
                port = int(port_range)
                if not (1 <= port <= 65535):
                    return False
            except ValueError:
                return False
        else:
            # handle range of ports
            try:
                start, end = map(int, port_range.split('-'))
                if not (1 <= start <= 65535 and 1 <= end <= 65535 and start <= end):
                    return False
            except ValueError:
                return False
    
    return True
```

This function validates the examples as mentioned in the issue ticket. It will also make sure that when given a range, the start port should be less than the end port, because that's something `nmap` also validates:
```bash
ammar in OpenKAT ❯ docker run instrumentisto/nmap:latest localhost -p 12-10
Starting Nmap 7.94 ( https://nmap.org ) at 2024-09-11 19:59 UTC
Your port range 12-10 is backwards. Did you mean 10-12?
QUITTING!
```

underdarknl:
> Approved because it's implemented as suggested in the issue ticket and includes well written tests. But in my opinion, we're overcomplicating it by using regex. The only advantage I see using regex is that we get free validation using `jsonschema` on the frontend side. This is a valid and reasonable advantage, but we might want to reconsider the complexity trade-off in future iterations.

I agree a regex is a horrible solution for this. however, its the only solution that fits inside the Schema, is somewhat declarative (our end goal for plugin definitions) and is usable in the backend, and frontend.

stephanie0x00:
## Checklist for QA:

### What works:
Seems to work for the ports and ranges I tested. The ports I skipped also seem to be skipped in the raw file. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

underdarknl (approved):

PR #3529, opened 2024-09-16
Add boefje interval and cron check for deadline in scheduler

jpbruinsslot:
### Changes

- Support to explicitly set a cron schedule of a boefje such that the recurrence of the tasks of that  boefje can be scheduled with a cron expression.

- Support for overriding the default time-out (grace period)  of tasks based on the boefje that was used. By specifying the interval of a boefje. Tasks using that boefje will inherit the timeout specified.

### Details

Depends on: https://github.com/minvws/nl-kat-coordination/pull/3555

- When a task is created and pushed onto the scheduler queue, it will now _not_ create a cron expression (field `schedule`) for a `Schedule` by default. But will determine its `deadline_at` by calling `calculate_deadline()` . When a `schedule` is explicitly set, the `deadline_at` calculation based on the cron expression will take precedence. This gives us a better indication of rescheduled tasks whether they are automatically scheduled by calculating their next run (`deadline_at`), and when they are explicitly scheduled on an interval (the `schedule` cron expression)

- When a boefje supplies a `cron` schedule the scheduler will add this `cron` schedule to all the resulting task `Schedule` objects. This allows a user to specify a `cron` schedule to all the ooi's that boefje will run on.

  - For this we override the `post_push`() of the `Scheduler` base class in the `BoefjeScheduler`  this is done since a `Schedule` for a task is created on this method. This allows us to create `Schedule` objects for other schedulers in the system. We override this in the `BoefjeScheduler` since there is a specific request to create a `Schedule`  on the presence of a `cron` field on the plugin representation of a boefje.
  - We retrieve the plugin for the boefje that is being referenced in the task that has been pushed onto the queue. We check  if we already have a `Schedule` for this `Task` in our system and create or update it accordingly
  
- The rescheduling of a `BoefjeTask` (or when the task is allowed to run again), is by default set by the 'grace period' of a scheduler installation (default 24 hours). Setting the `interval` value for an individual boefje a user will be able to set the timeout for tasks that that particular boefje will generate.

  - We override the `calculate_deadline()` method of the `Scheduler` base class, and check for the presence of the `interval` field in the plugin representation of a boefje. We calculate the deadline for the `Schedule` of the task (when a task needs to run again). When this isn't present we fallback to the default deadline calculate which is the grace period.

### QA notes

The functionality that this PR exposes will be used in https://github.com/minvws/nl-kat-coordination/pull/3555, specific QA notes in order to test this new functionality can be done when those changes are merged. So the main things that can be tested is that normal operations are still working. So scanning, rescanning should still work without any errors or exceptions.

---

### Code Checklist

---

---

## Checklist for QA:


mula/scheduler/schedulers/boefje.py
```diff
@@ -997,3 +997,19 @@ def get_boefjes_for_ooi(self, ooi) -> list[Plugin]:
...
+        Args:
+            task: The task to calculate the deadline for.
+
+        Returns:
+            The calculated deadline.
+        """
+        # Does the boefje have an interval defined?
+        interval = utils.deep_get(task.data, "boefje", "interval")
+        if interval is not None:
+            return datetime.now(timezone.utc) + timedelta(seconds=interval)  # FIXME: check if it is seconds or minutes
```
underdarknl, line 1013: The expected interval in the boefje schema is in minutes, as that is fine gained enough for our monitoring purposes.

mula/scheduler/schedulers/boefje.py
```diff
@@ -997,3 +998,50 @@ def get_boefjes_for_ooi(self, ooi) -> list[Plugin]:
         )
 
         return boefjes
+
+    def post_push(self, item: Task) -> Task:
+        """Override Schedule.post_push() when a boefje specifies a schedule for
+        execution (cron expression) we schedule for its execution"""
+        # Does a boefje have a schedule defined?
+        plugin = self.ctx.services.katalogus.get_plugin_by_id_and_org_id(
```
jpbruinsslot, line 1006: Check if this returns None when a plugin isn't found

mula/scheduler/schedulers/boefje.py
```diff
@@ -997,3 +998,50 @@ def get_boefjes_for_ooi(self, ooi) -> list[Plugin]:
...
+            utils.deep_get(item.data, ["boefje", "id"]),
+            self.organisation.id,
+        )
+        if plugin is None or plugin.cron is None:
+            return super().post_push(item)
```
Donnype, line 1011: Flow of control is a bit confusing here: if we have no `cron`, we let the parent handle the item, but there are checks like `if self.create_schedule` there that we don't do here, and the other important field being the `interval` here is handled because we also override the `calculate_deadline` method.

In short I feel like we should be able to override smaller methods here, perhaps override a `create_cron(item)` method so that you don't have to override the `post_push`?

mula/tests/integration/test_boefje_scheduler.py
```diff
@@ -816,6 +833,70 @@ def test_post_push(self):
...
+        # Schedule should be in datastore
+        schedule_db = self.mock_ctx.datastores.schedule_store.get_schedule(task_db.schedule_id)
+        self.assertIsNotNone(schedule_db)
+        self.assertEqual(schedule_db.id, task_db.schedule_id)
+
+        # Schedule deadline should be set
+        self.assertIsNotNone(schedule_db.deadline_at)
+
+        # Schedule cron should NOT be set
+        self.assertIsNone(schedule_db.schedule)
```
Donnype, line 864: :+1:

mula/tests/integration/test_boefje_scheduler.py
```diff
@@ -830,6 +911,203 @@ def test_post_push(self):
...
+        # Schedule cron should be set
+        self.assertIsNotNone(schedule_db.schedule)
+        self.assertEqual(schedule_db.schedule, cron)
+
+        # Check if the deadline_at is set correctly, to the next
+        # day at midnight
+        self.assertEqual(
+            schedule_db.deadline_at,
+            datetime.now(timezone.utc).replace(hour=0, minute=0, second=0, microsecond=0) + timedelta(days=1),
+        )
```
Donnype, line 931: :ok_hand:

mula/tests/integration/test_boefje_scheduler.py
```diff
@@ -830,6 +911,203 @@ def test_post_push(self):
...
+
+        # Schedule cron should NOT be set
+        self.assertIsNone(schedule_db.schedule)
+
+        # Check if the deadline_at is set correctly with the interval
+        # set to 1500 minutes (25 hours) to at least the next day
+        self.assertGreater(
+            schedule_db.deadline_at,
+            datetime.now(timezone.utc) + timedelta(days=1),
+        )
```
Donnype, line 1109: :+1:

mula/tests/integration/test_boefje_scheduler.py
```diff
@@ -830,6 +911,203 @@ def test_post_push(self):
...
+
+        # Schedule deadline should be set
+        self.assertIsNotNone(schedule_db1.deadline_at)
+        self.assertIsNotNone(schedule_db2.deadline_at)
+
+        # Schedule cron should be set
+        self.assertIsNotNone(schedule_db1.schedule)
+        self.assertEqual(schedule_db1.schedule, cron)
+        self.assertIsNotNone(schedule_db2.schedule)
+        self.assertEqual(schedule_db2.schedule, cron)
```
Donnype, line 1033: Perhaps there is some redundancy here given your other tests, although I haven't read these very carefully. It's tricky but my preference is always to keep the tests as orthogonal as possible as well :slightly_smiling_face:

mula/scheduler/schedulers/boefje.py
```diff
@@ -997,3 +998,50 @@ def get_boefjes_for_ooi(self, ooi) -> list[Plugin]:
...
+
+        schedule_db = self.ctx.datastores.schedule_store.get_schedule_by_hash(item.hash)
+        if schedule_db is None:
+            # Create schedule when not found
+            schedule = Schedule(
+                scheduler_id=self.scheduler_id,
+                hash=item.hash,
+                deadline_at=self.calculate_deadline(item),
+                data=item.data,
+                schedule=plugin.cron,
```
Donnype, line 1021: :+1:

Donnype (approved):
Looks good! Some refactoring ideas here and there, nothing blocking.

mula/scheduler/schedulers/boefje.py
jpbruinsslot: Yeah I agree, during our meeting it also was difficult to convey the purpose and why the overrides were necessary. I like your suggestion, and will try update it to make easier to reason about.

mula/tests/integration/test_boefje_scheduler.py
jpbruinsslot: Yeah I agree, I think I'll delete this test. It doesn't add more than the test with a single item.

mula/scheduler/schedulers/boefje.py
```diff
@@ -997,3 +997,29 @@ def get_boefjes_for_ooi(self, ooi) -> list[Plugin]:
         )
 
         return boefjes
+
+    def set_cron(self, item: Task) -> str | None:
```
ammar92, line 1001: Given the complexity and the implementation below, I believe using this as a setter instead of property is a good choice 👍

ammar92 (commented):
Nice work, I don't have much to add since there were already good remarks by other reviewers.

I do have a more general remark regarding naming: I believe `cron_expression` would be a more appropriate name in the models/ backend as it better describes the function and purpose. However, I think we can still use *cron* in the frontend if that aligns better with the expectations of our end users, but using *cron expression* internally/ in the code base would make more sense to me. What do you think?

jpbruinsslot:
> Nice work, I don't have much to add since there were already good remarks by other reviewers.
> 
> I do have a more general remark regarding naming: I believe `cron_expression` would be a more appropriate name in the models/ backend as it better describes the function and purpose. However, I think we can still use _cron_ in the frontend if that aligns better with the expectations of our end users, but using _cron expression_ internally/ in the code base would make more sense to me. What do you think?

Yeah I totally I agree, for a backend perspective it will more accurately describe what the field is for, `schedule` is a bit disambiguous and confusing in relation to the `Schedule` model. I'm opting to change it 1.18 since it will entail a database migration and some changes in several other services that I likely want to tackle in one go.

mula/scheduler/schedulers/boefje.py
```diff
@@ -997,3 +997,29 @@ def get_boefjes_for_ooi(self, ooi) -> list[Plugin]:
...
+        if plugin is None or plugin.cron is None:
+            return super().set_cron(item)
+
+        return plugin.cron
+
+    def calculate_deadline(self, task: Task) -> datetime:
+        """Override Scheduler.calculate_deadline() to calculate the deadline
+        for a task and based on the boefje interval."""
+        # Does the boefje have an interval defined?
+        plugin = self.ctx.services.katalogus.get_plugin_by_id_and_org_id(
```
underdarknl, line 1018: Dont we already have the plugins available somewhere? Or does this not trigger a call to the Katalogus api?

jpbruinsslot: The katalogus service in the scheduler has a cache implemented, it should not trigger a call when it isn't necessary

stephanie0x00:
## Checklist for QA:

### What works:
Looks good. All scheduled tasks complete (apart from one, but thats a different issue). Added an interval to a boefje variant, scheduled this and then rescheduled. Scheduled a bunch of other boefjes that also complete All normalizers also run and complete.

### What doesn't work:
Nothing found. 

### Bug or feature?:
- One of the boefjes fails, but works after reschedule. I created a new ticket for this as it is outside of the schedule #3598.

underdarknl (approved):

PR #3506, opened 2024-09-10
Check if the task is still running according to the scheduler before changing the status

Donnype:
### Changes

Checks if the task is still running according to the scheduler before changing the status. @Souf149 Could you perhaps verify the fix?

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/3426

### Demo

-

### QA notes

Please check if the tasks are kept to FAILED for a failing nmap-tcp!

---

### Code Checklist

- [x] All the commits in this PR are properly PGP-signed and verified.
- [x] This PR only contains functionality relevant to the issue.
- [x] I have written unit tests for the changes or fixes I made.
- [x] I have checked the documentation and made changes where necessary.
- [x] I have performed a self-review of my code and refactored it to the best of my abilities.

---

## Checklist for code reviewers:

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_qa.md) into your comment.

PR #3548, opened 2024-09-17
Improve boefje runner error messages on container failure

dekkers:
### Changes

The code logged an exception with stacktrace when the container returned a non-zero exit status, but this is confusing because nothing actually goes wrong inside the boefje runner when docker returns an error. To make it more confusing the message of exception that is logged might also contain a stack trace when the output of the docker container was a stacktrace. This change will just log an error message if docker returns an error instead of an exception with the strack trace.

Reraising the exception is not necessary anymore because this was already fixed by #3506 last week.

### QA notes

For testing you can force the container to return nonzero by changing the boefje code, for example change the run function of kat_dnssec/main.py to:

```
def run(boefje_meta: dict):
    import sys
    sys.exit(1)
```

Then run "make build" to build a new dnssec container image. Running dnssec should then result in only a single error message in the boefje runner. The task should also have failed status.

---

---

## Checklist for QA:

### DIFF
--- a/boefjes/boefjes/docker_boefjes_runner.py
+++ b/boefjes/boefjes/docker_boefjes_runner.py
@@ -62,7 +62,11 @@ class DockerBoefjesRunner:
                 # have to raise exception to prevent _start_working function from setting status to completed
                 raise RuntimeError("Boefje did not call output API endpoint")
         except ContainerError as e:
-            logger.exception("Container error")
+            logger.error(
+                "Container for task %s failed and returned exit status %d, stderr saved to bytes",
+                task_id,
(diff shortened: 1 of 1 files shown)


ammar92 (approved):
This was indeed confusing. Nice work

stephanie0x00:
## Checklist for QA:

### What works:
The normalizer and boefje both fail after editing the file as mentioned in the QA notes. Seems to work as expected and doesn't seem to break anything obvious. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

underdarknl (approved):

PR #3562, opened 2024-09-19
Set default findingtype risk in model instead of in bit

noamblitz:
### Changes

Set default findingtype risk in model instead of in bit

### Demo

No functional changes

### QA notes

Turn of kat-finding-types boefje
Create Hostname with L1
See that finding types appear
Make sure that they are "Pending"
Make sure that no reports are broken with these FindingTypes

Also check with migrations: do the same as above in main and then switch to this branch and "make" again. Rerun bits. Everything should still work

### Code Checklist

---

---

## Checklist for QA:


underdarknl:
Hmm, can we still store an 'unknown' severity for findings that we could not hydrate / have no information due to an embargo?
We opted to communicate findings with an unknown severity as being 'critical' due to the nature of embargoed findings. If we change the default to 0.0, they wont be communicated as critical anymore I'd guess?

noamblitz:
Before, the default was unset and immediately after creation, the bit would set it to 0.0. In the time between creating the finding and setting the value to 0.0, the finding would be invisible (this would be very very short/ unnoticeable). This behavior is now better.

After a while, the boefje would try to find the score, if it didn't succeed, it would set the value to 10.0. This behavior is unchanged.

octopoes/octopoes/models/ooi/findings.py
```diff
@@ -47,8 +47,8 @@ class FindingType(OOI):
     impact: str | None = None
     recommendation: str | None = None
 
-    risk_score: float | None = None
-    risk_severity: RiskLevelSeverity | None = None
+    risk_score: float | None = 0.0
```
ammar92, line 50: Pydantic v2 interprets this as:
- `risk_score`: a nullable field with default `0.0`, meaning we can explicitly set and initialize it to `None`, otherwise it will be assigned the default value
- `risk_severity`: a nullable field with default `RiskLevelServerity.PENDING`, meaning we can explicitly set and initialize it to `None`, otherwise it will be assigned the default value

Is this still the intended behavior? If so, leave as is. If not, for example if we don't accept `None`s anymore for the `risk_score` then we should just change the type to `float` only

ammar92 (commented):
Just one question to confirm. Looks good otherwise

octopoes/octopoes/models/ooi/findings.py
noamblitz: Why I did was that I'm afraid that during updating, there might be KAT installs that still have `None`s in them. Although the chance is tiny, this would break their install.

originalsouth: Nice catch, indeed I do not think we need the `None`'s anymore; @noamblitz?

noamblitz: I do think we need for backwards compatibility

originalsouth: Addressed in cad51b0ee169995bbadb33b783017f11d8b2c06c

noamblitz: As mentioned I **do not** agree with this commit.

noamblitz: This will possibly brake installs.

originalsouth: I misread, my bad, undoing...

ammar92 (approved):

stephanie0x00:
## Checklist for QA:

### What works:
Seems to work as described from what I can see. Tried both a migration scenario and tested on a clean branch. The findings are Pending and they are resumed when I enable the 

### What doesn't work:
n/a

### Bug or feature?:
n/a

underdarknl (approved):

PR #3571, opened 2024-09-23
Make API usable by non-admin users and check specific permissions

dekkers:
### Changes

When introducing the REST API we configured that only the admin user can use it because that was the simplest way to get started. This change makes the API check the right permissions for each endpoint. 

Note that the API does not check whether there is an OrganizationMember or use those permissions. Permissions checks are on the user object. If there is are clear usecases we can look into implementing that, but for now this doesn't look necessary and saves us time.

### QA notes

Create a new user without any permissions and an API token for the user. The user shouldn't have permission to access the API. After giving specific permissions to the user it should be able to access the API endpoints that match the permission.

---

---

## Checklist for QA:


ammar92 (approved):
## Checklist for QA:

### What works:

- I was able to confirm that a user with missing permissions wasn't able to access the specific API endpoints (I used the *can view organization* permission as an example)

rocky/tools/viewsets.py
```diff
@@ -31,6 +38,9 @@ def indemnification(self, request, pk=None):
 
     @indemnification.mapping.post
     def set_indemnification(self, request, pk=None):
+        if not request.user.has_perm("tools.add_indemnification"):
+            raise PermissionDenied()
```
underdarknl, line 42: Could we list/output the missing permission here? To aid in the correct setup for the used user?

underdarknl:
Do we also need to output the various audit trail events? Or are those handled by the various methods on the models already?

underdarknl (approved):

PR #3590, opened 2024-09-26
Use stdlib instead of dateutil to parse ISO datetime

dekkers:
### Changes

In #3521 dateutil is used to parse an ISO datetime but dateutil isn't added as a dependency. Dateutil is available in dev environment because it is an indirect dev dependency, but not available in non-dev environments. See also https://github.com/minvws/nl-kat-coordination/actions/runs/11051578484/job/30701994096

An ISO datetime can be parsed using `datetime.fromisotimestamp` from the standard library so there is also no need to introduce a dependency on dateutil, so I have changed the code to use `datetime.fromisotimestamp`.

### QA notes

Check if the changes of #3521 still work.

---

---

## Checklist for QA:

### DIFF
--- a/rocky/katalogus/views/plugin_detail.py
+++ b/rocky/katalogus/views/plugin_detail.py
@@ -2,7 +2,6 @@ from datetime import datetime, timezone
 from typing import Any
 
 from account.mixins import OrganizationView
-from dateutil import parser
 from django.contrib import messages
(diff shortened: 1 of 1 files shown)


ammar92 (approved):
Nice find

underdarknl (approved):

PR #3643, opened 2024-10-08
Fix jsonb 'contained by' query

### LINKED ISSUE #3627: Fix jsonb filtering in scheduler
_Originally posted by @Souf149 in https://github.com/minvws/nl-kat-coordination/issues/3299#issuecomment-2396755698_

jpbruinsslot:
### Changes

Queries using lists as filter values when using it in combination with a 'contained_by' operator (`<@` ) were incorrectly interpreted as text, and as such would create an incorrect SQL query.

### Issue link

Closes #3627
References https://github.com/minvws/nl-kat-coordination/pull/3299#issuecomment-2396755698

### QA notes

Api functionality from scheduler to filter on the `/tasks` and `/queues` endpoint using jsonb operator are mainly addressed here. Since we don't yet use this, this pr should not intervene with current functionality. So mainly filtering of tasks in rocky should be main area where something might happen. Checking that out would probably be a good thing.

I'm trying to find some good qa examples, to test the functionality in this pr. But having issues with using curl and how characters are getting escaped and interpreted by fastapi. So stay tuned for that.

---

### Code Checklist

---

---

## Checklist for QA:


mula/scheduler/storage/filters/casting.py
```diff
@@ -43,9 +43,9 @@ def cast_expression(expression: BinaryExpression, filter_: Filter) -> BinaryExpr
         # if the value can be decoded.
         try:
             decoded_value = json.loads(filter_.value)
-            if isinstance(decoded_value, dict):
-                # If it's a JSON object, return the expression as is. We don't
-                # need to cast it.
+            # If the string is a JSON object, return the expression as is.
+            # We don't need to cast it.
+            if isinstance(decoded_value, dict | list):
```
ammar92, line 48: Just to be sure; maybe you meant to use `isinstance(decoded_value, (dict, list))` instead?

As far as I know, `isinstance` doesn't know how to handle these union types

ammar92 (commented):
Nice work. Tiny remark, looks good otherwise

mula/scheduler/storage/filters/casting.py
jpbruinsslot: Yeah I got two different linter suggestions. First the one you mentioned and second the one that is implemented. Both cases should be covered in the test. But I'll check this explicitly

jpbruinsslot:
> Nice work. Tiny remark, looks good otherwise

Thanks for reviewing 💪

mula/scheduler/storage/filters/casting.py
ammar92: Ah I see. I would strongly recommend using a class type instead as that's what the signature suggests. I haven't seen unions being used in this context before. Not sure if this is the right example, but checking such types is also difficult, e.g.:
```
>>> type(2) is int
True
>>> type(2) is (int | str)
False
```

I'll leave it up to you

ppvg: > _Changed in version 3.10_: _classinfo_ can be a [Union Type](https://docs.python.org/3/library/stdtypes.html#types-union).

https://docs.python.org/3/library/functions.html#isinstance

I don't want to get involved, just out of curiosity because I don't know much about Python: how would classes be used to solve this? In TypeScript unions are a great zero-cost/compile-time abstraction. Are they unidiomatic in Python?

dekkers: In Python `is` returns whether it is exactly the same object in memoy, so that `type(2) is (int | str)` returns False doesn't really say anything. Since Python 3.10 you can indeed use the union type in `isinstance`:

```
In [1]: isinstance(["a"], dict|list)
Out[1]: True

In [2]: isinstance({"a": "b"}, dict|list)
Out[2]: True
```

stephanie0x00:
## Checklist for QA:

### What works:
Seems to work. Scheduled a bunch of tasks and then filtered on each of the task filters to see if it would work. Also checked the logs for the scheduler, rocky, octopoes_api, octopoes_api_worker and bytes to check for any error messages that might be side effects. No obvious bugs found. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

underdarknl (approved):

PR #3662, opened 2024-10-09
Fix empty vulnerability reports

### LINKED ISSUE #3637: Vulnerability Report - Sometimes doesn't show data or text
**Describe the bug**
In some cases, the vulnerability report doesn't show any data nor helper text, while other chapters do show a 
helper text or data.

**To Reproduce**

1. Create vulnerability report for multiple oois
2. Check the chapters

**Expected behavior**
I would expect every chapter to contain at least a helper text.

**Screenshots**
Image

**OpenKAT version**
main

madelondohmen:
### Changes
This PR adds a template tag to the vulnerability report, so it doesn't show empty reports anymore.

### Issue link
Closes #3637

### Demo
afbeelding

### QA notes
- Create a normal vulnerability report with multiple OOIs (with and without vulnerabilities)
- Check if the reports show report data if expected and "No vulnerabilities have been found on this system." if no vulnerabilities are expected. 
- Do the same for an aggregate report

---

### Code Checklist

---

---

## Checklist for QA:


noamblitz (approved):
As discussed, I think that at the moment this is the cleanest solution. Its a bit much logic for a template, but adding it to the report data would be messy and not backwards complatible.

rocky/tools/templatetags/findings_count.py
```diff
@@ -0,0 +1,13 @@
...
+
+register = template.Library()
+
+
+@register.filter
+def total_findings(data: dict[str, Any]) -> int:
+    total_findings = 0
+    for ip in data.values():
+        total_findings += int(ip["summary"]["total_findings"])
+    return total_findings
```
Rieven, line 13:
```suggestion
   return sum([ int(ip["summary"]["total_findings"]) for ip in data.values()])
```

ammar92: Nice suggestion @Rieven 

An even better approach would be using a generator expression instead. This saves an extra iteration by `sum`, because you don't have to construct an entire list to loop over it again.
```python3
return sum((int(ip["summary"]["total_findings"]) for ip in data.values()))
```

rocky/tools/templatetags/findings_count.py
```diff
@@ -0,0 +1,13 @@
+from typing import Any
+
+from django import template
+
+register = template.Library()
+
+
+@register.filter
+def total_findings(data: dict[str, Any]) -> int:
```
ammar92, line 9: Just out of curiosity: although this looks like a useful filter, it's currently only being used bye one template to check for emptiness?

ammar92 (commented):
Nice fixes. Tiny suggestion

rocky/tools/templatetags/findings_count.py
madelondohmen: Yes, it is only used by one template. As @noamblitz said, it's the best fix for now. Changing it in the report.py file would unfortunately not be backward compatible.

madelondohmen: I've been looking through the code and I only then notices the file report_extra.py. This is a better place for the filter than in a seperate file, so I've moved it to there.

underdarknl: I think @ammar92 means this isn't generic enough to warrant a template function, and instead should just be a part of the view / logic that generates the report-data / context-data.
Somewhere in between, this summing could be a python property of the object that we are sending over which would be accessibly from the template, and only executed when needed.

stephanie0x00:
## Checklist for QA:

### What works:
Works. There are some other bugs around the vulnerability report, but those are picked up in different tickets (like #3679) as these problems also occur on main.

### What doesn't work:
n/a

### Bug or feature?:
n/a

rocky/tools/templatetags/findings_count.py
noamblitz: You are right @underdarknl that this would be a bit better during generation of the report, but this method makes it backwards compatible with older reports.

noamblitz: The object is just a json of which it's unknown whether this part of the report exists

underdarknl (approved):

Issue #3598, opened 2024-09-30
Object reference not found in Octopoes


stephanie0x00:
**Describe the bug**
When scanning mispo.es I get an object reference not found error in the boefje logs for the finding 'KAT-No-certificate'. 

**To Reproduce**

1. Go through onboarding for mispo.es.
2. Enable nmap tcp + the SSLCertificates
3. Wait

**Expected behavior**
No error messages in the logs when boefjes are enabled.

**Screenshots**

**OpenKAT version**
main

**Additional context**
```
boefje-1  | {"event": "Set status to TaskStatus.COMPLETED in the scheduler for task[id=e44582f0-8a77-43ec-986b-e087e1b4b093]", "level": "info", "timestamp": "2024-09-30T11:40:59.953143"}
boefje-1  | HTTP Request: PATCH http://scheduler:8000/tasks/eba7db28-2f4c-4815-a0f5-a7d234c7a28d "HTTP/1.1 200 OK"
boefje-1  | {"event": "Handling boefje kat-finding-types[task_id=eba7db28-2f4c-4815-a0f5-a7d234c7a28d]", "level": "info", "timestamp": "2024-09-30T11:40:59.961040"}
boefje-1  | HTTP Request: GET http://octopoes_api/aa/object?reference=KATFindingType%7CKAT-NO-CERTIFICATE&valid_time=2024-09-30%2011%3A41%3A00.090250%2B00%3A00 "HTTP/1.1 404 Not Found"
boefje-1  | {"exc_info": true, "event": "An error occurred handling scheduler item[id=eba7db28-2f4c-4815-a0f5-a7d234c7a28d]", "level": "error", "timestamp": "2024-09-30T11:41:00.099727"}
boefje-1  | Traceback (most recent call last):
boefje-1  |   File "/app/boefjes/octopoes/connector/octopoes.py", line 51, in _verify_response
boefje-1  |     response.raise_for_status()
boefje-1  |   File "/usr/local/lib/python3.11/site-packages/httpx/_models.py", line 763, in raise_for_status
boefje-1  |     raise HTTPStatusError(message, request=request, response=self)
boefje-1  | httpx.HTTPStatusError: Client error '404 Not Found' for url 'http://octopoes_api/aa/object?reference=KATFindingType%7CKAT-NO-CERTIFICATE&valid_time=2024-09-30%2011%3A41%3A00.090250%2B00%3A00'
[...]

PR #3686, opened 2024-10-14
Silence KATFindingType not found error in JobHandler

originalsouth:
### Changes

Silence KATFindingType not found error in JobHandler in an attempt to propose a solution to #3598
(The issue is likely with KATFindingType because the get deleted and affirmed continuously) 

### Issue link

#3598

### QA notes
Mimic #3598

---

### Code Checklist

---

---

## Checklist for QA:

### DIFF
--- a/boefjes/boefjes/job_handler.py
+++ b/boefjes/boefjes/job_handler.py
@@ -106,7 +106,7 @@ class BoefjeHandler(Handler):
                 )
             except ObjectNotFoundException as e:
                 if reference.class_type == KATFindingType:
-                    logger.exception("Object {%s} not found in Octopoes", e)
+                    logger.error("Object %s not found in Octopoes", reference)
                     return
                 else:
                     raise ObjectNotFoundException(f"Object {reference} not found in Octopoes") from e
--- a/boefjes/boefjes/job_handler.py
+++ b/boefjes/boefjes/job_handler.py
@@ -22,7 +22,6 @@ from octopoes.api.models import Affirmation, Declaration, Observation
 from octopoes.connector.octopoes import OctopoesAPIConnector
 from octopoes.models import Reference, ScanLevel
 from octopoes.models.exception import ObjectNotFoundException
-from octopoes.models.ooi.findings import KATFindingType
 
 MIMETYPE_MIN_LENGTH = 5  # two chars before, and 2 chars after the slash ought to be reasonable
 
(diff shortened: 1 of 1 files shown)


boefjes/boefjes/job_handler.py
```diff
@@ -104,7 +105,11 @@ def handle(self, boefje_meta: BoefjeMeta) -> None:
                     reference, valid_time=datetime.now(timezone.utc)
                 )
             except ObjectNotFoundException as e:
-                raise ObjectNotFoundException(f"Object {reference} not found in Octopoes") from e
+                if reference.class_type == KATFindingType:
+                    logger.exception("Object {%s} not found in Octopoes", e)
+                    return
+                else:
+                    raise ObjectNotFoundException(f"Object {reference} not found in Octopoes") from e
```
dekkers, line 112:
```suggestion
                logger.info("Can't run boefje because OOI does not exist anymore", boefje_id=boefje_meta.boefje.id, ooi=reference, task_id=boefje_meta.id)
                return
```

The problem is not specific for KATFindingType, a boefje for every object type can be scheduled and then be deleted before the runner can fetch the input OOI. So I don't see why we should special case `KATFindingType`.

We also shouldn't use logger.exception because that would still log the exception just like we did before. What needs to be changed is that we should not log an exception because nothing actually goes wrong.

originalsouth: The idea here is to not crash on objects that we "expect" to have been deleted; these are typically objects like KATFindingTypes whom are being affirmed while being deleted. As we have no access to the origin that is being to be yielded we can deduce from the object type that in this case deletion is expected. For other case we might want to crash because we have found abnormal behavior?

stephanie0x00:
## Checklist for QA:

### What works:
Looks good. The error doesn't show up in the logs and the findings look good.

### What doesn't work:
n/a

### Bug or feature?:
n/a

boefjes/boefjes/job_handler.py
dekkers: The race condition can happen for every object. Every object can be deleted and if the object can be used as input OOI, it is possible that the deletion happens at the same moment the scheduler sends the OOI as input to the boefje runner.

That we see it more often and/or can reproduce it with KATFindingType doesn't really matter, it's just nature of race conditions that you don't always see them or can reproduce them.

originalsouth: Right, but this does not solve the question on whether you would like to crash in certain cases.... In any case, I'll yield to your insight on this matter.

dekkers (approved):

PR #3703, opened 2024-10-18
Fix bash equalsigns in entrypoints

trideeindhoven:
Simple fix for bash equal signs in entrypoint.sh


dekkers:
What was wrong with the equal signs? As far as I understand it `=` and `==` are equivalent in bash.

cloudgitaware:
Using the same equal convention is a good thing i believe:
`if [ "$DATABASE_MIGRATION" = "1" ] || [[ $DATABASE_MIGRATION == "true" ]]; then`
here both options are used...

ammar92:
After reading [this](https://stackoverflow.com/a/2237103) I'm not sure the implemented changes are the right approach and why the current implementation didn't work for you. I'll play around with this and test it out tomorrow just to make sure

Issue #3697, opened 2024-10-17
Refactor: Rocky KATalogus clients


Donnype:
Some APIs of those clients are not organization-specific though, and hence forcing an organization as a property of the API client becomes a bit ugly. I believe this client is not the place to force handing over the organization as that makes the clients do to much and make them not reusable in context where we should have full access, such as with the runner. To have a nice layered design, we could introduce a layer where we properly manage access, such as a new client class that can only call organization specific methods of the base client.

_Originally posted by @Donnype in https://github.com/minvws/nl-kat-coordination/pull/3666#discussion_r1797747056_

underdarknl:
I'd propose to split the client into two classes.
The base class that handles generic (eg, non orga or user bound) requests, which can be initiated without providing the organization code. This layer could also handle most generic functonality (session building, http requests as a private method) of the expected exceptions (eg, timeouts, connection errors) and streamline them into Exception classes that subclass from the Generic KatalogusException or BytesException.

The second class subclasses from this first class and adds the organization or user arguments to the init. It also adds the specific methods that only work when an organzation or client is required. This layer also adds permission checks if needed, and the needed exceptions. (eg, dot bleed a rw http 404 or http 403 back to the user but instead throw something like a KatalogusPermissionDenied or KatalogusObjectNotFound error. (which can be subclassed from the KatalogusException AND the generic objectNotfoud classes available in Django. This helps catching these exceptions everywhere upstream.

PR #3717, opened 2024-10-23
Refactor KATalogus client in Rocky

Donnype:
### Changes

- Handles client errors generically in a base client that can be organization-agnostic for some methods. 
- Introduces a new intermediate "adapter" called KATalogus that forces you to set an organization.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/3697

### Demo

-

### QA notes

Please verify that we can still properly create organizations, add plugins, settings and enable/disable them from Rocky!

---

### Code Checklist

---

---

## Checklist for QA:


rocky/katalogus/client.py
```diff
@@ -47,8 +47,13 @@ class Plugin(BaseModel):
...
     description: str | None = None
     related: list[str] = Field(default_factory=list)
-    enabled: bool
     type: str
+    # TODO: this is the only field making a Plugin organization-specific. If we could separate the use of the enabled
+    #  field from other uses, we would be able to drop the organization_code as an argument in a lot of places, hence
+    #  simplifying the usage of the KATalogus for installation-wide operations (as plugins are not specific to an
+    #  organization). One could argue that having this field here should mean we need an organization field as well to
+    #  make sense out of it: for which organization is this plugin in fact enabled?
+    enabled: bool
```
underdarknl, line 56: Looking at the future, I can see two types of plugins. Those (base plugins) listed on the remote repos (orga agnostic, not enabled), and those copied to a local organizations XTDB namespace, in which they can be enabled, forked, configured etc.

By doing this, it also becomes easier to manage variants and ownership of these installed plugins per organization, as deleting/changing them only affects the local organization instead of all organizations in the same install.

Sharing the local changes back to other organizations (local or remote) can be done by providing the base plugins either to the other organizations or by uploading them to the remote or public repos.

rocky/katalogus/client.py
```diff
@@ -70,6 +75,17 @@ class Boefje(Plugin):
     def serialize_consumes(self, consumes: set[type[OOI]]):
         return {ooi_class.get_ooi_type() for ooi_class in consumes}
 
+    @field_validator("boefje_schema")
+    @classmethod
+    def json_schema_valid(cls, schema: dict) -> dict:
+        if schema is not None:
```
underdarknl, line 81:
```suggestion
        if schema is None:
          return None
```

rocky/katalogus/client.py
```diff
@@ -70,6 +75,17 @@ class Boefje(Plugin):
...
+            try:
+                Draft202012Validator.check_schema(schema)
+            except SchemaError as e:
+                raise ValueError("The schema field is not a valid JSON schema") from e
```
underdarknl, line 85:
```suggestion
        if schema is None:
            return None
        try:
            Draft202012Validator.check_schema(schema)
        except SchemaError as e:
            raise ValueError("The schema field is not a valid JSON schema") from e
```

rocky/katalogus/client.py
```diff
@@ -121,181 +144,200 @@ def __init__(self):
...
+        raise KATalogusHTTPStatusError(error) from error
 
-    def create_organization(self, name: str):
-        response = self.session.post("/v1/organisations/", json={"id": self.organization, "name": name})
-        response.raise_for_status()
 
-        logger.info("Created organization", name=name)
+class KATalogusClient:
+    def __init__(self, base_uri: str):
+        self.session = httpx.Client(base_url=base_uri, event_hooks={"response": [verify_response]})
```
underdarknl, line 168: nice!

rocky/katalogus/client.py
```diff
@@ -121,181 +144,200 @@ def __init__(self):
...
-    def get_plugin_schema(self, plugin_id: str) -> dict | None:
-        plugin_id = quote(plugin_id)
-        response = self.session.get(f"{self.organization_uri}/plugins/{plugin_id}/schema.json")
-        response.raise_for_status()
+        logger.info("Created organization", name=organization.name)
 
-        schema = response.json()
-        if not schema:
-            return None
+    def delete_organization(self, organization_code: str):
```
underdarknl, line 188: The following methods should be available only on the Katalogus Class I'd say. We could then allow the user to instantiate a Katalogus for a given organization *only* if they have the correct rights / memberships, and make sure we have only one codepath to get there.

I'd expect a factory on the user object that allows for the instantiation of the Katalogus instance bound to the organization requested.

rocky/katalogus/client.py
```diff
@@ -121,181 +144,200 @@ def __init__(self):
...
+
+class KATalogus:
+    """
+    An adapter between the full KATalogusClient and the organization-specific context of most views. This restricts
+    the set of available methods on the KATalogusClient and simplifies the interface by "currying" the organization
+    into the relevant methods as an argument. We should use this class in the views to avoid making calls exposing
+    information from other organizations users are not allowed to see.
+    """
+
+    def __init__(self, katalogus_client: KATalogusClient, organization_code: str):
```
underdarknl, line 292: Why not subclass from KATalogusClient, and add the functionality that is specific to the organization here?

rocky/katalogus/utils.py
```diff
@@ -5,4 +5,6 @@
 
 
 def get_enabled_boefjes_for_ooi_class(ooi_class: type[OOI], organization: Organization) -> list[Boefje]:
-    return [boefje for boefje in get_katalogus(organization.code).get_enabled_boefjes() if ooi_class in boefje.consumes]
+    katalogus = get_katalogus(organization.code)
```
underdarknl, line 8: Could we provide a property on the view that provides us with the katalogus?
This property can be added in the Organization Mixin, and uses the user to instantiate the Katalogus based on their memberships / rights and the organization from the request/url.

rocky/katalogus/client.py
Rieven: Can we not pass the member as argument, as a member has many orgs from the one too many relationship. Then we can list all orgs that belongs to member, then we can check if a specific org is in this list.

rocky/katalogus/client.py
Donnype: I think at least we should try to work with "base plugins" while/where we can since managing cross-organization plugins will be significantly more work

rocky/katalogus/client.py
Donnype: Because that would still expose the non-organization-specific methods to users of the class. It also implies that this class has the same interface of the parent, but we actually change the interface because we remove the organization as an argument. These were the main two considerations for not extending the parent class and creating a bit of "duplication" although not literal duplication.

rocky/katalogus/client.py
Donnype: I was also looking into passing a user to the base client and do permission checking. However the complexity involved is that the permissions are mostly defined per _method_, not per instance.  Do we check for `"tools.can_scan_organization"`? Or `"tools.can_set_katalogus_settings"`? Perhaps `"tools.can_enable_disable_boefje"` as well? I could add those checks on the method calls instead, then probably add a helper factory-method to the OrganizationView.

Currently, most permissions are defined at the View level, which we should maintain properly to treat these new checks as a fallback. (On a side note: having the option to check permissions before moving forward is in my eyes another benefit of not extending the KATalogusClient as a parent and give it a bit more responsibility, having different behavior than a client.)

rocky/katalogus/utils.py
Donnype: Yes

Donnype:
@stephanie0x00 I've made several changes at this point and cannot reproduce this. Perhaps you could retry this? I did notice the error refers to an onboarding endpoint?

rocky/katalogus/client.py
```diff
@@ -117,185 +143,248 @@
...
+        logger.info("Deleted organization", organization_code=organization_code)
 
-        return None
+    def get_plugins(self, organization_code: str, **params) -> list[Plugin]:
+        response = self.session.get(f"/v1/organisations/{organization_code}/plugins", params=params)
+
+        return [parse_plugin(plugin) for plugin in response.json()]
+
+    def get_plugin(self, organization_code: str, plugin_id: str) -> Plugin:
+        response = self.session.get(f"/v1/organisations/{organization_code}/plugins/{quote(plugin_id)}")
```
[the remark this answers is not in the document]
underdarknl: We should quote or sanitize the organization_code in this url.
Better yet, we should provide this function only for an instance of the client bound to a user (and as such can check for the user's permissions for the requested orga), or do we also need this functionality at locations where we dont have a user?

Donnype:
@stephanie0x00 (hopefully) fixed!

rocky/katalogus/client.py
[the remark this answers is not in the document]
Donnype: I think the KATalogus wrapper serves this purpose, and I (currently) like this service-based design better than forcing backend-choosing-logic behind model specific methods. It would require us to either mock or juggle some global state in testing context (that Django gets away with because they put the necessary effort into managing it). Also: there is nothing preventing me from importing an Organization class, instantiate it with an arbitrary id string, and then access forbidden data through this object, just like there is nothing preventing you from using the "wrong" client and access the wrong data. In essence it's hence a design (pattern) choice, and I think the service-based approach has some advantages of which consistency with other services (e.g. Octopoes repositories, the Bytes implementation, the KATalogus) is not the least.

Perhaps a subsequent issue and discussion topic would be a good idea?

stephanie0x00:
## Checklist for QA:

### What works:
Looks good now. Onboarding works, can enable and disable boefjes and normalizers. Tasks are scheduled and complete. Findings are created. Generated a normal report and that works too. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

underdarknl (approved):

PR #3721, opened 2024-10-23
Rework workflow for variable python version, add python 3.11

sigio:
### Changes

Update packaging-workflow used by RDO to build against python 3.10 and 3.11

---

### Code Checklist

- [X] This PR only contains functionality relevant to the issue.
- N/A I have written unit tests for the changes or fixes I made.
- N/A I have checked the documentation and made changes where necessary.
- [X] I have performed a self-review of my code and refactored it to the best of my abilities.

### DIFF
--- a/.github/workflows/build-rdo-package.yml
+++ b/.github/workflows/build-rdo-package.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix: 
-        python_version: ['3.8','3.10','3.11']
+        python_version: ['3.10','3.11']
 
     # Generic bits
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-rdo-package.yml
```diff
@@ -26,7 +29,7 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.10"
+          python-version: ${{ matrix.python_version }}
           cache: "pip"
```
Donnype, line 33: Because of this change/step setting the right version, I believe you could probably replace all _invocations_ of `python${{ matrix.python_version }}` with just `python`. But I'll leave it up to you :+1:

Donnype (dismissed):

.github/workflows/build-rdo-package.yml
sigio: Probably.... but it worked / packages are build with this (in another orga-repo for testing)

Donnype (approved):

dekkers (approved):

PR #3755, opened 2024-10-29
Fix auth token middleware with wrong format header

### LINKED ISSUE #3754: Exception when authorization header has wrong format
```
rocky-1  | Internal Server Error: /api/v1/organization/
rocky-1  | Traceback (most recent call last):
rocky-1  |   File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
rocky-1  |     response = get_response(request)
rocky-1  |                ^^^^^^^^^^^^^^^^^^^^^
rocky-1  |   File "/app/rocky/rocky/middleware/auth_token.py", line 12, in middleware
rocky-1  |     user, token = authenticator.authenticate(request)
rocky-1  |     ^^^^^^^^^^^
rocky-1  | TypeError: cannot unpack non-iterable NoneType object
```

dekkers:
### Changes

Fix exception when the authorization header has the wrong format

### Issue link

Fixes #3754

### QA notes

Check if API and token auth still works. The problem happened when authorization header didn't have a token:

```
http GET http://127.0.0.1:8000/api/v1/report/ "Authorization: notatoken"
```

---

---

## Checklist for QA:


ammar92 (approved):

rocky/tests/test_api.py
```diff
@@ -11,3 +11,9 @@ def test_api_2fa_enabled(client, settings, admin_user):
 
     response = client.get("/api/v1/organization/", headers={"Authorization": f"Token {token}"})
     assert response.status_code == 200
+
+
+# Regression test for https://github.com/minvws/nl-kat-coordination/issues/3754
+def test_auth_header_wrong_format(client, settings, admin_user):
+    response = client.get("/api/v1/organization/", headers={"Authorization": "Not a token"})
+    assert response.status_code == 401
```
ammar92, line 19: I suggest also (partially) checking the error message/ response. During QA I noticed variations in messages for this scenario compared to cases where for example an incorrect token is passed

ammar92 (approved):
## Checklist for QA:

### What works:

- Seems to work as advertised
- Authentication still works
- API still works

rocky/rocky/middleware/auth_token.py
```diff
@@ -9,12 +9,13 @@ def middleware(request):
         if not request.user.is_authenticated and "authorization" in request.headers:
             authenticator = TokenAuthentication()
             try:
-                user, token = authenticator.authenticate(request)
+                user_and_token = authenticator.authenticate(request)
```
underdarknl, line 12: Why not just use _ for the token as it isn't used anywhere in this context?

dekkers: That would have been better in the old code, but we can't unpack the return value if it is None, so we have to assign it to a variable first to prevent the exception.

rocky/tests/test_api.py
dekkers: What the goal of this test is that we don't regress and make sure that this code path keeps doing the right thing and will return a 401. We could also test the message, but I think it's not really that important to spend time on.

underdarknl (approved):

PR #3774, opened 2024-10-31
add more precise error when rabbitMQ cannot be reached when storing raw files

underdarknl:
### Changes

Specifically catches the rabbitMQ errors which might occur if (even after the retries) it still cannot be reached.
The original 500 does not show the stacktrace in de logs, and as such there is no easy way for the user to know rabbitMQ might be at fault.

### Issue link

```
bytes-1                | $IP:$PORT - "POST /bytes/raw?boefje_meta_id=$SOME_UUID HTTP/1.1" 500
bytes-1                | fastapi.exceptions.HTTPException: 500: Could not save raw data
```

Closes ...

### Demo

### QA notes

---

### Code Checklist

---

---

## Checklist for QA:

### DIFF
--- a/bytes/bytes/api/router.py
+++ b/bytes/bytes/api/router.py
@@ -1,6 +1,7 @@
 from base64 import b64decode
 from uuid import UUID
 
+import pika.exceptions
 import structlog
 from cachetools import TTLCache, cached
 from fastapi import APIRouter, Depends, HTTPException, Query
@@ -190,6 +191,11 @@ def create_raw(
                 raw_data=RawDataMeta(id=raw_id, boefje_meta=raw_data.boefje_meta, mime_types=raw_data.mime_types),
             )
             event_manager.publish(event)
+        except pika.exceptions.AMQPError:
+            logger.exception("Error sending 'new raw data' event to RabbitMQ")
+            raise HTTPException(
+                status_code=codes.INTERNAL_SERVER_ERROR, detail="Error sending 'new raw data' event to RabbitMQ"
+            ) from error
         except Exception as error:
             logger.exception("Error saving raw data")
             raise HTTPException(status_code=codes.INTERNAL_SERVER_ERROR, detail="Could not save raw data") from error


bytes/bytes/api/router.py
```diff
@@ -190,6 +191,11 @@ def create_raw(
                 raw_data=RawDataMeta(id=raw_id, boefje_meta=raw_data.boefje_meta, mime_types=raw_data.mime_types),
             )
             event_manager.publish(event)
+        except pika.exceptions.AMQPError as error:
+            logger.exception("Error sending 'new raw data' event to RabbitMQ")
+            raise HTTPException(
```
Donnype, line 196: Is this something we want to expose to someone using the API?

Donnype (approved):

bytes/bytes/api/router.py
underdarknl: I'd say, yes? Its an authenticated user, and their request failed.

bytes/bytes/api/router.py
```diff
@@ -190,6 +191,11 @@ def create_raw(
                 raw_data=RawDataMeta(id=raw_id, boefje_meta=raw_data.boefje_meta, mime_types=raw_data.mime_types),
             )
             event_manager.publish(event)
+        except pika.exceptions.AMQPError as error:
```
ammar92, line 194: I'm not sure if this works, because Byte's `RabbitMQEventManager.publish` already handles the `AMQPError` exceptions and doesn't reraise them. That could explain the missing stack traces in the logs. However, while the `RabbitMQEventManager.publish` handles these errors, it could potentially still produce an `AMQPError` due to subsequent API calls via .e.g `basic_publish` or `queue_declare` within the exception handler. But in that case you would still see the warning "Reconnected to RabbitMQ because connection was closed" in the logs before the exception.

But imo the API router shouldn't be dealing with AMQP related stuff. As part of the separation of concerns principle, it's the EventManager that needs to be dealing with this technology details

PR #3786, opened 2024-10-31
Add configurable httpx request timeout and increase default

dekkers:
### Changes

This makes the timeout for outgoing requests configurable and increases the default timeout to 30.

### QA notes

This touches most requests between services, so we should just test if OpenKAT in general still works.

---

---

## Checklist for QA:


ammar92 (commented):
Nice work 👍 Just a few things:
- The setting for Rocky seems to be named differently, why's that?
- I suggest using a nullable float type for the timeout, since that's what `httpx` and most other request libraries accept. A float gives a bit more freedom and an int isn't necessarily easier, in fact you could still pass an integer and it will be implicitly cast to a float. And nullable for passing `None`s, which explicitly means no timeout and could still be a useful setting.

underdarknl:
I wonder if there is any good reasoning behind the magic 3 and 6 retries.
In any case we could make those configurable too I'd say.

stephanie0x00:
## Checklist for QA:

### What works:
Seems to work. Enabling boefjes works, reports can be scheduled and opened. Tasks complete and run without failing. Not seeing any errors or warnings in the logs from rocky or octopoes. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

PR #3809, opened 2024-11-07
Add start time to scheduled reports

### LINKED ISSUE #3738: Timestamp of scheduled reports
**Describe the issue**
During QA of #3701 a discussion started on how timestamps of scheduled reports should work. At the day of writing it's October 28th 2024, which helps to interpret the screenshots below. 

Snippet from the discussion in this ticket: 
> Scheduled a report with start date today, the report was scheduled for 24 oct, midnight. I expect that to be coming night, and not previous night, report was created though. That signals to me that the start date is not correctly passed or something like that?

With the response: 

>That's what you'd expect, right? If you want to schedule it for today, then it should be scheduled for 25-10-2024 00:00, not for the next day (26-10-2024 00:00). Otherwise you should select the 26th.

**The screenshots below show how it currently works for that PR:**

We schedule a report for todays date. According to the current text it should run at midnight, as we use the date as the reference date for the report. 

image

When the report is scheduled. It shows up as shown below with the 28th at midnight setting: 

image

However, when waiting some time, the report is picked up and processed by the scheduler: 

image

So the question now is: how is it supposed to work from a user point of view?

madelondohmen:
### Changes
This PR adds a start time to the scheduled reports. 
When scheduling a report, a user can now choose a start time as well. (Default is the current time, UTC)

### Issue link
Closes #3738

### Demo
afbeelding

afbeelding

### QA notes
- Schedule a report
- Change the scheduled date and time
- Check if everything is as expected on the Scheduled Reports page

---

### Code Checklist

---

---

## Checklist for QA:


rocky/rocky/views/scheduler.py
```diff
@@ -262,28 +267,31 @@ def run_boefje_for_oois(self, boefje: Boefje, oois: list[OOI]) -> None:
...
-                "yearly": f"0 0 {day} {month} *",  # Recurres every year on the {day} of the {month} at 00:00
+                "daily": f"{minute} {hour} * * *",  # Recurres every day at the selected time
+                "weekly": f"{minute} {hour} * * {week}",  # Recurres every week on the {week} at the selected time
+                "yearly": f"{minute} {hour} {day} {month} *",
+                # Recurres every year on the {day} of the {month} at the selected time
             }
 
             if 28 <= day <= 31:
-                cron_expr["monthly"] = "0 0 28-31 * *"
+                cron_expr["monthly"] = f"{minute} {hour} 28-31 * *"
```
underdarknl, line 290: We could just set this to 28th for now. To make sure we don't execute on 28,29,30 and 31.

madelondohmen: I've created a new issue for this: #3810

Rieven: You can use L notation for last day of the month

underdarknl: Only *IF* our cron lib understands the L.
And then we should still check if the cosen date is the last day or not.

rocky/reports/forms.py
ammar92, line 1: I'm a bit confused about how forms are used here. Django Forms are a good choice for generating forms in templates, but it looks like the form logic, like validation and cleanup (such as checking the format, but also return a parsed type such as a `datetime`), is handled in the view instead. This adds unnecessary complexity to the view. Moving this form logic into the form would make the view cleaner and less complex.

Also, since the template optically shows the start date, start time and recurrence as a form, why can't this be a single Django Form with these fields instead of 3 forms?

ammar92 (dismissed):

rocky/reports/forms.py
```diff
@@ -47,25 +57,32 @@ class ReportScheduleStartDateForm(BaseRockyForm):
...
 
-
-class ReportRecurrenceChoiceForm(BaseRockyForm):
-    choose_recurrence = forms.ChoiceField(
-        label="",
-        required=False,
-        widget=forms.RadioSelect(attrs={"class": "submit-on-click"}),
-        choices=(("once", _("No, just once")), ("repeat", _("Yes, repeat"))),
-        initial="once",
+    start_time = forms.TimeField(
```
Rieven, line 60: there is also `models.DateTimeField` to have both Date and time together in the widget.

rocky/reports/forms.py
```diff
@@ -47,25 +57,32 @@ class ReportScheduleStartDateForm(BaseRockyForm):
...
         choices=[("daily", _("Daily")), ("weekly", _("Weekly")), ("monthly", _("Monthly")), ("yearly", _("Yearly"))],
     )
 
+    def clean(self):
+        cleaned_data = super().clean()
+        start_date = cleaned_data.get("start_date")
+        start_time = cleaned_data.get("start_time")
+
+        if start_date and start_time:
+            start_datetime = datetime.combine(start_date, start_time)
```
Rieven, line 81: Don't need to add this if using `models.DateTimeField` to have both Date and time together in the widget.

rocky/reports/forms.py
Rieven: image

Rieven: Are we converting the local time to UTC? I think it is strange that when using now time time the user needs to think of 2 hours in the past ?

rocky/reports/forms.py
madelondohmen: See other comment. We'll keep those separated.

underdarknl (approved):

Issue #3811, opened 2024-11-07
Delete many in octopoes takes any string argument


Rieven:
`self.octopoes_api_connector.delete_many(report_references, datetime.now(timezone.utc))`

`delete_many` with any type of ooi reference will not throw an exception.

For example:
`self.octopoes_api_connector.delete_many(["hey"], datetime.now(timezone.utc))` will fail silently

PR #3813, opened 2024-11-07
Do not fail silently when deleting non-existing objects in octopoes

Donnype:
### Changes

This changes the API behavior to fail when we delete references that do not exist.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/3811

### Demo

See the updated integration test!

### QA notes

Please verify that you do not get a 200 back when using the delete-many API in Octopoes on non-existing references!

---

### Code Checklist

---

---

## Checklist for QA:


ammar92 (commented):
Nice work 👍 While this approach works, having two delete methods (for removing single entities) is unusual. A more common approach is to use a single delete method with an argument to control the behavior on failure (e.g., `fail_on_not_found: bool = True` or `on_error: Literal["ignore", "fail"] = "fail"`). This keeps the API simpler and avoids duplication. Setting a default value is also convenient for users who may not need to customize this behavior and in this case it also helps dealing with backwards compatibility

Donnype:
@ammar92 I also considered this, but I expect a delete method to fail and thought this new method mirrored the djangoesque approach. A matter of taste perhaps at this point.

ammar92:
Good points and thanks for clearing up. Here's some thoughts I still have about this.

> @ammar92 I also considered this, but I expect a delete method to fail and thought this new method mirrored the djangoesque (`update_or_create`, `get_or_create`) approach and hence is not that unusual.

I can see where the `get_or_create` or `update_or_create` comes from, but I think delete operations differ. Django’s approach is typically oriented around creating or updating data if it doesn’t exist. In Django, you can also only delete (or apply a delete method) on existing data, meaning you would only be able to use delete if a query set actually returns anything.

But regardless of that, there's no specific need of borrowing Django conventions since this module is not related to Django.

> I disagree that your suggestions avoids more duplication as I do not see any duplication being introduced.

Even if no code is duplicated, duplicating conceptual functionality adds maintenance burden. Any changes in delete logic or documentation now have to be mirrored between two methods which could add unnecessary upkeep.

> I also disagree that the API would be more simple per se, since I could argue that a method of which the interface changes (as the exceptions thrown is in my eyes part of the interface) based on the arguments is more difficult than two methods of which the failure modes are well-defined.

The API would be simpler in the sense that users won’t need to decide between two similar delete methods, reducing cognitive load. A single, flexible delete method can handle all scenarios while keeping the API surface lean.

Having a parameter-based approach allows for flexible error handling, and similar patterns are found in popular libraries (see for example panda's way of dealing with n/a values, coercion errors and existence/ condition checks). Changing behavior based on parameters (like `fail_on_not_found`) doesn’t introduce complexity as long as the behavior is well documented and of course if given a good and descriptive argument name. Users have a predictable method signature and just need to decide on error handling. This also aligns with Python’s philosophy of being explicit, using arguments to clarify intent instead of creating potentially overlapping methods

> An interesting discussion on this: https://softwareengineering.stackexchange.com/questions/362900/parameter-to-control-whether-to-throw-an-exception-or-return-null-good-practic 🤔

On its own, like you said, an interesting discussion, but imo not too relevant for this. This topic mostly discusses whether to throw an exception based on a control parameter, or to return a `None` or other placeholders instead. In our case, a delete method doesn't return anything because it's a binary operation (either an item exists to delete or it doesn’t). And although the question is universal and language agnostic, the answers tend to describe specific languages and their conventions and not necessarily how Python-land would deal with this.

Anyways it's obviously also a matter of taste, but ultimately in this case it comes down to the question whether we accept silent errors and what kind of style and conventions do we want to follow. Imho most silenced errors, and specifically this case, is a bug. But perhaps an interesting topic for the discussion meet or we could take this offline with some coffee soon?

Thanks again.

Donnype:
>  can see where the get_or_create or update_or_create comes from, but I think delete operations differ. Django’s approach is typically oriented around creating or updating data if it doesn’t exist. In Django, you can also only delete (or apply a delete method) on existing data, meaning you would only be able to use delete if a query set actually returns anything.
>
> But regardless of that, there's no specific need of borrowing Django conventions since this module is not related to Django.

Those are indeed design choices of Django, no right or wrong, but just a quick commonly known use case of the choice to add a method instead of parametrizing (given you called the approach of a helper method instead of adding a parameter unusual). On a side note, Django actually does not provide the option to not throw an error, which to me seems like an argument that parametrize a delete method is not that common.

> Even if no code is duplicated, duplicating conceptual functionality adds maintenance burden. Any changes in delete logic or documentation now have to be mirrored between two methods which could add unnecessary upkeep.
>
> The API would be simpler in the sense that users won’t need to decide between two similar delete methods, reducing cognitive load. A single, flexible delete method can handle all scenarios while keeping the API surface lean.
>
> Having a parameter-based approach allows for flexible error handling, and similar patterns are found in popular libraries (see for example panda's way of dealing with n/a values, coercion errors and existence/ condition checks). Changing behavior based on parameters (like fail_on_not_found) doesn’t introduce complexity as long as the behavior is well documented and of course if given a good and descriptive argument name. Users have a predictable method signature and just need to decide on error handling. This also aligns with Python’s philosophy of being explicit, using arguments to clarify intent instead of creating potentially overlapping methods

I think that the two-method-approach actually could save you from having to look at the documentation in most cases here because the method names are expressive. Again a tradeoff. (There is also no need to duplicate changes in the delete logic since the `delete_if_exists` calls the `delete` method, and that could also be the only documentation needed.) I could also argue that the new method's name is actually more explicit because it expresses intend even before we need to look at arguments.

> On its own, like you said, an interesting discussion, but imo not too relevant for this. This topic mostly discusses whether to throw an exception based on a control parameter, or to return a None or other placeholders instead. In our case, a delete method doesn't return anything because it's a binary operation (either an item exists to delete or it doesn’t). And although the question is universal and language agnostic, the answers tend to describe specific languages and their conventions and not necessarily how Python-land would deal with this.

I thought the points about the errors thrown were relevant to show there's no right or wrong in (both) the approaches. 
 
> Anyways it's obviously also a matter of taste, but ultimately in this case it comes down to the question whether we accept silent errors and what kind of style and conventions do we want to follow. Imho most silenced errors, and specifically this case, is a bug. But perhaps an interesting topic for the discussion meet or we could take this offline with some coffee soon?

Yes the silenced error in the API is now removed with this PR 👌. I wanted to challenge you on the statements mainly because I think we should be careful stating certain approaches are unusual, less simple or do not deal with backward compatibility, when they are often not binary and in this case imho especially a grey area.

We could discuss if we have a (personal or perhaps team-wide) preference for such conventions in PRs and then indeed put them up for the discussion meet, but if we can agree that this is not that big of an impact change we could just settle on one implementation. 🙂 Is the implementation choice blocking for you?

ammar92:
> We could discuss if we have a (personal or perhaps team-wide) preference for such conventions in PRs and then indeed put them up for the discussion meet, but if we can agree that this is not that big of an impact change we could just settle on one implementation. 🙂 Is the implementation choice blocking for you?

Yes sure thing 👍 The current implementation choice isn't blocking for me and most importantly, fixes what needed to be fixed

ammar92 (approved):

stephanie0x00:
## Checklist for QA:

### What works:
Works. The API throws a 404 when an unknown object is deleted. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

underdarknl (approved):

PR #3822, opened 2024-11-11
feat: toggle

HeleenSG:
### Changes
Adds toggle element. Test case is placed within "generate report > Select objects"

Todo:

- [ ] Check accessibility
- [ ] Add JS to set aria-current on current view

### Issue link
Closes ...

### Demo
Screenshot 2024-11-11 at 13 42 18

Hover: 
Screenshot 2024-11-11 at 13 42 41

### QA notes
Please review: 

- [ ] Implemented as designed: https://www.figma.com/design/MDbpc9K0qjyaoRGAGiDX6t/%F0%9F%8E%A8-KAT-%7C-Design-System-%7C-Components?m=auto&node-id=2066-1144&t=T92bk9ydYK4CnI0k-1
- [ ] States work as expected
   - [ ] Hover
   - [ ] Current view vs non selected

---

### Code Checklist

---

---

## Checklist for QA:

### DIFF
--- a/rocky/reports/templates/partials/report_ooi_list.html
+++ b/rocky/reports/templates/partials/report_ooi_list.html
@@ -17,21 +17,6 @@
             <li>
                 <a href="#">Manueel</a>
             </li>
-            <li>
-                <a href="#">Option</a>
-            </li>
-            <li>
-                <a href="#">Option</a>
-            </li>
-            <li>
-                <a href="#">Option</a>
-            </li>
-            <li>
-                <a href="#">Option</a>
-            </li>
-            <li>
-                <a href="#">Option</a>
-            </li>
         </ul>
         <p>
             {% blocktranslate %}
--- a/rocky/reports/templates/partials/report_ooi_list.html
+++ b/rocky/reports/templates/partials/report_ooi_list.html
@@ -10,6 +10,29 @@
                 Select objects ({{ total_oois }})
             {% endblocktranslate %}
         </h2>
+        <ul class="toggle-switch">
+            <li aria-current="true">
+                <a href="#">Live set</a>
+            </li>
+            <li>
+                <a href="#">Manueel</a>
+            </li>
+            <li>
+                <a href="#">Option</a>
+            </li>
+            <li>
+                <a href="#">Option</a>
+            </li>
+            <li>
+                <a href="#">Option</a>
+            </li>
+            <li>
+                <a href="#">Option</a>
+            </li>
+            <li>
+                <a href="#">Option</a>
+            </li>
+        </ul>
         <p>
             {% blocktranslate %}
                 Select which objects you want to include in your report. You can either continue
--- a/rocky/assets/css/components/toggle-button.scss
+++ b/rocky/assets/css/components/toggle-button.scss
@@ -1,0 +1,70 @@
+button,
+a,
+input[type="button"],
+input[type="submit"],
+input[type="reset"] {
+  $breakpoint: 24rem !default;
+
+  &.toggle-button {
+    /* Alignment */
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    align-self: flex-start;
+    gap: 0;
+    margin: 0;
+    box-sizing: border-box;
+
+    /* Styling */
+    border-radius: var(--border-radius-l);
+    padding: var(--spacing-grid-150);
+    width: 2.75rem;
+    height: 2.75rem;
+    border: 1px solid var(--colors-purrple-300);
+
+    /* Text and icon styling */
+    background-color: var(--colors-white);
+    color: var(--colors-purrple-500);
+    text-decoration: none;
+    font-size: 0;
+
+    &::before {
+      color: var(--colors-purrple-500);
+      padding: 0;
+    }
+
+    /* Behaviour */
+    cursor: pointer;
+    overflow-wrap: break-word;
+
+    /* States */
+    &[aria-current="page"] {
+      background-color: var(--colors-purrple-200);
+      border-color: var(--colors-purrple-200);
+
+      &::before {
+        color: var(--colors-purrple-700);
+      }
+    }
+
+    &:hover,
+    &:active {
+      border-color: var(--colors-purrple-500);
+    }
+
+    &:focus {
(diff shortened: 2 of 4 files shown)


ammar92 (commented):
Nice feature 👍 Just a tiny remark about the texts used in the partial template

rocky/reports/templates/partials/report_ooi_list.html
```diff
@@ -10,6 +10,14 @@ <h2>
                 Select objects ({{ total_oois }})
             {% endblocktranslate %}
         </h2>
+        <ul class="toggle-switch">
+            <li aria-current="true">
+                <a href="#">Live set</a>
+            </li>
+            <li>
+                <a href="#">Manueel</a>
+            </li>
```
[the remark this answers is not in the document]
HeleenSG: I did. Thanks for pointing that out. Fixed now. :)

TwistMeister:
Since this solution is very implementation specific I've created a new [issue](#3905) to refactor this to a generic solution.

ppvg (commented):
As discussed offline, this component seems a great fit for applying the [ARIA APG Tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs). It would make sense to follow the [Tabs with Manual Activation](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual) example. Note for example how the `aria-controls` attribute would be a great semantic alternative to the current `data-target-id`.

I understand the time pressure but as discussed before, in its current form it is not fully accessible.

rocky/assets/js/toggleSwitch.js
```diff
@@ -0,0 +1,51 @@
...
+
+      // Add click listener to switch options.
+      option.addEventListener("click", (event) => {
+        toggle(event.target, options);
+      });
+    }
+
+    // Initially show first option contents from toggle-switch.
+    document
+      .getElementById(options[0].getAttribute("data-target-id"))
```
underdarknl, line 28: Dont asume 0 is the 'current active' one when initing. the developer might have added an aria-current to one of the switches.

rocky/assets/js/toggleSwitch.js
```diff
@@ -0,0 +1,51 @@
...
+  for (let i = 0; i < toggle_switches.length; i++) {
+    let options = toggle_switches[i].querySelectorAll(".toggle-switch-button");
+
+    for (let j = 0; j < options.length; j++) {
+      let option = options[j];
+
+      // Hide all elements linked to toggle switch options.
+      document
+        .getElementById(option.getAttribute("data-target-id"))
+        .classList.add("hidden");
```
underdarknl, line 18: I'd say, Hide all of them by applyting a className to it.
And then use the :target selector and the url to show the one being active. that moves the state-management from JS to Css for most of this.
It would not be compatible with nested versions of this component, as selecting a nested 'tab' would hide the parent tab, But I don't think we have that usecase anyway.

ppvg: While it would be nice if state management didn't require JS, it's unfortunately not possible to make this fully accessible without using JS to manage ARIA attributes.

@underdarknl could you please verify whether the APG's [Tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) aligns with the expected use here, as I suspect? If so I would implore you to follow one of the APG's example implementations. If not, let's look together for fitting ARIA roles and behaviours.

underdarknl: I do think the Tabs pattern fits this nicely. Do you have a preference on which example to follow? Especially if we want to move this to Manon upstream at some point?

rocky/assets/js/toggleSwitch.js
TwistMeister: Good point...

rocky/assets/js/toggleSwitch.js
ppvg: I don't know the exact context. If switching accidentally would be a problem, manual activation would make sense. For upstream we could add both and make it configurable.

rocky/assets/js/toggleSwitch.js
```diff
@@ -0,0 +1,51 @@
+import { onDomReady } from "../js/imports/utils.js";
+
+onDomReady(initToggleSwitches);
+
+function initToggleSwitches() {
+  // Switches get stitches ;-)
+  let toggle_switches = document.querySelectorAll(".toggle-switch");
+
+  for (let i = 0; i < toggle_switches.length; i++) {
+    let options = toggle_switches[i].querySelectorAll(".toggle-switch-button");
```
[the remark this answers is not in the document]
TwistMeister: Done

stephanie0x00:
The 'Live filter' doesn't seem to pick up the chosen filtered items. The list of items remains empty with the following filter applied: 

image

image

The manual selection does pick up the applied filter: 

image

RomijnHumanoids:
I just performed an accessibility check with Apple's VoiceOver & tab navigation on Chrome browser, and I have some remarks that need to be assessed before merging this. I chose to review the toggle component in the 'Generate report' flow toggling 'live set' vs. 'manual selection'.

- I'm missing a spoken label explaining what the component is or does, as it just reads 'list 2 items'. We could add a functional descriptive Aria label like 'toggle', 'radio button', 'options' or 'switcher' to the component.
Screenshot 2025-03-19 at 14 22 31
      - This is an example of a single select toggle group component on shadcn/ui
Screenshot 2025-03-19 at 16 02 35
Screenshot 2025-03-19 at 16 02 46
      - Additionally, for more clarity we could add a visible label 'select by' (like in most Figma designs often with label 'show').
Screenshot 2025-03-19 at 16 26 28

- Toggle items themselves navigate clearly but read out as 'links' which does not communicate the functionality of a radio button or toggle. e.g. it is not clear that users have to choose between two options.
Screenshot 2025-03-19 at 14 22 47
Screenshot 2025-03-19 at 14 22 48


- The list of filtered items disappears when the toggle is set to 'live set'.

- Not in this component, but I would also suggest to change the order so there is an explanation about live sets followed by a toggle to select 'live set'. First explain different ways of selecting objects, and than ask users' preference.

Issue #3838, opened 2024-11-13
Combine all schedulers for all organisations


jpbruinsslot:
## Proposal

The current situation is that we create individual 'schedulers' for every organisation, this amounts to 3 schedulers for every organisation. 

The addition of queues in the scheduler being persisted as a database table allows us to explore the possibility of running a dedicated schedulers for all organisations.

Create one `BoefjeScheduler`,  `NormalizerScheduler` and `ReportScheduler` for all organisations instead of individual schedulers for every organisation. One message queue for all scan profile mutations to which scan profile mutations are posted of all organisations for the `BoefjeScheduler`, and one message queue for raw file creation of all organisations for the `NormalizerScheduler`, the `ReportScheduler` will reference its internal database table.

## Advantages

* Less overhead of creating and removing schedulers for every organisation
* Stops continuous checking of available organisations
* Allows to replicate scheduler application instances when used with a blocking messaging queue
* Additionally we need to support popping multiple tasks from the endpoint

## Disadvantages

* the `allow_update`s, `allow_replace`, `allow_priority_updates` for individual organisations can't be used, however this function doesn't seem to be used and is for every type scheduler the same
* we should investigate how we solve when organisations wan to disable a scheduler, and even if this has a use-case

## Impact

* now we have several different rabbitmq messaging queues for every organisation, this needs to be reduced to one
* organisation ids need to be passed in the messages on the queue
* changes to task runner to pop off tasks
* rocky likely needs to be update to interface with changes from the scheduler API

### Next steps and impact

* Pop endpoint has changed from `/queues` to `/schedulers/{id}/pop` , additionally it will return a paginated result instead of a single `Task` , this is because the pop endpoint now supports filtering with multiple tasks returns. Services that rely on the scheduler pop endpoint need to update their interfaces (https://github.com/minvws/nl-kat-coordination/issues/3961)
* Push endpoint changed from `/queues` to `schedulers/{id}/push`, services that interface with the push endpoints (rocky) need to update their interfaces. (https://github.com/minvws/nl-kat-coordination/issues/3962)
* scan profile mutation message queue, for every organisation a message queue is created for scan profile mutations, this needs to be updated and all scan profile mutations for every organisation needs to be relayed on a single scan profile mutations message queue (https://github.com/minvws/nl-kat-coordination/issues/3963)
* raw data file received message queue (the as with scan profile mutations) (https://github.com/minvws/nl-kat-coordination/issues/3964)
[...]

Issue #3357, opened 2024-08-13
Katalogus caching in the scheduler


jpbruinsslot:
The sophisticated caching and filtering in the scheduler can perhaps be optimized by:

* Adding more API filtering options in the Katalogus that would resolve creating 3 separate caches in the scheduler
* Adding ETag (Entity Tag) or `Last-Modified` headers to katalogus to invalidate internal katalogus state in the scheduler instead of relying on the `ExpiringDict` functionality
* Potentially resolve https://github.com/minvws/nl-kat-coordination/issues/856 since the katalogus state in the scheduler will always be updated to the latest modified timestamp
* Retrieval of new boefjes is now a call per organisation to diff internal cache of present boefjes to enabled boefjes. Could be optimized by a signal to the scheduler when a boefje gets enabled/added

Related: https://github.com/minvws/nl-kat-coordination/issues/856

underdarknl:
In general, I want to move in the direction of storing the enabled/configured plugins in XTDB, instead of the katalogus.
For the following point:
>     * Retrieval of new boefjes is now a call per organisation to diff internal cache of present boefjes to enabled boefjes. Could be optimized by a signal to the scheduler when a boefje gets enabled/added

When we move the plugin config to XTDB, octopoes should already send out signals to the Scheduler for changes to OOI's. We can use the same mechanism to pick up changes on plugins.

PR #3839, opened 2024-11-13
Combined schedulers

jpbruinsslot:
### Changes

* `app.py`: Removal of creation of organisational schedulers. We don't gather all available organisations from the katalogus and create individual boefje, report, and normalizer schedulers.
* `app.py`: Removal of continuous checking of added or removed organisations. We removed the `monitor_organisations` method that was running in a thread to check if we added or removed organisation and create schedulers for it.
* `models/` : `Task`  and `Schedule` get a specific `organisation` field. Since we don't differentiate tasks that are pushed on an organisation queue anymore, we need to be able to determine for what organisation an item on a queue is from.[^1]
* `schedulers/`: Since we've consolidated the organisational schedulers to 3 types of schedulers, and we assume that we get task creation from consolidated message queues, we need to update the schedulers to handle that.
  * `BoefjeScheduler`
    * retrieve scan profile mutations from one message queue
    * iterate over organisation to check for enables boefjes[^2]
    * refactor of method names and exception handling
* `server/` 
  * Merger of `queues/` and `schedulers/` endpoint. The notion of a queue and scheduler are used interchangeably. Therefore it makes more sense at the moment to merge those endpoints to avoid confusion and simplify the code.
  * Support to pop off more than one `Task` from a scheduler. Task runner will now be able to batch pop tasks from a scheduler using filters.

### Next steps and impact

* Pop endpoint has changed from `/queues` to `/schedulers/{id}/pop` , additionally it will return a paginated result instead of a single `Task` , this is because the pop endpoint now supports filtering with multiple tasks returns. Services that rely on the scheduler pop endpoint need to update their interfaces (https://github.com/minvws/nl-kat-coordination/issues/3961)
* Push endpoint changed from `/queues` to `schedulers/{id}/push`, services that interface with the push endpoints (rocky) need to update their interfaces. (https://github.com/minvws/nl-kat-coordination/issues/3962)
* scan profile mutation message queue, for every organisation a message queue is created for scan profile mutations, this needs to be updated and all scan profile mutations for every organisation needs to be relayed on a single scan profile mutations message queue (https://github.com/minvws/nl-kat-coordination/issues/3963)
* raw data file received message queue (the as with scan profile mutations) (https://github.com/minvws/nl-kat-coordination/issues/3964)
* Model definitions updates: `organisation` fields are added to `Task` , `Schedule` , services using these models need to update their specifications. (https://github.com/minvws/nl-kat-coordination/issues/3965)
* Batch status updates. Several places in the scheduler we can consider batch updating status field of task. Potentially exposing this as a endpoint. This because the task runner will be able to pop off multiple task from a scheduler, it might therefore be beneficial to batch update the tasks. (
* Deleting of organisations needs to be addressed, and what the protocol needs to be with regards of queued tasks, and schedules. (https://github.com/minvws/nl-kat-coordination/issues/3966)

### Issue link

https://github.com/minvws/nl-kat-coordination/issues/3838

### QA notes

The process of scanning tasks, rescheduling of tasks should remain the same from a users standpoint of view. Testing this would be the main focus point. Additionally since all organisational schedulers have been combined to one, particular emphasis should be made to check if users will only see, or are able to interact with tasks that are for the organisation that they are allowed to.

It would be interesting to see if with high volumes the execution of tasks will be done as expected by the user.

---

### Code Checklist

---

---

## Checklist for QA:

[^1]: The attentive reader will know that the organisation information is present in the `data` specification in a `Task` , and can be filtered on by using a `FilterRequest`. However in practice the scheduler is mainly used in the context of multiple organisations and filtering on this id is deemed developer friendly when interfacing with the scheduler. Hence the choice explicitly defining the `organisation` field on a top level.
[^2]: as mentioned in the code this iterating over all organisations in order to check if there are enabled plugins is very inefficient, please refer to issue #3357


mula/scheduler/schedulers/schedulers/report.py
```diff
@@ -16,195 +15,107 @@
...
-        super().__init__(
-            ctx=ctx,
-            queue=self.queue,
-            scheduler_id=scheduler_id,
-            callback=callback,
-            create_schedule=True,
-            auto_calculate_deadline=False,
-        )
+    ID: str = "report"
+    ITEM_TYPE: Any = models.ReportTask
```
ammar92, line 21: `Any` should be a `type`?

mula/scheduler/schedulers/schedulers/report.py
```diff
@@ -16,195 +15,107 @@
...
 
-        super().__init__(
-            ctx=ctx,
-            queue=self.queue,
-            scheduler_id=scheduler_id,
-            callback=callback,
-            create_schedule=True,
-            auto_calculate_deadline=False,
-        )
+    ID: str = "report"
```
ammar92, line 20:
```suggestion
    ID: Literal["report"] = "report"
```

ammar92 (commented):
Looks good in general. Just a few remarks while skimming through it, not a detailed look yet

mula/scheduler/schedulers/scheduler.py
```diff
@@ -331,17 +298,18 @@ def post_push(self, item: models.Task) -> models.Task:
         # based on the item.
         if schedule_db.schedule is not None:
             schedule_db.deadline_at = cron.next_run(schedule_db.schedule)
-        elif self.auto_calculate_deadline:
+        else:
```
jpbruinsslot, line 301: Check if this is deleted intentionally. Doesn't seems it should

jpbruinsslot: It shouldn't have been deleted.

stephanie0x00:
Current state (16/12) of this PR.

- Tasks don't seem to run. I did the onboarding and no tasks for boefjes or normalizers show up on the Tasks page. However, some findings are shown on the Findings page (which remain pending).

No tasks on the tasks page:
image

Findings page for this state: 
image

 
```
boefje-1  | HTTPStatusError: Client error '404 Not Found' for url 'http://scheduler:8000/queues'
boefje-1  | For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
boefje-1  | Traceback (most recent call last):
boefje-1  |   File "/app/boefjes/boefjes/app.py", line 89, in _fill_queue
boefje-1  |     queues = self.scheduler_client.get_queues()
boefje-1  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
boefje-1  |   File "/app/boefjes/boefjes/clients/scheduler_client.py", line 71, in get_queues
boefje-1  |     self._verify_response(response)
boefje-1  |   File "/app/boefjes/boefjes/clients/scheduler_client.py", line 67, in _verify_response
boefje-1  |     response.raise_for_status()
boefje-1  |   File "/usr/local/lib/python3.11/site-packages/httpx/_models.py", line 763, in raise_for_status
boefje-1  |     raise HTTPStatusError(message, request=request, response=self)
boefje-1  | httpx.HTTPStatusError: Client error '404 Not Found' for url 'http://scheduler:8000/queues'
boefje-1  | For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404
boefje-1  | HTTP Request: GET http://scheduler:8000/queues "HTTP/1.1 404 Not Found"
boefje-1  | 2024-12-16T11:47:16.426262 [error] Getting the queues from the scheduler failed
boefje-1  | ╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
boefje-1  | │ /app/boefjes/boefjes/app.py:89 in _fill_queue                                                    │
boefje-1  | │                                                                                                  │
boefje-1  | │    86 │   │   │   return                                                                         │
boefje-1  | │    87 │   │                                                                                      │
boefje-1  | │    88 │   │   try:                                                                               │
boefje-1  | │ ❱  89 │   │   │   queues = self.scheduler_client.get_queues()                                    │
boefje-1  | │    90 │   │   except HTTPError:                                                                  │
boefje-1  | │    91 │   │   │   # Scheduler is having issues, so make note of it and try again                 │
boefje-1  | │    92 │   │   │   logger.exception("Getting the queues from the scheduler failed")               │
boefje-1  | │                                                                                                  │
boefje-1  | │ ╭────────────────────────────────── locals ──────────────────────────────────╮                   │
boefje-1  | │ │ queue_type = <Queue.BOEFJES: 'boefje'>                                     │                   │
boefje-1  | │ │       self = <boefjes.app.SchedulerWorkerManager object at 0x74d6402b2e10> │                   │
boefje-1  | │ │ task_queue = <AutoProxy[Queue] object, typeid 'Queue' at 0x74d6406533d0>   │                   │
boefje-1  | │ ╰────────────────────────────────────────────────────────────────────────────╯                   │
boefje-1  | │                                                                                                  │
boefje-1  | │ /app/boefjes/boefjes/clients/scheduler_client.py:71 in get_queues                                │
boefje-1  | │                                                                                                  │
boefje-1  | │   68 │                                                                                           │
boefje-1  | │   69 │   def get_queues(self) -> list[Queue]:                                                    │
boefje-1  | │   70 │   │   response = self._session.get("/queues")                                             │
boefje-1  | │ ❱ 71 │   │   self._verify_response(response)                                                     │
boefje-1  | │   72 │   │                                                                                       │
boefje-1  | │   73 │   │   return TypeAdapter(list[Queue]).validate_json(response.content)                     │
boefje-1  | │   74                                                                                             │
boefje-1  | │                                                                                                  │
boefje-1  | │ ╭───────────────────────────────────────── locals ──────────────────────────────────────────╮    │
boefje-1  | │ │ response = <Response [404 Not Found]>                                                     │    │
boefje-1  | │ │     self = <boefjes.clients.scheduler_client.SchedulerAPIClient object at 0x74d6402b2890> │    │
boefje-1  | │ ╰───────────────────────────────────────────────────────────────────────────────────────────╯    │
boefje-1  | │                                                                                                  │
boefje-1  | │ /app/boefjes/boefjes/clients/scheduler_client.py:67 in _verify_response                          │
boefje-1  | │                                                                                                  │
boefje-1  | │   64 │                                                                                           │
boefje-1  | │   65 │   @staticmethod                                                                           │
boefje-1  | │   66 │   def _verify_response(response: Response) -> None:                                       │
boefje-1  | │ ❱ 67 │   │   response.raise_for_status()                                                         │
boefje-1  | │   68 │                                                                                           │
boefje-1  | │   69 │   def get_queues(self) -> list[Queue]:                                                    │
boefje-1  | │   70 │   │   response = self._session.get("/queues")                                             │
boefje-1  | │                                                                                                  │
boefje-1  | │ ╭─────────────── locals ────────────────╮                                                        │
boefje-1  | │ │ response = <Response [404 Not Found]> │                                                        │
boefje-1  | │ ╰───────────────────────────────────────╯                                                        │
boefje-1  | │                                                                                                  │
boefje-1  | │ /usr/local/lib/python3.11/site-packages/httpx/_models.py:763 in raise_for_status                 │
boefje-1  | │                                                                                                  │
boefje-1  | │    760 │   │   }                                                                                 │
boefje-1  | │    761 │   │   error_type = error_types.get(status_class, "Invalid status code")                 │
boefje-1  | │    762 │   │   message = message.format(self, error_type=error_type)                             │
boefje-1  | │ ❱  763 │   │   raise HTTPStatusError(message, request=request, response=self)                    │
boefje-1  | │    764 │                                                                                         │
boefje-1  | │    765 │   def json(self, **kwargs: typing.Any) -> typing.Any:                                   │
boefje-1  | │    766 │   │   return jsonlib.loads(self.content, **kwargs)                                      │
boefje-1  | │                                                                                                  │
boefje-1  | │ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
boefje-1  | │ │   error_type = 'Client error'                                                                │ │
boefje-1  | │ │  error_types = {                                                                             │ │
boefje-1  | │ │                │   1: 'Informational response',                                              │ │
boefje-1  | │ │                │   3: 'Redirect response',                                                   │ │
boefje-1  | │ │                │   4: 'Client error',                                                        │ │
boefje-1  | │ │                │   5: 'Server error'                                                         │ │
boefje-1  | │ │                }                                                                             │ │
boefje-1  | │ │      message = "Client error '404 Not Found' for url 'http://scheduler:8000/queues'\nFor     │ │
boefje-1  | │ │                more inf"+76                                                                  │ │
boefje-1  | │ │      request = <Request('GET', 'http://scheduler:8000/queues')>                              │ │
boefje-1  | │ │         self = <Response [404 Not Found]>                                                    │ │
boefje-1  | │ │ status_class = 4                                                                             │ │
boefje-1  | │ ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
boefje-1  | ╰──────────────────────────────────────────────────────────────────────────────────────────────────╯

```

jpbruinsslot:
> Looks good to me now. There are still some test errors, I think some of the test still use the older `create_item` utility function.
> 
> And about the QA test results: I'm not sure (yet) how the 404 for the `/queues` endpoint is related to this branch. If there was an error on that list endpoint, it shouldn't be a 404

The PR is at its current state not able to be QA'ed. In the description I've added the next steps that need to be taken in order to integrate this PR into openkat. There are substantial changes to both the API and the message queues, these changes need to be integrated in the systems that rely and use those.

In the next steps section I also created the issues related to those integrations and services that need to be picked up and developed.

mula/scheduler/app.py
```diff
@@ -272,23 +143,23 @@ def stop_threads(self) -> None:
...
+    def _collect_metrics(self) -> None:
+        """Collect application metrics throughout the application."""
 
-        Args:
-            scheduler_id: The id of the scheduler to remove.
-        """
-        with self.lock:
-            if scheduler_id not in self.schedulers:
-                return
+        # FIXME:: can be queries instead of a loop
```
jpbruinsslot, line 156: TODO

boefjes/boefjes/clients/scheduler_client.py
```diff
@@ -66,20 +78,24 @@ def __init__(self, base_url: str):
...
         response.raise_for_status()
 
-    def get_queues(self) -> list[Queue]:
-        response = self._session.get("/queues")
+    def pop_item(self, scheduler_id: str) -> Task | None:
+        response = self._session.post(f"/schedulers/{scheduler_id}/pop?limit=1")
         self._verify_response(response)
 
-        return TypeAdapter(list[Queue]).validate_json(response.content)
+        page = TypeAdapter(PaginatedTasksResponse | None).validate_json(response.content)
```
underdarknl, line 85: If the limit is hard coded to 1, do we really need a paginated response, especially when we then return only the first Task on line 89

jpbruinsslot: This is specifically left in to make the new changes work with the current setup, of the task runner handling one task at a time. The endpoint now allows for batched tasks to be popped from the endpoint as was requested in feature requests for the scheduler. I made @Donnype aware of these changes and he will be thinking about the necessary changes to the task runner to leverage this new functionality.

mula/scheduler/models/schedule.py
```diff
@@ -57,21 +52,14 @@ class ScheduleDB(Base):
...
-
     data = Column(JSONB, nullable=False)
-
     enabled = Column(Boolean, nullable=False, default=True)
-
     schedule = Column(String, nullable=True)
-
     tasks = relationship("TaskDB", back_populates="schedule")
 
     deadline_at = Column(DateTime(timezone=True), nullable=True)
```
underdarknl, line 63: Should this (and created_at and modified_at not be set to func.utcnow() to make sure we don't rely on the possibly missing UTC flag on the connection?

mula/scheduler/models/task.py
```diff
@@ -46,19 +46,13 @@ class Task(BaseModel):
...
-
     status: TaskStatus = TaskStatus.PENDING
-
     type: str | None = None
-
     hash: str | None = Field(None, max_length=32)
-
     data: dict = Field(default_factory=dict)
 
     created_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
```
underdarknl, line 58: also use func.utcnow() here.

boefjes/boefjes/clients/scheduler_client.py
jpbruinsslot: However, come to think about it I need to check whether pagination will work correctly with tasks being popped and set to dispatched correctly.

mula/scheduler/schedulers/schedulers/boefje.py
```diff
@@ -96,242 +54,179 @@ def run(self) -> None:
...
-            mutation.primary_key,
-            ooi_primary_key=mutation.primary_key,
-            organisation_id=self.organisation.id,
-            scheduler_id=self.scheduler_id,
-        )
-
-        # There should be an OOI in value
-        ooi = mutation.value
-        if ooi is None:
+        try:
```
underdarknl, line 79: I'd rather keep the try except for the ValidationError smaller and only around the actual validation code.

mula/scheduler/schedulers/schedulers/boefje.py
```diff
@@ -963,30 +673,52 @@ def get_boefjes_for_ooi(self, ooi: OOI) -> list[Plugin]:
...
+    def get_oois_for_boefje(self, boefje: models.Plugin, organisation: str) -> list[models.OOI]:
+        oois = []
+
+        oois_by_object_type = self.ctx.services.octopoes.get_objects_by_object_types(
+            organisation,
+            boefje.consumes,
+            list(range(boefje.scan_level, 5)),  # type: ignore
+        )
+
+        # Filter OOIs based on permission
```
underdarknl, line 683: This should be done in the remote API / query. Lets add a #TODO for this.

mula/scheduler/schedulers/schedulers/boefje.py
```diff
@@ -342,247 +237,170 @@ def push_tasks_for_rescheduling(self):
...
-        """Given a Boefje and OOI create a BoefjeTask and push it onto
-        the queue.
+                        # TODO: do we want to disable the schedule when a
+                        # boefje is not allowed to scan an ooi?
 
-        Args:
-            boefje: Boefje to run.
-            ooi: OOI to run Boefje on.
-            caller: The name of the function that called this function, used for logging.
+                        # Boefje allowed to scan ooi?
```
underdarknl, line 323: This should be done in the remote API / query. Lets add a #TODO for this.

mula/scheduler/schedulers/schedulers/report.py
```diff
@@ -1,210 +1,125 @@
...
-
-        if task_db is not None and task_db.status not in [TaskStatus.FAILED, TaskStatus.COMPLETED]:
-            self.logger.debug(
-                "Task is still running, according to the datastore",
-                task_id=task_db.id,
-                organisation_id=self.organisation.id,
-                scheduler_id=self.scheduler_id,
-            )
-            return True
+    def has_schedule_permission_to_run(self, schedule: models.Schedule) -> bool:
```
underdarknl, line 101: Can we somehow move this to the original query that pops our schedules, instead of evaluating this in python time?

mula/scheduler/schedulers/schedulers/report.py
```diff
@@ -1,210 +1,125 @@
...
-                organisation_id=self.organisation.id,
-                scheduler_id=self.scheduler_id,
+    @tracer.start_as_current_span(name="process_rescheduling")
+    def process_rescheduling(self):
+        schedules, _ = self.ctx.datastores.schedule_store.get_schedules(
+            filters=filters.FilterRequest(
+                filters=[
+                    filters.Filter(column="scheduler_id", operator="eq", value=self.scheduler_id),
+                    filters.Filter(column="deadline_at", operator="lt", value=datetime.now(timezone.utc)),
+                    filters.Filter(column="enabled", operator="eq", value=True),
```
underdarknl, line 49: Can we add the correct joins/filters here to limit this result set to only schedules that are allowed to produce a task?

mula/scheduler/models/schedule.py
jpbruinsslot: 1. For the `deadline_at` it is significant that it is possible to set it to `None` , setting the default on inserting a `Schedule` in the database has consequences on code that relies on this. This should remains as is.
2. Since this a model definition that translates to a SQL migration the `func.now()` function generates the SQL `NOW()`, which retrieves the current timestamp from the database server, if the database server is configured to use UTC , `NOW()` will return the current time in UTC. As far as I know there isn't a `func.utcnow()` function in sqlalchemy. We can however in the migrations force the database server to use UTC as its default timezone with:
```sql
SET timezone = 'UTC';
```
alternatively, I think we can also might be able to set the time on using Python's `datetime.now(timezone.utc)`

```python
modified_at = Column(
    DateTime(timezone=True),
    nullable=False,
    default=lambda: datetime.now(timezone.utc),
    onupdate=lambda: datetime.now(timezone.utc)
)
```
However this likely will be evaluated in Python instead of the database.

@ammar92 you have any particular thoughts about this?

mula/scheduler/schedulers/schedulers/report.py
jpbruinsslot: These filters should do that. Other evaluations of whether a task should be executed or not depend on the scheduler and subsequent cross-references checks to other services. E.g. for a boefje scheduler, does the ooi still exist, is the plugin still enabled, are we still allowed to scan, etc. These factors can't be evaluated by a database query.

mula/scheduler/models/schedule.py
ammar92: > 2. Since this a model definition that translates to a SQL migration the `func.now()` function generates the SQL `NOW()`, which retrieves the current timestamp from the database server, if the database server is configured to use UTC , `NOW()` will return the current time in UTC. As far as I know there isn't a `func.utcnow()` function in sqlalchemy. 

As far as I know the internally stored value for `timestamp with time zone` is always in UTC. So even if there was function like a `UTCNOW()`, it wouldn't matter for this non-naive date time type.

> However this likely will be evaluated in Python instead of the database.

It's indeed evaluated in Python land, but the performance impact is negligible. While it's a good alternative, using SQLAlchemy's function is generally preferred and more common for this use case

boefjes/boefjes/clients/scheduler_client.py
underdarknl: Makes sense, I was thinking it might have been because of the multi-pop support. But, I'm guessing that would either introduce an extra param (limit?) or a new method?

underdarknl: popping and changing the status (for any number of taskst) should be handled in the same database-transaction I'd say.  Hopefully relying on table locking to ensure a task is not popped twice simultaneously.

mula/scheduler/schedulers/schedulers/report.py
underdarknl: Can they be set on the update events of those changes though? Eg, disabling / removing tasks for boefjes that are being disabled seems like the correct way forward instead of keeping them around and checking them for an enabled boefje every time they could be executed.

underdarknl: Which is clearly out of scope for this PR, but still worth investigating.

mula/scheduler/schedulers/schedulers/report.py
jpbruinsslot: original method has been removed

mula/scheduler/schedulers/schedulers/boefje.py
```diff
@@ -343,247 +233,168 @@ def push_tasks_for_rescheduling(self):
...
-        with futures.ThreadPoolExecutor(
-            thread_name_prefix=f"BoefjeScheduler-TPE-{self.scheduler_id}-rescheduling"
-        ) as executor:
+        with futures.ThreadPoolExecutor(thread_name_prefix=f"TPE-{self.scheduler_id}-rescheduling") as executor:
             for schedule in schedules:
-                boefje_task = BoefjeTask.model_validate(schedule.data)
-
-                # Plugin still exists?
                 try:
+                    boefje_task = models.BoefjeTask.model_validate(schedule.data)
```
Donnype, line 248: I found an issue here that the 'data' field in the schedule contains the uuid of the boefje_meta, which is passed all the way to bytes that has a uniqueness constraint on this id. Hence, when we create new boefje tasks from the `data` field, we should make sure the id is a _new_ uuid, because rescheduled tasks now all fail for me.

Donnype (commented):
Good stuff! I think this could have a big impact for performance but above all simplicity. I found two issues that do not easily pop up when doing a QA with a reset of the data, and wanted to comment these already before checking further, please have a look!

mula/scheduler/storage/migrations/versions/0009_add_organisation.py
```diff
@@ -0,0 +1,30 @@
...
+revision = "0009"
+down_revision = "0008"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("schedules", sa.Column("organisation", sa.String(), nullable=False))
+    op.add_column("tasks", sa.Column("organisation", sa.String(), nullable=False))
```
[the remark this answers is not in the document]
jpbruinsslot: main reason was to add the organisation id to the tasks and schedules for filtering purposes, since this information was most of the time buried in the `data` field, and needed special filtering to get those filtering results. But yeah the not null constraint needs to be solved.

mula/scheduler/schedulers/schedulers/boefje.py
jpbruinsslot: This is strange, it should create a `new_boefje_task` that should have a new unique uuid. Perhaps you can show me what happened?

Donnype: Update: probably an issue on my system after a rebuild, although a strange one that might have to do with the creation of raw files failing after creation of a boefje_meta succeeded from the job_handler. Resolving for now.

underdarknl (approved):

Issue #3850, opened 2024-11-18
Static boefjes in the database get a static schema as well


Donnype:
## Implementation
When we enable a static boefje, the boefje gets added to the database and its schema as well using its `schema.json`. When the `schema.json` gets updated, the changes are not reflected in the API, probably because we take the value from the database.

### Possible solution
There are two straightforward solutions:
1. Make sure the database is in sync with the boefje definitions and use the database for the API as well
2. Let the local files take precedence over the database for static boefjes.

And I would opt for the second one.

### Alternatives considered
Well, see point 1 above!

PR #3858, opened 2024-11-19
Let local plugins (files) take precedence over database entries

Donnype:
### Changes

This makes sure we return info from the `boefjes.json` and `schema.json` instead of returning data in the database. This solves issues where we updated these files locally but did not see the changes reflected due to a stale database.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/3850

### Demo

-

### QA notes

Please enable `dns-records` and then delete the `schema.json` file in the `kat_dns` directory in the boefjes module. In the KATalogus, we should see no more settings for `dns-records` (perhaps after a restart of the boefjes container: `docker compose restart boefje katalogus`).

---

### Code Checklist

---

---

## Checklist for QA:

### DIFF
--- a/boefjes/boefjes/storage/interfaces.py
+++ b/boefjes/boefjes/storage/interfaces.py
@@ -56,8 +56,17 @@ class CannotUpdateStaticPlugin(NotAllowed):
 
 
 class DuplicatePlugin(NotAllowed):
-    def __init__(self, key: str):
-        super().__init__(f"Duplicate plugin {key}")
+    FIELD_NAMES = {"boefje_plugin_id": "id", "boefje_name": "name"}
+
+    def __init__(self, *, field: str | None = None, error_message: str = ""):
+        if not field:
+            for name, field_name in self.FIELD_NAMES.items():
+                if name not in error_message:
+                    continue
+
+                field = field_name
+
+        super().__init__(f'Duplicate plugin "{field}"')
 
 
 class OrganisationStorage(ABC):
--- a/boefjes/boefjes/dependencies/plugins.py
+++ b/boefjes/boefjes/dependencies/plugins.py
@@ -94,7 +94,7 @@ class PluginService:
             self.set_enabled_by_id(plugin_id, to_organisation, enabled=True)
 
     def upsert_settings(self, settings: dict, organisation_id: str, plugin_id: str):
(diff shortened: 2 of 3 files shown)


boefjes/boefjes/storage/interfaces.py
```diff
@@ -56,8 +56,17 @@ def __init__(self, plugin_id: str):
 
 
 class DuplicatePlugin(NotAllowed):
```
ammar92, line 58: Imho this is too complex for an exception class. A few points to consider:
- The user would need to know specific field values (which could be simplified if typed as a `Literal` with choices to reduce guessing)
- The `error_message` parameter is ignored unless `field` is passed, which might be unexpected behavior
- This was my confusion about the use of the `error_message` since I expected that to be an error message that I can assign to the exception. But the use of `error_message` as a proxy for another exception's message feels unusual. Usually if an exception is based on another you’d pass the exception object itself

Perhaps to make it simpler, introduce `DuplicatePluginID` and `DuplicatePluginName` subclasses with a (default) error message?
```python3
class DuplicatePluginID(DuplicatePlugin):
    def __init__(self, plugin_id: str):
        super().__init__(f"Duplicate plugin ID: {plugin_id}")

class DuplicatePluginName(DuplicatePlugin):
    def __init__(self, plugin_name: str):
        super().__init__(f"Duplicate plugin name: {plugin_name}")
```

ammar92 (commented):
Nice work on fixing the bug 👍

I only have one remark/ suggestion, let me know what you think

boefjes/boefjes/storage/interfaces.py
Donnype: @ammar92 I updated the logic so that it should work out of the box for any field

ammar92 (approved):

stephanie0x00:
## Checklist for QA:

### What works:
Works as mentioned in the QA notes. Below are screenshots of the settings section disappearing. When the `schema.json` is gone the boefje can still be enabled and disabled and adding a new hostname and getting its dns records works too. 

### What doesn't work:
n/a

### Bug or feature?:
n/a

### Screenshots

The Settings indeed disappear: 

image


main: 

image

underdarknl (approved):

PR #3895, opened 2024-11-27
Bug fix: KAT-alogus parameter is now organization member instead of organization code

Rieven:
### Changes
Fix wrong parameter passed to the KAT-alogus client


rocky/reports/views/generate_report.py
```diff
@@ -114,7 +114,7 @@ def post(self, request, *args, **kwargs):
             messages.error(request, _("You do not have the required permissions to enable plugins."))
```
underdarknl, line 114: This check is already handled by the katalogusclient.
We should try/except the enable_boefje_by_id and catch error thrown on https://github.com/minvws/nl-kat-coordination/blob/ff6d154b9d9ba3e805c5b33e21a36125d04960db/rocky/katalogus/client.py#L366

rocky/reports/views/generate_report.py
```diff
@@ -114,7 +114,7 @@ def post(self, request, *args, **kwargs):
...
             return PostRedirect(self.get_previous())
 
-        client = get_katalogus(self.organization.code)
+        client = get_katalogus(self.organization_member)
         for selected_plugin in selected_plugins:
             try:
                 client.enable_boefje_by_id(selected_plugin)
```
underdarknl, line 120: I'd expect this HTTPError to be caught in the client, and instead a more specialized KatalogusClientHttpError subclassed from the KatalogusError or something to be raised.

underdarknl, line 120: We need to emit the relevant Audit events when enabling boefjes.
Possibly these logging events could be moved to the Katalogus?

Issue #3904, opened 2024-11-28
KAT-alogus Audit events and errors handlers should be handled by the client


Rieven:
We need to emit the relevant Audit events when enabling boefjes.
Possibly these logging events could be moved to the Katalogus?

_Originally posted by @underdarknl in https://github.com/minvws/nl-kat-coordination/pull/3895#discussion_r1860626096_

      This check is already handled by the katalogusclient.
We should try/except the enable_boefje_by_id and catch error thrown on https://github.com/minvws/nl-kat-coordination/blob/ff6d154b9d9ba3e805c5b33e21a36125d04960db/rocky/katalogus/client.py#L366

_Originally posted by @underdarknl in https://github.com/minvws/nl-kat-coordination/pull/3895#discussion_r1860623277_

              I'd expect this HTTPError to be caught in the client, and instead a more specialized KatalogusClientHttpError subclassed from the KatalogusError or something to be raised.

_Originally posted by @underdarknl in https://github.com/minvws/nl-kat-coordination/pull/3895#discussion_r1860624888_

PR #3956, opened 2024-12-11
Move event codes logging to KATalogus client

Donnype:
### Changes

Log event codes in the KATalogus client.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/3904

### QA notes

Please verify that we see logs still when we enable/disable, create/delete plugins including event codes!

---

### Code Checklist

---

---

## Checklist for QA:


ammar92:
Looks good to me, but I'm unsure about the responsibilities here and perhaps I missed a good discussion about this. I'm also unsure how it solves the linked issue.

Usually in an MVC app it's the controller that should be focussed on business-level logs/ audit trails to track a user's action and intent (e.g. "User X initiated action Y for resource Z"). The client should be focussed on technical-level logs to track details of calls to external resources (e.g. "API call to /resource failed with status 404").

The only advantage I see in the suggested approach is consistent logging output regardless of which controller or view invokes the client. But I think we shouldn't be making the client aware of business-level context, as it doesn't fit the responsibilities of the client. It should be focused on handling external interactions and technical concerns and shouldn't need to know about user actions or business-level intents.

underdarknl:
I agree that the Client itself should not be dealing with these specifics. Audit-events are side-effects of affecting the data. With that logic they should be in the Models themselves. A model could emit an event when one of the crud actions is performed. You could even say these events are part of the data-warehouse and as such could be implemented with stored-procedures, as a record in the database should not be able to exists without the associated audit-record.

As we don't really have controller (In the MVC sense of things) we end up with a problem. In Django most of the controller work is handled in the views, but as we can see, views get copied, get duplicated and are written without al these side-effects handled properly.

In an earlier discussion I proposed to make all of these side effects (eg loading an object for a user, saving it to the db (with side effects), and checking permissions) part of the models, as the model in a sense a logical place to put this if we look at keeping everything close to where the action is, but that was countered by the now-used repository model in which the client and the business logic get bundled together a bit more.

Donnype:
I still think that given that we don't change the contents of the audit logs or introduce a new layer of business logic, this is the best we can do for now. In my experience loggers and event dispatchers tend to pop up at many layers of your application and tend to clutter it to a certain extend. (Because they don't care about layering: I may want to log both my requests and the query statements going to a specific database.) We should at least try to keep it coherent and I think logging where we are sure the action occurs, i.e. in the client if we are limited to Rocky, is more coherent given that we had these logs all over the place and could easily forget them. I agree with @ammar92 that it might be even better if these particular logs were done within at the start of the KATalogus request handling behind the API. But then again, you still have the regular Django models.

Note that the solution for the Django models boils down to the same design: we save a model, which triggers a signal in the background, that is handled by a signal handler in which we log the event code. For the KATalogus, we save a model and directly do the log of the event code, but we could create a signal there called "plugin_saved" and listen on that to do the logging, and then we are not that far from the other event code implementations.

ammar92 (approved):

rocky/katalogus/client.py
```diff
@@ -213,18 +222,21 @@ def get_plugin_settings(self, organization_code: str, plugin_id: str) -> dict:
         return response.json()
 
     def upsert_plugin_settings(self, organization_code: str, plugin_id: str, values: dict) -> None:
+        logger.info("Adding plugin settings", event_code=800023, plugin=plugin_id)
```
underdarknl, line 225: Can we add the values?

rocky/katalogus/client.py
```diff
@@ -213,18 +222,21 @@ def get_plugin_settings(self, organization_code: str, plugin_id: str) -> dict:
...
         self.session.put(f"/v1/organisations/{quote(organization_code)}/{quote(plugin_id)}/settings", json=values)
 
         logger.info("Upsert plugin settings", plugin_id=plugin_id)
 
     def delete_plugin_settings(self, organization_code: str, plugin_id: str) -> None:
+        logger.info("Deleting plugin settings", event_code=800024, plugin=plugin_id)
         self.session.delete(f"/v1/organisations/{quote(organization_code)}/{quote(plugin_id)}/settings")
 
-        logger.info("Delete plugin settings", plugin_id=plugin_id)
+        logger.info("Deleted plugin settings", plugin_id=plugin_id)
```
underdarknl, line 234: Seems redundant?

stephanie0x00:
Enable boefje:
```
rocky-1                | {"event_code": 800021, "plugin": "fierce", "event": "Enabling plugin", "user_id": 1, "request_id": "5a1e78c3-4b21-4fcc-8a79-45b0c4bf69ef", "ip": "172.30.0.1", "organization_code": "aa", "level": "info", "timestamp": "2024-12-16T11:04:49.405002"}
```

Disable boefje:
```
rocky-1                | {"event_code": 800022, "plugin": "fierce", "event": "Disabling plugin", "user_id": 1, "request_id": "5a0f4a67-0bd7-4b96-9df0-1bb7d6c1cc54", "ip": "172.30.0.1", "organization_code": "aa", "level": "info", "timestamp": "2024-12-16T11:05:46.783598"}
```

Creating boefje:
```
rocky-1                | {"event_code": 800025, "boefje": "Boefje(id='b9a1720d-74cd-42d8-8e8e-981ba95e758f', name='MyBoefje', version=None, authors=None, created='2024-12-16 08:55:47.201936', description='', related=[], type='boefje', enabled=False, scan_level=SCAN_LEVEL.L2, consumes={<class 'octopoes.models.ooi.network.IPAddressV6'>, <class 'octopoes.models.ooi.network.IPAddressV4'>}, produces={'boefje/nmap'}, options=None, runnable_hash=None, interval=None, boefje_schema={'title': 'Arguments', 'type': 'object', 'properties': {'TOP_PORTS': {'title': 'TOP_PORTS', 'type': 'integer', 'minimum': 1, 'maximum': 65535, 'description': 'Scan TOP_PORTS most common ports. Defaults to 250.'}}, 'required': []}, oci_image='ghcr.io/minvws/openkat/nmap:latest', oci_arguments=['--open', '-T4', '-Pn', '-r', '-v10', '-sV', '-sS'])", "event": "Creating boefje", "user_id": 1, "ip": "172.30.0.1", "request_id": "800ef452-cfea-4b9f-9961-72ba5427f5fc", "organization_code": "aa", "level": "info", "timestamp": "2024-12-16T08:55:47.242614"}
```

underdarknl (approved):

Issue #4000, opened 2024-12-31
Current `__hash__` implementation of OOI is broken


originalsouth:
The current `__hash__` implementation of OOI's:
https://github.com/minvws/nl-kat-coordination/blob/8730e188e9dad6276a363aaeaead8fc1deb82ac9/octopoes/octopoes/models/__init__.py#L242-L243
is broken because it only considers the primary key; meaning that OOI's with fields _not_ recorded in the primary key are erroneously deemed to be the same objects, causing Python's built-in hash dependent structures to find collapses.

Since we are dealing with OOI based on Pydantic BaseModel's we can easily generate a dict of the object using `model_dump`. Assuming that this is the best object to start to begin our `__hash__` implementation on, the question becomes how to best hash a dict (as python still hasn't figured out how to do this natively). 

The natural question arises why not hash `model_dump_json`? Because there is no guarantee it is stable see https://github.com/pydantic/pydantic/discussions/10343.

Hence, here I compare two algorithms with benchmarks:
```
#!/usr/bin/env python

from typing import Iterable, Any
import jcs
import random
import string
import time

N = 8
MAX_DEPTH = 4
K = 10**6


def random_string():
    return ''.join(random.choice(string.ascii_letters) for _ in range(random.randint(1, N)))


def random_obj(depth: int = 0, k: int = 7) -> Any:
    if depth < MAX_DEPTH:
        rand_choice = random.randint(0, k)
        if rand_choice == 0:
            return random.random()  # float
        elif rand_choice == 1:
            return random.randint(0, 10**6)  # int
        elif rand_choice == 2:
            return random_string()  # string
        elif rand_choice == 3:
            return [random_obj(depth + 1, 2) for _ in range(random.randint(1, N))]  # list
        elif rand_choice == 4:
            return [random_obj(depth + 1, 2) for _ in range(random.randint(1, N))]  # list
            # Non JSON compatible so broken for hasher_2 but hasher_1 digests it
            # return {random_obj(depth + 1, 2) for _ in range(random.randint(1, N))}  # set
        else:
            return {random_string(): random_obj(depth + 1) for _ in range(random.randint(1, N))}  # dict[str, Any]
            # Non JSON compatible so broken for hasher_2 but hasher_1 digests it
            # return {random_obj(depth + 1, 2): random_obj(depth + 1) for _ in range(random.randint(1, N))}  # dict[Any, Any]
    else:
        return random_string()


targets = [random_obj() for _ in range(K)]


[...]

PR #4004, opened 2025-01-01
RFD 0003: OOI Hashing

jpbruinsslot:


rfd/0003-ooi-hashing.md
```diff
@@ -0,0 +1,125 @@
...
+this natively).
+
+The natural question arises why not hash `model_dump_json`? Because there is no
+guarantee it is stable see
+https://github.com/pydantic/pydantic/discussions/10343.
+
+## Evaluation
+
+Hence, here I compare two algorithms with benchmarks:
+
```
jpbruinsslot, line 35:
```suggestion
1. hashing the ooi using the [`jcs`](https://pypi.org/project/jcs/) package
2. hashing the ooi using a custom `freeze()` function

Code:
```

rfd/0003-ooi-hashing.md
```diff
@@ -0,0 +1,125 @@
...
+hasher_1: 2.213041571e-05s
+hasher_2: 3.159127834e-05s
+```
+
+## Determinations
+
+Personally, I would opt for `hasher_1` as it more flexible and faster, but
+`hasher_2` is easier to maintain; also open to other suggestions.
+
+So how do we proceed to solve this problem?
```
jpbruinsslot, line 125:
```suggestion
So how do we proceed to solve this problem?

## References
- Issue #4000
```

rfd/0003-ooi-hashing.md
```diff
@@ -0,0 +1,125 @@
...
+## Background
+
+The current `__hash__` implementation of OOI's:
+
+https://github.com/minvws/nl-kat-coordination/blob/8730e188e9dad6276a363aaeaead8fc1deb82ac9/octopoes/octopoes/models/__init__.py#L242-L243
+
+is broken because it only considers the primary key; meaning that OOI's with
+fields _not_ recorded in the primary key are erroneously deemed to be the same
+objects, causing Python's built-in hash dependent structures to find collapses.
+
```
jpbruinsslot, line 19:
```suggestion
Additionally, we'll have to consider whether we want for:

d1 = {'a': 1, 'b': 2}
d2 = {'b': 2, 'a': 1}

to have different hashes (ie. `hash(d1) == hash(d2)` or `hash(d1) != hash(d2)`).
(this because python dicts are ordered)
```

jpbruinsslot: having problems with the code suggestion of github and backticks, will add those later when this is committed

rfd/0003-ooi-hashing.md
```diff
@@ -0,0 +1,125 @@
...
+## Background
+
+The current `__hash__` implementation of OOI's:
+
+https://github.com/minvws/nl-kat-coordination/blob/8730e188e9dad6276a363aaeaead8fc1deb82ac9/octopoes/octopoes/models/__init__.py#L242-L243
+
+is broken because it only considers the primary key; meaning that OOI's with
+fields _not_ recorded in the primary key are erroneously deemed to be the same
+objects, causing Python's built-in hash dependent structures to find collapses.
```
underdarknl, line 18:
```suggestion
objects, causing Python's built-in hash dependent structures to find coliisions.
```

rfd/0003-ooi-hashing.md
```diff
@@ -0,0 +1,125 @@
+---
+authors: (@originalsouth)
+state: draft
+discussion: https://github.com/minvws/nl-kat-coordination/pull/4004
+labels: hashing, pydantic, ooi, octopoes
+---
+
+# RFD 0003: OOI Hashing
+
+## Background
```
underdarknl, line 10: I'd like a little more explanation on why we chose to override the default hashing and implement hashing only on the primary-key fields: Because we cannot have two different objects in the graph with the same primary key, it makes sense to also have the rest of the codebase 'think' they are the same by hashing only those fields that would lead to different objects in the graph.
I'd also like a little more info on why and where this is problematic for the python code itself if we disregard the notion that we can only keep one version of two objects in the database If their primary key is the same.
Furthermore a bit more details on why this specifically is a problem in the context of dealing with Objects in nibbles where we want to 'see' changes on any field that is used/consumed/read/queried  by the nibble, regardless of what the primary key's values are. For example, having TTL for a DNSRecord in the primary key is useless for our Graph, but you could still envision a nibble that alerts users of DNSRecords with TTL's that are either too long, or too short. Not being able to see TTL changes (because from the outside the hash of the ooi stayed the same) would mean we only get to run the Nibble once, which is obviously not what you'd want.

n.b. A test nibble doing exactly this (min / max on TTL based on a config) would be a great way of demonstrating this, and adding functionality.

originalsouth: https://github.com/minvws/nl-kat-coordination/issues/4000#issuecomment-2567694801

rfd/0003-ooi-hashing.md
```diff
@@ -0,0 +1,142 @@
...
+fields _not_ recorded in the primary key are erroneously deemed to be the same
+objects, causing Python's built-in hash dependent structures to find collisions.
+Additionally, we'll have to consider whether we want for:
+
+d1 = {'a': 1, 'b': 2}
+d2 = {'b': 2, 'a': 1}
+
+to have different hashes (ie. `hash(d1) == hash(d2)` or `hash(d1) != hash(d2)`).
+(this because python dicts are ordered)
+
```
jpbruinsslot, line 28:
```suggestion
These issues are particular problematid for nibbles:

- Nibbles xxhash OOI's for recalculation, the underlying serialization (necessary for the resolution of this problem) must be reliable for hashing.
- Nibbles are written with uniqueness of objects in mind and uses dicts and sets as such; they will likely suffer performance impact. Both because the data structures get slower and because nibbles will run/rerun unexpectedly.
```

originalsouth:
Now that there is a democratically fortified chosen solution elected, I think we can close this one.

jpbruinsslot:
> Now that there is a democratically fortified chosen solution elected, I think we can close this one.

If by closing you mean disregarding I would advise against this.

The course of action for the RFD would be to formalize the document with the chosen path and its considerations. Then the file will be committed to the project such that we have a record of architectural decisions and how we got there.

originalsouth:
I was referring to the latter. It may, however, be more considerate to refrain from considering the varying levels of in depth understanding on the topic at hand among the constituency.

originalsouth:
I think this issue can be closed, the issue was resolved with a kind of work around to the original question... see #4000.

Donnype:
> I think this issue can be closed, the issue was resolved with a kind of work around to the original question... see #4000.

I am closing this one then!

PR #4016, opened 2025-01-08
Octopoes combined schedulers integration

### LINKED ISSUE #3838: Combine all schedulers for all organisations

jpbruinsslot:
### Changes

Main PR: https://github.com/minvws/nl-kat-coordination/pull/3839

### Issue link

Closes #3838


octopoes/octopoes/events/manager.py
```diff
@@ -26,6 +26,7 @@ class ScanProfileMutation(BaseModel):
     operation: OperationType
     primary_key: str
     value: AbstractOOI | None = None
+    client_id: str
```
jpbruinsslot, line 29: I named this `client_id`  since organisations are referred to as clients in octopoes. So not sure if this should be organisation or client

Donnype (approved):
👌

PR #4028, opened 2025-01-15
Test the https-availability nibble-query

Donnype:
### Changes

Small fix merging in a feature branch.

### DIFF
--- a/octopoes/tests/integration/test_https_availability_nibble.py
+++ b/octopoes/tests/integration/test_https_availability_nibble.py
@@ -1,5 +1,6 @@
 import os
 from datetime import datetime
+from itertools import permutations
 from unittest.mock import Mock
 
 import pytest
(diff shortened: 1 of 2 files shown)


ammar92 (approved):
Nice work. I'm no expert in the queries, but did take a look at the tests.

While looking at this I had a couple of thoughts for potential improvements or something to think about (for later):
- `ip_generator` could be a nice fixture on 'function' scope level to be used by multiple tests
- `test_https_availability_query` might be a good example for test parameterization, but in this case maybe too complex/ abstract

octopoes/tests/integration/test_https_availability_nibble.py
```diff
@@ -11,14 +12,61 @@
 from octopoes.models.ooi.network import IPAddressV4, IPPort, Network, Protocol
 from octopoes.models.ooi.service import IPService, Service
 from octopoes.models.ooi.web import HostnameHTTPURL, HTTPHeader, HTTPResource, WebScheme, Website
+from octopoes.repositories.ooi_repository import XTDBOOIRepository
 
 if os.environ.get("CI") != "1":
     pytest.skip("Needs XTDB multinode container.", allow_module_level=True)
 
-STATIC_IP = ".".join((4 * "1 ").split())
+STATIC_IP = ".".join(4 * ["1"])
```
underdarknl, line 20: Could we just write 1.1.1.1 here?

originalsouth: The reason we write it that way is not to have SonarCloud report a "static ip security vulnerability".

octopoes/tests/integration/test_https_availability_nibble.py
```diff
@@ -11,14 +12,61 @@
...
+    for _ in range(20):
+        create_port(xtdb_ooi_repository, next(ip), 80, valid_time)
+    assert xtdb_ooi_repository.session.client.query(query)[0][-1] == 0
+
+    create_port(xtdb_ooi_repository, str(first_ip.address), 443, valid_time)
+    assert xtdb_ooi_repository.session.client.query(query)[0][-1] == 1
+
+    for _ in range(12):
+        create_port(xtdb_ooi_repository, next(ip), 443, valid_time)
+    assert xtdb_ooi_repository.session.client.query(query)[0][-1] == 13
```
Donnype, line 66: @originalsouth Or should this be 1 since the ips we created are not attached to the hostname through the website?

Donnype: Perhaps the ipaddress should be a parameter in the or-join, I'll check.

originalsouth: I think the query brakes with more than one targets, let me check...

PR #4047, opened 2025-01-24
Fix/reuse report ooi entities

### LINKED ISSUE #3729: [EPIC] We should use a ReportOOI entity per Report, or update a Report in-place and traverse the history API
### First steps

- [x] #4009
- [x] #4011
- [x] #4012

## Subsequent issues
- [x] #4013
- [ ] https://github.com/minvws/nl-kat-coordination/issues/4044

## Follow up issues:
- [ ] https://github.com/minvws/nl-kat-coordination/issues/4014

Notes:

- This also means that the report recipe can contain a query, meaning that the Report OOI will have a changing `input_oois` field. Hence it is not guaranteed in this case that a recipe will keep updating a single AssetReport.

---

## To be discussed

Given some new idea/input on reporting, and especially given the fact that we want to re-use ReportOOIs as per the conclusion at the bottom of this ticket, some practical issues have arisen from the current data model. The parent-report vs. subreport dynamic is a bit strange in the new scenario because the parent-child relation if no longer one-to-many, but many-to-many. This is because I could collect the same dns-report on example.org in multiple parent reports. Also, would we have to re-use parent OOIs as well? This only makes sense if the parent points to (a list of) child reports.

Questions arise:
- Should we refactor the ReportOOIs data model?
- And should we somehow refactor so we could stop checking if we should create a parent ReportOOI or not? We could perhaps do that by making every Report have a parent, or always have a list of "child" reports that in some cases is just 1 report.
- I suppose we should not reuse ReportRecipes?
- Reports have a field called `input_oois: list[str]`. How do we re-use these in the case of multiple oois? Or should this be `input_ooi: str` now? Then we need to wrap a multi-report in a different way

---

## Notes frontend meeting:

@rieven:
- Perhaps use a timeline instead of pagination perhaps in the case pagination is going to be restructured?
- I worry about the use-case where we use a query and the assets we report on grow?
- If we update AssetReports in-place and hence has new Bytes-data, aggregated report data is no longer in-sync: the data in the aggregate depends on all its children, that can change continuously.

@Donnype:
Let's collect the affected features/pages and other aspects to refine.

---
First suggestion, perhaps ignore the input_oois being part of the natural key since this will become very unreadable:

```python3

class BaseReport(OOI):
    name: str
    template: str | None = None
    date_generated: datetime

    organization_code: str
    organization_name: str
    organization_tags: list[str]
    data_raw_id: str

    observed_at: datetime
    report_recipe: Reference | None = ReferenceField("ReportRecipe", default=None)

    @classmethod
        return f"Report {reference.tokenized.name}"


class Report(BaseReport):
    object_type: Literal["Report"] = "Report"
    report_type: Literal["concatenated-report", "aggregate-organisation-report", "multi-organization-report"]

    input_oois: list[str]

    _natural_key_attrs = ["report_recipe"]


class AssetReport(BaseReport):
    object_type: Literal["AssetReport"] = "AssetReport"
    report_type: str

    input_ooi: str

    _natural_key_attrs = ["input_ooi", "report_type"]
```

---

Per Report, we now create a [ `ReportOOI`](https://github.com/minvws/nl-kat-coordination/blob/main/octopoes/octopoes/models/ooi/reports.py):

```python3
class Report(OOI):
    object_type: Literal["Report"] = "Report"

    name: str
    report_type: str
    template: str | None = None
    date_generated: datetime

    input_oois: list[str]

    report_id: UUID

    organization_code: str
    organization_name: str
    organization_tags: list[str]
    data_raw_id: str

    observed_at: datetime
    parent_report: Reference | None = ReferenceField("Report", default=None)
    report_recipe: Reference | None = ReferenceField("ReportRecipe", default=None)
    has_parent: bool

# as a reference:
class ReportRecipe(OOI):
    object_type: Literal["ReportRecipe"] = "ReportRecipe"

    recipe_id: UUID

    report_name_format: str
    subreport_name_format: str | None = None

    input_recipe: dict[str, Any]  # can contain a query which maintains a live set of OOIs or manually picked OOIs.
    parent_report_type: str | None = None
    report_types: list[str]

    cron_expression: str

    _natural_key_attrs = ["recipe_id"]
```

This means that every run of a recipe, we add a new object to XTDB (and Bytes) that has its own history, which shows it as a seperate row in the current table overview.

Another option is to update a `ReportOOI` in-place, reducing the list but meaning we should query the history API more often. The question is if we should change the implementation to perform 

### Pros
- We have a smaller reports overview page.
- Seeing changes in a single over time is more natively supported through the history API.
- We could re-use this per-entity history view logic for OOI detail pages.

### Cons

(@Donnype: When I try to model for XTDB, I first ask myself if it makes sense to save an entity over time in a regular relational database, which in the case of a Report I have to answer with "Yes". Older versions of reports are too important to hide in a historic version.)

- Any query involving looking for older versions of reports become more complicated and/or slow(er):
  - "Give me the reports generated between 10-10-2024 and 21-10-2024" implies fetching the history of _every_ report.
- You always need to provide timestamps or timestamped urls when sharing or talking about historic reports.
- You cannot provide a proper per-generated-report overview (i.e. query across historic reports) unless you start filtering and querying in memory over the  history API.

## Conclusion after discussion on 25-10-2024

After a vote we concluded that we will re-use the Report OOI. We expect XTDB 2.0 to be able to resolve any use-cases that pop up and use the history API to traverse any other queries/use-cases. Both @dekkers and @underdarknl think using the history for new versions of a Report is a more adequate representation of the actual situation. (@Donnype thinks creating new objects is a more adequate representation that would also save us from potentially not being able to handle more intricate queries across reports.)

Donnype:
### Changes

This PR changes the data structure around reporting. The main changes have been described in #3729, but this also includes:
- A new Bytes API endpoint where we can request multiple raw files at once
- An Octopoes API update where we now add scan profiles to the load_bulk endpoint for OOIs

Note that the tests are still a work in progress in Rocky!

### Issue link

Closes #3729 

### Demo
Scherm­afbeelding 2025-01-24 om 13 20 56

Scherm­afbeelding 2025-01-24 om 13 21 31

Scherm­afbeelding 2025-01-24 om 13 21 47

### QA notes

The main changes are around reporting, so please verify that:
- Scheduling (aggregate) reports still works properly, especially given the other updates around schedule once
- The report overview pages still work
- You see the right amount of asset reports
- You can render both aggregate, concatenated, and single asset reports
- Of course, that the jobs still run properly given the changes in Octopoes and Bytes!

---

### Code Checklist

---

---

## Checklist for QA:


octopoes/octopoes/repositories/ooi_repository.py
```diff
@@ -292,7 +304,10 @@ def load_bulk(self, references: set[Reference], valid_time: datetime) -> dict[st
         return {ooi.primary_key: ooi for ooi in oois}
 
     def load_bulk_as_list(self, references: set[Reference], valid_time: datetime) -> list[OOI]:
-        query = generate_pull_query(FieldSet.ALL_FIELDS, {self.pk_prefix: list(map(str, references))})
+        if not references:
+            return []
+
+        query = Query().where_in(OOI, id=references).pull(OOI, fields="[* {:_reference [*]}]")
```
underdarknl, line 310: I'd rather see the extra referenced objects loaded optionally. If you look at the queries and datasets that we encode/decode and transport over the various api's it becomes clear we need to be much more precise in our resultsets as in many cases we don't use most of what we return.

underdarknl (approved):

PR #4050, opened 2025-01-28
optimize various bits around scan profiles

underdarknl:
### Changes

I noticed we do a periodic recalculate_scan_profiles, which by itself loops over the list of profiles three times.

### QA notes
Scan profiles should still be handled the exact same way. The code is functionally the same as before.

### Code Checklist

---

---

## Checklist for QA:


underdarknl:
Further improvements:
possibly only fetch the state of the scanlevels IF there have been any changes. Can we select the highest transactionID from xtdb somehow?

ammar92 (dismissed):
Nice work! It's clever to do `all_declared_scan_profiles` and `assigned_scan_levels` in one pass. There could be one step more for little optimization, and that is by creating `source_scan_profile_references` after the loop. You gain a little performance improvement because it's faster to setup a set on an existing list (or any sequence or iterable) than constructing it element by element via `set.add` (e.g. previous implementation `source_scan_profile_references = {sp.reference for sp in all_declared_scan_profiles}`)

Example of benchmark code below.
```python3
import timeit

code1 = """
x = set()
l = []
for i in range(50000):
    l.append(i)
    x.add(i)
"""

print(timeit.timeit(code1, number=2000)) # 2.3204293749295175

code2 = """
l = [i for i in range(50000)]
x = set([i for i in l])
"""

print(timeit.timeit(code2, number=2000)) # 2.087212207959965
```

But the performance gain is negligible and the suggested implementation is already fast and understandable, so I'll leave it up to you.

underdarknl:
> Nice work! It's clever to do `all_declared_scan_profiles` and `assigned_scan_levels` in one pass. There could be one step more for little optimization, and that is by creating `source_scan_profile_references` after the loop. You gain a little performance improvement because it's faster to setup a set on an existing list (or any sequence or iterable) than constructing it element by element via `set.add` (e.g. previous implementation `source_scan_profile_references = {sp.reference for sp in all_declared_scan_profiles}`)

Yes, I was pondering this too, as set operations 'one by one' are slower than doing an update, or mass assignment. There's a bunch of other optimizations that can be done, but I'm focussing on *not* running the entire loop if its needed first, as thats the main win that would safe 90% or more for most installs.

Furthermore Im guessing all of this will be undone once the nibbles start taking care of propagation, as they do all this taint tracking in a *much* more efficient way.

octopoes/octopoes/repositories/scan_profile_repository.py
```diff
@@ -41,6 +41,7 @@ def get_bulk(self, references: set[Reference], valid_time: datetime) -> list[Sca
 class XTDBScanProfileRepository(ScanProfileRepository):
     object_type = "ScanProfile"
     pk_prefix = "xt/id"
+    typeadapter = TypeAdapter(ScanProfile)
```
underdarknl, line 44: image

underdarknl: This reduces the runtime for a loop that does no changes by about 30%

ammar92: Nice catch! Reusing the adapter will indeed prevent recreation of schemas. Unfortunately we do this on other places as well, I'll pick them up in a following PR.

Regarding the line, it's probably better to rename it to `scan_profile_adapter` and move it out of the class, either in the same module or in the `octopoes.models` module where the different scan profile types are declared. This way it can also be reused.

octopoes/octopoes/models/__init__.py
```diff
@@ -11,8 +11,8 @@
 class Reference(str):
     @classmethod
     def parse(cls, ref_str: str) -> tuple[str, str]:
-        object_type, *natural_key_parts = ref_str.split("|")
-        return object_type, "|".join(natural_key_parts)
+        object_type, *natural_key_parts = ref_str.split("|", 1)
+        return object_type, natural_key_parts[0] if natural_key_parts else ""
```
dekkers, line 15: This isn't functionally the same and also not really merging different loops, is this supposed to be in this PR? If so I think we need to look at this more carefully as this seems to change reference objects everywhere and would result in objects that have multiple natural keys parts to only return the first part as natural key?

The Reference unit tests seem to only test objects that have a natural key with a single part.

octopoes/octopoes/repositories/scan_profile_repository.py
underdarknl: We could add a more generic Dict / lookup table of the various models? We could even create a caching proxy/factory for all these adapters which only instantiates them if they are used.

octopoes/octopoes/models/__init__.py
underdarknl: Sorry, no, it was an optimization that I spotted and chose to include in this PR, its probably better of in a separate Pr.
Functionally, this does not change the outcome whether or not multiple pipes are present in the natural_key.

dekkers, line 15: I read the code incorrectly, it indeed doesn't change the functionality. Given that a reference always contains the class and the natural key, we can also simplify this to

```suggestion
        return ref_str.split("|", 1)
```

dekkers: I don't think we really have to split simple optimizations like this, but I think it would be good to mention in the PR title and description that there are multiple optimizations in a PR next time.

underdarknl:
the tests are now failing on ill-formatted test-references.

ammar92 (approved):

Donnype (approved):

PR #4051, opened 2025-01-28
RFD 0003: Deduplication of boefjes between organization

dekkers:
### Changes

Add deduplication RFD

### Issue link

Closes ...

### Demo

### QA notes

---

### Code Checklist

---

---

## Checklist for QA:


rfd/0003-deduplication.md
```diff
@@ -0,0 +1,214 @@
...
+Deduplication can only be done when the input of the boefje is exactly the same.
+This includes all the boefje settings. A few boefjes need to have the
+organization, so those boefjes can't be duplicated.
+
+When doing deduplication between organizations it would be possible for a user
+to see that a boefje task for a certain OOI was already done for an organization
+in the same KAT install earlier while the user might not have access to this
+organization. This might be a problem for certain usage of OpenKAT, so
+deduplication should be a setting that can be turned off.
+
```
underdarknl, line 19: We should mention that we can only use previous scan raw files *if* they are not older than the requested interval in the requesting organization.

rfd/0003-deduplication.md
```diff
@@ -0,0 +1,214 @@
...
+#### Usage of input_ooi field
+
+Only the leakix boefje used the input_ooi field. For the sake of simplicy we
+should probably not provide this to the boefje. It should be easy to refactor
+this so it used the serialized input OOI.
+
+#### Usage of environment field
+
+This is not used directly from the BoefjeMeta, but this
```
underdarknl, line 128: Missing end of sentence

rfd/0003-deduplication.md
```diff
@@ -0,0 +1,214 @@
...
+
+#### Usage of started_at, endated_at, boefje and runnable_hash fields
+
+This isn't used by any boefje and can be safely removed.
```
underdarknl, line 132: The runnable hash should not be made available to the boefje itself, however, they are used by the rest of OpenKAT / can be used by the rest.

rfd/0003-deduplication.md
```diff
@@ -0,0 +1,214 @@
...
+#### Whether boefje needs organization
+
+The scheduler already fetches and caches all the available boefjes for an
+organization, so the scheduler can use this data to easily check if the boefje
+requires the organization and can't be deduplicated.
+
+#### Environment settings from previous task in other organization
+
+This data is stored in bytes and needs to be fetched by the scheduler to
+determine whether the environment settings are the same.
```
underdarknl, line 169: Or, we could compare the runnable hash? Which should contain all these fields / settings.
The problem however is that The scheduler *can* look up the previous runnable hash, it cannot (since it cannot see the current boefje workers environment settings), compute the current runnable hash.

rfd/0003-deduplication.md
```diff
@@ -0,0 +1,214 @@
...
+passed directly to the boefje runner as environment variables. There is
+currently no easy way to fetch those and adding an API for this to the boefje
+runner doesn't seem to be a good idea.
+
+The best course of action seems to be to disallow the environment variables
+passed to the boefje runner when deduplication is turned on. If both
+deduplication is turned on and environment variables are passed to the boefje
+runner we need to return an error because the user needs to know that the passed
+variables aren't used.
+
```
underdarknl, line 184: I think another option would be to populate a list of previous valid runnable hashes, and check against those in the worker.
Eg, 
* Here's a new job: boefje + input + setting. 
* We have the following previous raw files on file [runnable_hash1, runnable_hash2]
* The worker then computes its own runnable hash from the given job + local env vars, and only proceeds to execute when the list of available jobs does not contain the runnable_hash for its current job.

rfd/0003-deduplication.md
```diff
@@ -0,0 +1,214 @@
...
+#### Deduplication algorithm
+
+A possible algorithm for deduplication could be:
+
+- The scheduler wants to scheduler a boefje for a specific organization and
+  input OOI.
+- The scheduler checks if the boefje needs the organization. If so it can
+  continu with scheduling the boefje task.
+- The scheduler queries the stored tasks to see if a different organization has
+  already run on this input OOI.
```
Donnype, line 194: What is the time window for fetching those tasks? Since no job will rerun at all strictly reading this algorithm

rfd/0003-deduplication.md
```diff
@@ -0,0 +1,214 @@
+# Deduplication of boefjes between organization
```
jpbruinsslot, line 1:
```suggestion
---
authors: Jeroen Dekkers (@dekkers )
state: draft
discussion: https://github.com/minvws/nl-kat-coordination/pull/4051
labels: boefjes
---

# RFD 0004: Deduplication of boefjes between organization
```

Note there was already a rfd 0003, only changing the file name from 3 to a 4 should suffice

rfd/0003-deduplication.md
```diff
@@ -0,0 +1,214 @@
...
+
+- The scheduler wants to scheduler a boefje for a specific organization and
+  input OOI.
+- The scheduler checks if the boefje needs the organization. If so it can
+  continu with scheduling the boefje task.
+- The scheduler queries the stored tasks to see if a different organization has
+  already run on this input OOI.
+- If there is a match, the scheduler fetches the environment of the previous
+  boefje task and the environment of the new boefje task. If they are the same
+  they can be deduplicated. If they differ the new boefje needs to be scheduled.
```
jpbruinsslot, line 197: This could be a nice solution, and from the scheduler point of view this can be implemented. A few questions:

1. When the scheduler deems the `Task` to be a duplication, do we still persist the `Task` for that organisation but its results to be linked to the results of another organisation? If so do we need to communicate this to the user, that their `Task` has been resolved by the results of another `Task` from another organisation?
2. The `BoefjeScheduler` as-is, is complex in relating to checking whether or not a `Task` should be pushed on the queue, do you have any ideas or suggestions for adding these suggestions while making sure that we don't add too much additional complexity?

rfd/0003-deduplication.md
```diff
@@ -0,0 +1,214 @@
...
+on an OOI.
+
+It might be better to provide the serialized ooi as `input_ooi` argument to the
+boefje `run` to make it more clear what the contents of the field is, see below
+for the proposed signature.
+
+#### Usage of organization field
+
+The external DB boefje currently needs the organization to fetch the data from
+the external database. Obviously this means the boefje can't be deduplicated.
```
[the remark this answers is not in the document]
underdarknl: DB_ORGANIZATION_IDENTIFIER can be 'optionally' set via the settings. The filtered env for each call would still be different though, and as such the signature would indeed be different.

rfd/0003-deduplication.md
underdarknl: Ah check. Then we should add a hash (we need that for our proofs anyway) I think?

rfd/0003-deduplication.md
underdarknl: 1. For an initial implementation I'd say the user doesn't need to notice anything, except maybe that the job was resolved faster than usual.

rfd/0003-deduplication.md
underdarknl: Max age of possible raw files should be the interval on the requesting organisations boefje. So if the requestion org has an interval of 1 day, files no older than 1 day could be used from another organisation.

dekkers:
I've added a few clarifications and listed two options for how to implement this. My preference is for option 2, checking for possibility of deduplication before/after a boefje has run. The reason is that we don't have to either use a valid time which is different from when the boefje has run or a valid time that is in the past.

dekkers:
I added possible implementation, see commit https://github.com/minvws/nl-kat-coordination/pull/4051/commits/dbec163f9e64ba6b61aef71f486fa3094c16834f

rfd/0004-deduplication.md
```diff
@@ -0,0 +1,251 @@
...
+obvious place to implement deduplication. Because the data necessary is
+currently spread out over multiple services / databases this means it would need
+to do requests to other services to fetch this information. This is unavoidable
+because there is currently no single place with all the information.
+
+#### Whether boefje needs organization
+
+The scheduler already fetches and caches all the available boefjes for an
+organization, so the scheduler can use this data to easily check if the boefje
+requires the organization and can't be deduplicated.
```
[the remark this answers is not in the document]
underdarknl: True, but the organization is going to be part of the arguments / env that the container can access, and as such will be listed in the schema if its required. If its required, the various arguments wont be the same between jobs from different orga's and as such wont de deduplicated based on their different signature.
* This is true once we drop env support and move all the configurables to be sources from the KATalogus.

Donnype (approved):

PR #4054, opened 2025-01-29
Add min_scan_level to nibbles with phantom result toggle

originalsouth:
Add min_scan_level to nibbles with phantom result toggle


underdarknl:
Im curious to why this phantom result is needed, what does it store, and why?

noamblitz:
@underdarknl its an interesting idea. Its because we do have to query to know which objects are relevant for the nibble. But after changing the scan level, there is no change in data so saving the phantom results in the first run saves a query when changing the scan level. If the scan level is high enough now, we can just write the phantom results without having to do any querying or calculation.

noamblitz:
I will add this to the docs including thought process as soon as I’m back.

underdarknl:
Could another possibility be to 'cache' the results of these phantom bits in a shared Cache somewhere, instead of adding them to XTDB? This would allow the subsequent runs of nibbles to utilize the cache without burdening XTDB with the possible data. Scaling this cache, or even clearing it after a few minutes on inactivity could be a good way to ensure proper cache hit rates without consuming loads of memory / storage when the graph isn't moving hat much anymore.

octopoes/octopoes/core/service.py
```diff
@@ -395,6 +395,17 @@ def recalculate_scan_profiles(self, valid_time: datetime) -> None:
         )
         logger.info("Recalculated scan profiles")
 
+    def nibblet_toggle_phantom_result(self, origin_id: str, valid_time: datetime) -> None:
```
Donnype, line 398: This is never invoked as far as I can see? Or is the PR hence still in draft?

noamblitz: Yes :)

originalsouth:
> Could another possibility be to 'cache' the results of these phantom bits in a shared Cache somewhere, instead of adding them to XTDB? This would allow the subsequent runs of nibbles to utilize the cache without burdening XTDB with the possible data. Scaling this cache, or even clearing it after a few minutes on inactivity could be a good way to ensure proper cache hit rates without consuming loads of memory / storage when the graph isn't moving hat much anymore.

Yes, that would be possible but we would indeed have to make sure it centralized. AFAIK the only way to this at this time is through XTDB. We could make, however, make a phantom nibble repository -- or we can start using a new service like redis or some other kv-store with REST interface?
I do not think it will matter much in performance, however, as the heavy part of this will be when we dephantomize the object.

originalsouth:
After extra testing, ready for merge was achieved by @noamblitz and @originalsouth. Proceeding to merge to feature/nibbles branch see #3808

Issue #4055, opened 2025-01-30
Overlapping dropdowns language and user settings


stephanie0x00:
**Describe the bug**
Switching between the language and user profile dropdown will result in an overlap of the dropdowns. 

**To Reproduce**

1. Click on the User profile icon in the top right corner. 
2. Click on language switcher icon
3. See that you have two dropdowns overlapping

**Expected behavior**
Only one of the dropdowns is shown at all times. 

**Screenshots**

![Image](https://github.com/user-attachments/assets/fd62e8ee-b5e2-43ce-980f-ed469126735d)

**OpenKAT version**
main

Issue #3402, opened 2024-08-26
Logout does not function as expected


underdarknl:
**Describe the bug**
clicking logout seems to return an error.

**To Reproduce**

1. Click Logout
2. See error
3. Notice you are not logged out

**Expected behavior**
The user should be logged out

**OpenKAT version**
Main

PR #4056, opened 2025-01-30
Fix dropdowns and logout 

Rieven:
### Changes

- Fixed double dropdown open due to inconsistency in templates.
- Fixed Logout button. The logout was not working due to a GET request instead of POST
- Correcting CSS for buttons and links with form wrapped or without.
- Correcting width and positioning of dropdowns.

### Issue link
https://github.com/minvws/nl-kat-coordination/issues/4055

Closes https://github.com/minvws/nl-kat-coordination/issues/4055

### Demo

https://github.com/user-attachments/assets/a9d0f265-2151-4383-a649-54ed8ef17bb3

### QA notes
you have to run yarn to test out these changes:
`cd rocky`
`nvm use 18`
`yarn`
`yarn dev`

---

### Code Checklist

---

---

## Checklist for QA:

### DIFF
--- a/rocky/assets/css/components/dropdown.scss
+++ b/rocky/assets/css/components/dropdown.scss
@@ -8,7 +8,6 @@
     white-space: nowrap;
     text-overflow: ellipsis;
     overflow: hidden;
-    min-width: 0;
     display: flex;
     width: 100%;
   }
--- a/rocky/assets/css/components/dropdown.scss
+++ b/rocky/assets/css/components/dropdown.scss
@@ -1,6 +1,4 @@
 .dropdown {
-  max-width: 50%;
-  width: auto;
   padding: 0;
   position: relative;
   align-self: center;
@@ -21,8 +19,6 @@
     position: absolute;
     top: 3.25rem;
     z-index: 2;
-    width: max-content;
-    max-width: 20rem;
     margin-block-end: 0;
     margin-block-start: 0;
     padding-inline-start: 0;
@@ -31,7 +27,11 @@
     border: 1px solid var(--colors-grey-200);
     background-color: var(--colors-white);
     max-height: 90vh;
-    overflow-x: scroll;
+
+    &.dropdown-right {
+      right: 0;
+      left: auto;
+    }
 
     > ul {
       margin: 0;
@@ -53,7 +53,23 @@
         padding: 0;
         width: 100%;
 
-        > a {
+        form {
+          width: 100%;
+          padding: 0;
+        }
+
+        button {
+          background: transparent;
+          border-radius: 0;
+          border: 0;
+          font-weight: normal;
+          line-height: var(--header-navigation-link-line-height);
+          width: 100%;
+          max-width: 100%;
+        }
+
+        > a,
+        button {
           display: flex;
           flex-direction: row;
           align-items: center;
@@ -65,6 +81,7 @@
           padding: var(--spacing-grid-150);
           text-decoration: none;
           width: 100%;
+          min-width: max-content;
 
           .icon {
             display: flex;
@@ -138,6 +155,5 @@
   [aria-expanded="true"] + & {
     display: flex;
     flex-direction: column;
-    min-width: 100%;
   }
 }
--- a/rocky/rocky/templates/partials/language-switcher.html
+++ b/rocky/rocky/templates/partials/language-switcher.html
@@ -1,38 +1,32 @@
 {% load i18n %}
 
 {% get_language_info for LANGUAGE_CODE as lang %}
-<div class="language-selector">
-    <div class="dropdown">
-        <button type="button"
-                aria-controls="language-switcher"
-                aria-expanded="false"
-                class="dropdown-button ghost">
-            {{ lang.name_local|title }}
-            <span aria-hidden="true" class="icon ti-chevron-down"></span>
-        </button>
-        <ul id="language-switcher" role="listbox" class="dropdown-list">
+<div class="dropdown">
+    <button type="button"
+            aria-controls="language-switcher"
+            aria-expanded="false"
+            class="dropdown-button ghost">
+        {{ lang.name_local|title }}
+        <span aria-hidden="true" class="icon ti-chevron-down"></span>
+    </button>
+    <div id="language-switcher" role="listbox" class="dropdown-list">
+        <ul>
             {% for language in languages %}
                 {% get_language_info for language as lang %}
                 {% if LANGUAGE_CODE != language %}
                     <li>
-                        <a hreflang="{{ language }}"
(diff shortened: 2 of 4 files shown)


rocky/assets/css/components/dropdown.scss
```diff
@@ -31,7 +26,11 @@
     border: 1px solid var(--colors-grey-200);
     background-color: var(--colors-white);
     max-height: 90vh;
-    overflow-x: scroll;
```
underdarknl, line 34: This was added for a reason. (eg, very long lists of organizations)

rocky/assets/css/components/dropdown.scss
```diff
@@ -1,6 +1,4 @@
 .dropdown {
-  max-width: 50%;
-  width: auto;
```
underdarknl, line 3: Are you sure you need to remove all of these lines to make the clicking work again? It seems unlikely.

rocky/rocky/templates/partials/language-switcher.html
```diff
@@ -1,38 +1,32 @@
...
-          action="{% url "set_language" %}"
-          method="post"
-          class="inline">
-        {% csrf_token %}
-    </form>
 </div>
+<form id="set-language"
+      action="{% url "set_language" %}"
+      method="post"
+      class="inline">
```
underdarknl, line 30: I guess the form could be class="hidden" as it contains no elements anyway?

rocky/rocky/templates/partials/secondary-menu.html
```diff
@@ -2,15 +2,18 @@
...
-         data-open-label="{% translate "User navigation" %}"
-         data-close-label="{% translate "Close user navigation" %}"
-         data-media="(min-width: 100%)"
-         data-button-classes="user-icon"
-         aria-label="{% translate "User navigation" %}"
-         class="collapsible user-nav">
-        <div class="collapsing-element">
+    <div class="dropdown">
+        <button type="button"
+                aria-controls="user-profile"
```
underdarknl, line 7: We seem to have lost the data-open en data-close labels?

rocky/assets/css/components/dropdown.scss
Rieven: Yes, but we do not need to scroll in the x direction

rocky/assets/css/components/dropdown.scss
Rieven: There were a lot of inconsistencies when it comes to width, having the items in the dropdown to expand for `a` and `buttons` they must expand to full width and where the min width is the width of the `max-content`.

rocky/rocky/templates/partials/language-switcher.html
```diff
@@ -1,38 +1,32 @@
...
-                        <a hreflang="{{ language }}"
-                           href="#"
-                           role="option"
-                           data-value="{{ language }}"
-                           lang="{{ language }}">
-                            <button form="set-language" type="submit" name="language" value={{ language }}>
-                                {{ lang.name_local.title }}
-                            </button>
-                        </a>
+                        <button form="set-language" type="submit" name="language" value={{ language }}>
```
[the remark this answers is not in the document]
noamblitz: I think this is how it should be done according to the Django docs

rocky/assets/css/components/dropdown.scss
underdarknl: Hm true, that should have been overflow-y
Odd that it seems to have worked anyway, probably because scroll is on by default..

rocky/rocky/templates/partials/language-switcher.html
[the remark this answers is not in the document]
ammar92: I assume because we're using Django's `set_language` redirect view that handles this switch. According to the [docs](https://docs.djangoproject.com/en/5.1/topics/i18n/translation/#the-set-language-redirect-view): The view expects to be called via the POST method, with a language parameter set in request.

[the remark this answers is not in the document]
Rieven: Here is an explanation:
https://github.com/django/django/blob/stable/5.1.x/django/views/i18n.py#L30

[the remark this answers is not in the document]
underdarknl: POST requests are for State changes. 
A switch to a different language is not a state change (or shoudn't be) as the used language is present in each url, and cannot rely on stateful magic to select the language anyway.
For example, what would be the state if you where to select English. And then go to an url with dutch listed as the language. Would that GET url magically do the same state change, or does the language not reply on state, but instead just rely on the language listed in the url?

rocky/rocky/templates/partials/secondary-menu.html
Rieven: Ahh, yes, I will add them again. This code is a bit ols as we were waiting for the new Manon. I don't know if this is the case now, or I can just add it for the new styling of manon. I know that there were some differences in rocky why we use buttons. reference: https://minvws.github.io/nl-rdo-manon/components/collapsible

Rieven: Now I know why we choose the button styling. using styling of Manon we can only add text for `data-open-label` and `data-close-label` Here is an example:
image

The text will be added as dropdown text, but we use icons as well, the alternative was to use buttons instead so we can have more flexibility. But I think we can find a mid way between both. I think I can just add these labels and change text when it opens for screen readers.

rocky/rocky/templates/partials/language-switcher.html
Rieven: I moved this code together with the language selector button. It has 1 button in it that submits the value or we can create an inpput type hidden to set the value.

rocky/rocky/templates/partials/secondary-menu.html
ppvg: I'm not sure I know what the question is so I'd have to look into this.

I can say though that:
1. The reason for Manon's dropdown to rely on `data-` attributes for the button label was for progressive enhancement, i.e. so that Manon can "inject" the button if (and **only** if) JS is available.
2. I'm not convinced changing the label was a good idea, and it would probably be better to use the label only to identify (e.g. "profile menu") and use ARIA attributes to convey state (e.g. `aria-controls` + `aria-expanded=` true/false). We should probably revise Manon's dropdowns in a future version.

underdarknl:
closes https://github.com/minvws/nl-kat-coordination/issues/3402

Donnype:
@Rieven What's the status of this PR?

Rieven:
> @Rieven What's the status of this PR?

@HeleenSG is going to look into it.

PR #4061, opened 2025-02-04
Add run-on to Boefje Setup page

### LINKED ISSUE #3929: Implement runon in user interface / boefjes variant flow
## About this feature

We should show he user if a boefje is using the runon flag instead of the interval flag.
We should allow the user to set these options on boefje variants.

See: https://github.com/minvws/nl-kat-coordination/pull/3906

### Detailed description

Since we have now added runon create and runon update boefje flags in the scheduler, we should allow the user to set these flags on new boefje variants they produce, and also see which boefjes carry this flag.

I want to configure a new boefje variant to run only on new ooi's, or updated ooi's.

### Specifications

The implementation should include…

Either the interval is selected, Or a runon flag is iset.
The runon flag can have on of three values:
* create
* update
* create, update

runon is none or a list.

## Design

Do we want an icon / color for these boefjes? @RomijnHumanoids 

### Screenshots

### Figma link

## Implementation

### Possible solution

_Outline your proposed solution for implementing the feature. You can include any specific ideas, designs, or functionalities here._

### Alternatives considered

_Describe any alternative approaches or solutions you've considered, and why you believe the proposed solution is superior._

madelondohmen:
### Changes
This PR adds the possibility for a user to select "Run On" to the Boefje (variant).
The user can choose one out of two options for when the Boefje should run:
1. On a specific **interval**
2. Every time an object is **created and/or changed**

If the user selects one of the options, the other option should be empty/None.
It is also possible for the user to not select any of the options. In that case, the default system scan frequency will be used.

### Issue link
Closes #3929

### Demo
https://github.com/user-attachments/assets/f2a3de2f-5d9a-4650-a0da-062ee322a1ec

### QA notes
Check if all of this works:
- Create a new Boefje, with:
    - An interval per minute/hour/day/week/year
    - Run on create and/or change 
    - Nothing selected (should work as well)

Also check these two flows:
- Edit an existing Boefje
- Create a Boefje variant

---

### Code Checklist

---

---

## Checklist for QA:


boefjes/boefjes/sql/db_models.py
```diff
@@ -72,6 +72,7 @@ class BoefjeInDB(SQL_BASE):
     schema = Column(types.JSON(), nullable=True)
     cron = Column(types.String(length=128), nullable=True)
     interval = Column(types.Integer, nullable=True)
+    run_on = Column(types.ARRAY(types.String(length=128)), nullable=True)
```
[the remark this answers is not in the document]
underdarknl: And add the enum somewhere logical and resuable.

class RunonEnum(StrEnum):
    Create =0 
    Update = 1
    CreateUpdate = 2

We could even opt for an FlagEnum if we feel binary competent.

rocky/assets/js/boefjeSetupRunOn.js
```diff
@@ -0,0 +1,36 @@
+function toggleElements(value) {
```
underdarknl, line 1: If feel this name is a bit too general. Lets make it more specific to avoid collisions in the future.

rocky/assets/js/boefjeSetupRunOn.js
```diff
@@ -0,0 +1,36 @@
...
+    intervalFrequency.classList.add("hidden-input");
+    runOn.classList.remove("hidden-input");
+  } else {
+    intervalNumber.classList.add("hidden-input");
+    intervalFrequency.classList.add("hidden-input");
+    runOn.classList.add("hidden-input");
+  }
+}
+
+document.addEventListener("DOMContentLoaded", function () {
```
underdarknl, line 25: Make the internal anonymous function not anonymous. That way If we get to a situation where we want to add another set of form elements, we can just re-call the init function that sets up the event listeners.

rocky/assets/js/boefjeSetupRunOn.js
```diff
@@ -0,0 +1,36 @@
+function toggleElements(value) {
+  const intervalNumber = document.querySelector(
+    "label[for='id_interval_number']",
+  ).parentElement;
+  const intervalFrequency = document.querySelector(
+    "label[for='id_interval_frequency']",
```
underdarknl, line 6: Can we select the labels via a class selector? That way we don't get ID conflicts when adding more than one set of toggleable inputs.

rocky/katalogus/views/boefje_setup.py
```diff
@@ -232,3 +254,22 @@ def create_boefje_with_form_data(form_data, plugin_id: str, created: str | None)
...
     )
+
+
+def get_interval_minutes(interval_number, interval_frequency) -> int:
+    interval_minutes = 0
+    if interval_number == 0:
+        return None
+
+    if interval_frequency == "minutes":
+        interval_minutes = interval_number
```
underdarknl, line 265: You could simply return the int()ed result immediately. No need for all the elif branches.

boefjes/boefjes/sql/db_models.py
```diff
@@ -14,6 +15,36 @@ class ScanLevel(Enum):
     L4 = 4
 
 
+class RunOnDB(Enum):
+    CREATE = "create"
+    UPDATE = "update"
+    CREATE_UPDATE = "create_update"
+
+    @classmethod
+    def from_run_ons(cls, run_ons: list[RunOn] | None):
```
ammar92, line 24:
```suggestion
    def from_run_ons(cls, run_ons: list[RunOn] | None) -> RunOn | None:
```

rocky/katalogus/views/boefje_setup.py
```diff
@@ -232,3 +254,18 @@ def create_boefje_with_form_data(form_data, plugin_id: str, created: str | None)
         oci_image=form_data["oci_image"],
         oci_arguments=arguments,
     )
+
+
+def get_interval_minutes(interval_number, interval_frequency) -> int | None:
```
ammar92, line 259:
```suggestion
def get_interval_minutes(interval_number: int, interval_frequency: str) -> int | None:
```

- It would be nicer to throw a `ValueError` on invalid parameter (e.g. unknown `interval_frequency` argument)
- You can use `Literal["minutes", "hours", ...]` to type hint this literal of choices

boefjes/boefjes/models.py
```diff
@@ -34,6 +47,7 @@ class Boefje(Plugin):
     boefje_schema: dict | None = None
     cron: str | None = None
     interval: int | None = None
+    run_on: list[RunOn] | None = None  # {"run_on": "create_update"} -> {"run_on": ["create", "update"]}
```
ammar92, line 50: IMO the type suggestion is already clear on how to use this

```suggestion
    run_on: list[RunOn] | None = None
```

ammar92 (commented):
Looks good, minimal remarks

boefjes/boefjes/sql/db_models.py
Donnype: @madelondohmen This should be `Self` from typing_extensions (probably) it should actually return `RunOnDB | None` which won't work because it references itself. 
```suggestion
    def from_run_ons(cls, run_ons: list[RunOn] | None) -> Self | None:
```

underdarknl:
I've made the Javascript more general, and reusable for other variants of the same logic.

boefjes/boefjes/sql/db_models.py
ammar92: `typing.Self` is introduced in Python 3.11, but we still use 3.10 for development. Therefore:
`def from_run_ons(cls, run_ons: list[RunOn] | None) -> "RunOnDB" | None:`

ammar92:
## Checklist for QA:

### What works:

- I can select either option and it will be displayed correctly in the variants list
- I can change the option and it will be updated accordingly

### Bug or feature?:

- I still get to see both forms regardless which option I choose for the scan type:
image

underdarknl:
> ### Bug or feature?:
> 
>     * I still get to see both forms regardless which option I choose for the scan type:


Oops, copy paste error. Is now fixed.

dekkers (approved):

PR #4081, opened 2025-02-10
Update build-rdo-package.yml

sigio:
Add python-version number to whl package

### Changes

Make sure package name is unique between workflow-runs when building against multiple python versions in one workflow.


Donnype (approved):
👍

PR #4093, opened 2025-02-13
Fix/default finding type nibble not in cache

noamblitz:
Dont run this nibble in cache as to avoid https://github.com/minvws/nl-kat-coordination/pull/3808#issuecomment-2656269568


noamblitz:
To do:

Only works for Katfindingtypes for now.

Nibble still returns findingtype without score... should be fixed by deduplication.

PR #3892, opened 2024-11-27
Improve the KATalogus `/plugins` endpoint performance

Donnype:
### Changes

And add a bulk endpoint for the declarations of the CSV upload. The main fixes in the KATalogus are:
- Do not query the database per boefje id on each request to fetch the `enabled` state, just do 1 query per request
- Query both boefje and normalizers at the same time
- Do not query the existence of the organization on each request: this is added as a filter parameter anyways and this should be checked in Rocky
- Cache boefje/normalizer module loading not on an instance-level but with a function @cache decorator, since then we have the cache available for other requests as well where before we had a new instance of the local repository per request.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/3903
Closes https://github.com/minvws/nl-kat-coordination/issues/2524

#### Before

Snapshot of `htop` (also notice the postgres process because we did 100+ queries per second, because we managed to handle 2 requests to the `/plugins` endpoint per second):
Scherm­afbeelding 2024-12-02 om 09 54 43

Requests per second (RPS) of the `plugin` endpoint, tested using [Locust](https://locust.io/):
Scherm­afbeelding 2024-11-29 om 11 34 50

#### After
Requests per second (RPS) of the `plugin` endpoint, tested using [Locust](https://locust.io/):
Scherm­afbeelding 2024-11-29 om 11 32 31

Snapshot of `htop` (the boefje api processes where nowhere near the top):
Scherm­afbeelding 2024-12-02 om 08 30 43

### QA notes

Please verify that the CSV upload still works and the scheduler picks up tasks, also from duplicated boefjes! Perhaps use the [test server](https://kattest03.openkat.dev/) that I used to verify the interface works OK with 300 organizations.

---

### Code Checklist

- [x] All the commits in this PR are properly PGP-signed and verified.
- [x] This PR only contains functionality relevant to the issue.
- [x] I have written unit tests for the changes or fixes I made.
- [x] I have checked the documentation and made changes where necessary.
- [x] I have performed a self-review of my code and refactored it to the best of my abilities.

---

## Checklist for code reviewers:

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_qa.md) into your comment.

PR #4108, opened 2025-02-19
Remove caches for the KATalogus in the scheduler

jpbruinsslot:
### Changes

- Add a filter parameter in the KATalogus Plugins API to allow filtering on what the plugin consumes and produces to support it
- Removal of katalogus caches in the scheduler

### Issue link

Reference https://github.com/minvws/nl-kat-coordination/issues/3357

### QA notes

Interaction between the scheduler and katalogus has been optimized. Scanning of large volumes of items should function better, and without any issues monitored in either the katalogus or the scheduler.

---

### Code Checklist

---

---

## Checklist for QA:


boefjes/boefjes/katalogus/plugins.py
```diff
@@ -70,6 +70,12 @@ def list_plugins(
     # filter plugins by scan level for boefje plugins
     plugins = [plugin for plugin in plugins if plugin.type != "boefje" or plugin.scan_level >= filter_params.scan_level]
 
+    if filter_params.consumes is not None:
+        plugins = [plugin for plugin in plugins if filter_params.consumes.issubset(set(plugin.consumes))]
+
+    if filter_params.produces is not None:
+        plugins = [plugin for plugin in plugins if filter_params.produces.issubset(set(plugin.produces))]
```
underdarknl, line 77: Does this actually allow for plugins that only produce one of the filtered produce values?
Eg, I filter on produces, X or Y. this plugin produces Y and Z.
X and y is not a subset of Y and Z. It does produce an intersect result which Is probably what we are looking for as filters are usually OR, not AND?

underdarknl:
```suggestion
        plugins = [plugin for plugin in plugins if filter_params.produces.intersection(set(plugin.produces))]
```

boefjes/boefjes/katalogus/plugins.py
```diff
@@ -70,6 +70,12 @@ def list_plugins(
     # filter plugins by scan level for boefje plugins
     plugins = [plugin for plugin in plugins if plugin.type != "boefje" or plugin.scan_level >= filter_params.scan_level]
 
+    if filter_params.consumes is not None:
+        plugins = [plugin for plugin in plugins if filter_params.consumes.issubset(set(plugin.consumes))]
```
underdarknl, line 74:
```suggestion
        plugins = [plugin for plugin in plugins if filter_params.consumes.intersection(set(plugin.consumes))]
```

boefjes/boefjes/katalogus/plugins.py
jpbruinsslot: Perhaps not relevant to the discussion, but the scheduler will only request one type.

dekkers:
I don't really see any discussion about the scalability impact of removing the cache. The cache was implemented to speed things up. Won't this result in lots of requests to the katalogus in large installations that would previously hit the cache?

boefjes/boefjes/katalogus/plugins.py
Donnype: The idea when writing this was to do an AND, but an OR indeed makes more sense

Donnype:
> I don't really see any discussion about the scalability impact of removing the cache. The cache was implemented to speed things up. Won't this result in lots of requests to the katalogus in large installations that would previously hit the cache?

Unfortunately, there are also discussions or benchmarks that justify the cache. As far as I understand the biggest factor to building the internal data structure was not being able to filter the plugins by what they produce and consume. Moreover, the katalogus has been sped up significantly in [this PR](https://github.com/minvws/nl-kat-coordination/pull/3892) and will be hit even less given the new "global" scheduler.

---

But to at least reason about the change a bit going through processing of the scan profile mutations (the same arguments roughly hold for rescheduling). There are about 4 calls made to process scan profile mutations of which 3 use the same boefje cache: set_cron, calculate_deadline and has_boefje_task_grace_period_passed. (Note that the calls from set_cron and calculate_deadline could be merged if necessary.) Given a warm cache this introduces 3 extra calls per pushed boefje task.

If this becomes a bottleneck, processing one task is less expensive than those calls. In that case, it's safe to assume there are many mutations to process at once. In my eyes, a simpler optimization would be processing batches of the mutations and making just one call to the katalogus per batch. We would also be able to do batch updates to the database and octopoes where needed, meaning this might also be a more effective optimization.

Hence I'd say this PR both improves the katalogus API and simplifies the scheduler without any major cost or risks. It could also allow scheduling to be more responsive to changes in the katalogus such as adding/enabling boefjes.

jpbruinsslot:
Fair point, however at the moment we don't have data (at least not to my knowledge) to either support or refute that caching was/is necessary in the scheduler. Preliminary tests shows that with this PR the katalogus is more than capable of handling the load that the scheduler puts on it with a high volume of requests.

One of the reasons of moving away from a caching in the scheduler is that it adds additional complexity and that it dilutes the core purpose of the scheduler. Additionally if we want to be able to independently scale the scheduler we are likely to run into issues of stale data / divergent views of the same data.

In addition to the suggestions @Donnype made I would add if/when we deem it necessary to implement caching to offload the katalogus, to look at alternative caching methods like distributed caches.

boefjes/boefjes/katalogus/plugins.py
underdarknl: Doing an AND would change the behaviour of boefjes I'd guess?

underdarknl:
> Fair point, however at the moment we don't have data (at least not to my knowledge) to either support or refute that caching was/is necessary in the scheduler. Preliminary tests shows that with this PR the katalogus is more than capable of handling the load that the scheduler puts on it with a high volume of requests.
> 
> One of the reasons of moving away from a caching in the scheduler is that it adds additional complexity and that it dilutes the core purpose of the scheduler. Additionally if we want to be able to independently scale the scheduler we are likely to run into issues of stale data / divergent views of the same data.
> 
> In addition to the suggestions @Donnype made I would add if/when we deem it necessary to implement caching to offload the katalogus, to look at alternative caching methods like distributed caches.

We do actually know that on Z-cert's install with ~300 organizations, the continous katalogus update loop in the scheduler was taking longer than the interval. It also created a very long startup time for the scheduler (which the caching did not fix).
Since then the katalogus has sped up considerably, but I'd still want to not do any work we don't have to do. I think we should (in tandem with simplifying the code as proposed), look at E-tag headers set by the katalogus (based on the hash of katalogus data for a given orga) so we can skip fetching and processing unchanged katalogus info.
Beyond that, this all is to be removed / replaced by an event driven update model once we get the katalogus info stored in the graph anyway.

Donnype:
@underdarknl I've looked into Etag headers for the KATalogus shortly two weeks ago and recall concluding that these make more sense to use for file-like resources where data is mostly relatively large and static, so that there are a lot of cache hits and computing the hash is significantly faster than the bandwith used to actually send over the data. The concept of "the hash of katalogus data for a given orga" however is not that trivial to compute. We would have to fetch (though from a cache now) all boefje data as read from the disk, but also query the database for all the data of the organization before we would have all the information to compute a consistent hash.

As a rough indication, I did some quick local checks:

#### Etag approach

Fetching the local plugin info from a warm cache and hashing it: 1 ms
Querying the Boefje and BoefjeConfig models from the database: 27 ms
Parsing and Hashing the database data: perhaps 1 ms (harder to isolate and time)?
HTTP overhead: based on timing a random head requests to the katalogus: about 6 ms

Total: 35 ms

#### No Etag
Just doing `requests.get("http://localhost:8003/v1/organisations/test/plugins")` gives:

Total: 25 ms

boefjes/boefjes/katalogus/plugins.py
Donnype: How so? Given that the scheduler only uses one element in its queries, both implementations would have the same result from the schedulers point of view

underdarknl:
> @underdarknl I've looked into Etag headers for the KATalogus shortly two weeks ago and recall concluding that these make more sense to use for file-like resources where data is mostly relatively large and static, so that there are a lot of cache hits and computing the hash is significantly faster than the bandwith used to actually send over the data. The concept of "the hash of katalogus data for a given orga" however is not that trivial to compute. We would have to fetch (though from a cache now) all boefje data as read from the disk, but also query the database for all the data of the organization to compute a adequate hash.
> 
> As a rough indication, I did some checks locally to estimate the performance impact:
> #### Etag approach
> 
> Fetching the local plugin info from a warm cache and hashing it: 1 ms Querying the Boefje and BoefjeConfig models from the database: 27 ms Parsing and Hashing the database data: perhaps 1 ms (harder to isolate and time)? HTTP overhead: based on timing a random head requests to the katalogus: about 6 ms
> 
> Total: 35 ms
> #### No Etag
> 
> Just doing `requests.get("http://localhost:8003/v1/organisations/test/plugins")` gives:
> 
> Total: 25 ms
> 
> Hence, Etag headers might even reduce performance unless we actively update the hash whenever data changes in the KATalogus (degrading write performance, introducing extra bookkeeping, etc.).

That only takes into account the time used by the katalogus to generate the result and or etag. Maybe keeping the hash of the previous result in the scheduler to make sure we dont reset the internal data-structured when there is not change is a better way of dealing with this. Since we would then still need to implement hashing of the output stream we could choose to do this hashing either in the client (scheduler), or the server (katalogus). Doing so in the katalogus is as fast, with the added bonus that we could then use the hash as the etag, making sure the scheduler can do a head request, or only consume the header portion of a get request then the same etag value is repeatedly returned by the katalogus. The Scheduler could even cache the resulting internal data structures and re-use them if two orga's have the same hash. (as is common for large installations) by deep copying the whole thing.

I'm not sure why you include the database in your comment, as the etag send out with the initial GET response could just be the hash of the data in the response.

Donnype:
> That only takes into account the time used by the katalogus to generate the result and or etag. Maybe keeping the hash of the previous result in the scheduler to make sure we dont reset the internal data-structured when there is not change is a better way of dealing with this. Since we would then still need to implement hashing of the output stream we could choose to do this hashing either in the client (scheduler), or the server (katalogus). Doing so in the katalogus is as fast, with the added bonus that we could then use the hash as the etag, making sure the scheduler can do a head request, or only consume the header portion of a get request then the same etag value is repeatedly returned by the katalogus. The Scheduler could even cache the resulting internal data structures and re-use them if two orga's have the same hash. (as is common for large installations) by deep copying the whole thing.
> 
> I'm not sure why you include the database in your comment, as the etag send out with the initial GET response could just be the hash of the data in the response.

The current change actually means we do not need to keep track of an internal data structure, which largely is the simplicity-win. The Etags as far as I understand should communicate whether the data in the KATalogus has changed. That depends on data from the database (copied boefjes, boefjes being enabled/disabled, etc.). So the only way not to query the database on each HEAD request would be to either update the Etag on writes so it is readily available, or cache the Etag in the KATalogus and only update it e.g. every minute. But at that point we have 2 caches instead of 0, the latter being what this PR proposes (and I'd advocate given the current performance).

dekkers:
I also don't see how ETags are going to help us much here. To compute the ETag you need to do the same things as you need to just fulfill the request. If you want to pre-compute or cache the ETag everything gets very complex and I am also not sure if we are really going to win much with it.

I think 30 ms for the request won't be a big problem. If it is a problem the solution is also clear and simple in my opinion, just let the scheduler query PostgreSQL directly. I just tested and I get 30 ms for http request to the katalogus service, but PostgreSQL only takes 0.4 ms for an SQL query. But I think that's not a discussion for this ticket.

stephanie0x00:
## Checklist for QA:

### What works:
Seems to work. Didn't compare speed to the main branch. 
Tasks run and complete, normalizers work again, can enable boefjes in de katalogus. 
Reports can be scheduled, created and opened.

### What doesn't work:
n/a

### Bug or feature?:
n/a

dekkers (approved):

Issue #325, opened 2023-01-17
As a user I would like to see the content of the logs without downloading it.


HeleenSG:
**===== Do not implement until it has been approved and specified. =======**

**Feature request** 
As a user I would like to see the content of the logs without downloading it.

**Thoughts / questions to discuss**
Is it safely possible to show the content of the logs within the ui? It will make it easier for users to see why jobs have failed and what jobs have collected/done. But there are risks to showing unparsed remote data. 

![Screenshot 2023-01-17 at 14 10 04](https://user-images.githubusercontent.com/11405133/212907367-9bc57442-89da-48ed-895c-500d66562105.png)

For more information see the figma and the related comments. 
https://www.figma.com/file/QRL8085Z827bNdO0bIDdFL/KAT-%7C-Design-%7C-For-review?node-id=1%3A3&t=lZE8qsKdKNgKvk9Y-1

brennodewinter:
Consolidated into #2695 during backlog triage. The relevant scope and any unique context have been summarized on #2695. Closing this issue as a duplicate.

PR #4190, opened 2025-03-19
Feat/raw files in tasklist

underdarknl:
### Changes

Adds limited length in-line viewing of raw files on the task list. Still needs work to adhere to the designs: https://www.figma.com/file/QRL8085Z827bNdO0bIDdFL/KAT-%7C-Design-%7C-For-review?node-id=1%3A3&t=lZE8qsKdKNgKvk9Y-1, but @stephanie0x00 's security considerations have been applied.

### Issue link

Closes (https://github.com/minvws/nl-kat-coordination/issues/325

### Demo

image

### QA notes

Raw files should be visible when opening the expando-row on a task in the tasklist.
Boefjes wilt multiple raw files should show all.
Massive raw files should be limited to 1MB.
binary raw files will probably still break things.

### Code Checklist

---

---

## Checklist for QA:

### DIFF
--- a/rocky/assets/js/renderBoefjeOutput.js
+++ b/rocky/assets/js/renderBoefjeOutput.js
@@ -1,0 +1,118 @@
+const htmlElement = document.getElementsByTagName("html")[0];
+const language = htmlElement.getAttribute("lang");
+const organization_code = htmlElement.getAttribute("data-organization-code");
+
+const buttons = document.querySelectorAll(
+  ".expando-button.boefjes-task-list-table-row",
+);
+
+buttons.forEach((button) => {
+  const raw_task_id = button.closest("tr").getAttribute("data-task-id");
+  const task_id = button
+    .closest("tr")
+    .getAttribute("data-task-id")
+    .replace(/-/g, "");
+  const organization =
+    organization_code ||
+    button.closest("tr").getAttribute("data-organization-code");
+  const json_url =
+    "/" +
+    language +
+    "/" +
+    organization +
+    "/bytes/" +
+    encodeURI(task_id) +
+    '/raw?format=json';
+
+  const getJson = (url, callback) => {
+    var xhr = new XMLHttpRequest();
+    xhr.open("GET", url, true);
+    xhr.responseType = "json";
+    xhr.onload = function () {
+      var status = xhr.status;
+
+      const messages = button
+        .closest("tr")
+        .nextElementSibling.querySelectorAll(".error, .explanation");
+      messages.forEach((element) => {
+        element.remove();
+      });
+
+      if (status === 200) {
+        callback(xhr.response);
+      } else {
+        // Create HTML elements to build a snippet containing the error message.
+        let error_element = document.createElement("div");
+        let error_text_element = document.createElement("p");
+
+        error_text_element.innerText =
+          "Retrieving yielded raw files resulted in a Sever Error: Status code ";
+        error_element.appendChild(error_text_element);
+        error_element.classList.add("error");
+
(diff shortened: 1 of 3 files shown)


underdarknl:
Somewhere, we need to add some CSS that makes the pre inside the rawfiles list not be max-width: 100% but max-width: 100vw;

underdarknl:
image
Binary files do show up, however, they should be rendered as a hex display I'd say.

ammar92:
I've fixed some errors, linting and language files. Moving it to QA

stephanie0x00:
## Checklist for QA:

### What works:
- Raw files are shown \o/
- Boefjes with multiple raw files show all raw files.


### What doesn't work:
- Tried to test a binary file upload, however none of the binary files tried would be parsed by OpenKAT, so still looking into this. 
- See 4 points below. 

### Bug or feature?:

1. Displaying of data could be more user friendly to enhance readability. 
   - There are two headers (yielded objects and yielded raw files). I believe Yielded objects only belongs on the Normalizers tab and this boefjes tasks page should only show the yielded raw files? 
   - The hashes of the files seem sort of out of nowhere and can throw off non-technical users. So maybe each new raw file should start with something like: "Raw file: X", to make more clear what is going on. 
   - The secure hash seems to be always signed by null. Is this a bug? 
image



2. When expanding a row and then collapsing it again the interface seems to move a little.


3. The tasks list for boefjes shows often show a scroll bar instead of a code block. A code block makes it more user friendly. HTML does seem to get somewhat parsed (not wrapped though), but JSON is just a one-liner without any formatting. DNS records data seems to be parsed as a code block as well (with the scroll bar). 

image


4. The normalizer page seems to have some testing code 

image

underdarknl:
> ### What doesn't work:
> 
>     * Tried to test a binary file upload, however none of the binary files tried would be parsed by OpenKAT, so still looking into this.
Any image url should return a binary file.
> 
>     * See 4 points below.
> 
> 
> ### Bug or feature?:
> 
>     1. Displaying of data could be more user friendly to enhance readability.
>        
>        * There are two headers (yielded objects and yielded raw files). I believe Yielded objects only belongs on the Normalizers tab and this boefjes tasks page should only show the yielded raw files?
Fixed, Now only shows relevant header

>        * The hashes of the files seem sort of out of nowhere and can throw off non-technical users. So maybe each new raw file should start with something like: "Raw file: X", to make more clear what is going on.
Fixed, Its a Code block now its also no londer above the content.his

>        * The secure hash seems to be always signed by null. Is this a bug?
No, it depends on the signing provider. Its no longer shown if its null, its now also a link to the signing provider if available.

>     2. When expanding a row and then collapsing it again the interface seems to move a little.
Not noticed this, but I did fix any eronious unneeded reloading if the expando row had already been filled.

>     3. The tasks list for boefjes shows often show a scroll bar instead of a code block. A code block makes it more user friendly. HTML does seem to get somewhat parsed (not wrapped though), but JSON is just a one-liner without any formatting. DNS records data seems to be parsed as a code block as well (with the scroll bar).
I've changed the width of the code block to be less wide, it was still wider than the expando row before that.

>     4. The normalizer page seems to have some testing code

Now fixed, it was a left over part of the html.

I'm in the process of adding a hex set of tabs that show the content as plaintext, prettyprinted json and binary/hex. The first two are now implemented. The latter requires a bit more JS. Currently all views are visible continously, those should however be hidden by the toggle/tab feature above it. This tab feature relies on the mentioned PR.

underdarknl:
Next to plain text view also with Json view, (if json parseable)
image
and with HEX view if not json parseable.
image

PR #4211, opened 2025-03-25
Feat/cleanup add ooi form

### LINKED ISSUE #625: Improve URL form in onboarding flow
The onboarding flow currently shows an empty web_urls dropdown which can be confusing to new users. We should remove this dropdown from the onboarding flow by rendering this form differently.

underdarknl:
### Changes

This removes empty optional choice fields from the add_ooi form,
It also select the only valid option for required fields, if there is only one option.
It also removes the unneeded owner and hides the type field.

### Issue link

Closes #625

### Demo

image

### QA notes

See if we can still add / create ooi's of various types.
See if the owner/user is still saved.
See if the do in fact hide/show optional related OOI selectors based on the database contents (eg, no web-urls available, vs web-urls available).
See if the choice to select a required option is still available when two choices are present (eg, two networks exists and show up when adding a URL)
---

### Code Checklist

---

---

## Checklist for QA:

### DIFF
--- a/rocky/tools/forms/ooi_form.py
+++ b/rocky/tools/forms/ooi_form.py
@@ -39,24 +39,18 @@ class OOIForm(BaseRockyForm):
 
     def generate_form_fields(self, hidden_ooi_fields: dict[str, str] | None = None) -> dict[str, forms.fields.Field]:
         fields: dict[str, forms.fields.Field] = {}
+
         for name, field in self.ooi_class.model_fields.items():
+            if name in ("user_id", "user", "primary_key", "scan_profile"):
+                continue
+
             annotation = field.annotation
             default_attrs = default_field_options(name, field)
             # if annotation is an Union, get the first non-optional type
             optional_type = get_args(annotation)[0] if get_origin(annotation) == Union else None
 
-            if name == "primary_key":
-                continue
-
-            if name == "user_id":
-                continue
-
             # skip literals
             if hasattr(annotation, "__origin__") and annotation.__origin__ == Literal:
-                continue
-
-            # skip scan_profile
-            if name == "scan_profile":
                 continue
 
             if hidden_ooi_fields and name in hidden_ooi_fields:
@@ -66,6 +60,8 @@ class OOIForm(BaseRockyForm):
                 fields[name] = generate_select_ooi_field(
                     self.api_connector, name, field, get_relations(self.ooi_class)[name], self.initial.get(name, None)
                 )
+                if not fields[name]:
+                    continue
             elif annotation in [IPv4Address, IPv6Address]:
                 fields[name] = generate_ip_field(field)
             elif annotation == AnyUrl:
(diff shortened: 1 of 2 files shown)


underdarknl:
Url fields did also not show a label. Now also Fixed.

rocky/tools/forms/ooi_form.py
```diff
@@ -109,25 +106,35 @@ def generate_select_ooi_field(
...
 
-    option_text = "-- " + _("Optionally choose a {option_label}").format(option_label=option_label) + " --"
+    manytext = "one or more:" if is_multiselect else "a"
     if field.is_required():
-        option_text = "-- " + _("Please choose a {option_label}").format(option_label=option_label) + " --"
+        option_text = _("Please choose {manytext} {option_label}").format(option_label=option_label, manytext=manytext)
+    else:
+        option_text = _("Optionally choose {manytext} {option_label}").format(
+            option_label=option_label, manytext=manytext
+        )
```
ammar92, line 121: I think using `django.utils.translation.ngettext` would be better. I'm not sure if in the current implementation the `many_text` variable will be translated correctly

rocky/tools/forms/ooi_form.py
```diff
@@ -109,25 +106,35 @@ def generate_select_ooi_field(
...
         select_options.append((initial, initial))
 
     oois = api_connector.list_objects({related_ooi_type}, datetime.now(timezone.utc)).items
     select_options.extend([(ooi.primary_key, ooi.primary_key) for ooi in oois])
 
+    # Remove the selection option when only 1 OOI is available, and required
+    if field.is_required() and not initial and len(oois) == 1:
+        del select_options[0]
+    # Don't show select fields without options if they are not required
+    elif not field.is_required() and not len(oois):
```
ammar92, line 135:
```suggestion
    elif not field.is_required() and not oois:
```

While looking at this I was also wondering:
- It's a bit odd that this is the only generator function that could yield a `None` instead of a field, usually you want to decide wether or not te render this field *outside* of this function
- Do we really need to skip rendering this field when there's nothing to choose? Imo displaying it with e.g. "no options available" or making it unselectable is a more consistent approach

rocky/tools/forms/ooi_form.py
```diff
@@ -39,37 +39,34 @@ def get_fields(self) -> dict[str, forms.fields.Field]:
...
-
             if hidden_ooi_fields and name in hidden_ooi_fields:
                 # Hidden ooi fields will have the value of an OOI ID
                 fields[name] = forms.CharField(widget=forms.HiddenInput())
             elif name in get_relations(self.ooi_class):
-                fields[name] = generate_select_ooi_field(
+                select_field = generate_select_ooi_field(
                     self.api_connector, name, field, get_relations(self.ooi_class)[name], self.initial.get(name, None)
                 )
+                if not select_field:
```
ammar92, line 63:
```suggestion
                if select_field is None:
```

But see point below

rocky/tools/forms/ooi_form.py
```diff
@@ -39,37 +39,34 @@ def get_fields(self) -> dict[str, forms.fields.Field]:
 
     def generate_form_fields(self, hidden_ooi_fields: dict[str, str] | None = None) -> dict[str, forms.fields.Field]:
         fields: dict[str, forms.fields.Field] = {}
+
         for name, field in self.ooi_class.model_fields.items():
+            if name in ("user_id", "primary_key", "scan_profile"):
+                continue
```
ammar92, line 45: Tidy 👍

ammar92 (commented):
Nice cleanup. Just had a question regarding the URL field generation. Looks good otherwise

rocky/tools/forms/ooi_form.py
underdarknl: Since the user cannot select a value, and the field is optional my gut feeling says its a wrong move to ask the user to make a choice. Doing the detection of an empty field outside this function in in the loop that handles all the fields would add the need for another communication channel in which we tell the outer function about the selection choice length.

underdarknl: @HeleenSG @RomijnHumanoids what do you guys think?

RomijnHumanoids: What are the specific fields that are hidden or shown in pre-filled select with one option?

It depends on the context I would say, what users expect of the content. 
- If a network field is not expected in the form of a certain ooi type and the subject field only adds confusion to the form; it would beter to leave it out after that ooi type is selected. 
- If users expect the field to be there, maybe even expect to be able to make a selection and for some reason that is not possible in this form; it would be better to show the pre-filled single option field, or read-only state. If read only, possibly explain why different value is not possible.

I find it hard to judge the context of this specific content, maybe @stephanie0x00 could help interpret what’s expected in which situation?

underdarknl: Fields that are not required, and have no selectable options will be hidden with this change. It depends on the ObjectType, and the current database contents. When the database has more selectable types, it will more often show these optional fields, whereas for a beginner/novice user their database is till empty (probably) and the questions they cannot answer (on that page) are no longer asked.

hasecon:
**CI triage — no current signal available.**

Both red checks (`pre-commit`, `makelang`) are from a run old enough that GitHub has expired the logs, so there is no way to see what actually failed:

```
{"message":"Server Error", ... "status":"410"}
```

Nothing can be concluded from the current red marks. Pushing to the branch (or re-running the workflows) would produce a fresh, readable result — worth doing before anyone spends review time here, since the failure may well have been fixed on `main` in the meantime.

PR #4073, opened 2025-02-07
Remove block-all-mixed-content

underdarknl:
This CSP directive is no longer needed, and deprecated. Its also throwing warnings in modern browsers because of its deprecated status.

### Changes

Removal of setting from rocky config.

### Issue link

image

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/block-all-mixed-content

### QA notes

No real changes expected.  
The warning from the screenshot above should not be present in the browsers console / error log.

---

### Code Checklist

- [ ] All the commits in this PR are properly PGP-signed and verified.
- [ ] This PR only contains functionality relevant to the issue.
- [ ] I have written unit tests for the changes or fixes I made.
- [ ] I have checked the documentation and made changes where necessary.
- [ ] I have performed a self-review of my code and refactored it to the best of my abilities.

- [ ] Tickets have been created for newly discovered issues.
- [ ] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [ ] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [ ] I have included comments in the code to elaborate on what is not self-evident from the code itself, including references to issues and discussions online, or implicit behavior of an interface.

---

## Checklist for code reviewers:

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_qa.md) into your comment.

PR #4330, opened 2025-04-14
Update check_csp_header.py

### LINKED ISSUE #4329: Fix possible Ddoses in regexes
**Describe the bug**
We have two regexes where the complexity of parsing is dependent on the input string and possibly leads to longer runtimes than we'd like.

https://sonarcloud.io/project/security_hotspots?id=minvws_nl-kat-coordination&hotspots=AZJXrSBQdScQYYDmeU0A

https://sonarcloud.io/project/security_hotspots?id=minvws_nl-kat-coordination&hotspots=AZJXrSBQdScQYYDmeU0B

**Expected behavior**
Clea regexes that do not have the possible high polynomial runtime

**OpenKAT version**
Main

underdarknl:
Fix redos issues

### Changes

This should fix the possible redos issues on CSP headers as suggested by our friends over at sonarcloud.io

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/4329

### QA notes

* CSP headers with various domains should still work.
* Wildcard domains should still raise a finding. (eg, *.mispo.es in any of the allowed sources)
* Malicious CSP headers should no longer take long to parse/validate:
* *  `script-src aaaaaaaaaaaaaaaaaaaaaaaaaaaaa*.......`

### Code Checklist

---

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/about-openkat/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/about-openkat/templates/pull_request_template_review_qa.md) into your comment.


octopoes/bits/check_csp_header/check_csp_header.py
```diff
@@ -43,12 +43,11 @@ def run(resource: HTTPResource, additional_oois: list[HTTPHeader], config: dict[
...
 
     # checks for a wildcard in domains in the header
-    # 1: one or more non-whitespace
-    # 2: wildcard
-    # 3: second-level domain
-    # 4: end with either a space, a ';', a :port or the end of the string
-    #              {1}{ 2}{  3  }{         4       }
-    if re.search(r"\S+\*\.\S{2,3}([\s]+|$|;|:\d+)", csp_header):
+    # 1 \*\. - literal wildcard and dot (*.)
+    # 2 [a-zA-Z0-9.-]+ - domain parts (e.g., example, cdn.example)
```
BramVWS, line 47: Deze kan denk ik vervangen worden door \w+

octopoes/bits/check_csp_header/check_csp_header.py
```diff
@@ -43,12 +43,11 @@ def run(resource: HTTPResource, additional_oois: list[HTTPHeader], config: dict[
...
+    # 3 \.[a-zA-Z]{2,} - final TLD, like .com, .org, .co.uk (optional to tweak further)
```
BramVWS, line 48: Is (\.[a-z]+)+ niet simpeler met zelfde effect?

octopoes/bits/check_csp_header/check_csp_header.py
```diff
@@ -151,8 +150,12 @@ def _create_kat_finding(header: Reference, kat_id: str, description: str) -> Ite
 
 def _source_valid(policy: list[str]) -> bool:
     for value in policy:
+        #1 (?:\*\.|\.)? - optional wildcard *. or dot prefix.
+        #2 [a-zA-Z0-9-]+ - (sub)domain
```
BramVWS, line 154: Deze kan denk ik vervangen worden door \w+

octopoes/bits/check_csp_header/check_csp_header.py
```diff
@@ -151,8 +150,12 @@ def _create_kat_finding(header: Reference, kat_id: str, description: str) -> Ite
...
+        #3 (?:\.[a-zA-Z0-9-]+)+ - Optionally one or more hostnames (e.g., .nl, .co.uk)
```
BramVWS, line 155: Is (.[a-z]+)+ niet simpeler met zelfde effect? In principe zijn deze nooit hoofdletter of bevatten cijfers.

octopoes/bits/check_csp_header/check_csp_header.py
ammar92: Werkt waarschijnlijk ook niet, want `\w` staat gelijk aan `[a-zA-Z0-9_]`. Voor zover ik weet kan je geen underscores gebruiken in (sub)domains?

ammar92 (commented):
Looks good to me. Just a few remarks on the suggested patterns. But I was also wondering, since we already use `tldextract` package in Octopoes, could that help with extracting and verifying (sub)domains instead of relying solely on pattern matching?

octopoes/bits/check_csp_header/check_csp_header.py
underdarknl: Correct, Underscores in hostnames mogen niet, wel in 'records', maar dat zijn geen resolving hosts in de context van CSP.
Wel heb ik alles lowercase gemaakt. zowel de hostnames als de directives zijn lowercase in CSP.

underdarknl:
Todo:

These regexes do not catch 
default-src *
default-src https://*

hasecon (changes requested):
## QA Review — PR #4330

### Test setup
- Branch `fix/redos-csp` checked out locally
- Stack running from previous review (`make clean && make kat`)
- Tests run inside Docker container: `docker exec openkat-reviev-octopoes_api-1 python -m pytest tests/test_bit_csp_header.py -v`

### Code review

The PR replaces vulnerable ReDoS regex patterns in the CSP header checker with safer alternatives. The approach is correct: lowercase the header first, then use more specific character classes instead of `\S+` which caused catastrophic backtracking.

**Changes reviewed:**

| Change | Assessment |
|--------|------------|
| `.lower()` on CSP header (line 35) | Correct — allows simpler lowercase-only patterns |
| Wildcard domain regex: `\*\.[a-z0-9._-]+\.[a-z]{2,}` | Fixed ReDoS, but too restrictive (see below) |
| Source validation regex: `(?:\*\.\|\.)?[a-z0-9_-]+(?:\.[a-z0-9_-]+)+(?::\d+)?` | Fixed ReDoS, looks correct |
| Comment #4 on `_source_valid` says "Optional port number" but the regex is for delimiter | Minor: comment doesn't match regex |
| Test updates (new test cases for `default-src *` and `default-src https://*.com`) | Good additions, but tests currently fail |

### Test results

| Test | Status | Details |
|------|--------|---------|
| `test_check_csp_headers` | **FAIL** | Assertion fails at `results[4]` — see below |
| `test_check_csp_headers_non_xss_capable` | PASS | |
| Stack build | PASS | All 13 containers running |
| Log analysis | PASS | No new errors |

### Outstanding TODO (from [author's comment](https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/pull/4330#issuecomment-2805125321))

The wildcard regex on line 50 (`\*\.[a-z0-9._-]+\.[a-z]{2,}`) does not catch:
- `default-src *` — bare wildcard as source value
- `default-src https://*` — scheme-qualified bare wildcard
- `default-src https://*.com` — wildcard with single domain part (no subdomain separator)

The test has already been updated to expect these to be caught (test cases at indices 4 and 5), but the regex hasn't been updated to match them. The test fails at `results[4]`:

```
Expected: "1. The wildcard * for the scheme and host part of any URL should never be used in CSP settings."
Got:      "1. frame-src has not been defined or does not have a fallback."
```

The regex needs to be extended to also match bare `*` and `https://*` patterns (e.g. with optional scheme prefix and optional domain parts after `*`).

### Also noted

The github-advanced-security bot previously flagged the character class `[a-z0-9.-_]` as having an overly permissive range (the `-` between `.` and `_` creates an unintended ASCII range). This appears to have been fixed in the latest commit by reordering to `[a-z0-9._-]` (hyphen at end), which is correct.

### Verdict

The ReDoS fix approach is solid, but the test is currently failing because the TODO hasn't been addressed yet. Requesting changes until the wildcard regex handles bare `*` and `https://*` patterns and the test passes.

hasecon (changes requested):
## QA Review — round 2

### Setup
- Branch: `fix/redos-csp` @ `ed32189` (main merged in today, 2026-08-28)
- Stack: running dev stack, `octopoes_api` / `octopoes_api_worker` / `octopoes_api_scanprofiles` restarted on the branch code (mounted source) — all up, `/health` 200, Rocky 200, no new errors in the Octopoes logs after the RabbitMQ reconnect settled
- Tests: `docker exec openkat-reviev-octopoes_api-1 python -m pytest` (same invocation as the `Octopoes tests` workflow)

### Is this PR still relevant?
Yes. `check_csp_header.py` on `main` has not been touched since #4073 (Feb 2025), the CSP-parser rewrite discussed in April 2025 never landed, and #4329 is still open. The branch diff against current `main` is only the two files in this PR, so nothing else needs rebasing. The two regexes it replaces are still the ones on `main`.

### Previous round (2026-03-19) — status

| Finding | Status |
|---|---|
| Wildcard regex did not catch `default-src *`, `https://*`, `https://*.com` (Jan's TODO) | **Fixed** in `64d058d` — all three now match (verified on the head) |
| `test_check_csp_headers` failing at `results[4]` | **Still failing** — different cause now, see F1 |
| Comment #4 in `_source_valid` says "Optional port number" but describes the delimiter | Still open (Low) |

### CI/CD Status
- ❌ **`Octopoes tests` → `test (3.10/3.11/3.12/3.13)`** — "Run unit tests" fails on all four. Reproduced in Docker: `tests/test_bit_csp_header.py::test_check_csp_headers`. Same test passes on `main` → caused by this PR. The `tests/test_api.py` errors you also see locally are identical on `main` (in-container XTDB URL) and are not related.
- ❌ **CodeQL** — 1 new *high* alert on line 51: "Inefficient regular expression … exponential backtracking on strings starting with `*.` and containing many repetitions of `-.`". This is a real finding, see F2.
- ✅ pre-commit, container images, debian builds, Rocky/Bytes/Mula tests.

Both red checks are Octopoes-relevant → blocking.

### Code Review Findings

| # | Sev | File | Finding |
|---|---|---|---|
| F1 | **High** | `octopoes/tests/test_bit_csp_header.py:73-108` | The expected strings for `results[4]`, `[5]` and `[6]` were not updated when the inputs got their `default-src ` prefix (`064ae83`). The **code is right, the test is wrong**: `results[4]` still expects "frame-src / script-src / default-src has not been defined", which cannot appear once `default-src` is present; `results[5]` (`default-src *`) is missing the 5th line `A wildcard source should not be used…` that line 114 emits for `policy[1] == "*"`; `results[6]` still expects the pre-prefix output. Actual outputs on the head are below under "Suggested test fix" — I verified they pass against this branch. |
| F2 | **High** | `octopoes/bits/check_csp_header/check_csp_header.py:51` | The new wildcard regex is **exponentially** backtracking, which is worse than the polynomial pattern it replaces. Cause: the `.` inside the class of `(?:\.[a-z0-9._-]+)*` makes the repetition ambiguous (a dot can be consumed by the class *or* start the next group), so a non-matching tail forces every partition to be tried. Measured in CPython: `default-src *` + `.-`×22 → 0.2 s, ×26 → 3.2 s; `*.` + `a.`×28 (58 chars) → **18.5 s**; with ×40 the bit does not return within 30 s. For comparison the *old* regex needs 30 000 chars to reach 1.2 s. Since the header comes from the scanned site, a hostile server can stall the Octopoes worker with a 60-byte CSP header. — Full disclosure: this exact pattern came from my own suggestion in March; sorry for that, I only timed the inputs from the TODO and not the adversarial case. |
| F3 | Low | `check_csp_header.py:154-157` | Comment block for `_source_valid` doesn't match the regex: #3 says "Optionally one or more hostnames … including the invalid `_`" (the `+` is mandatory, and `_` being invalid argues for dropping it from the class), and #4 labels the trailing delimiter as "Optional port number" while the port `(?::\d+)?` is unlabeled. |

`_source_valid`'s regex (line 159) is fine: `.` is not in its character class, so the repetition is unambiguous — 0 ms on the same adversarial inputs.

#### Suggested fix for F2 (verified)
Drop `.` (and `_`, invalid in hostnames as you noted in April) from the class so each `\.` can only be consumed by the group boundary:

```python
if re.search(r"(?:^|\s|;)(?:https?://)?\*(?:\.[a-z0-9-]+)*(?:\s+|$|;|:\d+)", csp_header):
```

Checked against 11 inputs (`default-src *`, `https://*`, `https://*.com`, `*.example.com`, `https://*.example.com:8443; x`, `*.mispo.es`, `*.a-b.example`, `'self' *.example.com 'unsafe-inline'` → match; `https://cdn.example.com`, `a*.example.com` → no match): identical results to the current regex, and linear: `a.`×20 000 (40 kB) → 1.7 ms, 30 000 `a`'s + `*` → 0.5 ms. It gives the same finding lists for the three test inputs in F1, so the corrected assertions below stay valid after this change.

#### Suggested test fix for F1 (verified on the head)
```python
    assert results[4][0].id == "KAT-CSP-VULNERABILITIES"
    assert (
        results[4][1].description
        == """List of CSP findings:
 1. The wildcard * for the scheme and host part of any URL should never be used in CSP settings.
 2. base-uri has not been defined, default-src does not apply.
 3. frame-ancestors has not been defined.
 4. default-src has not been correctly defined."""
    )

    assert results[5][0].id == "KAT-CSP-VULNERABILITIES"
    assert (
        results[5][1].description
        == """List of CSP findings:
 1. The wildcard * for the scheme and host part of any URL should never be used in CSP settings.
 2. base-uri has not been defined, default-src does not apply.
 3. frame-ancestors has not been defined.
 4. default-src has not been correctly defined.
 5. A wildcard source should not be used in the value of any type in the CSP settings."""
    )

    assert results[6][0].id == "KAT-CSP-VULNERABILITIES"
    assert (
        results[6][1].description
        == """List of CSP findings:
 1. base-uri has not been defined, default-src does not apply.
 2. frame-ancestors has not been defined.
 3. default-src has not been correctly defined.
 4. CSP setting has no value.
 5. Private, local, reserved, multicast, loopback ips should not be allowed in the CSP settings.
 6. CSP setting has no value."""
    )
```

### Suggested regression test
The PR's purpose is "no ReDoS", but nothing pins that. This guard fails on the current head (killed after 30 s) and passes with the F2 regex (~1 ms):

```python
def test_check_csp_headers_wildcard_regex_is_linear(http_resource_https):
    # Hostile header from a scanned site must not stall the worker (#4329)
    header = "default-src *" + ".-" * 40 + "!"
    start = time.perf_counter()
    list(
        run(
            resource=http_resource_https,
            additional_oois=[
                HTTPHeader(resource=http_resource_https.reference, key="Content-Type", value="text/html"),
                HTTPHeader(resource=http_resource_https.reference, key="content-security-policy", value=header),
            ],
            config={},
        )
    )
    assert time.perf_counter() - start < 1.0
```

### Test Results
| Component | Test | Result |
|---|---|---|
| octopoes | `tests/test_bit_csp_header.py` (Docker) | **1 failed, 1 passed** — `test_check_csp_headers` (F1) |
| octopoes | full unit suite (Docker) | 6 failed / 6 errors in `test_api.py` — identical set on `main`, pre-existing env issue; the CSP test is the only delta |
| octopoes | same test on `main` | 2 passed |

### Functional Test
| Check | Status | Details |
|---|---|---|
| Octopoes services on branch code | ✅ | api / worker / scanprofiles restarted, all up |
| Rocky / Octopoes reachable | ✅ | 200 / 200 |
| Log analysis | ✅ | only the RabbitMQ reconnect during restart; 0 errors after settling |
| Regex timing (old vs new vs proposed) | ❌ | old: polynomial (1.2 s @ 30 kB); new: exponential (18 s @ 58 B); proposed: linear (2 ms @ 40 kB) |

Full `make clean && make kat` was skipped this round because the verdict is already fixed by two deterministic blockers; I'll do the full rebuild + a live CSP scan on the next round.

### Verdict
**CHANGES REQUESTED**

Two blockers, both small: (1) the three stale expected strings in the test, and (2) `[a-z0-9._-]` → `[a-z0-9-]` on line 51 so the wildcard regex is linear and the CodeQL high alert goes away. With those two the PR does what #4329 asks and is ready to go.

underdarknl:
This PR has been sitting for a while now, and I'm thinking the following next steps would make more sense.
Create a BIT that parses the headers into CSP objects. Any parsing issues can be flagged as findings against the supplying header.
Create a BIT that consumes these CSP objects and an adjacent Config object telling us exactly what options are required, forbidden and possibly what their values can, or cannot include. 
This allows us to separate parsing logic from Policy / compliance logic.

hasecon:
Superseded by #5351, which implements the two-bit rearchitect from your comment above: a `parse-csp` bit turns the header into structured `CSPDirective`/`CSPSource` OOIs, and a config-driven `check-csp-policy` bit does the compliance checks on that structure. It **removes `check_csp_header` entirely**, so the ReDoS this PR patches is gone rather than fixed in place — meaning there's nothing left here to merge once #5351 lands.

Suggest closing this PR in favour of #5351 (which is set to close #4329). Left open for now so you can compare.

brennodewinter:
## This PR is superseded by #5351

PR #5351 ("Rearchitect CSP checking into a parser bit and a policy bit") has been approved and completely replaces the approach taken here:

- **#5351** removes the entire `check_csp_header` bit (the file this PR patches) and replaces it with two new bits: `parse-csp` (deterministic string-based parser, no regex) and `check-csp-policy` (set/string operations on parsed OOIs).
- The ReDoS vulnerability this PR fixes (`\S+\*\.\S{2,3}(...)`) is eliminated entirely in #5351 — no regex remains in either bit.
- #5351 includes a timing test proving the adversarial input that stalled the worker ~18s now completes in <0.1s.
- All checks the old bit performed are preserved, plus new ones (duplicate directive detection, stray-asterisk false-positive fix).

**Recommendation: close this PR in favour of #5351.** Once #5351 is merged, the `check_csp_header.py` file no longer exists, and these regex patches have no target. Issue #4329 will be resolved by #5351.

brennodewinter:
Closing as superseded by #5351, which has been merged. The entire  bit is replaced by  +  (no regex, no ReDoS). Issue #4329 is resolved.

Issue #4334, opened 2025-04-15
Katalogus exposes settings / environment hash of boefje in api


jpbruinsslot:
And this should ideally be returned for all organisations to avoid having to loop over katalogus calls.

PR #4436, opened 2025-05-07
Add configs endpoint with (optional) duplicate configs across organisations

Donnype:
### Changes

This adds a `/configs` endpoint for the boefje configs (writing this, perhaps this needs a rename considering there are also normalizer configs!). The filters included for now are the `organisation_id`, `boefje_id` and `enabled`.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/4334

### Demo

See the integration tests.

### QA notes

This feature is not in direct use, but needed for the full deduplication feature. I think this could best be QA'd with https://github.com/minvws/nl-kat-coordination/pull/4379 perhaps, especially as @jpbruinsslot might need some tweaks in this implementation.

---

### Code Checklist

---

---

## Checklist for QA:


boefjes/boefjes/models.py
```diff
@@ -75,6 +77,24 @@ class Config:
...
+    boefje_id: str
+    organisation_id: str
+
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def env_hash(self) -> str:
+        sorted_settings = dict(
+            sorted(self.settings.items())
+        )  # settings are assumed to be a one level deep mapping of strings to strings
+        settings_bytes = json.dumps(sorted_settings).encode()
```
underdarknl, line 93: why not use the sort_keys argument on dumps?
Also, if the result is a set of bytes that are deterministically sorted, just hash those without encoding them, and as a result we are not limited ourselves to any depth or types.

boefjes/boefjes/models.py
```diff
@@ -75,6 +77,24 @@ class Config:
...
 
 
+class BoefjeConfig(BaseModel):
+    id: int
+    settings: dict
+    enabled: bool
+    boefje_id: str
+    organisation_id: str
+
+    @computed_field  # type: ignore[prop-decorator]
```
underdarknl, line 87: I'm not a fan of doing this on every read, when we could (and I think should) do so on write. Storing the hash in a database column also gives us the opportunity to query for other organizations that have the same hash by using an fast and indexed column lookup.

boefjes/boefjes/models.py
Donnype: Ah yes I glimpsed over the `sort_keys` option! The `encode` makes sure we have bytes, which is I believe most hash implementations expect, including those in hashlib.

underdarknl: Ah true, encode can stay :p
For sort_keys, this means we are still not sorting lists present in the Json, which would mean we should move to a more deterministic string-builder.

Donnype: Yes so the assumption is that the settings are a "`dict[str, str]`", which is a safe assumption perhaps since we inject these values in the runtime environment, turning it roughly into a mapping from strings to strings anyways.

boefjes/boefjes/models.py
Donnype: This does raise some extra security questions, especially since the settings are encrypted at rest. Do we want to hash the database value, meaning the encrypted settings if encryption has been turned on? Note that with the boefje_id and enabled filters, the result set would at most be the number of organizations, where doing this on write means maintaining a computed field that should be kept in sync with the original column, which is not possible to do within the database since we use application-level encryption.

boefjes/boefjes/models.py
```diff
@@ -1,12 +1,14 @@
 import datetime
+import json
 from enum import Enum
 from functools import total_ordering
+from hashlib import sha1
```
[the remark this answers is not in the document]
Donnype: Well if someone has the hashes from the KATalogus in the current implementation, they would also have the keys themselves since they are not stored

boefjes/boefjes/katalogus/configs.py
```diff
@@ -0,0 +1,27 @@
...
+router = APIRouter(tags=["configs"])
+
+logger = structlog.get_logger(__name__)
+
+
+@router.get("/configs", response_model=list[BoefjeConfig])
+def list_configs(
+    organisation_id: str | None = Query(None),
+    boefje_id: str | None = Query(None),
+    enabled: bool | None = Query(None),
```
ammar92, line 18: You can leave out `Query` as FastAPI handles function arguments as query parameters by default (also to be consistent with the rest of the codebase)
```suggestion
    organisation_id: str | None = None,
    boefje_id: str | None = None,
    enabled: bool | None = None,
```

(`Query` can be handy for e.g. documentation or validation, but even then, with modern versions of FastAPI and Pydantic validation can also be done with (annotated) types)

ammar92 (commented):
Nice work and tests. I have one style suggestion, looks good otherwise

stephanie0x00:
## Checklist for QA:

### What works:
Doesn't seem to break anything obvious. Created multiple organisations, added a LeakIX API key to see if that was allowed and enabled the boefje. The boefje seems to run as expected. 

### What doesn't work:
Not tested due to current bug: copying all settings from one org to another. Once that bugfix is merged (and this PR too) I'll double check that.

### Bug or feature?:
n/a

underdarknl (approved):

Issue #4280, opened 2025-04-02
Cloning of boefjes config between organisations results in error


paulvandenbraken:
_Please add `bug`, the name of any relevant modules (e.g. `rocky`), and any other relevant labels to your issue._

**Describe the bug**
In KAT-alogus the boefjes configuration  can be cloned from one to an other organisation. However selecting this function and clicking the "clone"-button result in an error message: "Internal server error"

**To Reproduce**
See above.

**Expected behavior**
Working cloning function resulting in an identical boejes setup as the source (template)

**Screenshots**

![Image](https://github.com/user-attachments/assets/bd05922d-ef1c-4698-b798-f2e365585ab2)

![Image](https://github.com/user-attachments/assets/205e8056-f2a6-45fd-ae2a-24724c498bd1)

**OpenKAT version**
v1.18

**Desktop (please complete the following information if relevant):**

- OS: na
- Browser: na
- Version: na

**Additional context**

underdarknl:
@paulvandenbraken is there a tacktrace available from rocky, and or the Katalogus?

underdarknl:
The key line is: KATalogusNotAllowedError("User is not allowed to set plugin settings")
We should let the user know about this, or not not even show the button.
Finer details: this probably works ok if the user can enable boefjes, and there are no boefjes settings to be configured. We should check for both permissions.

underdarknl:
Hm, weird. Normalizers indeed do not have settings, so not sure why or how we are trying to duplicate settings from one normalizer to another.

PR #4457, opened 2025-05-13
Fix exceptions and permission checks in katalogus client

dekkers:
### Changes

Fix for checking permissions on clone katalogus settings. It would always raise a `KATalogusNotAllowedError` for superusers because of a logic error. It also did not check whether the user has permissions to set katalogus settings in the destination organization. I also removed unnecessary checks if the user is superuser. Superusers always have all permissions, so it is not necessary to check if user has permissions or is superuser, `has_perm` will always return `True` for superusers.

Fix for `KATalogusNotAllowedError` calling `_()` on a dynamic string. This doesn't work because those strings aren't made available for translation and results in that the exception itself fails with
```
katalogus.client.KATalogusNotAllowedError: <exception str() failed>
```

### QA notes

- Check if you can clone settings as superuser
- Check if you get a proper error message if you don't have permissions to clone settings

---

---

## Checklist for QA:

### DIFF
--- a/rocky/katalogus/client.py
+++ b/rocky/katalogus/client.py
@@ -152,7 +152,7 @@ class DuplicateIdError(KATalogusError):
 
 class KATalogusNotAllowedError(KATalogusError):
     def __init__(self, error_message: str):
-        super().__init__(_(error_message))
+        super().__init__(error_message)
 
 
 def verify_response(response: Response) -> None:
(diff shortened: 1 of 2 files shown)


ammar92 (dismissed):
Nice work

Donnype (dismissed):
That logic was indeed not very logical. 😁 

Perhaps tweaking `test_katalogus_clone_settings` to ensure this behavior is a quick win as well?

underdarknl:
Could this be the same as the following bug reported?
https://github.com/minvws/nl-kat-coordination/issues/4280

underdarknl:
Looks good to me. lets qa

PR #4466, opened 2025-05-17
Fix findings history in findings report

dekkers:
### Changes

The findings report called get_history with the reference of the input OOI instead of the finding reference.

The vulnerability report used finding.reference and finding.primary_key for the history cache. While this is not a problem because they compare equal, I've changed it to finding.reference so the code is more clear.

The vulnerability report would use the first_seen of a previous finding if there was no history. Although I also don't understand how this could happen, because as far as I understand it an object can't exist without a transaction that creates it.

I also removed the unnecessary creation of the time_history object which is a minor optimization.

### QA notes

Check the first seen of the findings in the vulnerability and findings reports.

---

---

## Checklist for QA:

### DIFF
--- a/rocky/reports/report_types/vulnerability_report/report.py
+++ b/rocky/reports/report_types/vulnerability_report/report.py
@@ -49,7 +49,6 @@ class VulnerabilityReport(Report):
                 total_findings = sum(occurrences.values())
                 terms = list(occurrences)
                 recommendations = []
-                first_seen = ""
 
                 # remove duplicate findings and finding types
                 finding_types = list(set(findings_data["finding_types"]))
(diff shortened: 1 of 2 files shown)


rocky/reports/report_types/vulnerability_report/report.py
```diff
@@ -62,15 +61,15 @@ def collect_data(self, input_oois: Iterable[Reference], valid_time: datetime) ->
...
                             history_cache[finding.reference] = self.octopoes_api_connector.get_history(
                                 finding.reference
                             )
 
-                        time_history = [transaction.valid_time for transaction in history_cache[finding.reference]]
-
-                        if time_history:
-                            first_seen = str(time_history[0])
+                        if history_cache[finding.reference]:
+                            first_seen = str(history_cache[finding.reference][0].valid_time)
```
ammar92, line 70: While the string representation of the `valid_time` attribute yields the same result, using the `isoformat()` method is more explicit (and imo appropriate) if the attribute is a `datetime` object. I'll leave it up to you

ammar92 (approved):
Nice fix. Left one tiny remark/ suggestion, but consider it approved

stephanie0x00:
## Checklist for QA:

### What works:
Seems to work and not break things.

### What doesn't work:
n/a

### Bug or feature?:
n/a

PR #4560, opened 2025-06-12
Fix robot tests

Rieven:
### Changes
Changes of some HTML attributes and the targetting of these attributes in robot tests.
Failing tests :https://github.com/minvws/nl-kat-coordination/actions/runs/15579282571/job/43956500190


rocky/account/forms/account_setup.py
```diff
@@ -147,6 +147,7 @@ class Meta:
         widgets = {
             "name": forms.TextInput(
                 attrs={
+                    "id": "id_name",
```
underdarknl: I find it more descriptive to create form id's in the following format:
#formaction_fieldname

So for this I'd expect an ID for the input with the name="name" to be "create_organizationform_name" or "form_create_organization_name" 
That wat multiple forms on the same page can have a name field, without us getting into trouble with duplicate ID's.  Duplicate Id's are not allowed, cannot be used as anchors in url, and cannot be correctly targetted by styling and selectors in stuff like robot framework.

rocky/account/forms/account_setup.py
ammar92, line 1: Just out of curiosity; the form elements rendered here should already have a name automatically generated (since forms rely on name values to submit data). Is there a specific reason we also need to manually assign IDs? I'd prefer if possible to rely on the existing name values instead of adding both name and id manually

underdarknl: A few reasons exists for Id's
Names are unique in the scope of the form itself (or rather, names are used when constructing the post request). Id's on the other hand provide a way to uniquely target elements on a page via CSS selectors, fragment-urls, and are required for label elements in their "for" attribute. 

While some platforms generate id's based on the context of the form and the name field (for sake of the required <label for> association. I find it better it explicitly name fields (and other elements) so that we know for sure multiple uses of the same form (for example when editing and creating new objects can be done on the same page) use logical and reasonable names.

In the tests, we relied on the number of forms, instead of targeting the actual field, as it had no unique ID that we could easily target. My usual approach is to uniquely name forms, fields, pages (on their body) and stuff like sections, just to allow future use and targeting.

underdarknl: Oddly enough, when developing Javascript or CSS, I prefer to not use Id's, expect for the containing element., (targeting initialization of rich features by Id or class), and instead make sure any dynamically generated Objects Don't use Id's to make sure we *can* have multiple instantiated elements on the same page without the fear of colliding Id's.

dekkers:
Was this fixed in another PR? Because the robot tests success on main now: https://github.com/minvws/nl-kat-coordination/actions/runs/15924003005/job/44917268846

PR #4592, opened 2025-06-24
Handle normalizer configs as well when cloning settings between organizations

Donnype:
### Changes
This fixes the error seen in https://github.com/minvws/nl-kat-coordination/issues/4280. I was able to reproduce the error message by disabling a normalizer first (which is probably the reason the error popped up in the first place, as often all normalizers stay enabled and hence have not config entry).

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/4280

### Demo

See the integration test!

### QA notes

Please verify that the clone-config feature still works, also when a normalizer is disabled in the frontend first for organization A, before cloning everything to organization B. The normalizer should be disabled in organization B as well after the clone.

---

### Code Checklist

---

---

## Checklist for QA:

### DIFF
--- a/boefjes/boefjes/sql/config_storage.py
+++ b/boefjes/boefjes/sql/config_storage.py
@@ -181,7 +181,7 @@ class SQLConfigStorage(SessionMixin, ConfigStorage):
             .join(OrganisationInDB)
             .filter(BoefjeConfigInDB.organisation_pk == OrganisationInDB.pk)
             .filter(OrganisationInDB.id == organisation_id)
-            .filter(BoefjeConfigInDB.enabled == False)
+            .filter(BoefjeConfigInDB.enabled == False)  # noqa: E712
         )
         return [x[0] for x in enabled_boefjes.all()]
 
@@ -193,7 +193,7 @@ class SQLConfigStorage(SessionMixin, ConfigStorage):
             .join(OrganisationInDB)
             .filter(NormalizerConfigInDB.organisation_pk == OrganisationInDB.pk)
             .filter(OrganisationInDB.id == organisation_id)
-            .filter(NormalizerConfigInDB.enabled == False)
+            .filter(NormalizerConfigInDB.enabled == False)  # noqa: E712
         )
 
         return [plugin[0] for plugin in enabled_normalizers.all()]


boefjes/boefjes/dependencies/plugins.py
```diff
@@ -86,17 +86,30 @@ def get_all_settings(self, organisation_id: str, plugin_id: str):
...
 
     def clone_settings_to_organisation(self, from_organisation: str, to_organisation: str):
-        # One requirement is that only boefjes enabled in the from_organisation end up being enabled for the target.
+        # One requirement is that only boefjes enabled in the from_organisation end up being enabled for the target,
+        # and only normalizers disabled in the from_organisation end up being disabled.
         for plugin_id in self.config_storage.get_enabled_boefjes(to_organisation):
             self.set_enabled_by_id(plugin_id, to_organisation, enabled=False)
 
-        for plugin in self.get_all(from_organisation):
+        # One requirement is that only boefjes enabled in the from_organisation end up being enabled for the target.
```
underdarknl:
```suggestion
```

boefjes/boefjes/sql/config_storage.py
```diff
@@ -173,6 +173,31 @@ def get_enabled_normalizers(self, organisation_id: str) -> list[str]:
...
+        enabled_boefjes = (
+            self.session.query(BoefjeInDB.plugin_id)
+            .join(BoefjeConfigInDB)
+            .filter(BoefjeConfigInDB.boefje_id == BoefjeInDB.id)
+            .join(OrganisationInDB)
+            .filter(BoefjeConfigInDB.organisation_pk == OrganisationInDB.pk)
+            .filter(OrganisationInDB.id == organisation_id)
+            .filter(BoefjeConfigInDB.enabled == False)
+        )
+        return [x[0] for x in enabled_boefjes.all()]
```
ammar92, line 186: Since you're not doing anything specifically with the `Row` objects, you can also use `scalars()` to get the elements you need. I think this is also cleaner because it gives you a list of values without tuple unpacking.

```suggestion
        return enabled_boefjes.scalars().all()
```

I've seen this in other places, but it just got my attention now. We can choose to optimize or clean it up in a future PR. I'll leave this up to you.

boefjes/boefjes/storage/memory.py
```diff
@@ -134,6 +134,20 @@ def get_enabled_normalizers(self, organisation_id: str) -> list[str]:
...
+            for plugin_id, enabled in self._enabled.get(organisation_id, {}).items()
+            if not enabled and "norm" not in plugin_id
+        ]
+
+    def get_disabled_normalizers(self, organisation_id: str) -> list[str]:
+        return [
+            plugin_id
+            for plugin_id, enabled in self._enabled.get(organisation_id, {}).items()
+            if not enabled and "norm" in plugin_id
+        ]
```
ammar92, line 149: I think checking for normalizers like this isn't reliable. For example, see the list produced by running:
`find boefjes/boefjes/plugins -name normalizer.json -execdir grep id {} \;`

You would see that most normalizer IDs end with `_normalize`. But there are exceptions like `kat_report_data`, `kat_answer_parser`, `adr-validator-normalize`, etc. With the current approach it's also possible that a boefje ID could contain the string 'norm' even if it's unrelated. I think we need to find a more explicit or consistent way to identify normalizers?

ammar92 (commented):
Looks good to me, just a few remarks/question

boefjes/boefjes/storage/memory.py
Donnype: This class is only used as a test double, so we generally control the data that get's fed into it. Having said that, it definitely isn't the most pretty solution, but perhaps good enough for the current test cases until we need a smarter solution?

boefjes/boefjes/sql/config_storage.py
Donnype: Let's do them all at once.

dekkers (approved):

PR #4629, opened 2025-07-03
Improve performance for muting findings

### LINKED ISSUE #4616: We do not mute findings in bulk, so it is slow
Muting findings from the findings page is slow for multiple findings, while I believe we could use bulk APIs here. Unmuting is fast however!

madelondohmen:
### Changes
Muting findings was a bulk action with a lot of loops. 
This PR creates a "create_oois" method, to bulk the creation of objects.

For example, when there are 4 oois we'd like to create...
We go from 4 times these 5 calls (total of 20 API calls):
- login
- save_boefje_meta
- save_raw
- save_normalizer_meta
- save_declaration

To only 5 calls in total:
- login
- save_boefje_meta
- save_raw
- save_normalizer_meta 
- save_many_declatations

So for 20 Findings, this will result in 5 calls instead of 100 calls.

### Issue link
Closes #4616 

### QA notes
Before checking out to this branch, go to main and:
- Go to Findings page
- Select all findings in the table
- Click on "Mute findings"
- Count the time it takes before the findings are muted
- Unmute the findings again (in the filter, select "Show muted findings", then select all and press the "Unmute findings" button)

Then checkout to this branch and do the same.
Check the difference in time.

---

### Code Checklist

---

---

## Checklist for QA:


Donnype:
> ### Changes
> Muting findings was a bulk action with a lot of loops. This PR creates a "create_oois" method, to bulk the creation of objects.
> 
> For example, when there are 4 oois we'd like to create... We go from 4 times these 5 calls (total of 20 API calls):
> 
> * save_declaration
> * login()
> * save_boefje_meta
> * save_raw
> * save_normalizer_meta
> 
> To only 8 calls in total:
> 
> * save_many_declatations (x1)
> * login() (x1)
> * save_boefje_meta (x1)
> * save_raws (x1)
> * save_normalizer_meta (x4)
> 
> So for 20 Findings, this will result in 24 calls instead of 100 calls.
> 

Clear description, couldn't have described it better myself.

rocky/tools/ooi_helpers.py
```diff
@@ -254,3 +255,24 @@ def create_ooi(
...
+    bytes_client: BytesClient,
+    oois: list[OOI],
+    observed_at: datetime,
+    end_valid_time: datetime | None = None,
+) -> None:
+    declarations: list[Declaration] = []
+    raws: list[tuple[uuid.UUID, bytes]] = []
+
+    for ooi in oois:
+        task_id = uuid4()
```
underdarknl: are multiple task Id's needed? It could just be one task with many ooi's, right?

Donnype: We can indeed output multiple Declarations per normalizer, but I think this does assume these Declarations stem from the same raw file. Right now we save a raw file per Declaration however.

But I just looked at the single_ooi normalizer and concluded we should be able to save all declarations in one raw file actually as well. Then we can also tackle the multiple-normalizer-meta issue, CC @madelondohmen 👌

underdarknl: The only reason why we'd want to split the declarations into multiple raw files would be if we expect a newer version of the same 'job' to be available at some point. That newer version could then append/overwrite the previous declarations causing garbage collection on any declarations not present in the new version.
This however seems unlikely in the scenario of manually muted objects. (for csv uploads, we allow this by specifying an anchor point / input ooi, so multiple consecutive uploads of the same source do trigger this cleanup if needed).

Donnype: Yes and more importantly: deletion propagation does not apply to declarations

rocky/rocky/bytes_client.py
```diff
@@ -128,6 +128,26 @@ def _save_raw(self, boefje_meta_id: uuid.UUID, raw: bytes, mime_types: Set[str]
...
         return response.json()[file_name]
 
+    def _save_raws(
+        self, boefje_meta_id: uuid.UUID, raws: list[tuple[uuid.UUID, bytes]], mime_types: Set[str] = frozenset()
+    ) -> dict[str, str]:
+        files = [
+            {
+                "name": str(normalizer_id),
+                "content": b64encode(raw).decode(),
+                "tags": list(mime_types) + [f"Declaration/{str(normalizer_id)}"],
```
Donnype: I think this can be removed now:
```suggestion
                "tags": list(mime_types),
```

Donnype:
## Checklist for QA:

### What works:

- The mute findings button now seems almost instant

### What doesn't work:

- 

### Bug or feature?:
-

PR #4646, opened 2025-07-07
Use uv when building Debian packages

### LINKED ISSUE #4627: Debian test workflow fails after UV PR
See
https://github.com/minvws/nl-kat-coordination/actions/workflows/test_debian_packages_on_ubuntu.yml

The workflow started failing after merging the uv PR.

dekkers:
### Changes

Use uv when building Debian packages. 

- This also disables dh_dwz and excludes /opt/vens from dh_shlibdeps. The later means we don't have to rebuild from source wheels that include binary libraries causes further speedup.
- Updates gunicorn that is installed in Debian package to latest version

This also fixes the issue we currently have on main with the Debian packages that causes #4627 

### Issue link

Closes #4627 

### QA notes

The Debian packages build by the CI need to be tested.

---

---

## Checklist for QA:


rocky/debian/rules
```diff
@@ -23,14 +18,17 @@ override_dh_fixperms:
...
+	touch /tmp/requirements-empty.txt
+	dh_virtualenv --requirements=/tmp/requirements-empty.txt --skip-install --preinstall "uv"
+	$(DH_VENV_DIR)/bin/python -m uv sync --locked
+	$(DH_VENV_DIR)/bin/python -m uv pip install .
 	cd /octopoes && /usr/bin/python3 setup.py bdist_wheel
-	$(DH_VENV_DIR)/bin/python -m pip install /octopoes/dist/octopoes*.whl
+	$(DH_VENV_DIR)/bin/python -m uv pip install --no-deps /octopoes/dist/octopoes*.whl
+
+# remove pip and uv to prevent mutation of venv
+	$(DH_VENV_DIR)/bin/python -m uv pip uninstall pip uv
```
ammar92, line 31: In some of the other rules files you also passed the `--yes` option. I don't think it matters much, just pointing it out for consistency

boefjes/debian/rules
```diff
@@ -22,14 +18,19 @@ override_dh_fixperms:
 	chmod 755 $(DESTDIR)/usr/bin/update-katalogus-db
 
 override_dh_virtualenv:
-	dh_virtualenv $(DH_VENV_ARGS)
+# We want to use uv but dh_virtualenv doesn't support that. So we create an
+# empty requirements file and call uv manually..
+	touch /tmp/requirements-empty.txt
+	dh_virtualenv --requirements=/tmp/requirements-empty.txt --skip-install --preinstall "uv"
+	$(DH_VENV_DIR)/bin/python -m uv sync --locked
+	$(DH_VENV_DIR)/bin/python -m uv pip install .
```
ammar92, line 26: I think this line is redundant, the previous line will already install the environment based on the uv.lock file?

boefjes/debian/rules
```diff
@@ -22,14 +18,19 @@ override_dh_fixperms:
...
+	$(DH_VENV_DIR)/bin/python -m uv pip install gunicorn==23.0.0
 
-	$(DH_VENV_DIR)/bin/python -m pip install gunicorn==20.1.0
 	cd /octopoes && /usr/bin/python3 setup.py bdist_wheel
-	$(DH_VENV_DIR)/bin/python -m pip install /octopoes/dist/octopoes*.whl
+	$(DH_VENV_DIR)/bin/python -m uv pip install --no-deps /octopoes/dist/octopoes*.whl
```
ammar92, line 30: Hopefully we can soon install octopoes from source as a package rather than building it

boefjes/debian/rules
dekkers: uv sync installs the environment, but does not install the package itself. This is necessary so that the package/our code also ships in the Debian package. Previously dh_virtualenv did that for us, but the `--skip-install` tells it not to so it can be done with uv.

rocky/debian/rules
dekkers: In two other files I forgot to change it to uv. Uv doesn't need and understand the --yes option. I've fixed this that it also uses uv to uninstall.

ammar92 (dismissed):

Donnype (approved):

Issue #4231, opened 2025-03-27
Rocky health API endpoint


ammar92:
### Detailed description

While debugging an issue together with Kennisnet we identified that there was no automated way to check Rocky's health. Online other modules, this module lacks a dedicated endpoint to verify its operational status.

During a debugging session with Kennisnet, we identified the absence of an automated health check for the Rocky module. Unlike other modules, Rocky lacks a dedicated endpoint to verify its operational status.

### Feature benefit/User story

As a developer/ sysadmin/ devops, I want to access a health endpoint so that I can reliably confirm its operational health and identify potential issues quickly.

### Specifications

- Implement a health endpoint at `/api/v1/health` that returns health status information in a format consistent with other modules

### Additional information

_Any additional information, considerations, or context that might be helpful for understanding or evaluating the feature request._

### Alternatives considered

Checking the index page is an alternative approach, however, this method:
- Provides limited health information compared to a dedicated endpoint
-  Is susceptible to false positives due to proxy interactions
- Does not offer detailed insights into internal dependencies

PR #4649, opened 2025-07-08
Feature/scoped auth token only for api

Donnype:
### Changes

Only explicitly authenticate the API with the auth tokens.

### Issue link

Closes https://github.com/minvws/nl-kat-coordination/issues/4640

### Demo

-

### QA notes

- Create an organization with code "test"
- Create an auth token in [the admin page](http://localhost:8000/nl/admin/account/authtoken/add/)
- Use it to fetch some page:
```shell
$ curl -X GET -H "Content-type: application/json"  -H 'Authorization: Token $TOKEN' "http://localhost:8000/en/test/" -v | head
```
- See that you get redirected to `/en/login/` and not a regular page. On main I get redirected e.g. to the onboarding for a newly created member and auth token.

---

### Code Checklist

---

---

## Checklist for QA:


rocky/rocky/settings.py
```diff
@@ -423,16 +422,13 @@ def immutable_file_test(path, url):
 BROWSABLE_API = env.bool("BROWSABLE_API", DEBUG)
 
 if BROWSABLE_API:
-    DEFAULT_AUTHENTICATION_CLASSES = [
-        "knox.auth.TokenAuthentication",
-        "rest_framework.authentication.SessionAuthentication",
```
dekkers: We should keep `knox.auth.TokenAuthentication` here so that all API views can be accessed using tokens. This PR adds it to `OrganizationViewSet`, but this means the `ReportViewSet` and `ReportRecipeViewSet` don't have this authentication anymore. Setting the `DEFAULT_AUTHENTICATION_CLASSES` is so that we only need to set it once.

Note that `DEFAULT_AUTHENTICATION_CLASSES`  is only used for django-restframework, not by the rest of Django.

Donnype: `ReportViewSet`and `ReportRecipeViewSet` do in fact have the `TokenAuthentication` class set as well. But if indeed `DEFAULT_AUTHENTICATION_CLASSES` only applies to DRF removing the middleware would've been enough 👍

madelondohmen:
## Checklist for QA:

### What works:
Redirects correctly to `/en/login/?next=/en/test/xxx`

### What doesn't work:
N/A

### Bug or feature?:
N/A

rocky/rocky/settings.py
dekkers: Yes, and we need to remove the explicit setting of the authentication class in the viewsets because else the browsable API won't work because that also needs session authentication which is done in the Django settings.

ammar92 (dismissed):
Nice work. Quickly QA'd this myself too and I noticed that it will redirect to the login page also when a wrong token is passed.

Probably not related to this one https://github.com/minvws/nl-kat-coordination/issues/4231 but would be nice to pick this one up soon

rocky/rocky/settings.py
Donnype: Yes just realized I forgot

dekkers (approved):

PR #4685, opened 2025-07-21
Fix 500, logging and downloading of task if boefje fails

dekkers:
### Changes

This fixes several issues when boefje fails and with logging in general:
- Boefjes: bytes client doesn't log an error for 404
- Boefjes: In job handler the log messages are improved to describe better what actually is done
- Boefjes: The SQL closing session is logged with priority debug
- Bytes: Return a 404 if boefje meta can not be found instead of 500 internal server error
- Bytes: Do not log error/warning when session is rolled back because of an exception. This can just happen with a 404 and does not have to mean something went wrong. The exception is reraised and will be logged by fastapi if necessary.
- Rocky: The error message if there is no raw meta is changed to say that there is raw instead of that getting the raw failed.
- Rocky: The `DownloadTaskDetail` doesn't to set task_type and doesn't need to, so make this optional in `SchedulerView` base class.
- Rocky: In `LazyTaskList`, change debug logging to priority debug instead of info and correctly use structlog.

---

---

## Checklist for QA:

### DIFF
--- a/rocky/rocky/views/bytes_raw.py
+++ b/rocky/rocky/views/bytes_raw.py
@@ -9,6 +9,7 @@ from django.http import FileResponse
 from django.shortcuts import redirect
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
+from httpx import HTTPError
 
 logger = structlog.get_logger(__name__)
 
@@ -17,7 +18,14 @@ class BytesRawView(OrganizationView):
     def get(self, request, **kwargs):
         self.bytes_client.login()
         boefje_meta_id = kwargs["boefje_meta_id"]
-        raw_metas = self.bytes_client.get_raw_metas(boefje_meta_id, self.organization.code)
+        try:
+            raw_metas = self.bytes_client.get_raw_metas(boefje_meta_id, self.organization.code)
+        except HTTPError:
+            msg = _("Getting raw data failed.")
+            logger.exception("Getting raw data failed")
+            messages.add_message(request, messages.ERROR, msg)
+            return redirect(reverse("task_list", kwargs={"organization_code": self.organization.code}))
+
         if not raw_metas:
             msg = _("The task does not have any raw data.")
             messages.add_message(request, messages.ERROR, msg)
--- a/rocky/rocky/locale/django.pot
+++ b/rocky/rocky/locale/django.pot
@@ -9,7 +9,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2025-07-21 12:33+0000\n"
+"POT-Creation-Date: 2025-07-21 13:13+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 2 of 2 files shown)


bytes/bytes/database/sql_meta_repository.py
```diff
@@ -227,9 +227,9 @@ def create_meta_data_repository() -> Iterator[MetaDataRepository]:
     try:
         yield repository
     except Exception as error:
-        logger.exception("An error occurred during the session.")
+        logger.debug("An error occurred during the session.")
```
Donnype, line 230: These exceptions are logged elsewhere?

rocky/rocky/views/bytes_raw.py
```diff
@@ -15,20 +15,18 @@
...
-            raws = {raw_meta["id"]: self.bytes_client.get_raw(raw_meta["id"]) for raw_meta in raw_metas}
-
-            return FileResponse(zip_data(raws, raw_metas), filename=f"{boefje_meta_id}.zip")
-        except Http404 as e:
-            msg = _("Getting raw data failed.")
-            logger.error(msg)
-            logger.error(e)
+        self.bytes_client.login()
+        boefje_meta_id = kwargs["boefje_meta_id"]
+        raw_metas = self.bytes_client.get_raw_metas(boefje_meta_id, self.organization.code)
```
Donnype: Shouldn't we catch other HTTPErrors from Bytes here and display an error message?

rocky/rocky/views/scheduler.py
```diff
@@ -56,7 +56,8 @@ class SchedulerView(OctopoesView):
     def setup(self, request, *args, **kwargs):
         super().setup(request, *args, **kwargs)
         self.scheduler_client = scheduler_client(self.organization.code)
-        self.scheduler_id = self.task_type
+        if hasattr(self, "task_type"):
+            self.scheduler_id = self.task_type
```
Donnype, line 60: This will only postpone the error to line 65? I've had to figure this structure out as well recently. I concluded that every "final" child class has a task_type except for perhaps 1 view (forgot which one..). Thoughts?

boefjes/boefjes/clients/bytes_client.py
```diff
@@ -53,7 +53,7 @@ def _verify_response(response: Response) -> None:
         try:
             response.raise_for_status()
         except HTTPStatusError as error:
-            if error.response.status_code != 401:
+            if error.response.status_code != 401 and error.response.status_code != 404:
```
Donnype, line 56: What's the scenario we can ignore here? This potentially breaks several `get_` methods below I think

dekkers: How would changing what is logged break anything?

rocky/rocky/views/scheduler.py
dekkers: No, because `get_task_filters` isn't called by every subclass. It's a mess I don't want to spend a lot of time on to refactor and fix properly, with this change `DownloadTaskView` that doesn't need `task_type` works again.

bytes/bytes/database/sql_meta_repository.py
dekkers: They should be logged by FastAPI if they aren't handled. If not we should fix that in general instead of only logging exceptions that happen inside the sql repository but then also logging exceptions for 404s and other things that should not be done.

rocky/rocky/views/bytes_raw.py
dekkers: I've added that, but we weren't doing that before, because we would only catch the `Http404` raised by bytes client.

rocky/rocky/views/scheduler.py
Donnype: Agreed that this isn't a thing to look into too much

bytes/bytes/database/sql_meta_repository.py
Donnype: Ah I see the `raise` below indeed

boefjes/boefjes/clients/bytes_client.py
Donnype: Thought this ignored the error, missed the `raise` on line 60 here

Donnype (approved):

PR #4704, opened 2025-07-31
Add boefje that checks Github workflows using Octoscan

Souf149:
### Changes

Added a boefje that makes use of [Octoscan ](https://github.com/synacktiv/octoscan) to scan github workflows for URLs added in OpenKAT.

### Demo

image
image
image

### QA notes

- Add a Github hostname (github.com)
- Add a HostnameHTTPURL. For example:
  - Scheme https
  - Port 443
  - Path /minvws/nl-kat-coordination
  - Net location pointing to the previously mentioned Github hostname
- Make Octoscan scan this object and vulnerabilities should be found.
---

### Code Checklist

---

---

## Checklist for QA:


brennodewinter:
## Review — Octoscan boefje

A few issues to address before merge:

### 1. Finding produced on empty results

```python
def scan_octoscan_output(data, ooi_ref):
    if data:
        finding_type = KATFindingType(id="KAT-VULNERABLE-GH-WORKFLOW")
        yield finding_type

    description = "Found vulnerabilities in GitHub workflow files.\n"
    for vuln in data:
        description += f"..."

    yield Finding(finding_type=finding_type.reference, ooi=ooi_ref, description=description)
```

The `Finding` is yielded unconditionally — when `data == []` (no vulnerabilities), a Finding is still created with only the header line as description, and it references `finding_type` which was never yielded (the `if data:` guard skipped it). This produces an orphan Finding referencing a non-existent FindingType.

Fix: move the `Finding` yield inside the `if data:` block, or guard the whole function with `if not data: return`.

### 2. `REPO_TOKEN` required in schema but default "" in code

`schema.json` marks `REPO_TOKEN` as `required`, but `main.py` does `os.getenv("REPO_TOKEN", "")` and passes the empty string to octoscan. If no token is configured, octoscan gets an empty credential and will likely fail on private repos. Either make the empty string an explicit skip (return early with an info message), or align the schema (remove `required` if public-only repos are a valid use case).

### 3. Dockerfile clones without pinning

```dockerfile
RUN git clone https://github.com/synacktiv/octoscan /octoscan && \
    cd /octoscan && \
    /usr/local/go/bin/go mod tidy && \
    /usr/local/go/bin/go build
```

No commit hash or tag pinned — every build fetches whatever `HEAD` is at clone time. This is a reproducibility and supply-chain concern. Pin to a specific commit:

```dockerfile
RUN git clone https://github.com/synacktiv/octoscan /octoscan && \
    cd /octoscan && git checkout <commit-sha> && \
    ...
```

### 4. Minor

- `boefje_meta` typed as `dict` instead of `BoefjeMeta` — inconsistent with other boefjes in the repo.
- `org, repo = url.split("/")[-2:]` is fragile — a trailing slash or query params on the URL will break it. A `urlparse` + path split would be safer.
- `normalizer.json` `produces` lists `Finding` and `KATFindingType` — correct, good.

PR #4744, opened 2025-08-22
Make it possible to configure logging using ROCKY_LOG_CFG

dekkers:
### Changes

Make it possible to configure logging by specifying a logging configuration file using ROCKY_LOG_CFG in the same way as we do in the other services.

---

---

## Checklist for QA:

### DIFF
--- a/rocky/rocky/settings.py
+++ b/rocky/rocky/settings.py
@@ -78,7 +78,7 @@ LOGGING = {
 
 
 def configure_logging(logging_settings):
-    log_cfg = env("ROCKY_LOG_CFG", default="")
+    log_cfg = env.path("ROCKY_LOG_CFG", default="")
     if log_cfg:
         with Path(log_cfg).open() as f:
             logging.config.dictConfig(json.load(f))
--- a/rocky/rocky/settings.py
+++ b/rocky/rocky/settings.py
@@ -78,7 +78,7 @@ LOGGING = {
 
 
 def configure_logging(logging_settings):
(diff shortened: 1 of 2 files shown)


rocky/rocky/settings.py
```diff
@@ -69,24 +71,45 @@ class SaferExceptionReporterFilter(SafeExceptionReporterFilter):
...
-        }
-    },
+    "formatters": {"default": {"format": "%(message)s"}},
+    "handlers": {"console": {"class": "logging.StreamHandler", "formatter": "default"}},
     "loggers": {"root": {"handlers": ["console"], "level": env("LOG_LEVEL", default="INFO").upper()}},
 }
 
+
+def configure_logging(logging_settings):
+    log_cfg = env("ROCKY_LOG_CFG", default="")
```
ammar92, line 81: You can also use `Env.path`. Doesn't do validation, but does return a `Path` object. See [docs](https://django-environ.readthedocs.io/en/latest/api.html#environ.Env.path)

ammar92 (dismissed):
Nice work. Tiny suggestion. Tested and seems to work. `LOGGING_FORMAT` is still configurable as well

rocky/rocky/settings.py
dekkers: I've changed it, but it would have been more useful if it returned a pathlib.Path object instead of an environ.Path object.

dekkers: This actually doesn't work if I the variable isn't set, because env.path then returns "/app/rocky", so I reverted it.

PR #4895, opened 2025-11-04
docs: using sphinx icore-open python package

jpbruinsslot:
Adding support for using the sphinx icore-open theme package (https://pypi.org/project/sphinx-icore-open-theme/). This is currently in beta and openkat is one of the targets we want to test against.

Related: https://github.com/minvws/nl-kat-coordination/pull/4376


jpbruinsslot:
Having issues with dependencies, but it seems that multiple PR's have the same problem

underdarknl:
Your pyproject.toml file moves from python 3.10 to 3.11 is this intentional/required?

PR #4909, opened 2025-11-13
JWT token auth for plugin callbacks

### LINKED ISSUE #4833: Investigate finer grained API token for the plugins
We should not create users with model-level permissions on each plugin run. Instead, we should use a (stateless?) token that is short-lived and encapsulates not only model-level permissions, but also scopes the API access. This should have sane defaults be editable per plugin. Examples:
- Plugin X may only add data or edit data it has created itself
- Plugin A can only read hostnames and ips, only write IPPorts
- Plugin B can only read/write Files
- Plugin C can only read the File with id 123, and write at most 2 Files in the API (?)
- Plugin D can only read hostname A and write DNS records **for hostname A**
- Plugin E can only read all related models to hostname A (network, dns records and ip through DNSArecords)

---

### Implementation

It is not clear if all of the above will be part of the future requirements, so a pragmatic approach would be to see if we can reduce the default permissions as much as possible and still be able to use our current plugins.

Donnype:
### Changes

Instead of creating fake users, now properly use stateless, short-lived tokens for the plugin callbacks to restrict their access.

### Issue link

Closes #4833


openkat/permissions.py
```diff
@@ -18,22 +17,84 @@ class KATModelPermissions(DjangoModelPermissions):
...
         if getattr(view, "_ignore_model_permissions", False):
             return True
 
+        return request.user.has_perms(self.view_perms(request, view))
+
+    def has_token_permission(self, request, view):
+        if not hasattr(request, "auth") or not isinstance(request.auth, dict):
+            return False
+
+        token_perms = request.auth.get("permissions", {}) or {}
```
ammar92:
```suggestion
        token_perms = request.auth.get("permissions", {})
```

openkat/settings.py
```diff
@@ -205,10 +205,11 @@ class SaferExceptionReporterFilter(SafeExceptionReporterFilter):
 
 ROOT_URLCONF = "openkat.urls"
 
+REDIS_PASSWORD = env.str("REDIS_PASSWORD", default=None)
 CACHES = {
     "default": {
         "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": f"redis://:{env.str('REDIS_PASSWORD')}@{env.str('REDIS_HOST')}:6379/1",
+        "LOCATION": f"redis://:{REDIS_PASSWORD}@{env.str('REDIS_HOST', 'localhost')}:6379/1",
```
ammar92: Nice catch. Perhaps for consistency `REDIS_HOST` next to `REDIS_PASSWORD`?

```suggestion
REDIS_HOST = env.str("REDIS_HOST", "localhost")
REDIS_PASSWORD = env.str("REDIS_PASSWORD", None)
CACHES = {
"default": {
"BACKEND": "django_redis.cache.RedisCache",
        "LOCATION": f"redis://:{REDIS_PASSWORD}@{REDIS_HOST}:6379/1",
```

objects/models.py
```diff
@@ -95,7 +95,7 @@ class Meta:
     objects = XTDBQuerySet.as_manager()
 
     def save(self, force_insert=False, force_update=False, using=None, update_fields=None):
-        super().save(force_insert=True, force_update=False, using=using, update_fields=update_fields)
+        super().save(force_insert=True, force_update=False, using=using, update_fields=None)
```
ammar92, line 99: What is the reason this argument isn't propagated to the parent's method?

plugins/models.py
```diff
@@ -46,7 +46,8 @@ class Plugin(models.Model):
     # Image specifications
     oci_image = models.CharField(max_length=256, null=True)
     oci_arguments = ArrayField(models.CharField(max_length=256, blank=True), default=list)
-    version = models.CharField(max_length=16, null=True)
+    version = models.CharField(max_length=16, null=True)  # TODO: drop or use
+    permissions = models.JSONField(default=dict)
```
ammar92, line 50: I was wondering if there's already a defined schema for this. I've only seen empty dicts or `dict[Literal['pks'], list[str]]`, but I assume there must be other supported formats?

openkat/settings.py
```diff
@@ -421,9 +422,12 @@ class SaferExceptionReporterFilter(SafeExceptionReporterFilter):
 
 # Turn on the browsable API by default if DEBUG is True, but disable by default in production
 BROWSABLE_API = env.bool("BROWSABLE_API", DEBUG)
+JWT_KEY = env.str("JWT_KEY", "")
```
dekkers: Why not reuse the already existing SECRET_KEY? We can copy the logic from drf-simplejwt because that also uses SECRET_KEY by default: https://django-rest-framework-simplejwt.readthedocs.io/en/latest/settings.html#signing-key

openkat/auth/jwt_auth.py
```diff
@@ -0,0 +1,33 @@
...
+class JWTTokenAuthentication(TokenAuthentication):
+    def authenticate_credentials(self, key):
+        try:
+            return AnonymousUser(), jwt.decode(
+                key, settings.JWT_KEY, algorithms=[settings.JWT_ALGORITHM], options={"require": ["exp", "iat"]}
+            )
+        except PyJWTError:
+            return None
+
+    @classmethod
```
dekkers: `cls` is not used so can be a static method

objects/models.py
Donnype: Because it throws an error when force_insert is `True` and there are `update_fields`

plugins/models.py
Donnype: There is also "search" and "limit" at the moment, but that's all

PR #4943, opened 2025-12-17
Fix nikto boefje: bug fixes and performance improvements

### LINKED ISSUE #4945: Nikto boefje crashes with TypeError and missing 'name' field
## Description

The Nikto boefje has two bugs that cause it to fail:

### Bug 1: TypeError in main.js
`main.js:91` tries to iterate over `data.vulnerabilities` but Nikto outputs an array of scans, not a single object with a `vulnerabilities` property.

```javascript
// Current (broken):
for (const vuln of data.vulnerabilities) { ... }

// Should be:
for (const scan of data) {
    for (const vuln of scan.vulnerabilities) { ... }
}
```

### Bug 2: Missing 'name' field in oci_adapter.js
The file output is missing the required `name` field, causing API validation errors when submitting results to Bytes.

```javascript
// Current (broken):
{ content: base64, tags: [...] }

// Should be:
{ name: "output.json", content: base64, tags: [...] }
```

## Additional Issues

- Scans take ~25 minutes without tuning parameters
- No timeout causes scans to run indefinitely on some targets

## Proposed Fix

PR #4943 addresses all these issues:
- Fix TypeError by iterating over scan array
- Add required `name` field
- Add `TUNING` parameter (default: `3b`) for targeted scans
- Add `MAX_TIME` parameter (default: `600s`) to prevent endless scans

## Related

- PR #4943 - Fix nikto boefje: bug fixes and performance improvements
- PR #4787 - Alternative fix approach by @Souf149

hasecon:
## Summary

- **Fix TypeError**: Iterate over scan array instead of direct data object in `main.js`
- **Fix API compatibility**: Add required `name` field to file output in `oci_adapter.js`
- **Add TUNING parameter** (default: `3b`) for targeted scans
- **Add MAX_TIME parameter** (default: `600s`) to prevent endless scans
- Apply prettier formatting to JS/JSON files

## Issue link

Closes #4945

## Details

The nikto boefje had two bugs:
1. `main.js:91` tried to iterate over `data.vulnerabilities` but nikto outputs an array of scans, not a single object
2. `oci_adapter.js` was missing the required `name` field in file output, causing API validation errors

Additionally, performance was improved by adding configurable tuning options:
- Default tuning `3b` focuses on Information Disclosure + Software Identification
- Default max time of 600 seconds prevents scans from running indefinitely

**Performance improvement**: Scan time reduced from ~25 minutes to ~5 minutes while maintaining detection of security headers and software identification.

## Related

See also #4787 which addresses similar issues with a different approach. This PR adds performance tuning parameters not present in #4787.

## Test plan

- [x] Tested locally with Docker build
- [x] Verified nikto scan completes without TypeError
- [x] Verified API accepts output with `name` field
- [x] Verified tuning parameter reduces scan time

### DIFF
--- a/boefjes/boefjes/plugins/kat_nikto/oci_adapter.js
+++ b/boefjes/boefjes/plugins/kat_nikto/oci_adapter.js
@@ -1,4 +1,3 @@
-import { execSync } from "node:child_process";
 import run from "./main.js";
 
 /**
@@ -9,14 +8,14 @@ function b64encode(inp) {
   return Buffer.from(inp).toString("base64");
 }
 
-function main() {
+async function main() {
   const input_url = process.argv[process.argv.length - 1];
 
   // Getting the boefje input
+  let boefje_input;
   try {
-    var boefje_input = JSON.parse(
-      execSync(`curl --request GET --url ${input_url}`).toString(),
-    );
+    const input_response = await fetch(input_url);
+    boefje_input = await input_response.json();
   } catch (error) {
     console.error(`Getting boefje input went wrong with URL: ${input_url}`);
     throw new Error(error);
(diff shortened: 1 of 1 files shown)


boefjes/boefjes/plugins/kat_nikto/oci_adapter.js
underdarknl:
```suggestion
  await fetch(output_url, {
    method: "POST",
    headers: {
      "Content-Type": "application/json",
    },
    body: out_json,
  }).catch((error) => {
    console.error(`Posting boefje output went wrong with URL: ${output_url}`);
    throw new Error(error);
  });
```

both the curl call, and the subsequent execSync(cmd);  can be removed.

underdarknl: If curl is dropped, we can remove this import.

underdarknl: Removing the usage of Curl, and using the node fetch command is saver, as it allows us to not use any cli, and thus avoid possible injection.

underdarknl (changes requested):
If we can remove the external call to curl and use Fetch instead, that would be great.
Beyond that, it looks good to me.

underdarknl (approved):

PR #5017, opened 2026-02-18
Pass katalogus settings to containerized boefjes and add local CVE API

hasecon:
## Summary

- **Pass katalogus settings as Docker env vars** when starting containerized (OCI) boefjes. `DockerBoefjeHandler` now passes `boefje_meta.environment` to `containers.run()`, so plugins receive their configured settings (e.g. `CVEAPI_URL`) without relying solely on the callback re-hydration.

- **Add local CVE API** as an optional Docker Compose service (`cveapi` profile). Downloads all CVE data from NVD and serves it as static JSON files, removing the dependency on the external `cve.openkat.dev` which is frequently unavailable.

### Commit 1: Pass katalogus settings to containerized boefjes
- `boefjes/boefjes/job_handler.py`: add `environment=boefje_meta.environment or {}` to `containers.run()`

### Commit 2: Add local CVE API
- `cveapi/Dockerfile` + `cveapi/entrypoint.sh`: HTTP server + NVD download with 24h update cycle
- `docker-compose.yml`: `cveapi` service with profile, connected to both `default` and `boefjes` networks
- `.env-defaults`: `BOEFJE_CVEAPI_URL` (the `BOEFJE_` prefix is stripped by `get_environment_settings()`)
- `docs/source/installation-and-deployment/cveapi.rst`: Docker Compose usage instructions

## How to use

```bash
COMPOSE_PROFILES=cveapi make kat
```

No additional configuration needed — `BOEFJE_CVEAPI_URL` in `.env-defaults` points to the local instance automatically.

## Test plan

- [x] Clean build with `COMPOSE_PROFILES=cveapi make kat`
- [x] Verify CVE data downloads from NVD (~300k CVEs, ~2 GB)
- [x] Push `cve-finding-types` task via scheduler API
- [x] Verify task completes successfully using local CVE API
- [x] Verify it works without per-org katalogus setting (only `BOEFJE_CVEAPI_URL`)
- [x] Verify containerized boefje receives `CVEAPI_URL` env var via Docker environment


cveapi/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.13-slim
+
+RUN pip install --no-cache-dir httpx
```
underdarknl: Putting this requirement into a requiremens.txt file is better, this also allows us to do version/hash pinning and allows dependabot et all to suggest updates.

cveapi/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+UPDATE_INTERVAL=${CVEAPI_UPDATE_INTERVAL:-86400}
+
+# Start HTTP server immediately so CVEs already downloaded are available
+cd /var/lib/kat-cveapi
+python -m http.server 8080 &
+
+# Initial download/update
+cd /app
+echo "Starting CVE download..."
+python -c "from cveapi import run; run()"
```
underdarknl, line 17: We should make sure to use the provided cache volume to store this data. This makes sure we only download when there is no cache available, or when the cache is too old.

boefjes/boefjes/job_handler.py
```diff
@@ -62,6 +62,7 @@ def handle(self, task: Task) -> tuple[BoefjeMeta, BoefjeOutput] | None | Literal
                 remove=True,
                 network=settings.docker_network,
                 volumes=[f"{self.CACHE_VOLUME_NAME}:{self.CACHE_VOLUME_TARGET}"],
+                environment=boefje_meta.environment or {},
```
underdarknl: This env might still miss various 'allowed' env vars, such as CA-BUNDLE (needed for accepting self-signed certs in httpx), and vars like http proxy etc. These are not configurables for the boefje itself, but allowed to be accesses from the host env.

underdarknl (changes requested):

underdarknl:
I think you forgot to add the requirements.txt file to the git commit.

underdarknl (approved):

PR #5111, opened 2026-04-10
Update expired FreeTSA certificate for RFC3161 timestamping

hasecon:
### Changes

The bundled `freetsa.crt` certificate expired on 2026-03-11. This replaces it with the current certificate from https://freetsa.org/files/tsa.crt (valid until 2040-02-02).

| | Old certificate | New certificate |
|---|---|---|
| Valid from | 2016-03-13 | 2026-02-15 |
| Valid until | **2026-03-11 (expired)** | 2040-02-02 |
| Email | busilezas@gmail.com | busilezas@mailbox.org |

### Known issue: `rfc3161ng` incompatible with `cryptography>=46`

**This PR alone is not sufficient to fix RFC3161 timestamping.** After updating the certificate, a second error occurs:

```
TypeError: ECPublicKey.verify() takes 3 positional arguments but 4 were given
```

The `rfc3161ng` library calls `ECPublicKey.verify()` with the old `cryptography` API signature. Since `cryptography` was upgraded to 46.0.7 (#5105/#5106), the `verify()` method signature changed.

Options to fix:
1. Upgrade `rfc3161ng` to a version compatible with `cryptography>=46` (if available)
2. Pin `cryptography<46` for bytes (not recommended — security regressions)
3. Fork/patch `rfc3161ng` to use the new API

This certificate update is still needed regardless of which option is chosen.

### Issue link

Found during v1.22.0rc1 release testing when enabling RFC3161 timestamp signing.

### QA notes

The certificate update can be verified independently:
```python
from cryptography import x509
cert = x509.load_pem_x509_certificate(open("bytes/bytes/timestamping/certificates/freetsa.crt", "rb").read())
print(cert.not_valid_after_utc)  # Should show 2040, not 2026
```

Full RFC3161 testing requires resolving the `rfc3161ng` / `cryptography` incompatibility first.


underdarknl (approved):

underdarknl:
Looks good to me, but we should also try to upstream the changes in Ec support.

underdarknl:
I'm also not sure (yet) what this means for previously signed files, as we'd need to make sure 'their proof' is still verifiable. Maybe we need to add tracking 'which' cert we used to sign against in bytes so we can provide the user the correct information when the need of verification arises. (again, its 7:30 in the morning, I'm not overseeing the whole picture here)

hasecon:
Two updates on this PR:

## 1. Upstream issue filed

Filed the `rfc3161ng` EC-key incompatibility upstream: [trbs/rfc3161ng#26](https://github.com/trbs/rfc3161ng/issues/26). The issue proposes a minimal `isinstance(public_key, ec.EllipticCurvePublicKey)` branch around the existing RSA verify call. If upstream merges a fix, the 64-line monkey-patch in `bytes/bytes/timestamping/rfc3161.py` can be reverted and we can just pin a new `rfc3161ng` release in `requirements.txt`.

Small correction from the PR description worth noting: this is **not** a `cryptography>=46` regression. `EllipticCurvePublicKey.verify()` has been 3-arg for many years. The real root cause is that `rfc3161ng.check_timestamp()` hard-codes the RSA verify signature and was never updated for EC keys. FreeTSA's cert rotation to ECDSA P-384 in 2026-02 simply exposed it.

## 2. @underdarknl — answer on previously-signed files

I looked into this. Short version: **yes, this PR breaks verification of any raw file that was signed before the cert rotation**, and we can demonstrate it concretely. The good news is that the blast radius may be zero in practice depending on who actually had RFC3161 enabled before 2026-02.

### Why it breaks

`RFC3161HashRepository.verify()` uses `self.certificate` (the bundled `freetsa.crt`) as the trust anchor for *every* stored token:

```python
def verify(self, link, secure_hash):
    time_stamp_token = base64.b64decode(str(link))
    return self.timestamper.check(time_stamp_token, data=secure_hash.encode())
```

After this PR, `self.certificate` is the new EC cert (valid 2026-02-15 → 2040-02-02). Tokens stored before 2026-02 were signed with the old RSA private key. Public-key crypto being what it is, the new EC public key cannot verify signatures made with the old RSA private key.

### Concrete reproduction

I ran this on the fix branch, against live FreeTSA, with both the new EC cert (bundled in this PR) and the old RSA cert (extracted from `main`):

```
[2] monkey-patch applied: _patched_check_timestamp
[3] store() against live FreeTSA with new EC cert... stored OK, token len: 1220 chars
[4] verify with new EC cert (matching signer):       True
[5] verify SAME token with OLD RSA cert (mismatch):  InvalidSignature
```

Step 5 is exactly the scenario you raised: same stored token, different trust cert → verification fails. The cross direction (token signed with old RSA key, verified against new EC cert) would fail identically — it's just signature/pubkey mismatch.

Side note while I was here: the monkey-patch itself **does** work correctly — I initially thought it was silently broken because I was running tests from `main` and got confused. The patch rebinds both `rfc3161ng.check_timestamp` and `rfc3161ng.api.check_timestamp`, and `RemoteTimestamper.check()` picks up the new binding at call time because Python resolves module-level names lazily.

### How bad is this in practice?

Depends on who actually ran RFC3161 signing before the rotation. Facts I can check from here:

- In `.ci/.env.test` the `RFC3161_PROVIDER` / `RFC3161_CERT_FILE` vars are commented out. CI has never exercised the real RFC3161 path.
- The `mock_hash_repository` fixture in `bytes/tests/conftest.py` falls back to `InMemoryHashRepository` when those vars are missing. So `test_rfc3161_external_api` in the test suite has always been testing in-memory behaviour despite its name.
- The bytes DB on my running stack has 345 `raw_file` rows with a `hash_retrieval_link` — all of them are UUIDs from `InMemoryHashRepository.store()`, not base64 TSTs. My local instance never exercised RFC3161 either.
- Per the PR description, "Found during v1.22.0rc1 release testing when enabling RFC3161 timestamp signing" — so RFC3161 is an opt-in feature that was only recently being dogfooded for v1.22.

So the question is: **do we know of any deployment that has `RFC3161_PROVIDER` set and has been running it long enough to have stored tokens from before 2026-02?** If not, this PR breaks nothing in practice. If yes, those deployments lose the verification chain on every pre-rotation raw file.

### Remediation options, ordered by cost

1. **Ship both certs in a trust bundle and iterate on verify failure.** Small code change to `RFC3161HashRepository`: accept `list[bytes]` instead of `bytes`, try each cert, return True on the first that verifies, raise on the last. Add `bytes/timestamping/certificates/freetsa_2016.crt` back alongside the new one. Easy, but the bundle keeps growing with each rotation and you have to keep expired certs around forever.

2. **Embed the TSA cert in each TST going forward (`include_tsa_certificate=True`).** Change `RFC3161HashRepository.store()` to pass this flag, and make `verify()` prefer the embedded cert over `self.certificate` when present. Solves rotations for new tokens only — old tokens still need option (1) or (3).

3. **Track the signer per raw file.** Add a `signer_cert_fingerprint` column to `raw_file`, record it at `store()` time, and look up the right cert from a bundle at `verify()` time. Schema migration, but fully future-proof. Also makes it possible to surface "this file was verified with cert X" to the user, which is I think what you were gesturing at.

4. **Accept the loss and document it.** If deployment telemetry says nobody had this enabled before 2026-02, just add a line to the PR description explaining that pre-rotation tokens are unverifiable and that nothing was lost in practice.

My recommendation: **option 1 now** (trivial), **option 3 for v1.23** as a proper fix. Option 2 is a nice-to-have but doesn't solve the historic-file problem alone. If you want, I can draft option 1 as a stacked PR on top of this one so we don't block the cert update itself.

Either way, we should not merge this PR *as-is* without at least knowing which deployments had RFC3161 enabled — that's the bit you need to confirm before we pick a remediation path.

underdarknl:
A migration as proposed under 3, would be good, however keeping track of all 'possible' keys in a separate lookup table (or even an Enum for now) would be better than just storing the key's id/hash in each raw-file record.
There also remains a question on what we want to do with multiple-raw files from a single job. It feels highly unlikely that we'd end up with different signing keys within a single job's raw files. 


Concretely, I'd say: 

* Add storage that binds the possibly used keys to a space-efficient identifier.
* keep track of which of these I'd should be stored for each job. 
* Add a migration script for the SQL migration that sets this ID for al jobs that have a signing provider FreeTSA (can we check that by the signing url?)

Bonus , If we are doing migrations anyway:
* migrate the signing provider away from the rawfiles to the jobs in a similar lookup style to minimize storage space.

This will probably also need some rework in the various bytes api's job models and rocky code / documentation.

hasecon:
Agreed — lookup table binding signing keys to a space-efficient identifier + per-job tracking is the right long-term shape. The per-rawfile fingerprint I proposed in option 3 was lazier than it needed to be.

Proposing we land this PR as-is now (cert + EC monkey-patch), because:

- The expired cert already breaks bytes for anyone who enables RFC3161 — I just hit it on the running stack during unrelated testing (#5117), had to flip `BYTES_EXT_HASH_REPOSITORY=IN_MEMORY` to get past it. Blocks any v1.22.0rc2 dogfooding that relies on signing.
- CI has never exercised real RFC3161 (vars commented out in `.ci/.env.test`), and the `mock_hash_repository` fixture silently falls back to `InMemoryHashRepository` when the provider vars are missing. No green-path coverage is being traded for this merge.
- The upstream EC-support issue is filed ([trbs/rfc3161ng#26](https://github.com/trbs/rfc3161ng/issues/26)). When that merges we can drop the 64-line monkey-patch in a follow-up.

I ran the fix live against FreeTSA inside the bytes container on this branch:

```
Cert valid until: 2040-02-02 19:44:22+00:00
Public key type: ECPublicKey
[OK] Stored TST: link_len=1220
[OK] Verified: True
```

And will open a separate v1.23 tracking issue for the full migration:

1. `signer` lookup table (provider URL + cert fingerprint → small id)
2. Per-job (not per-rawfile) column in `job` / `boefje_meta` referencing that id
3. Migration that backfills existing jobs with the FreeTSA-2016 identifier, conditioned on the old provider URL
4. As a bonus, migrate `signing_provider` away from raw-file rows into the same lookup scheme so we shrink the main table

Does that split feel right — ship the cert fix now, separate issue for the migration work targeting v1.23?

PR #5113, opened 2026-04-11
Add normalizer unit tests for kat_shodan, kat_wpscan and kat_nuclei_cve

hasecon:
## Summary

Adds unit tests for three boefje normalizers that had no test coverage. Part of the same coverage pass as #5112 (octopoes bits).

Normalizers covered:
- `kat_shodan` — open-port detection from Shodan host scans (CVE/`vulns` path is **not** tested here, see note below)
- `kat_wpscan` — CVE extraction from WPScan JSON output via `wpscan_out_parse`
- `kat_nuclei_cve` — CVE finding extraction from Nuclei JSONL output

All three tests use inline JSON payloads — no new fixture files needed under `boefjes/tests/examples/`.

## Bug discovered while writing tests

The original triage scope was `kat_shodan, kat_nmap_tcp, kat_wpscan`, but `kat_nmap_tcp` is already covered by `tests/plugins/test_nmap.py::test_normalizer`. I replaced it with `kat_nuclei_cve`.

While writing the Shodan test I discovered that `kat_shodan/normalize.py` is broken on the `vulns` code path:

```python
if "vulns" in scan:
    for cve in scan["vulns"].values():   # <-- iterates DETAIL DICTS
        ft = CVEFindingType(id=cve)      # <-- crashes: id is not a string
```

The Shodan REST API returns `vulns` as a dict keyed by CVE id (with detail dicts as values), so `.values()` yields dicts, not CVE strings. `CVEFindingType(id={...})` then raises a `pydantic ValidationError`. The bug is latent because most Shodan accounts do not return `vulns` data unless the CVE facet is enabled.

To keep this PR scoped to test additions only, the `vulns` test is intentionally omitted. A follow-up fix PR will:
1. Change `scan["vulns"].values()` → iterate keys (or `.items()` if we want the detail string as a finding description)
2. Add the missing test for the `vulns` path

## Test plan

- [x] `docker exec openkat-reviev-boefje-1 python -m pytest tests/plugins/test_shodan.py tests/plugins/test_nuclei_cve.py tests/plugins/test_wpscan.py` — **11/11 passed**
- [x] `pre-commit run --files ...` — passed
- [ ] CI (`boefjes-tests`) green

## Notes

- No production code changed; tests-only PR.
- Follows the existing style of `test_rdns.py` / `test_snyk.py` — direct `run()` invocation with constructed `input_ooi` dict and inline raw bytes.
- Draft until Jan has reviewed.

PR #5112, opened 2026-04-11
Add unit tests for five high-impact octopoes bits

hasecon:
## Summary

Adds unit tests for five octopoes bits that previously had no coverage. Part of a broader test coverage pass across the core components — follow-up PRs for boefjes normalizers, bytes API and mula rankers will be filed separately.

Bits covered:
- `dns_resolving` — emits `ResolvedHostname` per A/AAAA record
- `missing_dmarc` — yields `KAT-NO-DMARC` finding when apex domain has no DMARC record
- `missing_spf` — yields `KAT-NO-SPF` finding (same shape)
- `missing_dkim` — yields `KAT-NO-DKIM` finding (same shape)
- `https_redirect` — yields `KAT-NO-HTTPS-REDIRECT` when an HTTP URL has headers but no `Location`

Context: octopoes/bits had only 4 of 36 bits covered by unit tests (11%). These five were selected as high-impact, low-fixture-cost entry points — they power core email security and DNS inference findings that feed most KAT reports.

Each test file asserts:
1. **Happy path**: bit yields the expected OOI or finding
2. **Guard conditions**: subdomain skip, NXDOMAIN skip, presence of the required record prevents the finding (as applicable per bit)

## Test plan

- [x] `docker exec openkat-reviev-octopoes_api-1 python -m pytest tests/test_bit_dns_resolving.py tests/test_bit_missing_dmarc.py tests/test_bit_missing_spf.py tests/test_bit_missing_dkim.py tests/test_bit_https_redirect.py -v` — **21/21 passed** locally
- [x] `pre-commit run --files ...` — passed (ruff, ruff-format, codespell, vulture)
- [ ] CI (`test`, `rtest`) green

## Notes

- No production code changed; tests-only PR.
- Follows the existing style of `test_bit_spf_discovery.py` and `test_bit_missing_headers.py` — direct `run()` invocation with constructed OOIs, no XTDB fixtures needed.
- Draft until Jan has a chance to review.

PR #4994, opened 2026-02-03
Feat/bytes normalizer metas endpoint

underdarknl:
### Changes

Adds endpoints to allow clients to fetch multiple normalizer metas from bytes in one go.

### Issue link

_You have to create an issue to link to this PR. If this really is not possible, write a very detailed description here and add this PR to the project board directly._

_Please add the link to the issue after "Closes"._

Closes ...

### Demo

_Please add some proof in the form of screenshots or screen recordings to show (off) new functionality, if there are interesting new features for end-users._

### QA notes

This patch just adds the endpoints, rocky consumes this endpoint on the ooi detail page for ooi's that have tasks and related origins. Testing can be done by doing a curl againt the endpoint and listing multiple meta-id's.
You can gather meta-id's by tailing the bytes log.

### Code Checklist

- [x] All the commits in this PR are properly PGP-signed and verified.
- [x] This PR only contains functionality relevant to the issue.
- [x] I have written unit tests for the changes or fixes I made.
- [x] I have checked the documentation and made changes where necessary.
- [x] I have performed a self-review of my code and refactored it to the best of my abilities.

- [x] Tickets have been created for newly discovered issues.
- [x] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [x] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [x] I have included comments in the code to elaborate on what is not self-evident from the code itself, including references to issues and discussions online, or implicit behavior of an interface.

---

## Checklist for code reviewers:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_qa.md) into your comment.

PR #5114, opened 2026-04-11
Add bytes API coverage for /normalizer_metas and auth edge cases

hasecon:
## Summary

Adds 8 tests across two areas of bytes that had no coverage. Part of the same coverage pass as #5112 (octopoes bits) and #5113 (boefjes normalizers).

### Integration: `GET /bytes/normalizer_metas` (list-by-UUID)

This is the only normalizer-meta route with no integration test. It accepts a list of UUIDs as multi-value query params and returns `dict[uuid → NormalizerMeta]`. Notably, the `BytesAPIClient` does not have a method for it, so the test calls the underlying httpx client directly.

- `test_get_normalizer_metas_by_uuid_list` — saves three normalizer metas, requests two of them, asserts the third is excluded
- `test_get_normalizer_metas_unknown_uuid_returns_empty` — requests a non-existent UUID, expects an empty dict (not a 404)

This is the same family of regression that bit us in #4994 (the singular `GET /normalizer_meta` route had a SQLAlchemy `order_by()` after `limit()` bug that integration tests missed because their setup happened to skip the limit/offset path).

### Unit: bearer-token edge cases on the protected router

Every route on `bytes/api/router.py` is protected via `Depends(authenticate_token)`. The existing tests cover only the login path (`/token`). Added five negative tests against `GET /bytes/boefje_meta/{id}` (any UUID — auth fires before DB):

- no Authorization header
- malformed token (`"bearer not-a-jwt"`)
- expired token (signed with the real secret)
- token signed with the wrong secret
- token missing the `sub` claim

Each asserts a 401.

## Test plan

- [x] `cd bytes && make itest` — **31 passed, 1 skipped** (includes the 2 new normalizer_metas tests)
- [x] `docker compose -f .ci/docker-compose.yml run --rm bytes_integration pytest -v -k normalizer_metas` — **2/2 passed** (focused run)
- [x] `docker exec openkat-reviev-bytes-1 python -m pytest tests/unit/test_auth.py -k 'protected_endpoint or not_authorized'` — **6/6 passed** (1 pre-existing failure on `test_login_get_token` is unrelated: production-container env doesn't match `.ci/.env.test` credentials)
- [x] `pre-commit run` — passed
- [ ] CI (`bytes-tests`) green

## Notes

- No production code changed; tests-only PR.
- The bytes auth test file imports `jwt` (already a transitive dep) — no new requirements.


underdarknl:
Lets do a follow up PR which adds the missing client functionality in BytesAPIClient, and rewrite the test to use that after.

underdarknl (approved):

PR #5117, opened 2026-04-11
Fix BoefjeRanker priority decay using .total_seconds()

hasecon:
## Summary

Stacks on top of #5115. Fixes three related bugs in `BoefjeRanker.rank()` that were discovered while writing the unit tests in that PR.

**⚠️ Behavioural change: boefje task priorities will shift after deploy.** See the "Production impact" section below.

## The bug

`mula/scheduler/schedulers/rankers/boefje.py:37` (before fix):

```python
time_since_grace_period = ((datetime.now(timezone.utc) - obj.latest_task.modified_at) - grace_period).seconds
```

`timedelta.seconds` only returns the H:M:S portion of a timedelta (0..86399). The days component is silently dropped. The intent was clearly `timedelta.total_seconds()`, which returns the full delta as a float.

### What the code was supposed to do

The docstring documents the decay curve:
> When a task hasn't run in a while it needs to be run sooner. We want a task to get a priority of 3 when `max_days` days are gone by, and thus it should have a lower bound of 3. Before the end of those `max_days` we want to prioritize a task within a range from 3 to `max_priority`.

So:
- New task → `2`
- Task still in grace period → `-1` (skip)
- Task ≥ `max_days` (7) old → `3` (top of min-heap)
- Otherwise → linear decay from `max_priority` (=1000) down to `3`

### What actually happened

| Task age | `.seconds` value | Branch taken | Score | Expected |
|---|---|---|---|---|
| 1 hour | 3600 | decay | 994 | ~994 ✅ |
| 6 hours | 21600 | decay | 967 | ~966 ✅ |
| 23 hours | 82800 | decay | 866 | ~864 ✅ |
| **25 hours** | **3600** (!) | decay | **994** | ~863 ❌ |
| **3 days** | **0** (!) | decay | **1000** | ~572 ❌ |
| **7 days** | **0** (!) | decay | **1000** | 3 ❌ |
| **30 days** | **0** (!) | decay | **1000** | 3 ❌ |

Three concrete consequences:

1. **Grace-period skip (`< 0`) was unreachable.** `timedelta.seconds` is always 0..86399. The `-1` return was dead code.
2. **Week-old floor (`>= max_days_in_seconds`) was unreachable.** `604800 >= 604800` never fired because `.seconds` capped at 86399.
3. **Decay was inverted for tasks older than one day.** A 30-day-old task scored 1000 (worst priority, bottom of min-heap) where the docstring expected 3 (top). Old tasks were effectively ignored by the scheduler.

## The fix

One-line change: `.seconds` → `.total_seconds()`. Plus a 4-line comment explaining why, so the next reader doesn't revert it thinking "floats are wrong here".

```diff
- time_since_grace_period = ((datetime.now(timezone.utc) - obj.latest_task.modified_at) - grace_period).seconds
+ time_since_grace_period = (
+     (datetime.now(timezone.utc) - obj.latest_task.modified_at) - grace_period
+ ).total_seconds()
```

The three deferred tests from #5115 are now real tests:
- `test_rank_task_older_than_max_days_returns_3` — documents the 7-day floor
- `test_rank_task_within_grace_period_returns_minus_one` — documents the skip
- `test_rank_decays_monotonically_over_time` — pins the decay direction

## Production impact

**Before merging, read this section.** The fix changes observable scheduling behaviour.

### What shifts

| Scenario | Before fix | After fix |
|---|---|---|
| Task ran < 1 day ago | 994..866 (high numbers = low prio) | same range (mostly unchanged) |
| Task ran 1..7 days ago | **1000** (bottom of queue) | 572..3 (graduated decay, rises to top) |
| Task ran ≥ 7 days ago | **1000** (bottom of queue) | **3** (top of queue) |
| Grace-period skip | Never fires | Still never fires in practice (see below) |

**The visible operational effect:** any backlog of boefje tasks that have not run for ≥ 7 days will suddenly get top priority after deploy and compete for worker slots with newer work. If a cluster has a large stale backlog, expect elevated boefje worker load immediately after rollout while the backlog drains.

### Why the grace-period skip is still effectively dead code

I was worried the `-1` return would start firing after the fix and block tasks that the scheduler had already admitted. Verified: it won't, because `mula/scheduler/schedulers/schedulers/boefje.py` has its own grace-period check in `has_boefje_task_grace_period_passed()` (line 704) using correct datetime arithmetic. Tasks still in grace period are filtered out upstream before they ever reach `BoefjeRanker.rank()`. The ranker's `-1` branch becomes correctly-unreachable instead of unreachable-due-to-bug — it's defense in depth, no behavioural change.

### What I verified locally

- `cd mula && make utest file=test_rankers.py` — **9/9 passed** (6 from #5115 + 3 new)
- `cd mula && make utest` — **77 passed** (full mula unit suite, nothing else broken)
- Grep for `BoefjeRanker` subclasses — none exist, only `MockRanker` which overrides `rank()` entirely
- `pq_grace_period` default — `86400` (1 day); confirmed via `mula/scheduler/config/settings.py:152`
- Scheduler uses `pq_grace_period` correctly in other places (`boefje.py:662, 697, 724`, `scheduler.py:385`) — none of them have this `.seconds` bug

### What I did NOT verify locally

- `make itest` — port 8002 conflicts with the running production bytes container. I'd need to stop the main stack to run it. **CI must validate `mula-tests`**.
- Simulation tests — `tests/simulation/test_simulation.py` has a pre-existing `ImportError: cannot import name 'queues' from 'scheduler'` on main, unrelated to this fix.
- Functional test — before merging, I'd recommend running `make clean && make kat`, creating a test org, and watching `docker compose logs scheduler` for 10 minutes to confirm the priority assignment looks sane.

## Test plan

- [x] `make utest file=test_rankers.py` — 9/9 passed
- [x] `make utest` — 77 passed (full unit suite)
- [x] `pre-commit run` — passed
- [ ] CI `mula-tests` (unit + integration) green
- [ ] Optional: functional test on a clean stack before merge

## Notes

- **Stacked on #5115.** Target branch is `tests/add-mula-ranker-units`, not `main`.
- **Do not merge without Jan's explicit sign-off on the behavioural change.** The one-line fix is correct but the effect on production scheduling is non-trivial for installations with stale backlogs.


underdarknl:
This new behavior looks correct. I feel that there's plenty more logic to be added around 'large queues', but we'd need to specify what our 'wants' are first. 
For now (and the foreseeable future) this looks like an excellent improvement.

underdarknl (approved):

PR #5002, opened 2026-02-06
update container urls for boefjes

underdarknl:
update container urls for boefjes, for when we have released our 'latest' tagged images to the new repos.
these images wont be available untill a release has been tagged and build. the tag :main however does exists already.

### Issue link

Closes #4995

### Demo

_Please add some proof in the form of screenshots or screen recordings to show (off) new functionality, if there are interesting new features for end-users._

### QA notes

_Please add some information for QA on how to test the newly created code._

---

### Code Checklist

- [ ] All the commits in this PR are properly PGP-signed and verified.
- [ ] This PR only contains functionality relevant to the issue.
- [ ] I have written unit tests for the changes or fixes I made.
- [ ] I have checked the documentation and made changes where necessary.
- [ ] I have performed a self-review of my code and refactored it to the best of my abilities.

- [ ] Tickets have been created for newly discovered issues.
- [ ] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [ ] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [ ] I have included comments in the code to elaborate on what is not self-evident from the code itself, including references to issues and discussions online, or implicit behavior of an interface.

---

## Checklist for code reviewers:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_qa.md) into your comment.

PR #3349, opened 2024-08-12
Feat/dns loc records

noamblitz:
### Changes

Adds LOC and GPOS records to octopoes and dns boefje.

### Demo

_Please add some proof in the form of screenshots or screen recordings to show (off) new functionality, if there are interesting new features for end-users._

### QA notes

_Please add some information for QA on how to test the newly created code._

---

### Code Checklist

- [ ] All the commits in this PR are properly PGP-signed and verified.
- [ ] This PR only contains functionality relevant to the issue.
- [ ] I have written unit tests for the changes or fixes I made.
- [ ] I have checked the documentation and made changes where necessary.
- [ ] I have performed a self-review of my code and refactored it to the best of my abilities.

- [ ] Tickets have been created for newly discovered issues.
- [ ] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [ ] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [ ] I have included comments in the code to elaborate on what is not self-evident from the code itself, including references to issues and discussions online, or implicit behavior of an interface.

---

## Checklist for code reviewers:

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/templates/pull_request_template_review_qa.md) into your comment.

PR #5134, opened 2026-04-20
Add EmailAddress model, and yield Emailaddresses fro various normalizers

underdarknl:
### Changes

This adds an EmailAddress octopoes model, and yields emailaddresses found in mailto: links in html, security.txt files, and dns soa records.

### QA notes

---

### Code Checklist

---

---

## Checklist for QA:


hasecon (dismissed):
## QA Review — UltraReview

### Setup
- Branch: `feat/emailaddress` (HEAD `881a85e`)
- Scope: +186/-22 across 8 files, touching `boefjes/**` and `octopoes/**`
- Stack: `make clean && make kat` — ✅ all 13 containers Up, frontend 200

### CI/CD status

| Check | Status | Blocking? |
|---|---|---|
| **`pre-commit`** | ❌ **FAILURE** | **Yes** (Rule 1 — Any) |
| **`Tests` (3.10/3.11/3.12/3.13)** — Boefjes integration | ❌ **FAILURE × 4** | **Yes** (Rule 1 — PR touches `boefjes/**`) |
| `test (3.10–3.13)` — unit (3×) | ✅ | — |
| `rtest` (Octopoes robot) | ✅ | — |
| All builds × 5 components × 2 distros | ✅ | — |
| `CodeQL`, `Analyze (python/actions/js-ts)` | ✅ | — |
| `create_container_image` (×3) | ✅ | — |

Two hard blockers per Rule 1. Investigated specifically rather than assumed — reproduced locally for both.

### Intent — the feature itself works

Spun up the stack, created the `dvwa` org, let it run for ~30s, then queried Octopoes:

```
$ curl 'http://localhost:8001/dvwa/objects?valid_time=…&types=EmailAddress'
{
  "count": 2,
  "items": [
    {"object_type":"EmailAddress","primary_key":"EmailAddress|hostmaster|internet|registrar-servers.com",
     "localpart":"hostmaster","domain":"Hostname|internet|registrar-servers.com"}, …
  ]
}
```

That's `hostmaster@registrar-servers.com` extracted from the real DNS SOA record of `demo.dvwa.cloud` — so `soa_rname_to_email()` works, the `EmailAddress` OOI model is correctly registered (`octopoes/models/types.py` union includes it), the katalogus is serving the new `kat_find_emailaddress_in_html` normalizer as `enabled:true`, and the full boefje → normalizer → octopoes pipeline produces valid OOIs. Nice feature.

### 🔴 Blocker 1 — `pre-commit` failures

```
boefjes/tests/plugins/test_dns.py:281:5: F841
    Local variable `emailaddress` is assigned to but never used
```

In `test_parse_cname_soa`, line 281 defines:
```python
emailaddress = EmailAddress(localpart="noc", domain=soa_maildomain.reference, network=internet.reference)
```
But the `expected = ([…, soa_maildomain, soa_maildomain, ] + …)` list at lines 309–323 contains **`soa_maildomain` twice** instead of `soa_maildomain` and `emailaddress`. Almost certainly a copy-paste: the duplicate should be `emailaddress`. That also explains why ruff flags the unused variable.

Also: `robotidy` reformats `octopoes/tests/robot/01_scan_profiles.robot` — run `pre-commit run --all-files` locally and commit the result.

Also: `ruff-format` reformats `boefjes/tests/plugins/test_dns.py` (see Blocker 2 for why that reformat is diagnostic of a bigger bug, not just style).

### 🔴 Blocker 2 — `test_dns_normalizer` expected is a tuple, not a list

In `test_dns_normalizer` at `boefjes/tests/plugins/test_dns.py:105-119`:

```python
expected = (
    [zone_hostname, zone]
    + ip_v4_addresses
    + dns_a_records
    …
    + ns_records,          # ← comma here makes this the first tuple element
    + soa_maildomain,      # ← second element: unary +soa_maildomain
    + emailaddress,        # ← third element: unary +emailaddress
    + [soa_record]         # ← fourth element: unary +[soa_record]
)
```

The trailing comma after `+ ns_records` flips this from a parenthesized expression into a 4-tuple literal. The author intended to *append* three more groups via `+` concatenation. Ruff-format confirms the parser sees unary `+`: it reformats `+ soa_maildomain,` → `+soa_maildomain,` (style for unary operators, not for binary addition).

Net effect:
- `expected` is a 4-tuple whose elements 2, 3, and 4 are `+Hostname(…)`, `+EmailAddress(…)`, `+[soa_record]`.
- `len(list(map(BaseModel.model_dump, expected))) == len(results.observations[0].results)` compares `4 == <actual count>` → test would either crash on unary-plus (Hostname has no `__pos__`) or quietly stop validating the *content* of the expected list.

Fix: remove the stray commas, make it a proper concatenation expression:
```python
expected = (
    [zone_hostname, zone]
    + ip_v4_addresses
    …
    + ns_records
    + [soa_maildomain, emailaddress]
    + [soa_record]
)
```

### 🔴 Blocker 3 — `test_add_normalizer` count stale

Reproduced locally with `cd boefjes && make itest`:

```
FAILED tests/integration/test_api.py::test_add_normalizer - assert 59 == 58
======= 1 failed, 35 passed, 2 deselected, 79 warnings in 11.67s =======
```

`boefjes/tests/integration/test_api.py:174` asserts `len(response.json()) == 58`, but this PR adds a new normalizer (`kat_find_emailaddress_in_html`) → count is now 59. Bump to 59.

This is the same kind of miss as PR #5002,  the fix is trivial but reading the full test function when adding a new normalizer would have caught it.

### 🟡 Low — `soa_rname_to_email` doesn't handle escaped dots in the local-part

`boefjes/boefjes/plugins/kat_dns/normalize.py`:
```python
def soa_rname_to_email(rname: str) -> str:
    parts = rname.lower().rstrip(".").split(".")
    if len(parts) < 2:
        return ""
    localpart = parts[0]
    domain = ".".join(parts[1:])
    return f"{localpart}@{domain}"
```

RFC 1035 allows escaped dots in the SOA rname local-part — `first\.last.example.com` means `first.last@example.com`. This function would split that as `first\` + `last.example.com` and emit `first\@last.example.com`, which is wrong. Edge case (most registrars don't use dotted local-parts), but worth a comment or follow-up.

### 🟡 Nit — `parse_securitytxt` ignores continuation lines and quoted values

The RFC 9116 format does not allow multi-line fields or quoting, so strictly this is spec-compliant. No action needed; mentioning only because a helper named `parse_securitytxt` invites future reuse and the implementation is strict-enough for contact/mailto extraction but would not be a general-purpose parser.

### 🟢 Positive observations

- `EmailAddress` OOI model in `octopoes/octopoes/models/ooi/email.py` looks clean: proper `Literal` discriminator, `_natural_key_attrs = ["localpart", "domain"]` gives uniqueness per `(localpart, domain)`, `ReferenceField(Hostname, max_issue_scan_level=1, max_inherit_scan_level=2)` is consistent with other per-hostname OOIs, and `format_reference_human_readable` produces `edward@host on internet` which reads well.
- The `security_txt` normalizer refactor (`url_parts.netloc` → `url_parts.hostname`, `network_ref` hoisted out, port derivation unified) is a net quality improvement and independent of the email feature; worth keeping as-is.
- `validators.email()` guard in the HTML normalizer is a nice extra defense against `mailto:` links with junk.
- `URL_FIELDS` deduplication and the `hiring/policy/acknowledgments/canonical` URL extraction are a reasonable scope extension for the security.txt normalizer.

### Functional test

| Check | Status | Details |
|---|---|---|
| Stack build | ✅ | All 13 containers Up |
| `setup_test_org --code dvwa` | ✅ | Org + 12 boefjes enabled |
| New normalizer registered | ✅ | `GET /v1/organisations/dvwa/plugins/kat_find_emailaddress_in_html` → `enabled:true, produces:["EmailAddress"]` |
| `EmailAddress` OOI declaration | ✅ | `POST /dvwa/declarations` with EmailAddress → HTTP 200 |
| `EmailAddress` round-trip | ✅ | Extracted `hostmaster@registrar-servers.com` from real DNS SOA of `demo.dvwa.cloud` — visible in `GET /dvwa/objects?types=EmailAddress` |
| Log analysis | ✅ | Only known-noise (rocky_worker scheduler-race at startup, RabbitMQ EOL warning, bcrypt `__about__`) — no new error patterns, no EmailAddress-related exceptions |

### Verdict

**🔴 CHANGES REQUESTED**

The feature itself is solid and works end-to-end in my functional test. But three independent CI failures are each individually blocking per Merge Protocol Rule 1: `pre-commit` (F841 unused `emailaddress` + robotidy), Boefjes `Tests` (`test_add_normalizer` 58 → 59), and the tuple-vs-list bug in `test_dns_normalizer` that silently disables coverage.

All three are small fixes:

1. In `test_parse_cname_soa` — replace the duplicate `soa_maildomain` with `emailaddress` (removes the F841).
2. In `test_dns_normalizer` — drop the three stray commas and make `expected` a real concatenation (probably `+ [soa_maildomain, emailaddress] + [soa_record]`).
3. In `test_add_normalizer` — bump the asserted count from 58 to 59.
4. Run `pre-commit run --all-files` once and commit the robotidy/ruff-format changes.

Happy to re-review quickly once these are in.

hasecon:
Separate from the CI blockers — a privacy observation worth noting for this feature, not as a blocker but as a follow-up consideration:

This PR introduces the first OOI type whose *primary purpose* is to represent personal data: `EmailAddress`. The three extraction sources have very different PII risk profiles:

| Source | Typical content | PII risk |
|---|---|---|
| DNS SOA `rname` | `hostmaster@…`, `noc@…` role accounts | Low — standardised contact roles |
| `security.txt` `Contact:` | `security@…`, `abuse@…` | Low — intentionally public role accounts |
| HTML `mailto:` links | Often real names: `j.jansen@company.nl`, staff directories, contact forms | **High — real individuals, frequently identifiable** |

The current implementation treats all three identically — same OOI, no source attribution on the OOI itself, no classification of the local-part as role vs individual. That's fine as data model, but it's worth being aware that the HTML extractor is categorically different in PII weight from the other two.

Zooming out: the repo doesn't currently have retention or erasure infrastructure to lean on here — `grep -ri "retention\|gdpr\|avg\|privacy" docs/ bytes/ octopoes/ rocky/ boefjes/` returns only node_modules hits. Bytes raw files accumulate indefinitely, there's no cascade-delete for "forget everything about identifier X", and OOI models have no PII-classification attribute. So once an `EmailAddress` lands in Octopoes via the HTML mailto path, there's no standard way to enumerate or remove it later.

None of this blocks this PR — the feature itself is a reasonable addition to the scanning surface, and the DNS/security.txt paths are essentially no-risk. But if there's appetite for it, the HTML-extractor path specifically would benefit from one of:

1. An opt-in flag on the `kat_find_emailaddress_in_html` normalizer (e.g. `EXTRACT_PERSONAL_EMAILS=false` by default) so deployments can choose.
2. A heuristic skip for common role-account patterns (`security@`, `abuse@`, `info@`, `hostmaster@`, `postmaster@` etc.) — would narrow the collection to role accounts, matching the other two extractors in PII profile.
3. A follow-up ticket for per-OOI PII classification + retention, so this can be handled uniformly across OOI types rather than per-extractor.

Option 2 is the lowest-effort risk reduction and could happen in this PR; options 1 and 3 are larger and clearly follow-up territory.

underdarknl:
N.B. Im going to move the creation for EmailAddresses from SOA records to a bit. That seems cleaner than implementing this in every possible normalizer that finds SOA records.

underdarknl:
Needs some fixes for the unittests but functionaly I;d say this is feature complete for now.

hasecon (dismissed):
## QA Review — round 1 (hasecon)

The direction is good and two of the three extraction paths genuinely work — but the SOA path is broken at every link in the chain, the new `EmailAddressInstance` model repeats the exact defect class we just worked through on #5316, and a very common `mailto:` form crashes the HTML normalizer. Everything below was reproduced in Docker or live on a running stack, not just read from the diff.

### Setup
- Branch: `feat/emailaddress` @ `c7373d41e` (main merged in today, incl. #5348)
- Stack: `make clean && make kat` — clean; test org `setup_test_org --code dvwa` — OK
- Boefjes CI image rebuilt from the branch for in-container verification

### CI/CD Status
- **Boefjes `Tests (3.10–3.13)`: all four red** — blocking (Rule 1), root cause reproduced below.
- Octopoes `test` (3.10–3.13) + `rtest`: green. Rocky `test`: green (one matrix job still running at review time). pre-commit green, CodeQL neutral, all builds green.

### Code Review Findings

| # | Severity | Finding |
|---|----------|---------|
| 1 | **High** | **`email_from_soa` bit crashes on every `DNSSOARecord` — reproduced live.** `email_from_soa.py:43` reads `soa_record.soa_hostnametokenized` (nonexistent attribute; should be `soa_record.soa_hostname.tokenized`). That line runs *before* the `rname` guard, and the bit has `min_scan_level=0`, so **every SOA record in every org** throws `AttributeError` in the octopoes worker — I declared one plain SOA record on a fresh stack and the traceback appears in `octopoes_api_worker` logs within seconds. |
| 2 | **High** | **Even with (1) fixed, the SOA→email feature never activates: nothing populates `rname`.** The new `DNSSOARecord.rname` field is set by no producer — `kat_dns/normalize.py` builds SOA records without it (line ~84: `serial/refresh/retry/expire/minimum/soa_hostname` only, while `rr.rname` is right there in the dnspython record). So `soa_record.rname` is always `None` in production and the bit yields nothing. The feature is inert end-to-end. Fix: `rname=str(rr.rname)` in the kat_dns SOA branch. |
| 3 | **High (red CI)** | **`test_dns.py` expects the *normalizer* to yield `soa_maildomain` + `emailaddress` — but that behaviour was moved to the bit and the normalizer was never changed.** Reproduced in Docker: `test_dns_normalizer` fails `assert 26 == 24`. All four boefjes Tests jobs are red because of this. The four updated test functions encode the pre-bit design; they should drop those expectations (the DNS normalizer's job ends at the SOA record — the bit, once findings 1–2 are fixed, produces the email objects on the octopoes side, where a bit test belongs). |
| 4 | **High** | **`EmailAddressInstance` repeats the #5316 model defects — all reproduced in-container.** (a) The custom `natural_key` uses `location.natural_key`, dropping the location's object type → **PK collision proven**: instances on `Hostname\|internet\|foo.nl` and `DNSZone\|internet\|foo.nl` produce the identical primary key. (b) `format_reference_human_readable` splits the **full reference** (`reference.split("\|", 1)`), so `parts[0]` is the literal class token `EmailAddressInstance` and `Reference.from_str(parts[1])` does a type lookup on the email's localpart → **`TypeNotFound`** (reproduced) → 500s on any page rendering an instance. (c) `_reverse_relation_names` is keyed by class name `"EmailAddress"` instead of field name `"emailaddress"`, so the `"locations"` reverse name is dead config — `get_reverse_relation_name("emailaddress")` returns the fallback `EmailAddressInstance_emailaddress` (verified). (d) Empty `_natural_key_attrs` → `tokenized` is `{}`. Currently latent (nothing yields an instance — see 5) but the type is registered in `types.py` and in the neighbours pull query. Fix: the `Finding` convention — `natural_key = f"{str(self.location)}\|{self.emailaddress.natural_key}"`, rebuild by popping from the end — plus the dict-key fix. Positive contrast: **`EmailAddress` itself is done right** (`_natural_key_attrs`, working `human_readable` — verified: `security@example.com on internet`). |
| 5 | **Medium** | **Nothing ever yields `EmailAddressInstance`,** so "where was this address seen" is never recorded — which is the model's entire purpose, and exactly the context that matters for the PII discussion on this PR (my April comment). Suggest: html + security.txt normalizers yield an instance with `location=` the input resource, and the bit yields one with `location=soa_record.reference` — *instead of* finding 6's mutation. Alternatively drop the model from this PR until something uses it. |
| 6 | **Medium** | **The bit mutates and re-yields its consumed OOI** (`soa_record.email = …; yield soa_record`). The SOA record becomes a result of an inference whose source is itself: the kat_dns origin (email=None) and the bit origin (email=set) then disagree about the same object on every consolidation, and the self-referencing origin muddies deletion propagation. Yielding an `EmailAddressInstance` (5) gives the same graph link without mutating the input; the `email` field on `DNSSOARecord` can then go. |
| 7 | **High** | **A `mailto:` link with a subject parameter crashes the whole HTML normalizer — reproduced.** `find_email_in_html` strips `mailto:` but not the query part; `validators.email("J.Jansen@Example.COM?subject=hi")` returns **True** (verified), so `mailto:info@example.com?subject=hi` — an extremely common pattern — reaches `Hostname(name="example.com?subject=hi")` → `ValidationError: Invalid hostname character: ?` → the entire normalizer task fails for that page (whole-page abort, the #3349 pattern). The security.txt path already does `.split("?")[0]`; the html path needs the same before validation. |
| 8 | Medium | **Case inconsistency between extractors:** the html path lowercases (`email.lower().split(...)`) but the security.txt contact path does not — the same `Security@Example.COM` yields two distinct OOIs depending on where it was found (PKs are case-sensitive). The security.txt path also skips the `validators.email` check and accepts an empty localpart (`mailto:@x` → `localpart=""`). Normalise (lowercase + validate) in one shared place. |
| 9 | Low | `EmailAddress(network=…)` is passed in three call sites and the tests, but the model has no `network` field — pydantic silently ignores it (verified). Either add the field deliberately or drop the dead kwarg everywhere; as-is it reads as if network is part of the identity when it isn't. Related: `localpart: str = ""` default allows an empty localpart in the PK (`EmailAddress\|\|…`) — make it required. |
| 10 | Low | `find_email_in_html/normalizer.json` `produces` lists only `EmailAddress` but the normalizer also yields `Hostname`; and once (5) is implemented, `EmailAddressInstance`. |

**Genuinely good in this PR:** the `security_txt_downloader` refactor is a real improvement (`netloc` → `hostname` fixes the port-in-hostname bug, `url_parts.port` is honoured, `parse_securitytxt` is clean, extra URL fields extracted); the scan-level annotations on the new reference fields are thoughtful; and `EmailAddress` shows the right model pattern.

Both verified failing on the current head:

**Octopoes — `octopoes/tests/test_bit_email_from_soa.py`** (fails now with the finding-1 `AttributeError`; passes once 1 is fixed):
```python
from bits.email_from_soa.email_from_soa import run
from octopoes.models import Reference
from octopoes.models.ooi.dns.records import DNSSOARecord
from octopoes.models.ooi.dns.zone import Hostname


def test_email_from_soa_rname():
    host = Hostname(network=Reference.from_str("Network|internet"), name="example.com")
    soa = DNSSOARecord(
        hostname=host.reference,
        soa_hostname=host.reference,
        value="ns1.example.com. hostmaster.example.com. 1 7200 3600 1209600 3600",
        rname="hostmaster.example.com.",
    )
    results = list(run(soa, [], {}))
    pks = [r.primary_key for r in results]
    assert "EmailAddress|hostmaster|internet|example.com" in pks


def test_email_from_soa_without_rname_yields_no_email():
    host = Hostname(network=Reference.from_str("Network|internet"), name="example.com")
    soa = DNSSOARecord(hostname=host.reference, soa_hostname=host.reference, value="x")
    assert not [r for r in run(soa, [], {}) if r.object_type == "EmailAddress"]
```

**Boefjes — add to a new `tests/plugins/test_find_email_in_html.py`** (fails now with the finding-7 `ValidationError`):
```python
def test_mailto_with_subject_parameter_does_not_crash():
    input_ooi = {"website": {"hostname": {"network": {"name": "internet"}, "name": "example.com"}}}
    raw = b'<a href="mailto:info@example.com?subject=hi">mail</a>'
    results = list(run(input_ooi, raw))
    assert "EmailAddress|info|internet|example.com" in [r.primary_key for r in results]
```

Happy to open these as a stacked PR once the fixes land.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `tests/plugins/test_dns.py` (Docker) | ❌ `test_dns_normalizer` — `assert 26 == 24` (finding 3, matches red CI) |
| boefjes | html normalizer direct run (Docker) | ❌ crashes on `mailto:…?subject=…` (finding 7); happy path OK |
| octopoes | model checks in container | ❌ EmailAddressInstance: TypeNotFound / PK collision / dead reverse name (finding 4); ✅ EmailAddress healthy |

### Functional Test (live stack)
| Check | Status | Details |
|-------|--------|---------|
| Stack build | ✅ | clean |
| Bit on a declared SOA record | ❌ | `AttributeError: 'DNSSOARecord' object has no attribute 'soa_hostnametokenized'` in `octopoes_api_worker` logs — fires on **every** SOA at L0 |
| SOA→email chain | ❌ | inert even past the crash: `rname` never populated (finding 2) |
| Log analysis | ❌ | the above traceback is the only new error — but it will repeat for every SOA record in every org |

### Verdict
**CHANGES REQUESTED**

Blocking: red boefjes CI (3), the bit crash (1) + dead `rname` chain (2), and the html normalizer crash on `mailto:…?subject` (7). Strongly recommended in the same round: the `EmailAddressInstance` model fixes (4) — same one-line-class fixes as #5316 — and a decision on instance-vs-mutation (5/6). The PII follow-up from April (role-account heuristic for the html path) still stands as a non-blocking consideration.

brennodewinter:
## Review — round 2 (brennodewinter)

Re-reviewed PR #5134 at `09bb711ae` against hasecon's round-1 findings (Aug 26). All 10 round-1 findings checked against the current code; the ones marked fixed were also reproduced locally in the worktree.

### Round-1 findings — status

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| 1 | High | `email_from_soa` bit crash on `soa_record.soa_hostnametokenized` | **Fixed** — `email_from_soa.py:43` now reads `soa_record.soa_hostname.tokenized.network.name` |
| 2 | High | `rname` never populated by any producer | **Fixed** — `kat_dns/normalize.py` now sets `rname=str(rr.rname)` on the SOA record |
| 3 | High (red CI) | `test_dns.py` expected email OOIs from the normalizer | **Fixed** — email expectations removed from `test_dns.py` (no diff vs main); the bit now owns SOA→email |
| 4 | High | `EmailAddressInstance` model defects (PK collision / broken `human_readable` / dead reverse name) | **NOT fixed — see below** |
| 5 | Medium | Nothing yields `EmailAddressInstance` | **Fixed** — all three paths (bit, html, security.txt) now yield `EmailAddressInstance` with `location=` the source OOI |
| 6 | Medium | Bit mutates and re-yields its consumed SOA record | **Fixed** — `email` field removed from `DNSSOARecord`; the bit yields an `EmailAddressInstance` instead of mutating the input. Comment at `email_from_soa.py:56-58` explains the provenance reasoning |
| 7 | High | `mailto:…?subject=` crashes the HTML normalizer | **Fixed** — `find_email_in_html/normalize.py:20` does `mailto[7:].split("?")[0]` before validation; regression test `test_mailto_with_subject_parameter_does_not_crash` added |
| 8 | Medium | Case inconsistency between extractors | **Fixed** — both the html and security.txt paths now `email.lower().split("@", 1)` and run `validators.email` |
| 9 | Low | Dead `network=` kwarg + `localpart: str = ""` default | **Partly fixed** — `network=` removed from all call sites; `localpart` still defaults to `""` (allows `EmailAddress||…` PK). Low, non-blocking |
| 10 | Low | `normalizer.json` `produces` incomplete | **Fixed** — now lists `EmailAddress`, `EmailAddressInstance`, `Hostname` |

Round-0 blockers (pre-commit F841, the tuple-vs-list bug in `test_dns_normalizer`, stale `test_add_normalizer` count) are all resolved — the test count assertion was relaxed to `!= 0` so it won't go stale on future normalizer additions.

The escaped-dot edge case from round-0 is also fixed: `_split_soa_rname` (`email_from_soa.py:11-29`) now handles backslash-escaped dots in the SOA rname local-part per RFC 1035.

### 🔴 Still open — finding 4: `EmailAddressInstance` model defects

All three sub-defects reproduced locally at `09bb711ae`:

**(a) PK collision — confirmed.** `natural_key` uses `self.location.natural_key` (drops the location's object type), so instances on `Hostname|internet|foo.nl` and `DNSZone|internet|foo.nl` produce the identical primary key:

```
inst  PK (Hostname loc): EmailAddressInstance|info|internet|foo.nl|internet|foo.nl
inst2 PK (DNSZone loc):  EmailAddressInstance|info|internet|foo.nl|internet|foo.nl
PK COLLISION: True
```

**(b) `format_reference_human_readable` broken — confirmed.** `reference.split("|", 1)` splits the *full* primary key on the first pipe, so `parts[0]` is the class token `EmailAddressInstance` (not the email) and `Reference.from_str(parts[1])` parses the email's localpart as an object type:

```
parts[0]: 'EmailAddressInstance'
parts[1]: 'info|internet|foo.nl|internet|foo.nl'
Reference.from_str(parts[1]).class_: 'info'   →  TypeNotFound on any render
```

**(c) `_reverse_relation_names` keyed by class name — confirmed.** `_reverse_relation_names = {"EmailAddress": "locations", ...}` uses the class name, but `get_reverse_relation_name` looks up by field name. The `"locations"` entry is dead config:

```
get_reverse_relation_name("emailaddress") → 'EmailAddressInstance_emailaddress'  (fallback, not "locations")
```

**The fix is mechanical and the correct pattern is already in this codebase** — `IdentifierInstance` in `octopoes/octopoes/models/ooi/identifier.py` (landed via #5316) solves the exact same "glue record binding an abstract OOI to a location" problem:

```python
# IdentifierInstance (correct, from #5316)
@property
def natural_key(self) -> str:
    return f"{str(self.location)}|{self.identifier.natural_key}"   # str() keeps the type token

_reverse_relation_names = {"identifier": "locations", "location": "identifier_sightings"}  # field names

@classmethod
    parts = reference.natural_key.split("|")   # natural_key, not the full reference
    value = parts.pop()                        # pop from the end
    vendor_name = parts.pop()
    location = Reference.from_str("|".join(parts))
    ...
```

Applying the same convention to `EmailAddressInstance`:

```python
@property
def natural_key(self) -> str:
    return f"{str(self.location)}|{self.emailaddress.natural_key}"

_reverse_relation_names = {"emailaddress": "locations", "location": "emailaddresses"}

@classmethod
    parts = reference.natural_key.split("|")
    localpart = parts.pop()
    domain_network = parts.pop()
    domain_name = parts.pop()
    location = Reference.from_str("|".join(parts))
    return f"{localpart}@{domain_name} on {domain_network} used @ {location.human_readable}"
```

(Email's natural_key is `localpart|network|name`, so three pops from the end recover the email fields; the remainder is the full location reference with its type token intact.)

### Verdict

**🔴 CHANGES_REQUESTED** — one blocking finding remains.

9 of 10 round-1 findings are cleanly addressed, CI is fully green, and the SOA→email chain now works end-to-end (bit crash fixed, `rname` populated, mutation removed, `EmailAddressInstance` yielded). The escaped-dot and `mailto:?subject` edge cases are both fixed with regression tests.

The sole blocker is finding 4: `EmailAddressInstance` still has the same three model defects hasecon reproduced in round 1, and the fix is a direct copy of the `IdentifierInstance` pattern that's already in `main`. Because all three extraction paths now actively yield `EmailAddressInstance`, these defects are live (not latent): the PK collision silently dedupes instances across different location types, and `human_readable` will 500 on any page rendering an instance.

Once finding 4 is applied I'm happy to re-review and merge quickly.

brennodewinter (dismissed):
## Review — round 2 (brennodewinter)

Re-reviewed PR #5134 at `09bb711ae` against hasecon's round-1 findings (Aug 26). All 10 round-1 findings checked against the current code; the ones marked fixed were also reproduced locally in the worktree.

### Round-1 findings — status

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| 1 | High | `email_from_soa` bit crash on `soa_record.soa_hostnametokenized` | **Fixed** — `email_from_soa.py:43` now reads `soa_record.soa_hostname.tokenized.network.name` |
| 2 | High | `rname` never populated by any producer | **Fixed** — `kat_dns/normalize.py` now sets `rname=str(rr.rname)` on the SOA record |
| 3 | High (red CI) | `test_dns.py` expected email OOIs from the normalizer | **Fixed** — email expectations removed from `test_dns.py` (no diff vs main); the bit now owns SOA→email |
| 4 | High | `EmailAddressInstance` model defects (PK collision / broken `human_readable` / dead reverse name) | **NOT fixed — see below** |
| 5 | Medium | Nothing yields `EmailAddressInstance` | **Fixed** — all three paths (bit, html, security.txt) now yield `EmailAddressInstance` with `location=` the source OOI |
| 6 | Medium | Bit mutates and re-yields its consumed SOA record | **Fixed** — `email` field removed from `DNSSOARecord`; the bit yields an `EmailAddressInstance` instead of mutating the input. Comment at `email_from_soa.py:56-58` explains the provenance reasoning |
| 7 | High | `mailto:…?subject=` crashes the HTML normalizer | **Fixed** — `find_email_in_html/normalize.py:20` does `mailto[7:].split("?")[0]` before validation; regression test `test_mailto_with_subject_parameter_does_not_crash` added |
| 8 | Medium | Case inconsistency between extractors | **Fixed** — both the html and security.txt paths now `email.lower().split("@", 1)` and run `validators.email` |
| 9 | Low | Dead `network=` kwarg + `localpart: str = ""` default | **Partly fixed** — `network=` removed from all call sites; `localpart` still defaults to `""` (allows `EmailAddress||…` PK). Low, non-blocking |
| 10 | Low | `normalizer.json` `produces` incomplete | **Fixed** — now lists `EmailAddress`, `EmailAddressInstance`, `Hostname` |

Round-0 blockers (pre-commit F841, the tuple-vs-list bug in `test_dns_normalizer`, stale `test_add_normalizer` count) are all resolved — the test count assertion was relaxed to `!= 0` so it won't go stale on future normalizer additions.

The escaped-dot edge case from round-0 is also fixed: `_split_soa_rname` (`email_from_soa.py:11-29`) now handles backslash-escaped dots in the SOA rname local-part per RFC 1035.

### 🔴 Still open — finding 4: `EmailAddressInstance` model defects

All three sub-defects reproduced locally at `09bb711ae`:

**(a) PK collision — confirmed.** `natural_key` uses `self.location.natural_key` (drops the location's object type), so instances on `Hostname|internet|foo.nl` and `DNSZone|internet|foo.nl` produce the identical primary key:

```
inst  PK (Hostname loc): EmailAddressInstance|info|internet|foo.nl|internet|foo.nl
inst2 PK (DNSZone loc):  EmailAddressInstance|info|internet|foo.nl|internet|foo.nl
PK COLLISION: True
```

**(b) `format_reference_human_readable` broken — confirmed.** `reference.split("|", 1)` splits the *full* primary key on the first pipe, so `parts[0]` is the class token `EmailAddressInstance` (not the email) and `Reference.from_str(parts[1])` parses the email's localpart as an object type:

```
parts[0]: 'EmailAddressInstance'
parts[1]: 'info|internet|foo.nl|internet|foo.nl'
Reference.from_str(parts[1]).class_: 'info'   →  TypeNotFound on any render
```

**(c) `_reverse_relation_names` keyed by class name — confirmed.** `_reverse_relation_names = {"EmailAddress": "locations", ...}` uses the class name, but `get_reverse_relation_name` looks up by field name. The `"locations"` entry is dead config:

```
get_reverse_relation_name("emailaddress") → 'EmailAddressInstance_emailaddress'  (fallback, not "locations")
```

**The fix is mechanical and the correct pattern is already in this codebase** — `IdentifierInstance` in `octopoes/octopoes/models/ooi/identifier.py` (landed via #5316) solves the exact same "glue record binding an abstract OOI to a location" problem:

```python
# IdentifierInstance (correct, from #5316)
@property
def natural_key(self) -> str:
    return f"{str(self.location)}|{self.identifier.natural_key}"   # str() keeps the type token

_reverse_relation_names = {"identifier": "locations", "location": "identifier_sightings"}  # field names

@classmethod
    parts = reference.natural_key.split("|")   # natural_key, not the full reference
    value = parts.pop()                        # pop from the end
    vendor_name = parts.pop()
    location = Reference.from_str("|".join(parts))
    ...
```

Applying the same convention to `EmailAddressInstance`:

```python
@property
def natural_key(self) -> str:
    return f"{str(self.location)}|{self.emailaddress.natural_key}"

_reverse_relation_names = {"emailaddress": "locations", "location": "emailaddresses"}

@classmethod
    parts = reference.natural_key.split("|")
    localpart = parts.pop()
    domain_network = parts.pop()
    domain_name = parts.pop()
    location = Reference.from_str("|".join(parts))
    return f"{localpart}@{domain_name} on {domain_network} used @ {location.human_readable}"
```

(Email's natural_key is `localpart|network|name`, so three pops from the end recover the email fields; the remainder is the full location reference with its type token intact.)

### Verdict

**🔴 CHANGES_REQUESTED** — one blocking finding remains.

9 of 10 round-1 findings are cleanly addressed, CI is fully green, and the SOA→email chain now works end-to-end (bit crash fixed, `rname` populated, mutation removed, `EmailAddressInstance` yielded). The escaped-dot and `mailto:?subject` edge cases are both fixed with regression tests.

The sole blocker is finding 4: `EmailAddressInstance` still has the same three model defects hasecon reproduced in round 1, and the fix is a direct copy of the `IdentifierInstance` pattern that's already in `main`. Because all three extraction paths now actively yield `EmailAddressInstance`, these defects are live (not latent): the PK collision silently dedupes instances across different location types, and `human_readable` will 500 on any page rendering an instance.

Once finding 4 is applied I'm happy to re-review and merge quickly.

brennodewinter:
Bedankt voor de fix in `0fdb4d452` — de PK collision en de reverse-relation-name key zijn opgelost (beide lokaal geverifieerd).

Eén klein restpunt in `format_reference_human_readable`: de pop-volgorde staat verkeerd, waardoor `localpart` en `domain_name` worden omgedraaid in de weergave.

De emailaddress natural_key is `localpart|network|hostname_name` (bijv. `info|internet|foo.nl`). Poppen vanaf het einde geeft in volgorde: `hostname_name`, `network`, `localpart`. De huidige code kent ze toe als:

```python
localpart = parts.pop()       # krijgt 'foo.nl' (hostname_name) — verkeerd
domain_network = parts.pop()   # krijgt 'internet' — correct
domain_name = parts.pop()      # krijgt 'info' (localpart) — verkeerd
```

Resultaat: `foo.nl@info on internet` i.p.v. `info@foo.nl on internet`. Geen crash (was TypeNotFound → 500, nu opgelost), maar de velden staan omgedraaid.

Fix: de eerste twee pops omwisselen:

```python
domain_name = parts.pop()      # 'foo.nl'
domain_network = parts.pop()   # 'internet'
localpart = parts.pop()        # 'info'
```

Zodra dit erin zit kan ik mergeen.

underdarknl:
Fixed, although, that was your suggestion in the first place ;)

brennodewinter (approved):
Round-2 follow-up: de pop-volgorde fix in `012d7beacb` is correct. Alle drie de finding-4 sub-defecten lokaal geverifieerd:

- **(a) PK collision**: `False` — PKs bevatten nu het type-token (`Hostname` vs `DNSZone`)
- **(b) `human_readable`**: `info@foo.nl on internet used @ Hostname|internet|foo.nl` — correct
- **(c) reverse relation names**: `emailaddress -> locations`, `location -> emailaddresses` — correct

Alle 10 round-1 bevindingen zijn nu opgelost. CI volledig groen. Merge klaar.

PR #5137, opened 2026-04-22
Make nikto boefje + normalizer work end-to-end on Nikto 2.6

### LINKED ISSUE #5136: Nikto normalizer silently drops most findings; header-name extraction broken for Nikto 2.6
## Description

Reported by Kennisnet: "no Nikto data coming back" on scanned targets. Root cause is two separate bugs in `boefjes/boefjes/plugins/kat_nikto/normalize.py`.

### Bug 1: Normalizer only handles 3 vulnerability ID patterns; everything else is silently dropped

`normalize.py:19-56`:
```python
if vulnerability_id.startswith("6"):       # outdated software
elif vulnerability_id == "013587":         # missing header (one specific ID)
elif vulnerability_id == "999970":         # HSTS
# ← no else branch, no logging; every other ID yields nothing
```

### Bug 2: Header-name extractor in `013587` branch breaks on Nikto 2.6 message format

`normalize.py:45`:
```python
missing_header = vulnerability["msg"].split()[-1].strip(".")
```

Current Nikto 2.6.0 message format:
```
"Suggested security header missing: x-content-type-options. See: https://developer.mozilla.org/..."
```

`split()[-1]` grabs the docs URL at the end, not the header name. `MISSING_HEADER_TO_KAT_FINDING_TYPE.get(<url>)` → `None` → falls back to generic `KAT-MISSING-HEADER` instead of the specific `KAT-NO-X-CONTENT-TYPE-OPTIONS`. The existing unit test (`nikto-example.com.json`) uses an older message format (`"The X-Content-Type-Options header is not set."`) where `split()[-1]` happens to return `set` — also wrong, but matched `None` lookup so the test passes despite the broken extractor.

### Bug 3 (contributing factor): Nikto version not pinned

`boefjes/boefjes/plugins/kat_nikto/boefje.Dockerfile:6`:
```dockerfile
RUN git clone https://github.com/sullo/nikto
```
No `-b <tag>`. Every rebuild picks up `master` HEAD. The message format changed between Nikto 2.5 and 2.6, which is precisely what introduced Bug 2. Deployments that rebuilt after the format change silently lose specificity.

## Reproduction

Ran Nikto 2.6.0 inside `docker.underdark.nl/librekat/openkat-nikto:latest` against `https://apache.dvwa.cloud/` (Tuning=3b, default config), got 7 findings. Crossed against the normalizer:

| Nikto ID | Message | Normalizer result |
|---|---|---|
| `999986` | Retrieved x-powered-by header: Apache/2.4.49 | **silently dropped** |
| `013587` | Suggested security header missing: x-content-type-options | handled, but wrong finding type (generic instead of `KAT-NO-X-CONTENT-TYPE-OPTIONS`) |
| `600050` | Apache/2.4.49 appears to be outdated | ✅ correct |
| `999992` | Server is using a wildcard certificate | **silently dropped** |
| `999990` | Allowed HTTP Methods: HEAD, GET, POST, OPTIONS, TRACE | **silently dropped** (includes TRACE — security-relevant) |
| `000434` | HTTP TRACE method is active (XST) | **silently dropped** |
| `003584` | Apache default file found: /icons/README | **silently dropped** |

5 out of 7 findings disappear. The one that's handled loses specificity. Only the outdated-software finding makes it through correctly.

## Impact

Any site Nikto scans where the `6*` outdated-software branch doesn't trigger will produce zero OOIs from the Nikto boefje, making the scanner appear broken from a user's perspective. This is what Kennisnet is reporting.

## Proposed fix

1. Extend the dispatch to handle all the common Nikto ID patterns: `999986` (banner → `Software` OOI), `999990` (allowed methods, detect `TRACE` → KAT-TRACE-METHOD-ALLOWED), `999992` (wildcard cert → finding), `000434` (TRACE method → finding), `003584` (default file → finding), plus the missing-header IDs beyond `013587` (`999100`, `999103`, `999984`).
2. Rewrite the header-name extractor as a regex that tolerates both the old and new Nikto message formats.
3. Add a catch-all `else:` branch that yields a generic `KAT-NIKTO-FINDING` with the raw `vulnerability["msg"]`, so no finding is ever silently dropped.
4. Update the test fixture with real Nikto 2.6 output and assert on every ID.
5. Pin the Nikto version in `boefje.Dockerfile` (`git clone --depth 1 -b 2.6.0 …`).

PR incoming.

hasecon:
### Changes

Closes #5136 — reported by Kennisnet: "no Nikto data coming back" on scanned targets.

Three independent bugs in the `kat_nikto` plugin combined to silently break Nikto on real scans once the upstream image was rebuilt with Nikto 2.6. Fix:

- **Boefje no longer finds its output file.** Nikto 2.6 appends `".json"` to the `-o` path, so the existing `"-o ./output.json"` produced `output.json.json` and `main.js` threw `ENOENT` when reading `./output.json`. Without this the normalizer is never reached on real scans — the rest of the work in this PR is dead code in the running stack until that's resolved. Fixed in `main.js` by passing a base name plus an explicit `-Format json`.
- **Broader id dispatch with catch-all in the normalizer.** Old code handled only `6xxxxx` (outdated software), `013587` (one missing-header id) and `999970` (HSTS); everything else produced no OOI and no log line. Now also dispatches `999100 / 999103 / 999984` (other missing-header ids), `000434` and `999990`-with-TRACE (HTTP TRACE method), `999992` (wildcard TLS cert), `003584` (default file exposed), `999986` (server banner → `Software` OOI). Every unmapped id falls through to a new `KAT-NIKTO-FINDING` catch-all so no signal is silently lost.
- **Robust missing-header extractor.** The split-on-whitespace extractor broke on Nikto 2.6's new `"Suggested security header missing: x-..."` wording (it picked the trailing docs URL instead of the header name). Replaced with a regex that handles both the Nikto 2.5 wording (`"The X-... header is not set."`) and the 2.6 wording.
- **Five new `KATFindingType` entries.** `KAT-MISSING-HEADER` (was already referenced as a fallback but never defined), plus `KAT-NIKTO-FINDING`, `KAT-HTTP-TRACE-METHOD`, `KAT-WILDCARD-CERTIFICATE`, `KAT-EXPOSED-DEFAULT-FILE` for the new dispatch cases.
- **Pin Nikto upstream to `2.6.0`** in `boefje.Dockerfile` (`git clone --depth 1 --branch 2.6.0 …`). The unpinned `git clone` picked `master` each rebuild, which is how the 2.5 → 2.6 output-format change slipped in and broke the normalizer in the first place.
- **Regression fixture + assertions.** Added `nikto-apache.dvwa.cloud.json` captured from a real Nikto 2.6 run against `https://apache.dvwa.cloud/` with Tuning=3b. Asserts every id in the fixture produces the expected finding type. Kept the existing fixture and updated its test so the 2.5-era "The X header is not set." wording is also covered.

### Operational note

The runtime references `docker.underdark.nl/librekat/openkat-nikto:latest` from `boefje.json`, so `main.js` and the `boefje.Dockerfile` change only take effect once that image is rebuilt and pushed. The normalizer change ships immediately because normalizers run inside the boefjes container itself.

### Issue link

Closes #5136

### Demo

Reproduction before the fix, from `docker.underdark.nl/librekat/openkat-nikto:latest` against `https://apache.dvwa.cloud/`, Tuning=3b (7 findings in Nikto JSON output):

| Nikto id | Message | Normalizer (before) | Normalizer (after) |
|---|---|---|---|
| `999986` | Retrieved x-powered-by header: Apache/2.4.49 | *dropped* | `Software(Apache, 2.4.49)` + `SoftwareInstance` + `KAT-NIKTO-FINDING` |
| `013587` | Suggested security header missing: x-content-type-options | `KAT-MISSING-HEADER` (generic, wrong extractor) | `KAT-NO-X-CONTENT-TYPE-OPTIONS` |
| `600050` | Apache/2.4.49 appears to be outdated | ✅ `KAT-OUTDATED-SOFTWARE` | ✅ `KAT-OUTDATED-SOFTWARE` |
| `999992` | Server is using a wildcard certificate | *dropped* | `KAT-WILDCARD-CERTIFICATE` |
| `999990` | Allowed HTTP Methods: HEAD, GET, POST, OPTIONS, TRACE | *dropped* | `KAT-HTTP-TRACE-METHOD` (msg contains TRACE) |
| `000434` | HTTP TRACE method active (XST) | *dropped* | `KAT-HTTP-TRACE-METHOD` |
| `003584` | Apache default file found | *dropped* | `KAT-EXPOSED-DEFAULT-FILE` |

### QA notes

1. On this branch, enable the `nikto` boefje and point it at a Website OOI served by Apache (`https://apache.dvwa.cloud/` works well).
2. After the scan completes, verify that **every** finding id in the bytes raw output has produced at least one finding in octopoes. Before this fix only the `600050` outdated-software case made it through, and only when the boefje managed to write its output file at all.
3. `cd boefjes && make itest` — the new `tests/test_nikto_normalizer.py::NiktoNormalizerTest::test_nikto_2_6_full_coverage` asserts the full id coverage; `test_outdated_and_legacy_missing_header` keeps the old Nikto 2.5 wording exercised.
4. End-to-end functional test requires `docker.underdark.nl/librekat/openkat-nikto:latest` to be rebuilt with the `main.js` and `boefje.Dockerfile` changes from this PR.

### Code Checklist

- [x] Tickets have been created for newly discovered issues. (#5136)
- [x] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly. (none)

### DIFF
--- a/boefjes/boefjes/plugins/kat_kat_finding_types/kat_finding_types.json
+++ b/boefjes/boefjes/plugins/kat_kat_finding_types/kat_finding_types.json
@@ -613,5 +613,45 @@
     "risk": "recommendation",
     "impact": "Depending on what software is outdated this can be critical.",
     "recommendation": "Inspect the software version, determine if additional measures need to be taken and install updates to reduce the attack surface."
+  },
+  "KAT-MISSING-HEADER": {
+    "name":"Missing security header",
+    "description": "A recommended HTTP security header is not present on the response. The exact header was either not recognised by the parser or not specifically mapped to a dedicated finding type.",
+    "source": "https://owasp.org/www-project-secure-headers/",
+    "risk": "recommendation",
+    "impact": "Missing security headers may weaken the browser-side defences of the site against common web attacks such as XSS, clickjacking, or MIME-sniffing.",
+    "recommendation": "Configure the web server or application to emit the missing security header(s) with appropriate values per the OWASP Secure Headers project."
+  },
+  "KAT-NIKTO-FINDING": {
+    "name":"Nikto finding",
+    "description": "Nikto reported an issue that does not yet have a dedicated OpenKAT finding type. The raw message from Nikto is attached to the finding description.",
+    "source": "https://github.com/sullo/nikto",
+    "risk": "recommendation",
+    "impact": "Depends on the specific Nikto finding. Consult the finding description to assess impact.",
+    "recommendation": "Review the finding description and consult the Nikto documentation or your web server vendor for remediation guidance."
+  },
+  "KAT-HTTP-TRACE-METHOD": {
+    "name":"HTTP TRACE method is enabled",
+    "description": "The HTTP TRACE method is enabled on the web server. TRACE can be abused in cross-site tracing (XST) attacks to read security-sensitive headers such as cookies when combined with other client-side vulnerabilities.",
+    "source": "https://owasp.org/www-community/attacks/Cross_Site_Tracing",
+    "risk": "medium",
+    "impact": "An attacker who also exploits a client-side vulnerability (e.g. XSS) can use TRACE to read otherwise HttpOnly cookies or authorization headers.",
+    "recommendation": "Disable the TRACE method on the web server. For Apache: TraceEnable off. For Nginx: do not forward TRACE. For IIS: disable the verb in request filtering."
+  },
+  "KAT-WILDCARD-CERTIFICATE": {
+    "name":"Wildcard TLS certificate in use",
+    "description": "The server presents a wildcard TLS certificate. Wildcard certificates simplify operations but increase blast radius on compromise because the same private key authenticates every subdomain.",
+    "source": "https://en.wikipedia.org/wiki/Wildcard_certificate",
+    "risk": "recommendation",
+    "impact": "Compromise of the wildcard certificate private key allows an attacker to impersonate every matching subdomain.",
+    "recommendation": "Prefer per-host certificates when operationally feasible; automate issuance via ACME. Keep wildcards only where a single key/cert must cover many ephemeral subdomains and protect the key accordingly."
+  },
+  "KAT-EXPOSED-DEFAULT-FILE": {
+    "name":"Default file exposed",
+    "description": "A default file or directory shipped with the web server software is exposed and reachable. These files often disclose software versions or provide an information foothold for attackers.",
+    "source": "https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/01-Information_Gathering/",
+    "risk": "low",
+    "impact": "Information disclosure about the underlying software stack, assisting reconnaissance.",
+    "recommendation": "Remove default files from production deployments or deny-list them at the web server level."
   }
 }
--- a/boefjes/boefjes/plugins/kat_nikto/boefje.Dockerfile
+++ b/boefjes/boefjes/plugins/kat_nikto/boefje.Dockerfile
@@ -3,7 +3,7 @@ FROM perl:5.40
 WORKDIR /app
 RUN apt-get update && apt-get install -y --no-install-recommends git nodejs
 
-RUN git clone https://github.com/sullo/nikto
+RUN git clone --depth 1 --branch 2.6.0 https://github.com/sullo/nikto
 RUN cpanm --notest JSON XML::Writer
 
 ARG BOEFJE_PATH=./boefjes/plugins/kat_nikto
--- a/boefjes/boefjes/plugins/kat_nikto/main.js
+++ b/boefjes/boefjes/plugins/kat_nikto/main.js
@@ -12,7 +12,8 @@ function get_config_content(scheme) {
 
   // Setup config file
   try {
-    let config_contents = `PROMPTS=no\nUPDATES=no\nCLIOPTS=-404code=301,302,307,308 -Tuning ${TUNING} -maxtime ${MAX_TIME} -o ./output.json`;
+    // Nikto 2.6 appends ".json" to the -o path, so pass base name + explicit -Format json.
+    let config_contents = `PROMPTS=no\nUPDATES=no\nCLIOPTS=-404code=301,302,307,308 -Tuning ${TUNING} -maxtime ${MAX_TIME} -o ./output -Format json`;
 
     if (scheme == "https") config_contents += " -ssl";
     if (IS_USING_PROXY) config_contents += " -useproxy";
--- a/boefjes/boefjes/plugins/kat_nikto/normalize.py
+++ b/boefjes/boefjes/plugins/kat_nikto/normalize.py
@@ -1,4 +1,5 @@
 import json
+import re
 from collections.abc import Iterable
 from typing import Any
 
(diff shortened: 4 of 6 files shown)


underdarknl:
I'm not sure we need/want /nikto to add the header findings by itself. Ideall, we'd collect the headers, and then based on what we know `in the graph` combined with a requires/allowed/disallowed headers config decide on what findings to generate.
But, since we dont have that functionality for headers just yet, I'd be inclined to include this version.

hasecon:
**Real-world validation: Kennisnet sample**

A Kennisnet user shared raw Nikto output from a scan against one of their internal hosts. I ran his exact data through both `main` and this PR's normalizer to quantify the impact.

| id | msg (truncated) | main | this PR |
|---|---|---|---|
| 999962 | Server banner changed from 'granian' to 'Apache' | *dropped* | `KAT-NIKTO-FINDING` |
| 999966 | Content-Encoding: deflate → BREACH attack | *dropped* | `KAT-NIKTO-FINDING` |
| 999986 | Retrieved via header: 1.1 ... | *dropped* | `KAT-NIKTO-FINDING` |
| 013587 | Suggested security header missing: content-security-policy | `KAT-CSP-VULNERABILITIES` | `KAT-CSP-VULNERABILITIES` |
| 013587 | Suggested security header missing: permissions-policy | `KAT-NO-PERMISSIONS-POLICY` | `KAT-NO-PERMISSIONS-POLICY` |
| 740000 | Multiple index files responding | *dropped* | `KAT-NIKTO-FINDING` |
| 999992 | Server is using a wildcard certificate | *dropped* | `KAT-WILDCARD-CERTIFICATE` |
| 999967 | Web server accepts junk HTTP methods | *dropped* | `KAT-NIKTO-FINDING` |
| 999972 | DEBUG HTTP verb may show debugging info | *dropped* | `KAT-NIKTO-FINDING` |
| 000016, 000017, 000038, 000040, 000150, 000297 | Various legacy CGI / config-file lookups | *dropped* | `KAT-NIKTO-FINDING` |

**Net effect on this scan: 2/15 findings on `main` → 15/15 on this branch.** No more silent drops, every Nikto observation is at minimum surfaced via the catch-all.

---

**QA performed locally (PR commit `4234f02`)**
- All 30 GitHub checks green.
- `boefjes` unit tests pass in Docker (3/3, 94 % coverage on `kat_nikto/normalize.py`).
- Full stack starts cleanly (`make clean && make kat`); 13 containers up, no restart loops, no new errors after the startup-noise window.
- All 5 new `KATFindingType` entries register in katalogus on startup; total 82 finding types.
- Sample raw output piped through `boefjes.plugins.kat_nikto.normalize.run` produces the table above.

---

**Re: @underdarknl's architectural concern**

Fully agree the dispatch table in the normalizer is not the right long-term home for this logic. A cleaner pattern that already exists in the codebase: `Question` OOIs scoped per organisation (e.g. `Question|/bit/port-classification-ip|internet`) driving a graph bit that decides classification per-org. The same shape would fit Nikto well:

- Nikto emits raw observations (id + msg + url + references) without per-id mapping.
- A `nikto-finding-classification` bit with a `Question` OOI lets each organisation enable/disable specific Nikto ids and configure the finding type and severity per id.

That removes both the hardcoded dispatch *and* the silent-drop bug, and lets users mute the noisier 000xxx legacy CGI scans while keeping high-signal findings (BREACH, wildcard cert, exposed configs) active. Probably worth scoping for v1.23.

This PR is a tactical fix to stop the bleed today; the Question-OOI version is the right follow-up. Other concrete follow-ups suggested by the Kennisnet sample:

1. CVE extraction from Nikto's `references` field → `CVEFindingType` OOIs (generic, no per-id work).
2. Per-id mapping for the highest-signal categories currently in the catch-all: `KAT-BREACH-COMPRESSION` (999966), `KAT-HTTP-DEBUG-METHOD` (999972), `KAT-EXPOSED-CONFIG-FILE` (~50 ids in 000xxx range).
3. URL OOI emission for exposed paths so other boefjes can pick them up.

hasecon:
**Functional test caught a third Nikto 2.6 bug — fixed in 5f62440**

While running this branch end-to-end against a Website OOI we hit `Boefje caught an error: ENOENT: no such file or directory, open './output.json'` on every Nikto task. Tracing it: Nikto 2.6 appends `.json` to the `-o` path, so `-o ./output.json` produces `output.json.json` and `main.js:86` `readFileSync("./output.json")` throws.

Fixed by passing a base name plus explicit `-Format json`. Without this, the rest of the work in this PR is unreachable in the running stack — the normalizer never sees any input.

Re-tested end-to-end against `apache.dvwa.cloud` with a locally-built Nikto image off this branch:

| Nikto id | Result |
|---|---|
| 999986 banner | `Software\|Apache\|2.4.49` + `KAT-NIKTO-FINDING` ✅ |
| 013587 missing header | `KAT-NO-X-CONTENT-TYPE-OPTIONS` ✅ |
| 600050 outdated | `KAT-OUTDATED-SOFTWARE` ✅ |
| 999992 wildcard cert | `KAT-WILDCARD-CERTIFICATE` ✅ |
| 999990 trace methods | `KAT-HTTP-TRACE-METHOD` ✅ |
| 000434 trace XST | `KAT-HTTP-TRACE-METHOD` ✅ |
| 003584 default file | `KAT-EXPOSED-DEFAULT-FILE` ✅ |

7/7 Nikto observations now reach octopoes. Title and body of the PR updated to reflect that scope is now boefje + normalizer, not normalizer-only.

**@underdarknl** — to land this end-to-end you'll need to rebuild and push `docker.underdark.nl/librekat/openkat-nikto:latest` from this branch. The registry image is what the runtime actually pulls; the `boefje.Dockerfile` pin and `main.js` change in this PR have no effect until the image is rebuilt. Could you handle that when you have a moment?

underdarknl (approved):

underdarknl:
The :latest version of this image will be available as soon as we tag this as a released version.

PR #5146, opened 2026-05-02
crisis_room: skip dashboard items with deleted organization

### LINKED ISSUE #5140: Crisis room crashes with AttributeError after deleting an organization
## Describe the bug

After deleting an organization via the API (`OrganizationViewSet.destroy()`), the crisis room page crashes with:

```
AttributeError: 'NoneType' object has no attribute 'code'
```

at `rocky/crisis_room/views.py`, line 117, in `get_dashboard_items`.

## Root cause

The `Dashboard` model uses `on_delete=models.SET_NULL` for its `organization` foreign key (`crisis_room/models.py`, line 20):

```python
organization = models.ForeignKey(Organization, on_delete=models.SET_NULL, null=True)
```

When an organization is deleted:
1. `OrganizationViewSet.destroy()` cleans up Octopoes node and Katalogus organization
2. `super().destroy()` deletes the Django `Organization` object
3. Django's `SET_NULL` sets `crisis_room_dashboard.organization_id = NULL`
4. The `Dashboard` and `DashboardItem` records remain as orphans

The crisis room view then queries `DashboardItem.objects.filter(findings_dashboard=True)` without filtering out items whose dashboard has a `NULL` organization. When it hits a dashboard item with a recipe, it tries to access `dashboard_item.dashboard.organization.code` (line 117), which fails because `organization` is `None`.

## Steps to reproduce

1. Create an organization (this automatically creates a Findings Dashboard via the `organization_post_save` signal)
2. Delete the organization via the API or admin
3. Visit `/nl/crisis-room/`
4. Observe the `AttributeError`

You can also verify orphaned records with:

```sql
SELECT di.id, d.name, d.organization_id
FROM crisis_room_dashboarditem di
JOIN crisis_room_dashboard d ON di.dashboard_id = d.id
WHERE d.organization_id IS NULL;
```

## Expected behavior

The crisis room should either:
- **Option A**: Clean up dashboards and dashboard items when an organization is deleted (e.g., change `on_delete=SET_NULL` to `on_delete=CASCADE`, or add cleanup in `OrganizationViewSet.destroy()`)
- **Option B**: Skip dashboard items with `NULL` organization in `get_dashboard_items()` (defensive check)

## Workaround

Delete the orphaned records manually:

```sql
DELETE FROM crisis_room_dashboarditem WHERE dashboard_id IN (
    SELECT id FROM crisis_room_dashboard WHERE organization_id IS NULL
);
DELETE FROM crisis_room_dashboard WHERE organization_id IS NULL;
```

## Environment

- **OpenKAT version**: v1.22.0rc1
- **Python version**: 3.13.13
- **Django version**: 5.1.15

## Traceback

```
File "/app/rocky/crisis_room/views.py", line 314, in setup
    self.organizations_findings = dashboard_service.get_dashboard_items(dashboard_items)
File "/app/rocky/crisis_room/views.py", line 117, in get_dashboard_items
    report_filters.append((dashboard_item.dashboard.organization.code, str(dashboard_item.recipe)))
AttributeError: 'NoneType' object has no attribute 'code'
```

SAY-5:
Fixes #5140.

`Dashboard.organization` uses `on_delete=SET_NULL`, so deleting an organization leaves orphan dashboards. `get_dashboard_items` then accesses `dashboard.organization.code` and 500s the crisis room on the next visit. Skip orphan items in the loop, every downstream branch needs `organization.code`, so a single guard at the top covers them all.

Option B per the issue body. Option A (cascade-delete or explicit cleanup in `OrganizationViewSet.destroy`) would also clean up storage, but it's a bigger schema-migration call I'd leave to the maintainers; this fix gets users out of the immediate crash on existing orphans.

### DIFF
--- a/rocky/crisis_room/views.py
+++ b/rocky/crisis_room/views.py
@@ -106,6 +106,13 @@ class DashboardService:
 
         # First collect al data, if recipe id is found then fetch recipe ids to get reports later.
         for dashboard_item in dashboard_items:
+            # Dashboard.organization uses on_delete=SET_NULL, so a deleted
+            # organization leaves orphan dashboards/items. Skip them — every
+            # downstream branch needs `dashboard.organization.code`, and
+            # without this guard the crisis room 500s on the next visit
+            # after any organization deletion (#5140).
+            if dashboard_item.dashboard.organization is None:
+                continue
             if not dashboard_item.recipe and dashboard_item.source == "object_list":
                 item_data = DashboardItemView(dashboard_item, self.get_ooi_list(dashboard_item))
                 dashboard_items_with_data.append(item_data)
--- a/rocky/crisis_room/views.py
+++ b/rocky/crisis_room/views.py
@@ -107,7 +107,7 @@ class DashboardService:
         # First collect al data, if recipe id is found then fetch recipe ids to get reports later.
         for dashboard_item in dashboard_items:
             # Dashboard.organization uses on_delete=SET_NULL, so a deleted
-            # organization leaves orphan dashboards/items. Skip them — every
+            # organization leaves orphan dashboards/items. Skip them, every
             # downstream branch needs `dashboard.organization.code`, and
             # without this guard the crisis room 500s on the next visit
             # after any organization deletion (#5140).


underdarknl (approved):

underdarknl (dismissed):

hasecon (approved):
## QA Review

### Setup
- Branch: `say5/crisis-room-orphan-dashboard-5140` @ `a37b81a` (from fork)
- Approach: reused running stack on `main`, reproduced the bug end-to-end, applied the patch via `docker cp`, verified the fix, restored container state.

### CI/CD Status

`total_count: 0`, status `pending` for SHA `a37b81a`. Looks like CI has not been re-triggered since the DCO rebase on 2026-05-25. @underdarknl's earlier approval was on `53b0ca5` and was auto-dismissed when the branch was force-pushed for the signoff. Worth re-running checks before merge; the diff is functionally identical to what was approved before (single-line guard, only the commit signature changed).

### Code Review

The fix is exactly Option B from issue #5140 and the guard is correctly placed. I dereferenced `dashboard.organization.code` three times in `get_dashboard_items` on `main`:

- `views.py:116` — recipe branch (the line in the original traceback)
- `views.py:133` — report-section branch inside the recipe loop
- `views.py:152` — final merge branch where `OctopoesAPIConnector` is constructed

The `continue` at line 108 fires before all three. The two non-recipe branches (`object_list`, `finding_list`) also need an organization context to resolve OOIs/findings against the right Octopoes node, so skipping orphans there is the right behaviour too.

Comment in the patch is good — it names the FK semantics (`SET_NULL`), the failure mode (500 on next visit), and links back to #5140.

### Functional verification

Stack on `main`, then:

**1. Reproduced the bug** by creating org `orphantest`, deleting it (which fires `SET_NULL`), and running the failing code path directly:

```
Items to process: 3
Orphan items in set: 1
CRASH (expected on main): 'NoneType' object has no attribute 'code'
```

Matches the issue traceback exactly.

**2. Applied this PR's `views.py`** via `docker cp` into `openkat-reviev-rocky-1`, re-ran the same scenario:

```
Items in set: 3 (of which orphan: 1)
OK — returned 2 non-orphan items processed without crash
```

Crisis room view no longer 500s; the two healthy dashboard items still resolve.

**3. Cleaned up** the orphan dashboard row and restored the container file.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| rocky | reproduce bug on main | bug confirmed (`AttributeError`) |
| rocky | post-patch behavior | guard skips orphan, others process |

### Suggestion (non-blocking)

A small unit test in `rocky/tests/test_crisis_room.py` along the lines of "given a `DashboardItem` whose `dashboard.organization is None`, `get_dashboard_items` skips it and returns the other items" would lock in the regression. Not blocking — the change is 7 LoC defensive code with clear semantics — but I notice CI never caught this in the first place because no test covers the orphan path.

### Verdict

**APPROVED**

Clean defensive fix for the exact crash I described in #5140 (Option B). Bug reproduced and fix verified locally. Complementary to #5148 (Option A — schema-level cascade + data migration), as the author and Jan both noted; both can land independently, with #5146 unblocking users running on existing orphans before any migration is applied.

CI re-trigger needed before merge.

PR #5148, opened 2026-05-05
fix(crisis_room): cascade-delete dashboards with their organization

### LINKED ISSUE #5140: Crisis room crashes with AttributeError after deleting an organization

SAY-5:
### Changes

`Dashboard.organization` uses `on_delete=SET_NULL`, so deleting an organization leaves orphan dashboards and items. The crisis room then 500s on `dashboard.organization.code` (the surface PR #5146 patches defensively in the view). This is the storage-side root-cause fix per Option A in the issue: switch the FK to `CASCADE` and clean up any pre-existing orphans in the same migration. Complementary to #5146, not a replacement.

### Issue link

Closes #5140.

### QA notes

1. On main, create an organization and delete it via the API or admin; visit `/nl/crisis-room/` to reproduce the `AttributeError`.
2. Apply this branch and run `python manage.py migrate` in `rocky/`. Existing orphan dashboards are removed by the data migration.
3. Repeat the create/delete cycle; the crisis room loads cleanly. Verify `crisis_room_dashboard` rows are gone for the deleted org's id.

---

### Code Checklist

### DIFF
--- a/rocky/crisis_room/migrations/0008_alter_dashboard_organization.py
+++ b/rocky/crisis_room/migrations/0008_alter_dashboard_organization.py
@@ -1,0 +1,24 @@
+from django.db import migrations, models
+
+
+def delete_orphan_dashboards(apps, schema_editor):
+    Dashboard = apps.get_model("crisis_room", "Dashboard")
+    Dashboard.objects.filter(organization__isnull=True).delete()
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("crisis_room", "0007_alter_dashboarditem_source"),
+        ("tools", "0047_alter_organization_code_alter_organization_name"),
+    ]
+
+    operations = [
+        migrations.RunPython(delete_orphan_dashboards, migrations.RunPython.noop),
+        migrations.AlterField(
+            model_name="dashboard",
+            name="organization",
+            field=models.ForeignKey(
+                null=True, on_delete=models.CASCADE, to="tools.organization"
+            ),
+        ),
+    ]
--- a/rocky/crisis_room/models.py
+++ b/rocky/crisis_room/models.py
@@ -17,7 +17,7 @@ logger = structlog.get_logger(__name__)
 
 class Dashboard(models.Model):
     name = models.CharField(blank=False, max_length=126)
-    organization = models.ForeignKey(Organization, on_delete=models.SET_NULL, null=True)
+    organization = models.ForeignKey(Organization, on_delete=models.CASCADE, null=True)
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 


hasecon (approved):
## QA Review

### Setup
- Branch: `fix/crisis-room-org-delete-5140` @ `ee490ec` (from fork)
- Approach: reused running stack on `main`, created an orphan via real `Organization.delete()`, applied this PR's `models.py` + new `0008` migration via `docker cp`, ran `manage.py migrate crisis_room`, verified pre-existing orphan was deleted and CASCADE behaves on subsequent deletes, then rolled back with `manage.py migrate crisis_room 0007`.

### CI/CD Status

`total_count: 0`, status `pending` for SHA `ee490ec`. Same as #5146 — checks haven't been re-triggered, likely because the branch is from a fork and needs maintainer approval to run.

### Code Review

**Migration ordering is correct:**
```python
operations = [
    migrations.RunPython(delete_orphan_dashboards, migrations.RunPython.noop),   # 1. clean data
    migrations.AlterField(... on_delete=models.CASCADE ...),                     # 2. switch FK
]
```
Cleaning orphans **before** the AlterField is required — without it, Django's `AlterField` itself would not delete the orphans (Django doesn't backfill data when changing `on_delete` semantics), so they'd stay as zombies forever even after the schema flip.

**Cascade propagates to items:** verified `DashboardItem.dashboard = ForeignKey(Dashboard, on_delete=CASCADE)` (models.py:52), so the `Dashboard.objects.filter(organization__isnull=True).delete()` in the data migration also removes the associated DashboardItems via Django's cascade. No explicit DashboardItem cleanup needed.

**Migration dependencies are current:**
- `crisis_room.0007_alter_dashboarditem_source` — leaf in crisis_room
- `tools.0047_alter_organization_code_alter_organization_name` — leaf in tools (the new cross-app dep is needed because the FK now cascades into Organization)

**Reverse is `noop`** for the data part (can't recreate deleted orphans) and the `AlterField` is implicitly reversible by Django (CASCADE → SET_NULL on backward). Safe.

**`unique_together = ["name", "organization"]`** is more meaningful with CASCADE: previously you could end up with multiple `("Findings Dashboard", NULL)` rows from successive org deletions; now NULL rows can no longer accumulate.

### Functional verification

**1. Created orphan state** on main (SET_NULL): created `casctest` org → deleted org → `Orphan dashboards: 1, Orphan items: 1`.

**2. Applied this PR's files** and ran `manage.py migrate crisis_room`:
```
Applying crisis_room.0008_alter_dashboard_organization...
  Dashboard Dashboard object (5) deleted
  DashboardItem DashboardItem object (5) deleted
 OK
```
Orphan dashboard and item both deleted via cascade. Migration completed cleanly.

**3. Verified CASCADE on new deletion:** created `casctest2` → `Dashboards for casctest2 before delete: 1` → deleted org →
```
Orphans after delete: 0
casctest2 dashboards left: 0
```
Findings Dashboard and DashboardItem deleted with the org. No orphans created. CASCADE works end-to-end.

**4. Rolled back** with `manage.py migrate crisis_room 0007` — unapplied cleanly. Restored `models.py` and removed the migration file from the container.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| rocky | apply migration with pre-existing orphan | orphan dashboard + item deleted |
| rocky | delete org after migration | dashboard + item cascade-deleted, no orphan |
| rocky | reverse migration | unapply OK, schema back to SET_NULL |

### Suggestions (non-blocking)

1. **`null=True` is now vestigial.** With `on_delete=CASCADE` the FK is effectively required after this migration — no normal code path can create a row with `organization=NULL`. The `null=True` could be dropped in a follow-up migration (would also let you remove the `dashboard_item.dashboard.organization is None` guard from #5146 eventually). Not blocking this PR — keeping `null=True` avoids a NOT NULL constraint failure on any pre-existing NULL the migration somehow misses.
2. **No tests** for the migration. A small `rocky/tests/test_crisis_room_migrations.py` using `django_db.MigrationTestCase`-style helpers (or just `pytest-django`'s `django_db(transaction=True)`) that asserts an orphan gets removed when 0008 runs would lock in the behavior. Same suggestion as #5146 — neither PR adds tests, and the bug went undetected exactly because no test covers the orphan path.

### Relation to #5146

Complementary, as both you and Jan noted. #5146 is the defensive view-layer guard that protects users on existing orphans even before this migration runs; #5148 fixes the storage so no new orphans can appear and cleans up existing ones in a single migration. Order of landing doesn't matter — they don't conflict, and either one alone unblocks the immediate crash.

### Verdict

**APPROVED**

Correct cascade-and-cleanup migration with sound ordering and dependencies. Functional verification confirms orphan cleanup, new-delete cascade, and clean rollback. Per Jan's earlier comment leaning toward CASCADE, this is the right root-cause fix for #5140.

CI re-trigger needed before merge (same as #5146).

PR #5165, opened 2026-05-15
Add index on taskDB that allows quick filtering on org, type and sort on datetime, as this s a common pattern to search for.

underdarknl:
Add index on taskDB that allows quick filtering on org, type and sort on datetime, as this s a common pattern to search for.

### Changes

_Please describe the essence of this PR in a few sentences. Mention any breaking changes or required configuration steps._

### Issue link

Large installs / or mostly long running installs saw intermittent timeouts when requesting task lists, due to table scans in Mula's task api.
This was due to the ordering component of these queries having no index to rely on.
Since these calls usually involve the organization, the job type (eg, boefje or normalizer), and then sort on created_at. A compound index makes sense.

### Demo

_Please add some proof in the form of screenshots or screen recordings to show (off) new functionality, if there are interesting new features for end-users._

### QA notes

_Please add some information for QA on how to test the newly created code._

---

### Code Checklist

- [x] All the commits in this PR are properly PGP-signed and verified.
- [x] This PR only contains functionality relevant to the issue.
- [x] I have written unit tests for the changes or fixes I made.
- [x] I have checked the documentation and made changes where necessary.
- [x] I have performed a self-review of my code and refactored it to the best of my abilities.

- [x] Tickets have been created for newly discovered issues.
- [x] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [x] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [x] I have included comments in the code to elaborate on what is not self-evident from the code itself, including references to issues and discussions online, or implicit behavior of an interface.

---

## Checklist for code reviewers:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_qa.md) into your comment.

Issue #5151, opened 2026-05-06
upgrade rabbitMQ to new supported version


underdarknl:
## About this feature
We are currently using rabbitMQ 3.12, which is EOL. We should upgrade to ideally the newest version.

### Detailed description

Careful testing and migration-guides should be provided for existing installs.
4.3 seems like the current proper target to move to.

We might possibly need to migrate to intermediate versions first.

### Feature benefit/User story
Dont rely on EOL software.

### Specifications

_Please add some specifications for the implementation. What needs to be implemented to match the design?_

The implementation should include…

- [ ] Specification 1
- [ ] Specification 2

### Additional information

_Any additional information, considerations, or context that might be helpful for understanding or evaluating the feature request._

## Design

_This part should only be filled in by the OpenKAT design team._

### Screenshots

_Include screenshots of the proposed design changes here._

### Figma link

_Link to the Figma design for further visualization (if applicable)_

## Implementation

_This part should only be filled in by the developers._

### Possible solution

_Outline your proposed solution for implementing the feature. You can include any specific ideas, designs, or functionalities here._

### Alternatives considered

_Describe any alternative approaches or solutions you've considered, and why you believe the proposed solution is superior._

hasecon:
## Proposed plan: stage the upgrade across three OpenKAT releases

### Why staged

Per the [official RabbitMQ upgrade docs](https://www.rabbitmq.com/docs/upgrade) you cannot hop from 3.12 to 4.3 in one step on a persistent volume. The supported chain is:

- `3.12.x → 3.13.x` (with all stable feature flags enabled afterwards, but **not** `khepri_db` — Khepri in 3.13 is experimental and silently breaks the 4.x upgrade).
- `3.13.x → 4.2.x` (one step is supported).
- `4.2.x → 4.3.x` (only path into 4.3; requires `khepri_db` enabled while still on 4.2).

A single PR that bumps 3.12 → 4.3 directly would require operators with a persistent `rabbitmq_data` volume to do all the manual feature-flag / intermediate-image work themselves around our release. Staging it gives them ~30 days of soak time on each landing point.

### Roadmap

| Release | PR contents | Operator action after pulling |
|---|---|---|
| **1.23** | Bump all 7 compose files to `rabbitmq:3.13-management` + release notes | `docker compose exec rabbitmq rabbitmqctl enable_feature_flags all` (do **not** also enable `khepri_db`) |
[...]

PR #5166, opened 2026-05-18
Allow for bounded/unpartial counts to be returned from the scheduler …

underdarknl:
…endpoints, and use those in Rocky to avoid counting *all* rows when we only need to know if we can present up to 6 'next page' buttons

### Changes

This changes adds optional bounded counts to on the task and schedule api endpoints. 
This is useful when the full count for a given query isn't relevant, but instead we only want to know if we need/can display a number of 'page next' buttons. (as we do in rocky).
This heavily optimizes the time taken to return these results on very large databases (we saw databases with millions of tasks, which where spending most their time in the count query).

On the API result a new flag is introduced which signals if the count was partial. 

### Issue link

Closes ...

### Demo

image

### QA notes

Pages with many items should be faster (noticeably the task lists), and still show the correct number of page buttons.

### Code Checklist

---

---

## Checklist for QA:


hasecon (changes requested):
## QA Review — UltraReview

### Setup
- Branch: `feat/mula-partial-counts`
- Scope: 4 Mula files (`stores/task.py`, `stores/schedule.py`, `handlers/tasks.py`, `handlers/schedules.py`, `utils/pagination.py`) + 1 Rocky file (`rocky/scheduler.py`)
- Functional `make kat` test: **skipped** — issues are textual/structural and would not surface on a fresh stack with empty tables. The interesting case (large dataset, paginator beyond 6 pages) requires a populated DB that `make kat` doesn't provide.

### CI/CD Status

| Check | Status | Notes |
|---|---|---|
| `test (3.10/3.11/3.12/3.13)` — latest run | ✅ pass | Earlier run failed; superseded by 13:39 run — verified green. |
| `pre-commit` | ✅ pass | |
| `CodeQL`, `Analyze (python/js/actions)` | ✅ pass | |
| `makelang` | ❌ fail | Non-blocking per merge protocol, but unusual for a PR with no template/i18n changes — worth a quick check that it's not catching a stale `.po` regen rather than something the PR touched. |
| `build (debian12, mula)`, `build (ubuntu22.04, mula)`, `build (debian12, rocky)`, `build (ubuntu22.04, rocky)` | ❌ fail | Could not retrieve logs via API. Likely the known Cloudsmith/Debian-host infra issue (see #5151), but please confirm pre-existing on main before merging. |
| `create_container_image` — latest | ✅ pass | |

### Code Review Findings

#### 🛑 High — Rocky never reads `is_partial_count`, so the paginator will lie about totals

The whole point of `is_partial_count` is to tell the consumer that `count` is a lower bound, not a real total. Rocky doesn't currently read that signal:

- `rocky/rocky/scheduler.py:198` (`PaginatedTasksResponse`) and `:209` (`PaginatedSchedulesResponse`) do not declare `is_partial_count` — Pydantic silently drops the new field.
- No template references it (`grep -r is_partial_count rocky/` returns nothing).
- `LazyTaskList.count` (`rocky/rocky/scheduler.py:225-228`) uses the bounded `count` as ground truth and feeds it directly into Django's paginator.

Concrete impact on the views the PR targets (`rocky/views/tasks.py`):
- `paginate_by = 150` is used in two task list views (lines 32, 134).
- With `max_pages=6` (server default) and `limit=150`, `max_count = 0 + 6*150 + 1 = 901`.
- Rocky will report exactly **901 tasks** to the paginator regardless of whether the real total is 902 or 5 million.
- The paginator renders pages 1–6 and a stub page 7 containing 1 row. Pages 7+ are unreachable through the UI.

To honour the PR's stated intent ("we only need to know if we can present up to 6 'next page' buttons"), Rocky needs to:
1. Add `is_partial_count: bool = False` to `PaginatedTasksResponse` and `PaginatedSchedulesResponse`.
2. Propagate it into `LazyTaskList` (and the schedule equivalent) so templates can render "showing 1–150 of 900+" instead of an exact false number.
3. Either re-fetch with `allow_partial_count=False` on "go to last page", or hide the "last page" jump entirely while partial.

Alternative: keep `allow_partial_count=True` opt-in in Rocky (don't unconditionally set it on every list call from `rocky/scheduler.py:300` and `:361`) and turn it on only on the call sites where the UI has been adjusted.

#### 🛑 Medium — Type-annotation regression in `ScheduleStore.get_schedules`

`mula/scheduler/storage/stores/schedule.py`:

```python
-    ) -> tuple[list[models.Schedule], int]:
+    ) -> tuple[list[models.Task], int, bool]:
```

That should be `tuple[list[models.Schedule], int, bool]`. Looks like a copy-paste from `TaskStore`. Mypy/pyright won't catch this in many call sites because the values get model-validated downstream, but it actively misleads readers.

#### ⚠️ Medium — `max_pages` query parameter is dead on `GET /schedules`

In `mula/scheduler/server/handlers/schedules.py`, the `list()` handler adds:

```python
max_pages: int = 6,
allow_partial_count: bool = False,
```

…but the store call only forwards `allow_partial_count`:

```python
results, count, is_partial_count = self.ctx.datastores.schedule_store.get_schedules(
    ...
    offset=offset,
    limit=limit,
    allow_partial_count=allow_partial_count,   # max_pages missing
)
```

So `?max_pages=12` on `GET /schedules` is silently ignored — the store always falls back to its own default `max_pages=6`. Both `POST /schedules/search` and `POST /tasks` forward it correctly, so this is just an oversight in this one handler. Either add `max_pages=max_pages,` to the call or drop the parameter from the signature.

#### ⚠️ Medium — No tests added

The bounded count has subtle off-by-one behaviour (`max_count = offset + (max_pages * limit) + 1`, `is_partial_count = count == max_count`). Given:
- The `+1` sentinel pattern is easy to mis-edit.
- `limit` and `offset` are now silently clamped (`limit = min(max(limit, 1), 500)`, `offset = max(offset, 0)`) — also a behavioural change worth pinning.
- `max_pages` itself is clamped to `[1, 20]`.

…at least a handful of `TaskStore`/`ScheduleStore` unit tests would be valuable:
1. Real count smaller than the cap → `is_partial_count=False`, `count` exact.
2. Real count equals `max_count - 1` → `is_partial_count=False`, exact.
3. Real count equals `max_count` (i.e., real >= `max_count`) → `is_partial_count=True`.
4. `allow_partial_count=False` still produces an exact count (regression guard).
5. `limit > MAX_LIMIT` is clamped (regression guard for any caller that depended on uncapped limits — there's at least one: `LazyTaskList.HARD_LIMIT = 500` in Rocky, which now matches but the contract should be tested).

#### ℹ️ Low — Silent clamp of `limit=0`

`LazyTaskList.count` (`rocky/scheduler.py:227`) calls `self.scheduler_client.list_tasks(limit=0, ...)` purely to fetch the count. With the new clamp `limit = min(max(limit, 1), 500)` this now becomes `limit=1`, so the scheduler does one extra row materialisation per count check. Trivial cost, but worth noting that previously `limit=0` was a recognised "count only" idiom. Either:
- Document `limit=0` as no longer supported and update the Rocky caller to ask the scheduler more efficiently (e.g. a dedicated count endpoint), or
- Allow `limit=0` to flow through (skip the result query when zero).

#### ℹ️ Info — Implementation note (no change requested)

The bounded count construction is correct and a clean SQLAlchemy idiom:

```python
bounded_query = query.order_by(None).with_entities(literal(1)).limit(max_count)
count = session.query(func.count()).select_from(bounded_query.subquery()).scalar()
```

Projecting a constant and stripping ordering avoids unnecessary work; the surrounding `subquery() + COUNT(*)` materialises only up to `max_count` rows. The `+1` for partial-detection is sound. Nicely done.

### Verdict

**CHANGES REQUESTED**

The core Mula change is sensible and the SQL is correct, but as it stands the consumer-side (Rocky) doesn't know about partiality, so the UX regression — Rocky showing a wrong "total" and silently capping users at 6 pages on the very installs this PR targets — outweighs the perf win. Needs Rocky-side updates to propagate `is_partial_count`, plus the type-annotation and dead-`max_pages` fixes, before this can land.

Happy to re-review once those are in.

underdarknl:
* High — Rocky never reads is_partial_count, so the paginator will lie about totals
While is is true the is_partial_count is not used in the rocky models, rocky also does not need this info. There is no 'row count 'shown on the consuming templates, and the reasoning around only being able to click up to page 7 is nonsense, as for those new pages, a new request will be fired showing up with the new higher partial count: current offset + 6 pages (if available). This leads to always being able to click to all lower pages than the current one, and all higher pages within the range of 0 to 6 from the current page, if they are actually available). Since we don't show the row-count, and visually there is no difference in the resulting page there is no need to show 'count is partial' to the user.

* Medium — Type-annotation regression in ScheduleStore.get_schedules -> Fixed.
* Medium — max_pages query parameter is dead on GET /schedules -> Fixed.

hasecon (changes requested):
## QA Review — Collin Pipeline (Round 2)

### Setup
- Branch: `feat/mula-partial-counts` @ `bff08bb`
- Stack: `make clean && make kat` — ✅ 13/13 containers up
- Test orgs: `setup_test_org --code dvwa` + `--code org2` (second org needed to verify stats filtering) — ✅
- All findings below were **reproduced live against the running stack**, not just read from the diff.

### Round-1 status

| Round-1 finding | Status |
|---|---|
| Type annotation `tuple[list[models.Task], ...]` in ScheduleStore | ✅ Fixed |
| `max_pages` dead on `GET /schedules` | ✅ Fixed |
| Rocky doesn't consume `is_partial_count` | ◐ Rebuttal partially accepted — your sliding-window argument is correct **for the page slices** (real `limit`, fresh count per request). But the `limit=0` count-probe path breaks hard; see Critical 1. |
| No tests for bounded-count logic | ❌ Still none — and Critical 2 below is exactly the kind of bug those tests would have caught |
| `limit=0` clamp (was Low) | ⬆️ Escalated — root cause of Critical 1 |

### CI/CD Status

23/24 green on `bff08bb` (mula + rocky tests, builds, pre-commit, CodeQL). Only `makelang` is red — non-blocking per merge protocol, translation regen to be picked up separately.

### 🛑 Critical 1 — Task lists collapse to 7 rows / 1 page

`LazyTaskList.count` (`rocky/rocky/scheduler.py:227`) probes the count with the `limit=0` idiom. With this PR, `list_tasks` unconditionally adds `allow_partial_count=True`, and the store clamps `limit = min(max(0, 1), 500) = 1`, so `max_count = 0 + 6×1 + 1 = 7`.

Reproduced on the running stack (36 normalizer tasks in DB):

```
POST /tasks?scheduler_id=normalizer&limit=0                          → count: 36   (main behaviour)
POST /tasks?scheduler_id=normalizer&limit=0&allow_partial_count=true → count: 7, is_partial_count: true
```

And through Rocky's own code path (Django shell in the rocky container):

```python
l = LazyTaskList(c, scheduler_id='normalizer', filters={'filters': {'and': []}})
len(l)      # → 7    ← what Django's Paginator sees
l[0:150]    # → 36 tasks actually returned
```

Django's `Paginator` runs `count` **before** slicing (`page() → validate_number → num_pages → count`), then caps the slice at `top = count`. So every task list view (`paginate_by=150`) renders **at most 7 rows on exactly 1 page**, regardless of the real total. The cached_property means the later, larger count from `__getitem__` never wins. This inverts the PR's goal: instead of faster task lists, users get truncated ones.

**Fix direction:** the count probe needs a real page size, e.g. give `LazyTaskList` the view's `paginate_by` and probe with `limit=<page_size>&allow_partial_count=true` (count then caps at `6×150+1`, and your sliding-window behaviour works as intended). Forcing `allow_partial_count=False` just for the probe would work too, but re-introduces the full COUNT the PR is trying to kill.

### 🛑 Critical 2 — Unfiltered bounded count returns 1 (FROM clause lost)

`query.order_by(None).with_entities(literal(1))` drops the FROM clause when the query has no WHERE conditions — SQLAlchemy renders plain `SELECT 1`, which yields a single row:

```
POST /tasks?limit=150&allow_partial_count=true   (no filters)
→ count: 1, results: 68     ← count says 1, the same response carries 68 rows
```

```sql
SELECT count(*) FROM (SELECT 1 LIMIT 901) sub;  -- → 1
```

Rocky currently always sends `scheduler_id`, so it doesn't hit this in the UI today — but it's latent in both stores for any other consumer (API users, future callers). **Fix:** project a real column instead of a bare literal, e.g. `.with_entities(models.TaskDB.id)` (keeps FROM, still avoids wide rows), same in `ScheduleStore`.

### 🛑 High — `organisation_ids` rename silently disables org filtering of task stats

`rocky/rocky/scheduler.py:428` renames the query param to `organisation_ids`, but the mula endpoint still accepts `organisation_id` (singular — `mula/scheduler/server/handlers/tasks.py:112`). FastAPI ignores unknown query params, so the filter is dropped. Reproduced with two orgs in the stack:

```
GET /tasks/stats?scheduler_id=normalizer&organisation_id=dvwa    → total 36  (correct, dvwa only)
GET /tasks/stats?scheduler_id=normalizer&organisation_ids=dvwa   → total 48  (= unfiltered, includes org2)
```

Every org-scoped task page will now show statistics that include **other organisations' tasks** — wrong numbers plus a (mild) cross-org information leak. **Fix:** either revert to `organisation_id`, or — if the intent was to make multi-org stats actually work for `get_combined_schedulers_stats` — change the mula handler/store to accept `organisation_ids: list[str]` and filter with `IN`, in the same PR.

### ⚠️ Medium — Still no tests for the bounded-count logic

Repeat of round 1, now with evidence it matters: a store test calling `get_tasks(allow_partial_count=True)` **without filters** would have caught Critical 2 immediately, and a test for the `limit=0` contract would have caught Critical 1. Minimum set: no-filter case, exact-below-cap, at-cap (`is_partial_count=True`), `allow_partial_count=False` regression guard, clamp behaviour.

### ℹ️ Notes
- The `task_type` → `scheduler_id` swaps in `rocky/views/scheduler.py` are semantically neutral (`self.scheduler_id = self.task_type`, line 63) — fine.
- `ReportScheduler.process_rescheduling` 3-tuple unpack — fine.
- The bounded-count SQL construction itself (with the FROM fix) remains a good idiom; Critical 1 and 2 are integration issues, not a flaw in the approach.

### Test Results (local, in Docker)

| Component | Test | Result |
|-----------|------|--------|
| mula | `make utest` | ✅ 77 passed |
| mula | `make itest` | ✅ 198 passed, 3 skipped, 2 failed* |
| rocky | `make utest` | ✅ 470 passed, 1 skipped |

\* `test_clients.py::KatalogusTestCase` — same local Docker DNS race as documented in the PR #5165 round-3 review, reproduced identically on main earlier today; pre-existing environment issue, not this PR.

### Functional Test

| Check | Status | Details |
|-------|--------|---------|
| Stack build | ✅ | 13/13 containers up |
| Log analysis | ✅ | only known noise |
| Task list count path | ❌ | count capped at 7 (Critical 1) |
| Unfiltered count | ❌ | count=1 with 68 results (Critical 2) |
| Stats org filtering | ❌ | `organisation_ids` ignored, cross-org totals (High) |

### Verdict

**CHANGES REQUESTED**

The mula-side bounded count is a good idea and the handler plumbing is now correct, but the Rocky integration breaks the very pages it optimizes: task lists render 7 rows max, and task statistics lose their organisation filter. Both were reproduced live on a fresh stack. With the count-probe fix (page-size-aware), the `with_entities` FROM fix, and the stats param fix (+ a few store tests), this should be ready next round.

underdarknl:
*  High — organisation_ids rename silently disables org filtering of task stats -> Fixed

hasecon (changes requested):
## QA Review — Collin Pipeline (Round 3)

### Setup
- Branch: `feat/mula-partial-counts` @ `f1c2463`
- Stack: `make clean && make kat` — ✅ 13/13 containers up
- Test orgs: `dvwa` + `org2` (two orgs needed to verify stats filtering) — ✅
- All round-2 findings were **re-tested live against the running stack**.

### Round-2 status

| Round-2 finding | Status |
|---|---|
| 🛑 Critical 1 — task lists collapse to 7 rows (`limit=0` probe) | ❌ Not fixed — no commit touches this path; re-reproduced live |
| 🛑 Critical 2 — unfiltered bounded count returns 1 (FROM-loss) | ❌ Not fixed — `with_entities(literal(1))` unchanged in both stores; re-reproduced live |
| 🛑 High — stats org filter dropped | ❌ Still broken — the fix doesn't bind as a query param; see below |
| ⚠️ No tests for bounded-count logic | ❌ Still none |

### What changed since round 2 (7 commits)

- `list` → `get_paged` renames in the three handlers — ✅ correct and necessary: `def list` shadowed the `list` builtin in the class body, so the new `list[str]` annotation on `stats` would otherwise fail at import (`'function' object is not subscriptable`). All renamed endpoints smoke-tested 200.
- stats `organisation_id` → `organisation_ids: list[str]` + `.in_()` in the store — the store logic is right, but the handler binding is wrong:

### 🛑 High — stats fix doesn't work: `organisation_ids` binds as request body, not query param

FastAPI requires an explicit `Query()` for list-typed query parameters; a bare `list[str] | None = None` is interpreted as a **JSON request body** — on a GET endpoint. The generated OpenAPI spec confirms it: `scheduler_id` is `in: query`, but `organisation_ids` sits under `requestBody`.

Live repro with two orgs (ground truth: dvwa=5, org2=24 boefje tasks):

```
GET /tasks/stats?scheduler_id=boefje                        → total 29  (all orgs)
GET /tasks/stats?scheduler_id=boefje&organisation_ids=dvwa  → total 29  ← Rocky's call; filter silently ignored
GET /tasks/stats?scheduler_id=boefje   + JSON body ["dvwa"] → total 5   ← only the (unintended) body binding filters
```

Rocky sends it as a query param (`rocky/rocky/scheduler.py:428`, via httpx `params=`), so every org-scoped stats page still shows **cross-org totals** — the same end-user effect as round 2.

**Fix (one line):**

```python
organisation_ids: list[str] | None = fastapi.Query(None)
```

plus ideally a handler test asserting `?organisation_ids=x` actually filters — that's the test that would have caught both the round-2 and round-3 variants of this bug.

### 🛑 Critical 1 — unchanged, re-reproduced at `f1c2463`

```
POST /tasks?scheduler_id=boefje&limit=0&allow_partial_count=true → count: 7, is_partial_count: true
POST /tasks?scheduler_id=boefje&limit=0                          → count: 24
```

And through Rocky's own code path (Django shell in the rocky container):

```python
l = LazyTaskList(c, scheduler_id='boefje', filters={'filters': {'and': []}})
len(l)      # → 7    ← what Django's Paginator uses
l[0:150]    # → 29 tasks actually returned
```

Same mechanics as round 2: `list_tasks` unconditionally sets `allow_partial_count=true`, the count probe uses `limit=0`, the store clamps it to 1, so `max_count = 6×1+1 = 7`. Task list views still render **at most 7 rows on 1 page** regardless of the real total. The fix direction from round 2 stands: probe with the view's real page size (then the cap is `6×150+1` and the sliding-window pagination works as you intended), or keep `allow_partial_count` opt-in per call site.

### 🛑 Critical 2 — unchanged, re-reproduced at `f1c2463`, also in `ScheduleStore`

```
POST /tasks?limit=150&allow_partial_count=true   (no filters) → count: 1, results: 34
GET  /schedules?limit=1&allow_partial_count=true&max_pages=2  → count: 1   (real count: 50)
```

`query.order_by(None).with_entities(literal(1))` drops the FROM clause when the query has no WHERE → `SELECT 1` → count=1. Present in both `TaskStore` and `ScheduleStore`. Rocky always sends a filter today, so it's latent for the UI but broken for any other API consumer. **Fix:** project a real column, e.g. `.with_entities(models.TaskDB.id)` / `models.ScheduleDB.id`.

### ⚠️ Medium — still no tests for the bounded-count logic

Third round without them, and both criticals plus the stats binding are exactly the bugs the minimal set would catch: no-filter case, `limit=0` contract, at-cap (`is_partial_count=True`), below-cap exact, `allow_partial_count=False` regression guard, and a stats handler test for `organisation_ids`.

### CI/CD Status

24/25 green at `f1c2463` (mula + rocky tests, builds, pre-commit, CodeQL). Only `makelang` red — non-blocking, translation regen to pick up separately.

### Test Results (local, in Docker)

| Component | Test | Result |
|-----------|------|--------|
| mula | `make utest` | ✅ 77 passed |
| mula | `make itest` | ✅ 198 passed, 3 skipped, 2 failed* |
| rocky | `make utest` | ✅ 470 passed, 1 skipped |

\* same `is_host_available(ci_katalogus)` local Docker DNS race as documented in rounds 1–2 and the PR #5165 review — pre-existing environment issue, identical counts to round 2, not this PR. CI's own mula integration tests are green.

### Functional Test

| Check | Status | Details |
|-------|--------|---------|
| Stack build | ✅ | 13/13 containers up |
| Log analysis | ✅ | only known noise + transient startup DNS (self-recovered) |
| Renamed endpoints | ✅ | `GET /tasks`, `/schedules`, `/schedulers` all 200 |
| Task list count path | ❌ | count capped at 7 (Critical 1) |
| Unfiltered bounded count | ❌ | count=1 with 34 results; schedules count=1 vs 50 (Critical 2) |
| Stats org filtering | ❌ | query param ignored — binds as request body (High) |

### Verdict

**CHANGES REQUESTED**

The `get_paged` rename was needed groundwork and the store-side `.in_()` filtering is correct, but all three round-2 issues are still live on this head: Critical 1 and 2 weren't touched, and the stats fix doesn't reach the store because FastAPI binds the bare `list[str]` as a request body instead of a query param — `fastapi.Query(None)` is the missing piece. With that one-liner, the `with_entities(<real column>)` fix, the page-size-aware count probe, and a handful of store/handler tests, this should be ready next round.

hasecon (changes requested):
## QA Review — Collin Pipeline (Round 4)

### Setup
- Branch: `feat/mula-partial-counts` @ `40a5ac0d`
- Verified this round via: local `pre-commit run` on the changed files + direct code inspection of the current head (checked out `40a5ac0d`). I did **not** rebuild the full `make kat` stack this round — two blockers below are conclusive at the code level (one reproduced locally, one a plain code fact), so the PR needs another push regardless. Happy to do a full live functional pass once those are green.

### Round-3 status (each re-checked against `40a5ac0d`, not from comments)

| Round-3 finding | Status |
|---|---|
| 🛑 Critical 1 — task lists collapse to 7 rows (`limit=0` probe → clamp 1 → `max_count=7`) | ✅ Addressed at store level — `TaskStore.get_tasks` now early-returns `[], query.count(), False` for `limit==0`, so the count probe yields the real total (see note below) |
| 🛑 Critical 2 — unfiltered bounded count returns 1 (FROM-loss) | ◐ **Half-fixed** — `TaskStore` now projects `models.TaskDB.id`, but **`ScheduleStore` still uses `with_entities(literal(1))`** (`schedule.py:76`). Still broken there. |
| 🛑 High — stats `organisation_ids` binds as body not query | ✅ Fixed — `organisation_ids: list[str] \| None = fastapi.Query(None)` is now present, and Rocky sends it as a query param (`rocky/scheduler.py:428`). |
| ⚠️ No tests for bounded-count logic | ❌ Still none — the only test changes are 2-tuple → 3-tuple unpacks. 4th round without a test for this logic. |

### 🛑 Blocker 1 (new) — `pre-commit` CI is red, reproduced locally

`pre-commit` is a required check and it fails on this head. Reproduced with `pre-commit run --files …`:

- **trailing-whitespace** — `mula/scheduler/storage/stores/task.py` has two trailing-whitespace lines (the blank line and the `try:` line just above the `if limit == 0` block).
- **ruff (F401)** — `task.py:3` still does `from sqlalchemy import desc, exc, func, literal`, but `literal` is now unused there (TaskStore switched to `models.TaskDB.id`). Ruff removes it.
- **ruff-format** — the `stats(...)` signature in `handlers/tasks.py` gets collapsed back onto one line.

Fix: `pre-commit run --all-files` and commit the result. (This is component-relevant — `pre-commit` red = REQUEST_CHANGES by default per our merge protocol.)

### 🛑 Blocker 2 — Critical 2 still live in `ScheduleStore`

`mula/scheduler/storage/stores/schedule.py:76`:
```python
bounded_query = query.order_by(None).with_entities(literal(1)).limit(max_count)
count = session.query(func.count()).select_from(bounded_query.subquery()).scalar()
```
When the schedule query has no WHERE clause, `with_entities(literal(1))` drops the FROM → SQLAlchemy renders `SELECT 1` → `COUNT(*)` over it returns **1** regardless of the real total (round-3 repro: `GET /schedules?limit=1&allow_partial_count=true&max_pages=2 → count: 1`, real 50). You fixed exactly this in `TaskStore` (`models.TaskDB.id`) but left `ScheduleStore` on `literal(1)`. Apply the same fix here — `.with_entities(models.ScheduleDB.id)` — and drop the now-unused `literal` import from `schedule.py`.

While here, the two stores also diverge on `limit` handling: `TaskStore` clamps `min(max(limit, 0), MAX_LIMIT)` **and** early-returns a full count for `limit == 0`; `ScheduleStore` clamps `min(max(limit, 1), MAX_LIMIT)` with no `limit == 0` path. Worth making consistent so a `limit=0` count-probe behaves the same on both endpoints.

### ⚠️ Medium — still no tests (round 4)

Both criticals across rounds 2–4 are precisely what a small store-level test suite would have caught on the first push. Minimum set: no-filter case (guards the FROM-loss), `limit=0` contract, at-cap (`is_partial_count=True`), below-cap exact, `allow_partial_count=False` regression guard, and a `stats` handler test asserting `?organisation_ids=x` actually filters.

### Note on Critical 1 / the optimization (not a change request, just so we're aligned)

The `limit=0 → full count` early-return removes the round-3 `len()==7` behaviour. Because `RockyPaginator.page()` slices *before* reading `count`, `LazyTaskList.__getitem__` (real `limit`, `allow_partial_count=True`) populates `_count` with the bounded count first, so the paginator uses the bounded value and the `limit=0` full count acts as a fallback for direct `count` access. So the correctness bug looks resolved and the optimization path is still reachable — I'd just like to confirm this live in the next round once the stack is worth rebuilding (i.e. after the blockers above are fixed).

### CI/CD Status

| Check | Status | Notes |
|---|---|---|
| `pre-commit` | ❌ failure | Blocker 1 — trailing whitespace + unused import + format (reproduced locally) |
| `makelang` | ❌ failure | Non-blocking per merge protocol (translation regen, `make lang` follow-up) |
| `test (3.10–3.13)` mula + rocky | ✅ success | |
| `build (*)` all distros, `create_container_image`, `CodeQL`, `Analyze (*)` | ✅ success | |

### Verdict

**CHANGES REQUESTED**

Good progress — the stats query-param fix and the `TaskStore` FROM-fix + `limit=0` handling are correct. Two things block this round: `pre-commit` is red (trailing whitespace + a now-unused `literal` import in `task.py`), and the identical FROM-loss bug is still live in `ScheduleStore` (only `TaskStore` was fixed). With those two + the missing tests, this should be ready — ping me and I'll re-run with a full functional pass.

hasecon (approved):
## QA Review — Collin Pipeline (Round 5)

### Setup
- Branch: `feat/mula-partial-counts` @ `2b3bc1f`
- Stack: `make clean && make kat` — ✅ 13/13 containers up, Rocky HTTP 200
- Test orgs: `dvwa` + `org2` (two orgs to verify stats filtering) — ✅
- As promised in round 4: **full live functional pass** — every previous finding re-tested against the running stack, not just read from the diff.

### Round-4 status (each verified live at `2b3bc1f`)

| Round-4 finding | Status |
|---|---|
| 🛑 Blocker 1 — `pre-commit` CI red (trailing whitespace, unused `literal` import, format) | ✅ Fixed — `pre-commit` green in CI; both stores' imports clean at head |
| 🛑 Blocker 2 — FROM-loss still live in `ScheduleStore` | ✅ Fixed & verified live — `with_entities(models.ScheduleDB.id)`; `GET /schedules?limit=1&allow_partial_count=true&max_pages=2` now returns the real count (round-3 repro returned `1`) |
| Store consistency (`limit` clamp + `limit==0` path diverged between stores) | ✅ Fixed — both stores now share `min(max(limit, 0), MAX_LIMIT)` and the `limit == 0` → exact-count early return |
| ⚠️ No tests for bounded-count logic | ❌ Still none (5th round) — see recommendation below |

### Verified this round (live, on the running stack)

**ScheduleStore bounded count (Blocker 2):**
```
GET /schedules?limit=1                                        → count: 2, exact        (baseline)
GET /schedules?limit=1&allow_partial_count=true&max_pages=2   → count: 2, partial: false  (was 1 — fixed)
GET /schedules?limit=1&allow_partial_count=true&max_pages=1   → count: 2 = max_count, partial: true  (sentinel works)
GET /schedules?limit=0&allow_partial_count=true               → exact count, 0 results   (probe contract)
```

**TaskStore regression checks (rounds 2–3 criticals stay fixed):**
```
POST /tasks?limit=150&allow_partial_count=true  (no filters)  → count: 24 = real (was 1)
POST /tasks?scheduler_id=boefje&limit=0&allow_partial_count=true → count: 5 = ground truth (was 7-capped)
POST /tasks?limit=1&max_pages=1&allow_partial_count=true      → count = max_count, partial: true
```

**Rocky code path (Django shell in the rocky container):**
```python
l = LazyTaskList(c, scheduler_id='boefje', filters={'filters': {'and': []}})
len(l)      # → 5  = real count (rounds 2–3 gave the bogus 7 cap)
l[0:150]    # → 5 tasks, count stays consistent after slice
```
`RockyPaginator.page()` slices before reading `count`, so the paginator consumes the bounded count from the real-page-size request and the `limit=0` probe is only a fallback that now returns an exact count — the optimization path is reachable and correct, confirming the round-4 code-level analysis.

**Stats org filtering (round-2/3 High):** ground truth dvwa=5, org2=2 boefje tasks:
```
GET /tasks/stats?scheduler_id=boefje                                          → 7 (all orgs)
GET /tasks/stats?scheduler_id=boefje&organisation_ids=dvwa                    → 5 ✅
GET /tasks/stats?scheduler_id=boefje&organisation_ids=org2                    → 2 ✅
GET /tasks/stats?scheduler_id=boefje&organisation_ids=dvwa&organisation_ids=org2 → 7 ✅ (multi-value binds)
```

**Rocky views:** `/tasks/`, `/tasks/boefjes`, `/tasks/normalizers` (org-scoped) and `/en/tasks/` (all-orgs, exercises the multi-org stats path) all render 200 via authenticated Django test client, no error messages.

### CI/CD Status

| Check | Status | Notes |
|---|---|---|
| `pre-commit` | ✅ | Round-4 blocker cleared |
| `test (3.10–3.13)` mula + rocky, all builds, `create_container_image`, `CodeQL`, `Analyze (*)` | ✅ | |
| `makelang` | ❌ | Same translation-regen failure as rounds 1–4; non-blocking per merge protocol — please pick up `make lang` in a follow-up |

Note: the branch is currently behind `main`; a branch update before merge is advisable but CI ran on the merge ref and is green.

### Test Results (local, in Docker)

| Component | Test | Result |
|-----------|------|--------|
| mula | `make utest` | ✅ 77 passed |
| mula | `make itest` | ✅ 198 passed, 3 skipped, 2 failed* |
| rocky | `make utest` | ✅ 470 passed, 1 skipped |

\* Same `KatalogusTestCase` local Docker DNS race as documented in rounds 2–3 (pre-existing environment issue, identical counts; CI's mula integration tests are green on this head).

### Functional Test

| Check | Status | Details |
|-------|--------|---------|
| Stack build | ✅ | 13/13 containers up |
| Log analysis | ✅ | Only known noise (bcrypt warning, RabbitMQ EOL notice, one self-recovered startup DNS race in rocky_worker) |
| Schedule bounded count | ✅ | Real count, sentinel correct |
| Task bounded count | ✅ | Real count, no filters case fixed |
| `limit=0` probe contract | ✅ | Exact count, both stores |
| LazyTaskList / paginator | ✅ | Correct count through Rocky's own code path |
| Stats org filtering | ✅ | Single + multi-org query params filter correctly |
| Task list views | ✅ | 200, no errors |

### ⚠️ Remaining recommendation — tests for the bounded-count logic (not blocking, but please don't skip it)

Fifth round without unit tests for this logic, and rounds 2–4 are the argument for them: every critical found in this PR (FROM-loss ×2, the `limit=0` collapse, the stats param binding) would have been caught by a small store/handler test set. Concretely, in `mula/tests/integration/test_task_store.py` / `test_schedule_store.py`:
1. `get_tasks(allow_partial_count=True)` **without filters** → count equals real total (guards the FROM-loss);
2. `limit=0` → exact count, empty results;
3. at-cap → `is_partial_count=True`, count == `offset + max_pages*limit + 1`;
4. below-cap → exact count, `is_partial_count=False`;
5. `allow_partial_count=False` → exact count (regression guard);
6. a `stats` handler test asserting `?organisation_ids=x` filters.

Happy to contribute these as a follow-up PR if that helps.

### Verdict

**APPROVED**

All blockers from rounds 2–4 are fixed and were re-verified live on a fresh stack this round: both stores produce correct bounded counts (including the previously broken unfiltered case), the `limit=0` probe contract is restored, the partial-count sentinel behaves as designed, Rocky's paginator consumes the bounded count through the intended sliding-window path, and stats org filtering works for single and multiple organisations. CI is green apart from the long-standing non-blocking `makelang`. The missing test coverage is the one open item — strongly recommended as an immediate follow-up (offer stands to write it).

PR #5176, opened 2026-05-20
Add NoOperation Exception that allows bits to signal they will never …

underdarknl:
…act on the given inputset. This keeps the logic neatly in python, and limits bit runs to a single run, instead of repeatedly firing them of when ooi's are mutated.

### Changes

### Issue link

Closes ...

### Demo

### QA notes

---

### Code Checklist

---

---

## Checklist for QA:


hasecon (changes requested):
## QA Review — Collin Pipeline

### Setup
- Branch: `feat/bit-noop` @ `2018854d`
- Pre-commit reproduced locally — see mypy errors below
- Octopoes unit tests reproduced in `openkat-reviev-octopoes_api-1`
- Functional test (`make kat`) skipped — the PR is broken at unit/lint level and the BitNoOperation mechanism is not wired into the inference loop, so functional behaviour would not exercise the new path

### CI/CD Status

PR touches `octopoes/**`. Per Merge Protocol Rule 1, the following component-relevant checks **must** be green; per Rule 4 ("optimization PRs: any red test for the optimized component"), default verdict is automatically **REQUEST_CHANGES**:

| Check | Status |
|---|---|
| `pre-commit` | ❌ mypy 4 errors |
| `rtest` (Octopoes robot) | ❌ |
| Octopoes `test (3.10)` | ❌ |
| Octopoes `test (3.11)` | ❌ |
| Octopoes `test (3.12)` | ❌ |
| Octopoes `test (3.13)` | ❌ |
| (other components / builds) | ✅ |

Main is currently green on `octopoes_tests.yml` (last 2 runs `success`) — these failures are introduced by this PR and are not flake.

### Critical findings

#### 🔴 1. `BitNoOperation` is never caught — the entire noop mechanism is not wired up

`save_origin_as_noop()` is defined in `octopoes/octopoes/core/service.py:213` but has **0 callers** anywhere in the repo (verified by grep). `_run_inference` has only a generic handler at line 293:

```python
except Exception as e:
    logger.exception("Error running inference", exc_info=e)
    raise e
```

So when any of the 14 modified bits raises `BitNoOperation`:
1. It is caught by `except Exception` (BitNoOperation extends ValueError extends Exception).
2. **Logged as an ERROR** — every noop produces log spam.
3. **Re-raised** — propagates up to the event handler, leaving the origin's previous result intact rather than clearing it.

Concrete consequence: when a bit that previously returned `[]` (cleaning its old findings) now raises `BitNoOperation`, the previous findings **stay in the graph indefinitely**. Old findings are not cleared.

The correct shape would be:

```python
try:
    resulting_oois = BitRunner(bit_definition).run(source, parameters, config=config)
    self.save_origin(origin, resulting_oois, valid_time)
    return True
except BitNoOperation as e:
    logger.info("Bit noop: %s", e)
    self.save_origin_as_noop(origin, valid_time)
    return False
except Exception as e:
    logger.exception("Error running inference", exc_info=e)
    raise e
```

#### 🔴 2. Once 1 is fixed: deleting the origin breaks bit re-triggering when an additional_oois changes (Edward's question)

The QA concern Edward raised before this review was: *if a bit consumes multiple origins (input_ooi + additional_oois), does the bit re-fire when one of them changes?* Tracing through the dispatch:

`_on_update_ooi` (service.py:481-486) collects bits to re-run via:

```python
inference_params = self.origin_parameter_repository.list_by_reference(
    event.new_data.reference, valid_time=event.valid_time
)
for inference_param in inference_params:
    inference_origins.append(self.origin_repository.get(inference_param.origin_id, event.valid_time))
```

That is, the system looks up *which origins reference this OOI as a parameter*, then re-runs them. **Those origin_parameter rows live alongside the origin** — if `save_origin_as_noop` later deletes the origin, the origin_parameter rows are useless (their `origin_id` no longer exists; the `origin_repository.get(...)` call would raise `ObjectNotFoundException`, which is suppressed).

Concrete failure with `missing_caa`:
1. `Hostname|example.com` exists, `NXDOMAIN` for it exists → bit raises `BitNoOperation("We have NXDomains, as such we do not have CAA.")` → would delete the origin.
2. DNS recovers, `NXDOMAIN` is deleted → `_on_delete_ooi` fires.
3. `list_by_reference(nxdomain_ref)` returns whatever origin_parameter rows still exist — but the missing_caa origin is gone, so its parameter binding either doesn't exist or points to nothing.
4. **`missing_caa` is never re-evaluated** — the hostname is permanently stuck in "not evaluated" rather than transitioning to "missing CAA finding".

The PR description's claim (*"as an origin is a set of ooi+adjacent ooi's, meaning we don't need to cleanup, or trigger events"*) is incorrect for any bit whose decision depends on `additional_oois`. That is the majority of the modified bits (missing_caa, missing_dkim, missing_dmarc, missing_spf, nxdomain_flag, check_csp_header, check_hsts_header, missing_headers, ipv6_nameservers, https_redirect, website_discovery, disallowed_csp_hostnames, oois_in_headers, spf_discovery — all 14).

Suggested fix: keep the origin with empty result (current "return" behaviour preserves origin + origin_parameter rows so triggers keep working). The PR's optimisation goal — avoiding repeated executions when nothing changes — should be addressed by deduplicating events / debouncing inference runs, not by deleting origins.

#### 🔴 3. mypy errors (4) in `_run_inference` — all confirm dead/incorrect code

Local `pre-commit run --from-ref origin/main --to-ref HEAD` reproduces:

```
octopoes/octopoes/core/service.py:244: error: Invalid index type "str" for "Reference"; expected type "SupportsIndex | slice"  [index]
octopoes/octopoes/core/service.py:244: error: Invalid index type "str" for "str"; expected type "SupportsIndex | slice"  [index]
octopoes/octopoes/core/service.py:251: error: Unsupported operand types for > ("int" and "None")  [operator]
octopoes/octopoes/core/service.py:282: error: Item "None" of "OOI | None" has no attribute "model_dump_json"  [union-attr]
```

- L244 `if "scan_profile" in origin.source: level = int(origin.source["scan_profile"]["level"])` — `origin.source` is `Reference` (a string-like). `in` returns substring presence; `[...]` raises. This branch is dead in normal operation and crashes if ever reached.
- L251 `if level < bit_definition.min_scan_level:` — `level` can be `None` because the `if not level:` guard above doesn't actually set it in the `"scan_profile" in origin.source` branch when that branch is dead.
- L282 `source.model_dump_json()` — `source` declared `OOI | None`, and the `if not source:` guard at L264 may leave it None if neither passed-in nor refetched.

Also: `if not level:` (L251) treats `level=0` as falsy and re-fetches, defeating the optimisation when callers explicitly pass `level=0`. Use `if level is None:` instead.

#### 🔴 4. Unit tests for affected bits not updated → `test (3.10-3.13)` fail

Reproduced in `openkat-reviev-octopoes_api-1`:

```
FAILED tests/test_bit_csp_header.py::test_check_csp_headers
FAILED tests/test_bit_csp_header.py::test_check_csp_headers_non_xss_capable
```

The tests call `run(...)` directly and expect `[]` when no CSP header is present. The new code raises `BitNoOperation` instead. Same pattern likely affects `test_bit_missing_headers.py` and any other bit-level unit test that exercises a path the PR converted from `return` to `raise`. Either the tests need to assert `pytest.raises(BitNoOperation)`, or the bit signatures should keep returning `[]` for noop cases.

#### 🟡 5. `https_redirect` logic inversion is more permissive than before

```python
# OLD: only react to http
if input_ooi.scheme.value != "http" or not header_keys:
    return
# NEW: only short-circuit on https
if input_ooi.scheme.value == "https" or not header_keys:
    raise BitNoOperation("Already https")
```

Old: bit only ran for `scheme == "http"`. New: bit runs for any scheme that is not `"https"` — i.e. it now also runs for `ws`, `wss`, `file`, etc. URLs. Either intentional (then please document) or a regression.

#### 🟡 6. `missing_headers` behaviour change — Location-response now triggers HSTS/deprecated-headers checks

Old code did `if "location" in header_keys: return` at the top, skipping ALL header checks for redirect responses. New code only skips X-Frame-Options, CSP, Permissions-Policy, Referrer-Policy, X-Content-Type-Options on Location responses, but HSTS and deprecated-headers checks now also fire for redirects. That likely produces new findings on 301/302 responses that previously had none. Intentional?

#### 🟡 7. `missing_caa` description change is OK but the typo "fiding" in the comment ("only add fiding for rhe domain itself")

Minor cosmetic.

### Test Results

| Component | Test | Result |
|---|---|---|
| Octopoes | mypy via pre-commit | ❌ 4 errors in core/service.py |
| Octopoes | `tests/test_bit_csp_header.py::test_check_csp_headers` (docker exec) | ❌ AssertionError on BitNoOperation |
| Octopoes | `tests/test_bit_csp_header.py::test_check_csp_headers_non_xss_capable` (docker exec) | ❌ same root cause |
| Octopoes | rest of `tests/` (non-API, non-integration) | ✅ 130 passed |
| CI on main | `octopoes_tests.yml` recent runs | ✅ green — proves none of the failures are pre-existing |

### Verdict

**REQUEST_CHANGES** — three independent reasons, each individually sufficient per Merge Protocol Rule 1:

1. Pre-commit, rtest, and all 4 Python versions of `test` are red on a PR that touches `octopoes/**`.
2. This is explicitly an **optimization PR** for octopoes, with optimized-component tests red — Rule 4 red alert applies.
3. The headline mechanism (`BitNoOperation` → `save_origin_as_noop` → origin deletion) is **only partially implemented** — the bits raise but the service never catches the exception, so the optimisation it claims to deliver doesn't actually run, and instead every noop logs as an error and bubbles up.

Before re-review, please:

1. Wire `BitNoOperation` into `_run_inference` (catch before generic `Exception`, call `save_origin_as_noop`).
2. **Reconsider the deletion strategy** — origin deletion breaks the additional_oois change-trigger path (Edward's QA concern). Keeping the origin with empty result is the safer default; if the goal is to avoid repeated identical executions, deduplicate at the event-emission layer instead.
3. Fix the 4 mypy errors in `core/service.py` (dead `if "scan_profile" in origin.source` branch, `if not level:` should be `if level is None:`, type-narrow `source` / `level` before use).
4. Update the bit unit tests that now raise `BitNoOperation` to assert the exception (or keep the bits returning `[]`).
5. Verify the `https_redirect` and `missing_headers` semantic changes are intentional, and update the bit docs / changelog.
6. Re-run `rtest` locally (`cd octopoes && make rtest`) to confirm the integration story.

PR #5179, opened 2026-06-01
Add debian13 / Ubuntu 24.04 build targets with their respective python versions

underdarknl:
also fix maintainer and various monorepos changes

### Changes

This adds two new build targets, debian 13, and ubuntu24.04, which build with their respective Python versions.

### Issue link

On Newer ubuntu verions the python version 3.10 is no longer available and cannot be installed.

Closes ...

### Demo

_Please add some proof in the form of screenshots or screen recordings to show (off) new functionality, if there are interesting new features for end-users._

### QA notes

_Please add some information for QA on how to test the newly created code._

---

### Code Checklist

- [ ] All the commits in this PR are properly PGP-signed and verified.
- [ ] This PR only contains functionality relevant to the issue.
- [ ] I have written unit tests for the changes or fixes I made.
- [ ] I have checked the documentation and made changes where necessary.
- [ ] I have performed a self-review of my code and refactored it to the best of my abilities.

- [ ] Tickets have been created for newly discovered issues.
- [ ] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [ ] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [ ] I have included comments in the code to elaborate on what is not self-evident from the code itself, including references to issues and discussions online, or implicit behavior of an interface.

---

## Checklist for code reviewers:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_qa.md) into your comment.

PR #5185, opened 2026-06-03
Upgrade RabbitMQ to 3.13 (#5151 step 1 of 3)

hasecon:
### Changes

First of three staged PRs that move OpenKAT off RabbitMQ 3.12 (EOL since June 2025) toward 4.3, as proposed in #5151. RabbitMQ does not support a direct 3.12 → 4.3 hop on a persistent volume; the supported chain is 3.12 → 3.13 → 4.2 → 4.3, with feature-flag work in between.

| Release | Bump | Operator action after pull |
|---|---|---|
| **1.23** (this PR) | 3.12 → 3.13 | `rabbitmqctl enable_feature_flag all` (do not enable `khepri_db`) |
| 1.24 (follow-up) | 3.13 → 4.2 | `rabbitmqctl enable_feature_flag khepri_db` |
| 1.25 (follow-up) | 4.2 → 4.3 | nothing |

This PR contains:

- **All 7 compose files** (`docker-compose.yml`, `docker-compose.release-example.yml`, and the 5 `*/.ci/docker-compose.yml`) bumped to `rabbitmq:3.13-management`.
- **New named volume `rabbitmq-data`** mounted at `/var/lib/rabbitmq` in `docker-compose.yml` and `docker-compose.release-example.yml`. Previously the broker stored state in the container filesystem, so every recreate (including this image bump) silently wiped queues and any in-flight messages. With the volume in place, broker state survives the 3.13 → 4.2 → 4.3 upgrades.
- **`docs/source/about-openkat/release-notes/1.23.rst`** — staged migration plan, post-upgrade `enable_feature_flag all` step with the Khepri warning, explicit drain-queues procedure (loop on `list_queues messages = 0`, `export_definitions` backup), health-check commands and rollback notes.
- **`docs/source/about-openkat/release-notes/1.22.rst`** — stub to close the toctree gap (1.22 was released 2026-04-30 and never got a notes file).
- **`docs/source/installation-and-deployment/production-debian-environment.rst`** — replace the bare `apt install rabbitmq-server` with the Team RabbitMQ `deb1/deb2.rabbitmq.com` apt repository setup, pinned to `3.13.*`. The distro archives ship RabbitMQ 3.9 (Ubuntu 22.04), 3.10 (Debian 12), 3.12 (Ubuntu 24.04) and 3.13 (Debian 13), none of which can follow the 3.13 → 4.2 → 4.3 path on their own.
- **`.github/workflows/test_debian_packages_on_ubuntu.yml`** — switch the deb-install test from stock-apt RabbitMQ 3.9 to the same Team RabbitMQ repos pinned to `3.13.*`, so CI actually exercises the version we are shipping.

### Issue link

Implements step 1 of #5151.

### Dependency on #5179

This PR hardcodes `jammy` (Ubuntu 22.04) in the apt source URLs in `test_debian_packages_on_ubuntu.yml`, because that is currently the only `matrix.os` value in the workflow. Once #5179 (debian13 / ubuntu24.04 build targets) merges, the apt source line needs a small follow-up to make the codename conditional on `matrix.os` so the workflow also tests the new distros. This is intentionally not bundled here — the matrix expansion belongs with #5179.

The `production-debian-environment.rst` already lists `noble` / `bookworm` / `trixie` alongside `jammy`, so the operator docs are forward-compatible with #5179 today.

### QA notes

I ran the upgrade locally against a 6-day-old stack with the DVWA test organisation:

- `docker compose up -d rabbitmq` recreated the container; it came up healthy in ~15 s reporting `RabbitMQ version: 3.13.7`.
- All 13 dependent services (rocky, mula, bytes, octopoes_api, octopoes_api_worker, scheduler, boefje, normalizer, etc.) reconnected automatically with no restart loops.
- The durable queues (`octopoes`, `raw_file_received`, `scan_profile_mutations`) were re-declared by the services on reconnect; visible in `rabbitmqctl list_queues -p kat`.
- `rabbitmqctl enable_feature_flag all` enabled the stable feature flags (`message_containers`, `message_containers_deaths_v2`, etc.) and correctly **skipped** `khepri_db` because it is still experimental on 3.13. `rabbitmqctl list_feature_flags | grep khepri_db` confirmed `disabled`.
- No new errors in `docker compose logs` related to AMQP, channel close, or reconnect.

Note on the volume change: existing stock-compose operators upgrading to 1.23 will get a fresh empty `rabbitmq-data` volume on first pull. Their pre-upgrade queue state lived in the container filesystem and is not migrated, but in practice the OpenKAT services re-declare the durable queues on startup; only in-flight messages at the moment of upgrade are lost. This is documented in the release notes.

### Self-review caught and fixed before pushing

- The Cloudsmith `ppa1.rabbitmq.com` URLs the team historically used were shut down on 2025-11-01 (Team RabbitMQ apt repos migrated to `deb1/deb2.rabbitmq.com` with a single unified signing key `0A9AF2115F4687BD29803A206B73A36E6026DFCA`). The workflow and the operator docs use the current URLs.
- The release notes initially wrote `enable_feature_flags all` (plural) which RabbitMQ rejects with "Did you mean 'enable_feature_flag'?". Caught while running it against the live 3.13 broker; fixed to the singular form.

---

### Code Checklist


hasecon (commented):
## QA Review

Reviewing our own PR per the OpenKAT review protocol. Posted as `COMMENT` rather than `APPROVE` because (a) the protocol's "all component-relevant CI checks are green" requirement is still being satisfied (CI just restarted on the awk fixup commit), and (b) GitHub does not allow PR authors to formally approve their own PR. The Cloudsmith URL fix and the awk-loop fix below were both caught by this self-review before any external reviewer time was spent.

### Setup
- Branch: `chore/rabbitmq-3.13-upgrade` @ `9471b1336`
- Stack: tested in place on a 6-day-old running stack with DVWA test org — `docker compose up -d rabbitmq` recreated the broker cleanly on `rabbitmq:3.13.7-management`
- Test org: `setup_test_org --code dvwa` (already existed in baseline state)
- pre-commit: `pre-commit run --files <changed>` → all green (run twice across both commits)

### CI/CD Status

Per the awk fixup push, CI is re-running on `9471b1336`. Prior run on `a684d8765` reached 31/46 passed, 1 failed, 14 still running before being superseded. The 1 failed check is `makelang` which is documented in our CLAUDE.md as a non-blocking translation-regen issue (follow-up with `make lang` is sufficient).

| Check | Status (a684d8765) |
|---|---|
| `pre-commit` | success |
| `CodeQL` (python / javascript-typescript / actions) | success |
| `test (3.10/3.11/3.12/3.13)` — Octopoes | success |
| `Tests (3.11/3.12/3.13)` — Boefjes | success (3.10 was still in-progress at push time) |
| `build-artifact`, `create_container_image` (x4) | success |
| `build (ubuntu22.04, bytes/mula)` | success — relevant: confirms compose tag bump didn't break itest stack |
| `build (debian12, *)` + 5 more | still queued at push time |
| `makelang` | failed (non-blocking) |

`test_debian_packages_on_ubuntu` does **not** run on this PR — the workflow's `pull_request:` trigger only fires on PRs to `release**` branches, not `main`. This means our most critical change (the Cloudsmith URL swap in that workflow file) will only be exercised after the PR merges to main. Mitigation: I've separately verified the deb1/deb2 URLs return 200 from curl and match the canonical upstream `https://www.rabbitmq.com/docs/install-debian` snippet for jammy.

### Code Review Findings

Three rounds of `feature-dev:code-reviewer` agent passes were run; findings filtered after empirical verification:

| Severity | Finding | Status |
|---|---|---|
| Critical | `ppa1.rabbitmq.com` URLs are dead since 2025-11-01 (Cloudsmith mirror EOL) — caught in round 1 | **Fixed in `a684d8765`** — replaced with `deb1+deb2.rabbitmq.com` and single key `0A9AF...DFCA` per upstream docs |
| Critical | `enable_feature_flags all` (plural) typo — caught in functional test against live broker, rabbitmq rejected with "Did you mean 'enable_feature_flag'?" | **Fixed in `a684d8765`** — singular form throughout |
| Important | `until ... awk '{ s += $1 } END { print s }'` drain loop hangs forever when broker has no queues (empty input → awk prints `""`, not `"0"`) — caught in round 3 | **Fixed in `9471b1336`** — `print s+0` coerces empty to `0`, verified empirically |
| Important | Heredoc terminator may not be at column 0 in workflow yaml — agent claim in rounds 2 and 3 | **False positive** — verified empirically with `python -c "import yaml; yaml.safe_load(...)"` that YAML `\|` block scalar strips the leading whitespace from the body, EOF lands at col 0 in the resulting bash, heredoc terminates correctly |
| Nit | Stale `Debian 11 / bullseye` paragraph in `production-debian-environment.rst` lines 18-22 references a December 2023 end-of-support date | Acknowledged, out of scope for this PR — should be a follow-up cleanup |
| Nit | No explicit `rabbitmq-diagnostics check_running` gate before `enable_feature_flag all` in release notes | Acknowledged, low-impact — operators starting from a known-good stack don't need this; documented health-check at end of procedure covers post-upgrade verification |

### Functional Test

Performed against the live 6-day stack (not `make clean && make kat`, to preserve the existing DVWA test data and to better simulate a real operator upgrade):

| Check | Result | Detail |
|---|---|---|
| 3.12 → 3.13 image swap | ✅ | `docker compose up -d rabbitmq` recreated container; healthy in ~15s |
| Version reported | ✅ | `rabbitmqctl version` → `3.13.7` |
| Service reconnect | ✅ | All 13 dependent services stayed "Up 6 days"; rocky/mula/bytes/octopoes/boefje/normalizer/scheduler all reconnected to the new broker without restart |
| Durable queue redeclare | ✅ | `octopoes`, `raw_file_received`, `scan_profile_mutations` reappeared in `rabbitmqctl list_queues -p kat` |
| `enable_feature_flag all` | ✅ | Enabled `message_containers`, `message_containers_deaths_v2`, etc.; **`khepri_db` correctly stayed `disabled`** (auto-skipped because experimental on 3.13, exactly as the release notes warn) |
| Khepri non-activation | ✅ | `rabbitmqctl list_feature_flags \| grep khepri` → `khepri_db disabled` after enable-all |
| Error scan | ✅ | `docker compose logs --since 60s 2>&1 \| grep -iE '(error\|exception\|traceback\|amqp.*reconnect\|channel.*closed)'` → only normal rabbitmq startup info, no service-side AMQP errors |
| Smoke test 3.13 image standalone | ✅ | Pre-test before live upgrade: `docker run --rm rabbitmq:3.13-management` boots, reports 3.13.7 |

The volume change was not retroactively applied to the running container (which still uses the ephemeral container fs from the 6-day uptime); first proper test of the volume happens on the next operator pull-and-recreate, where the named volume will be created fresh and start surviving subsequent recreates.

### Verdict

**COMMENT** (waiting for CI on `9471b1336` to complete; expect APPROVE-equivalent state once green except `makelang`).

The PR is in a state I would recommend merging once CI completes on the awk fixup, with the standard `makelang` follow-up. Three real issues were caught by the self-review and fixed before any external reviewer time was spent.

@underdarknl — when you pick this up, the things worth a fresh look from your side:
1. The Cloudsmith URL/key swap (deb1+deb2.rabbitmq.com with `0A9AF...DFCA`) — second pair of eyes welcome.
2. The `test_debian_packages_on_ubuntu.yml` only triggers on PRs to `release**`, so the workflow's RabbitMQ install path is not validated by this PR's CI. Worth a manual `workflow_dispatch` run against this branch before merge if you want pre-merge confidence.
3. The pre-existing `rabbitmq-data` volume gap in compose files — included here because it's directly load-bearing for the staged migration plan (without it, every operator upgrade silently wipes in-flight messages). Scope is debatable; happy to split out if you'd rather keep this PR pure-tag-bump.

underdarknl:
Manually running the 'test_debian_packages_on_ubuntu' tests reveals the following for ubuntu 22.04:

```

The following packages have unmet dependencies:
 rabbitmq-server : Depends: erlang-base (< 1:27.0) but 1:27.3.4.11-1 is to be installed or
                            erlang-base-hipe (< 1:27.0) but it is not installable or
                            esl-erlang (< 1:27.0) but it is not installable
                   Depends: erlang-crypto (< 1:27.0) but 1:27.3.4.11-1 is to be installed or
                            esl-erlang (< 1:27.0) but it is not installable
                   Depends: erlang-eldap (< 1:27.0) but 1:27.3.4.11-1 is to be installed or
                            esl-erlang (< 1:27.0) but it is not installable
                   Depends: erlang-inets (< 1:27.0) but 1:27.3.4.11-1 is to be installed or
                            esl-erlang (< 1:27.0) but it is not installable
                   Depends: erlang-mnesia (< 1:27.0) but 1:27.3.4.11-1 is to be installed or
                            esl-erlang (< 1:27.0) but it is not installable
                   Depends: erlang-os-mon (< 1:27.0) but 1:27.3.4.11-1 is to be installed or
                            esl-erlang (< 1:27.0) but it is not installable
                   Depends: erlang-parsetools (< 1:27.0) but 1:27.3.4.11-1 is to be installed or
                            esl-erlang (< 1:27.0) but it is not installable
                   Depends: erlang-public-key (< 1:27.0) but 1:27.3.4.11-1 is to be installed or
                            esl-erlang (< 1:27.0) but it is not installable
                   Depends: erlang-runtime-tools (< 1:27.0) but 1:27.3.4.11-1 is to be installed or
                            esl-erlang (< 1:27.0) but it is not installable
                   Depends: erlang-ssl (< 1:27.0) but 1:27.3.4.11-1 is to be installed or
                            esl-erlang (< 1:27.0) but it is not installable
                   Depends: erlang-syntax-tools (< 1:27.0) but 1:27.3.4.11-1 is to be installed or
                            esl-erlang (< 1:27.0) but it is not installable
                   Depends: erlang-tools (< 1:27.0) but 1:27.3.4.11-1 is to be installed or
                            esl-erlang (< 1:27.0) but it is not installable
                   Depends: erlang-xmerl (< 1:27.0) but 1:27.3.4.11-1 is to be installed or
                            esl-erlang (< 1:27.0) but it is not installable
E: Unable to correct problems, you have held broken packages.
```

hasecon:
Thanks for running it — that's a real bug in the PR, fixed in `cd2c3f219`.

**Root cause:** the `rabbitmq-erlang` repo serves both Erlang 26.x and 27.x, and without a pin apt picks the newest (27.3.4.11). But `rabbitmq-server=3.13.*` depends on `erlang-base (< 1:27.0)`, so the install fails exactly as your log shows. RabbitMQ didn't gain Erlang 27 support until 4.0.

**Fix:** add an apt preferences pin for `erlang*` to the 26 series before installing:

```
Package: erlang*
Pin: version 1:26.*
Pin-Priority: 1000
```

The repo currently has `1:26.2.5.14-1` available for jammy (I checked the `Packages` index), so the pin resolves to a runtime 3.13 accepts. Applied in both places: `test_debian_packages_on_ubuntu.yml` and the operator docs (`production-debian-environment.rst`).

Could you re-trigger the `workflow_dispatch` run on this branch (`cd2c3f219`) to confirm green? I can't run that path locally on macOS.

**One open question (Debian 13):** the doc note currently says trixie doesn't need the `rabbitmq-erlang` sources because the distro ships a recent Erlang. But trixie ships Erlang **27**, which `rabbitmq-server` 3.13 would reject for the same reason. That path isn't exercised by this workflow (jammy only), so I left the note as-is rather than guess — but for the 3.13 step it looks like Debian 13 also needs the `rabbitmq-erlang` repo + the Erlang-26 pin. Want me to update that note too, or handle Debian 13 in the 1.24 PR where we move to 4.2 (which does support Erlang 27)?

underdarknl:
im running the workflow now in the following action:
https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/actions/runs/28615241181

underdarknl:
During the test, ubuntu 24.04 now errors out in the step 'setting up rabbitMQ' The install of both rabbitMQ and erlang seem to have completed without issue, however epmd seems to be not running.

```
Run echo "export ERL_EPMD_ADDRESS=127.0.0.1" | sudo tee -a /etc/rabbitmq/rabbitmq-env.conf
export ERL_EPMD_ADDRESS=127.0.0.1
export NODENAME=rabbit@localhost
epmd: Cannot connect to local epmd
Error: Process completed with exit code 1.
```
This test previously completed without issue but now repeatedly fails, but is also canceling the the ubuntu 22.04 test due to how the workflow is constructed.

hasecon:
Thanks for running it and catching this — pushed a fix in `09bd038a3`.

**Diagnosis:** this is the `#5179` follow-up flagged in the PR description. When you merged `main` in (`68608244`), `#5179`'s matrix expansion added the **ubuntu-24.04** runner to `test-deb-install`, but the workflow still hardcoded `jammy` in every apt source line. On the noble runner it therefore installs jammy-built erlang/rabbitmq debs; those come up "installed" but `rabbitmq-server` can't actually start, so epmd is never running by the time the *Prepare rabbitmq configurations* step reaches `sudo epmd -kill` → `Cannot connect to local epmd` → exit 1. Because the matrix was fail-fast, that also cancelled the (fine) 22.04 leg.

**Fix (`09bd038a3`, workflow only):**
1. **Codename from `matrix.os`** — new step maps `ubuntu-22.04 → jammy`, `ubuntu-24.04 → noble`, and the apt sources now use `${RMQ_CODENAME}` instead of hardcoded `jammy`. This is the actual 24.04 fix (the team repo publishes a `noble` suite; our operator docs already map 24.04 → noble). Erlang stays pinned to `1:26.*`.
2. **`sudo epmd -kill || true`** — the step's only purpose is to make epmd restart bound to `ERL_EPMD_ADDRESS`; a not-yet-running epmd shouldn't abort it.
3. **`fail-fast: false`** on the `test-deb-install` matrix — so one runner failing no longer cancels the other, and we get results for both distros in a single run.

I can't exercise this path locally (macOS, and it needs the runner + systemd), so could you re-trigger the `workflow_dispatch` on `09bd038a3` to confirm both legs go green? If noble surfaces anything new (e.g. a genuine rabbit start failure that the codename fix doesn't cover), the `epmd -kill` guard won't mask it — the later *Wait until rabbitmq is healthy* step would still catch it.

**Still open — Debian 13 / trixie docs note:** your 22-06 point stands. The doc currently says trixie doesn't need the `rabbitmq-erlang` sources, but trixie ships Erlang 27 which `rabbitmq-server=3.13.*` rejects, so for the 3.13 step Debian 13 also needs the erlang repo + the 26 pin. The CI here only exercises Ubuntu, so it's not blocking this workflow. Your call: fix that note in this PR, or handle Debian 13 in the 1.24 PR where we move to 4.2 (which does accept Erlang 27) — happy either way.

hasecon:
**Closing the open Debian 13 / trixie question** — I dug into the actual package availability, and the answer turns out to be different from both options I offered earlier ("fix the note with an Erlang pin" vs "defer to 1.24"). Neither works; here's why, and what I propose instead.

### The facts

1. RabbitMQ 3.13.x accepts Erlang 26.x at most (max 26.2.x); Erlang 27 is only supported from 4.0.4 onward ([which-erlang](https://www.rabbitmq.com/docs/which-erlang)).
2. Debian trixie ships **rabbitmq-server 4.0.5** (`4.0.5-6+deb13u2`) built against Erlang 27 ([packages.debian.org](https://packages.debian.org/stable/rabbitmq-server)). The PR description's claim that Debian 13 ships 3.13 is wrong — I'll correct that.
3. Decisive: Team RabbitMQ **builds no Erlang 26 packages for trixie at all**. Their [build configuration](https://github.com/rabbitmq/erlang-packages/blob/main/erlang-deb-configuration.json) only produces Erlang 27 for bookworm/trixie/jammy/noble. The 26.x debs that make our pin work on jammy are older builds that were never produced for trixie.

So there is no installable RabbitMQ 3.13 on Debian 13, period: the distro Erlang is 27 (rejected by 3.13), and no compatible Erlang 26 exists in any repo for trixie. The current doc note ("trixie doesn't need the erlang sources") followed by `rabbitmq-server=3.13.*` from the team repo would fail with exactly the unmet-dependencies error you hit on jammy — so deferring to 1.24 would ship broken instructions in 1.23.

### Proposal

**Debian 13 doesn't need 3.13 in the first place.** The staged 3.12 → 3.13 → 4.2 → 4.3 chain exists to migrate *existing broker state*. But Debian 13 is a new deployment target — it's being added in this same release via #5179. No OpenKAT install on trixie has a 3.12 datastore to migrate; fresh installs can start on 4.x directly. In fact, current main's docs say plain `apt install rabbitmq-server`, so a trixie operator today already gets 4.0.5 — OpenKAT-on-4.x is already the de facto state there. The services don't notice the difference (plain AMQP 0-9-1 via pika/celery; the 4.0 removals — classic queue mirroring, global QoS — aren't used by OpenKAT).

Concretely, in this PR:

- **Replace the trixie note**: on Debian 13, use the distro package (`apt install rabbitmq-server`, 4.0.x, security-maintained by Debian) — no team repos, no pin. The team-repo + 3.13-pin instructions apply to jammy/noble/bookworm.
- **Release notes**: correct the "3.13 (Debian 13)" claim to 4.0.5, and note that fresh Debian 13 installs skip the staged migration chain entirely. One edge case documented: an OS upgrade bookworm → trixie *with existing broker state* should enable all feature flags on 3.13 first (which is the 1.23 instruction anyway).

Somewhat pleasingly, the original note ("trixie doesn't need the rabbitmq-erlang sources") was accidentally correct — just for a different reason than it stated.

Unless you'd rather handle it differently, I'll push the doc + release-notes update to this branch. And when you get a chance: could you re-trigger the `workflow_dispatch` run on the current head? It includes the ubuntu-24.04 codename fix (`09bd038a3`) from your last run, plus your merge of main — both legs should now go green (with `fail-fast: false` they'll no longer cancel each other).

hasecon:
**Follow-up on the "vicious circle" feeling around Debian 13 — it led us somewhere more fundamental.** I walked through every migration scenario, and the honest conclusion is: the Debian 13 problem is a symptom. The real question is whether the 3.13 intermediate step — and the whole staged 3.13 → 4.2 → 4.3 chain — is protecting a scenario that actually exists. We don't think it is anymore. Laying out the reasoning so you can judge.

### First, the direct answer on Debian 12 → 13 migration

No deadlock in any scenario, provided one ordering rule is documented:

1. **Fresh install on Debian 13** — distro 4.0.5, enters the chain past the 3.13 waypoint. Fine.
2. **In-place OS upgrade bookworm → trixie with broker state** — do OpenKAT 1.23 first (3.13 + `enable_feature_flag all`, which is the documented 1.23 step anyway), *then* the OS upgrade; apt swaps in trixie's 4.0.5, and 3.13-with-flags → 4.0 is a supported path. Rule to document: **upgrade OpenKAT first, OS second.**
3. **New Debian 13 machine** — drain queues on the old box (procedure already in the 1.23 release notes), fresh 4.0.5. Broker state never travels.
4. **Wrong order (OS first)** — broker won't start; recovery is wiping `/var/lib/rabbitmq`. Services re-declare the durable queues on reconnect; only in-flight messages are lost. Same impact class as the named-volume note already in this PR. Annoying, not a trap.

The circle dissolves once you see 3.13 as a *state-migration waypoint*, not a requirement of OpenKAT 1.23. Debian 13 installs can't carry pre-1.23 broker state, because Debian 13 support ships in the same release.

### But that raises the bigger question: do we need the waypoint at all?

Four observations, all verified:

1. **Docker deployments (the majority) have no broker volume until this very PR.** State lives in the container filesystem and is wiped on every recreate — every past upgrade has already been "skipping the chain" by silently starting fresh. Bumping the image straight from 3.12 to 4.x and adding the named volume at that point is zero regression against today.
2. **For most deb-host installs the in-place path never existed anyway.** Distro versions are 3.9 (jammy) and 3.10 (bookworm), and 3.10 → 3.13 is not a supported hop. Only noble (3.12) could enter the chain in-place. Drain + fresh start was always the real procedure for the rest.
3. **Broker state is disposable in OpenKAT by design.** All three queues are durable and re-declared by the services on startup (mula's listener even auto-recovers from changed queue properties, `mula/scheduler/clients/amqp/listeners.py`). The real data lives in XTDB/PostgreSQL/Bytes; lost in-flight events self-heal via boefje reschedules.
4. **3.13 is itself EOL** — 1.23 would ship an end-of-life broker as a stepping stone away from an end-of-life broker. And all of the 3.13-specific complexity in this PR (the Erlang 26 pin, the CI fixes for it, the trixie anomaly) exists *only* because 3.13 refuses Erlang 27. With 4.x, default Erlang 27 works everywhere — jammy, noble, bookworm and trixie get identical instructions.

What the chain still buys: a deb-host operator on noble who insists on in-place upgrading without a one-minute drain window. That seems like a vanishingly small group with a documented alternative.

### The two options

**A. Land this PR as-is** — 3.13 in 1.23, then 4.2 (1.24) and 4.3 (1.25) per the #5151 plan. It works, it's two review rounds in, and the remaining changes are small (fix the trixie note + release-notes corrections from my previous comment).

**B. Pivot this PR to a single jump to 4.x** (latest stable) in 1.23. Upgrade procedure = drain queues (already documented) and start fresh; keep the named-volume addition, which makes in-place upgrades *actually* possible from 4.x onward. Uniform install instructions on every distro, no Erlang pins, no feature-flag choreography, no trixie special case, and the 1.24/1.25 follow-up PRs disappear.

We lean toward **B** — knowing what we know now, it's simply less machinery for the same outcome, and it answers the Debian 13 question by removing it. But it's a scope pivot on a PR you've already reviewed twice, and it rewrites part of the #5151 plan, so it's genuinely your call. If you pick B, I'll rework this branch (compose files to 4.x, docs without the pin, release notes around drain-and-fresh-start + the OS-ordering rule); if A, I'll push the trixie doc fix as described above. Either way the CI fixes already on the branch stay useful.

brennodewinter:
## Review — RabbitMQ 3.13 upgrade

hasecon's analysis of the staged 3.13 → 4.2 → 4.3 chain is compelling and I want to +1 option B (direct jump to 4.x) for @underdarknl to consider.

### Why I lean toward B (direct to 4.x)

The four points hasecon verified all point the same direction:

1. **Docker deployments (the majority) have no broker volume until this very PR.** Every past upgrade already silently started fresh — bumping straight to 4.x and adding the named volume is zero regression against today.
2. **Most deb-host installs never had an in-place path.** Distro versions are 3.9 (jammy) / 3.10 (bookworm); 3.10 → 3.13 isn't a supported hop. Only noble (3.12) could enter the chain in-place.
3. **Broker state is disposable by design.** All queues are durable and re-declared by services on startup. Real data lives in XTDB/PostgreSQL/Bytes; lost in-flight events self-heal via boefje reschedules.
4. **3.13 is itself EOL.** Shipping an EOL broker as a stepping stone away from an EOL broker is hard to justify when the alternative is simpler.

### What option B eliminates

- The Erlang 26 pin (which only exists because 3.13 refuses Erlang 27 — 4.x works with default Erlang 27 everywhere)
- The trixie special case (no Erlang 26 packages exist for trixie at all)
- The feature-flag choreography (`enable_feature_flag all`, the `khepri_db` warning)
- Two follow-up PRs (1.24 and 1.25)
- Uniform install instructions on every distro

### What the chain still buys

A deb-host operator on noble who insists on in-place upgrading without a one-minute drain window. That's a narrow group with a documented alternative (drain + fresh start, already in the release notes).

### If A (land as-is)

The remaining work is small: fix the trixie note (distro 4.0.5, skip the chain) and correct the release-notes claims. The CI fixes already on the branch stay useful either way.

### My call

B is less machinery for the same outcome. The named-volume addition (which makes in-place upgrades actually possible from 4.x onward) is the real improvement in this PR — the 3.13 intermediate step is the part worth questioning.

@underdarknl — your decision. Both are viable; B just has fewer moving parts and no EOL waypoint.

brennodewinter (commented):
Reviewed against head `095f5ac5` (post-merge of main). Two things block this from merging today:

## 1. `rtest` is red — and it looks real, not flaky

5 of 15 robot tests fail, all with the same signature: `POST /_dev/node` returns HTTP 500. That endpoint is `router.py:create_node` → `xtdb_session.client.create_node()` + `commit()` — the 500 means an `XTDBException` on commit. The same workflow is green on `update-tls-cipher-severities` as recently as yesterday, and `rtest` passed on #5430 on Monday, so this failure is specific to this branch. Given the PR's contents, the most plausible mechanism is the broker: the CI compose now runs `rabbitmq:3.13` and if octopoes' commit path publishes to the bus and can't reach it (version handshake, or the fresh 3.13 missing something the 3.12 image had), the commit fails exactly like this. Worth a run with broker logs captured — `rabbitmqctl status` passing in healthcheck doesn't prove AMQP works.

## 2. The strategy question is still open

The thread's last substantive exchange is hasecon's option-B proposal (skip 3.13 entirely, jump to a supported 4.x) plus my +1 — waiting on @underdarknl's call. If B wins, most of this PR (Erlang 26 pinning, `enable_feature_flag` choreography, the 3.13 docs) gets thrown away anyway, so merging now would land code we expect to revert.

Not approving until the rtest failure is explained — and I'd hold the merge until the 3.13-vs-4.x decision is made.

PR #5209, opened 2026-06-21
Fix nuclei-cve detection and harden nuclei normalizers

### LINKED ISSUE #5208: Nuclei CVE boefje: outdated pin + fragile normalizer + MIME cross-wiring — proposed remediation
## Context

Investigation into "I don't see CVEs from the Nuclei boefje in the results." Done live against a running stack plus the pinned image. Findings below; a 3-step remediation is proposed at the end with one build-method decision for the maintainer.

## What works (verified live)

The `nuclei-cve` boefje runs and completes correctly: templates download on the boefje network (4016 loaded), httpx reaches the target. Empty output on the test org's hosts is **legitimate** — `apache.dvwa.cloud` / `demo.dvwa.cloud` are not exploitable by any templated CVE. All four `nuclei-cve` raws in Bytes are 0 bytes.

To prove the pipeline against real data, a deliberately-vulnerable Apache 2.4.49 target was stood up (CVE-2021-41773, path traversal confirmed leaking `/etc/passwd`). nuclei v3.2.4 and v3.9.0 both detect it and emit valid JSONL.

### Why the existing apache.dvwa.cloud "isn't seen" by Nuclei
- Its `Server:` banner advertises `Apache/2.4.49`, which the Octopoes bit `check_cve_2021_41773` (origin_type: inference) flags as the CVE — a **version-string match, not verified exploitability**.
- An active path-traversal probe against the real host returns **404** — it is not actually exploitable *through the proxy* (see Live validation below for the precise reason).
- Nuclei actively verifies the exploit, gets 404, and correctly reports nothing.

So the CVE finding currently visible in OpenKAT comes from the banner-inference path, not Nuclei. Nuclei is the stricter (true-positive) engine; the banner bit is the lenient (potential false-positive) one. This is expected behaviour, not a bug.

### 1. Normalizer crashes on normal Nuclei output (highest impact)
`kat_nuclei_cve/normalize.py` hard-indexes:
```python
id_ = data["info"]["classification"]["cve-id"][0].upper()
description = data["info"]["description"]
curl_command = data["curl-command"]
```
Verified against the real normalizer module in the running normalizer container:
- a CVE line **without `info.description`** (common in Nuclei templates) → `KeyError: 'description'`
- `classification.cve-id = null` (CWE-only template) → `TypeError: NoneType`

`run()` is a generator over `splitlines()`; the first crashing line raises out of the task, so **zero findings are persisted — including valid CVEs on other lines.** This is the most likely cause of "I don't see the CVE" on a genuinely vulnerable host.

### 2. MIME cross-wiring (verified in production logs)
The `nuclei-cve` raw carries both `boefje/nuclei-cve` and `openkat/nuclei-output`. All three Nuclei normalizers consume `openkat/nuclei-output`:
```
cve            consumes: [boefje/nuclei-cve, openkat/nuclei-output]
exposed_panels consumes: [boefje/nuclei-exposed-panels, openkat/nuclei-output]   ← also
take_over      consumes: [boefje/nuclei-takeover, openkat/nuclei-output]         ← also
```
Normalizer logs confirm `kat_nuclei_exposed_panels_normalize` and `kat_nuclei_takeover_normalize` ran with `source_method='nuclei-cve'`. On a non-empty CVE raw they would emit `EXPOSED-PANELS` / `SUB-DOMAIN-TAKEOVER` findings from CVE data.

### 3. `EXPOSED-PANELS` finding type is not in the catalogue
`kat_nuclei_exposed_panels` yields `KATFindingType(id="EXPOSED-PANELS")`, which is absent from `kat_finding_types.json` (only `SUB-DOMAIN-TAKEOVER` exists). Such findings render without name/risk.

### 4. Outdated Nuclei pin + templates not bundled
- Pinned to **v3.2.4** (early 2024) in `kat_nuclei_cve/boefje.Dockerfile`; latest is **v3.9.0** (2026-06-10). ~7 minor versions behind.
- Neither the custom build nor the official `projectdiscovery/nuclei` image bundles templates — both auto-download ~4000 templates (`nuclei-templates v10.4.4`) on **every** ephemeral run. Slow, egress-dependent; a failed download = silent empty result.

### 5. Bare-hostname scheme handling (latent)
The boefje passes `-u <bare hostname>`. Nuclei/httpx prefers `https://`; for an http-only vulnerability it may probe the wrong endpoint. Also: Nuclei resolves names via its own resolvers (bypasses Docker DNS), honouring only `/etc/hosts` — relevant for internal test targets.

## Proposed remediation (sequenced)

Schema stability was verified: the JSON fields the normalizer uses are identical between v3.2.4 and v3.9.0, so the version bump is low-risk and should come first.

1. **Pin Nuclei to v3.9.0** and rebuild/republish the boefje image. Fix/test the normalizer against the version we will actually ship.
2. **Pin templates at build time** (`nuclei -update-templates` to a fixed template version in the Dockerfile) so scans are deterministic and don't re-download ~4000 templates per run.
3. **Fix the normalizer**: per-line guards (skip lines without `cve-id` via `continue`; treat `description`/`curl-command` as optional; catch per-line JSON/key errors so one bad line doesn't abort the task). Give each Nuclei boefje a unique output MIME (end cross-wiring). Add `KAT-EXPOSED-PANELS` to the catalogue (or rename to a `KAT-` id) and apply the same robustness to the panels/takeover normalizers.

## Live validation: Traefik path-handling + Log4Shell A/B

Two follow-up findings, both verified live against the deployed test lab (Traefik-fronted).

### A. apache.dvwa.cloud is a false positive because of Traefik, not a patched server
The Apache 2.4.49 backend is genuinely exploitable: the identical image+config run **directly** leaks `/etc/passwd` via `/icons/.%2e/...`. But the host sits behind **Traefik v3.7.5**, which decodes/normalizes the `.%2e/` path before forwarding, so the backend receives a clean path and returns 404. Proven side-by-side:
- direct to the vulnerable Apache → `root:x:0:0:...`
- same image behind Traefik → **404** (a normal request still returns 200, so routing works — only the traversal is neutralized)

Conclusion: path-traversal CVEs (41773/42013) are **not detectable through a normalizing reverse proxy** — only the banner-inference bit fires (false positive). Header/body/query-based CVEs are unaffected. To make a path-traversal target genuinely reachable you'd need TCP/TLS passthrough (backend terminates TLS).

### B. Log4Shell — concrete proof that remediation step 1 (v3.9.0) fixes real detection
A genuinely vulnerable, **header-based** CVE was added to the lab: `log4shell.dvwa.cloud` (CVE-2021-44228, JNDI via the `X-Api-Version` header). Header payloads survive the Traefik HTTP layer, so this is actively detectable end-to-end through the production proxy.

Running the **exact boefje command** (`-t http/cves/ -jsonl -u log4shell.dvwa.cloud`, bare hostname, full cves dir) against the live target:

| Nuclei version | Result on log4shell.dvwa.cloud |
|---|---|
| **v3.2.4** (current boefje pin) | **0 findings** → empty Bytes raw → no data |
| **v3.9.0**, identical args | **CVE-2021-44228 detected** (critical) |

The OpenKAT `nuclei-cve` task on this target completed in ~52s with a **0-byte raw**, matching the v3.2.4 column exactly. A plain version bump to v3.9.0 — no normalizer or OOB-tuning changes — makes the boefje detect it. Reproducible live evidence that remediation step 1 is both necessary and sufficient for this class of CVE.

hasecon:
### Changes

Fixes the Nuclei CVE boefje so it actually produces findings on genuinely vulnerable hosts. Three independent root causes (all verified live — see #5208):

1. **Outdated pin** — `kat_nuclei_cve/boefje.Dockerfile` pinned Nuclei **v3.2.4** (early 2024). It silently misses CVEs that current **v3.9.0** detects. Bumped to v3.9.0; the JSON output schema the normalizer reads is unchanged between the two, so the bump is low-risk.
2. **Fragile normalizer** — `kat_nuclei_cve/normalize.py` hard-indexed `info.classification.cve-id[0]`, `info.description` and `curl-command`. A CVE line without `description` (common) raised `KeyError`, and a null `cve-id` raised `TypeError`. Because `run()` is a generator over `splitlines()`, the first bad line aborted the whole task and **dropped every finding**. Now: lines without a `cve-id` are skipped, `description`/`curl-command` are optional, and per-line JSON errors are caught so one bad line cannot abort the task. The same hardening is applied to the exposed-panels and takeover normalizers.
3. **MIME cross-wiring** — all three Nuclei boefjes emit the shared `openkat/nuclei-output` MIME, and all three normalizers `consume` it, so every boefje's output was fed to all three normalizers (e.g. the panels/takeover normalizers ran on CVE output). Each normalizer now consumes only its own `boefje/nuclei-*` MIME.

Also adds the missing **`EXPOSED-PANELS`** finding type to `kat_finding_types.json` (the panels normalizer yielded it but it was absent from the catalogue, so it rendered without name/risk).

**No breaking changes.** The version bump only takes effect once the `openkat-nuclei` boefje image is rebuilt/republished.

Deferred to a follow-up (noted in #5208): bundling/pinning the Nuclei templates at build time so scans are deterministic instead of auto-downloading ~4000 templates per run.

### Issue link

Closes #5208

### Demo

A genuinely vulnerable, **header-based** target (`log4shell.dvwa.cloud`, CVE-2021-44228) was used to verify end-to-end through a production Traefik. Running the **exact boefje command** (`-t http/cves/ -jsonl -u log4shell.dvwa.cloud`, bare hostname, full cves dir):

| Nuclei version | Result on log4shell.dvwa.cloud |
|---|---|
| **v3.2.4** (old pin) | **0 findings** → empty Bytes raw → no data |
| **v3.9.0** (this PR), identical args | **CVE-2021-44228 detected** (critical) |

The OpenKAT `nuclei-cve` task on that host completed in ~52s with a 0-byte raw on v3.2.4, exactly matching the reported "no data" symptom. The version bump alone makes it detect.

### QA notes

- New unit tests: `boefjes/tests/plugins/test_nuclei_cve.py` (5 tests) cover the normalizer robustness — valid CVE, missing `description` (the regression), null/absent `cve-id` skipped, and one malformed line not dropping valid findings. Run in the boefje container: `docker exec <boefje-container> python -m pytest tests/plugins/test_nuclei_cve.py` (all pass).
- After rebuilding the nuclei boefje image to v3.9.0, scan a header-based vulnerable target (e.g. a Log4Shell app) and confirm a CVE finding is created end-to-end.
- Confirm a CVE scan no longer produces spurious `EXPOSED-PANELS` / `SUB-DOMAIN-TAKEOVER` findings (the MIME cross-wiring fix).

---

### Code Checklist

- [x] Tickets have been created for newly discovered issues. (template-bundling follow-up tracked in #5208)

### DIFF
--- a/boefjes/boefjes/plugins/kat_kat_finding_types/kat_finding_types.json
+++ b/boefjes/boefjes/plugins/kat_kat_finding_types/kat_finding_types.json
@@ -503,21 +503,13 @@
     "impact": "An attacker using your hosting provider may setup a virtual host for your domain and thus intercept and trick users. Since they control content available under your (sub-domain) they could reach user cookies/sessions and perform various other attacks without breaking the browsers same-site principles.",
     "recommendation": "To prevent subdomain takeover, organizations should regularly monitor their DNS records to identify and remove any unused subdomains. Additionally, they should ensure that all subdomains are properly configured and point to valid services that you control."
   },
-  "EXPOSED-PANELS": {
+  "EXPOSED-ADMIN-PANELS": {
     "name": "Exposed administrative panel",
     "description": "An administrative login panel or management interface was found exposed on this host. Such panels are intended for internal use and should not be reachable from the internet, as they widen the attack surface for brute-force and known-vulnerability attacks.",
     "source": "https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/05-Enumerate_Infrastructure_and_Application_Admin_Interfaces",
-    "risk": "medium",
+    "risk": "low",
     "impact": "An exposed administrative interface lets an attacker attempt credential brute-forcing, exploit known vulnerabilities in the management software, and use it as an entry point into the environment.",
     "recommendation": "Restrict access to administrative panels to trusted networks or a VPN, place them behind authentication, and do not expose them directly to the internet."
-  },
-  "EXPOSED-ADMIN-PANELS": {
-    "name":"Administrative login interface reachable",
-    "description": "Exposed login panels for services can pose security risks as they can be targeted by malicious actors for brute-force attacks, phishing attempts, and other forms of unauthorized access.",
-    "source": "https://resources.infosecinstitute.com/topics/application-security/dangers-web-management/",
-    "risk": "recommendation",
-    "impact": "Administrative interfaces may be easy ways for an attacker to gain access to your network.",
-    "recommendation": "Ideally to minimize the attack surface as much as possible these panels should not be directly exposed to the internet."
   },
   "KAT-CRITICAL-BAD-CIPHER": {
     "name":" Insecure SSL/TLS cipher(s) detected",
--- a/boefjes/boefjes/plugins/kat_kat_finding_types/kat_finding_types.json
+++ b/boefjes/boefjes/plugins/kat_kat_finding_types/kat_finding_types.json
@@ -503,6 +503,14 @@
     "impact": "An attacker using your hosting provider may setup a virtual host for your domain and thus intercept and trick users. Since they control content available under your (sub-domain) they could reach user cookies/sessions and perform various other attacks without breaking the browsers same-site principles.",
     "recommendation": "To prevent subdomain takeover, organizations should regularly monitor their DNS records to identify and remove any unused subdomains. Additionally, they should ensure that all subdomains are properly configured and point to valid services that you control."
   },
+  "EXPOSED-PANELS": {
+    "name": "Exposed administrative panel",
+    "description": "An administrative login panel or management interface was found exposed on this host. Such panels are intended for internal use and should not be reachable from the internet, as they widen the attack surface for brute-force and known-vulnerability attacks.",
+    "source": "https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/05-Enumerate_Infrastructure_and_Application_Admin_Interfaces",
+    "risk": "medium",
+    "impact": "An exposed administrative interface lets an attacker attempt credential brute-forcing, exploit known vulnerabilities in the management software, and use it as an entry point into the environment.",
+    "recommendation": "Restrict access to administrative panels to trusted networks or a VPN, place them behind authentication, and do not expose them directly to the internet."
+  },
   "EXPOSED-ADMIN-PANELS": {
     "name":"Administrative login interface reachable",
     "description": "Exposed login panels for services can pose security risks as they can be targeted by malicious actors for brute-force attacks, phishing attempts, and other forms of unauthorized access.",
--- a/boefjes/boefjes/plugins/kat_nuclei_cve/boefje.Dockerfile
+++ b/boefjes/boefjes/plugins/kat_nuclei_cve/boefje.Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.24-alpine AS build
+FROM golang:1.25-alpine AS build
 ARG NUCLEI_VERSION=v3.9.0
 
 RUN go install -v github.com/projectdiscovery/nuclei/v3/cmd/nuclei@${NUCLEI_VERSION}
--- a/boefjes/boefjes/plugins/kat_nuclei_cve/boefje.Dockerfile
+++ b/boefjes/boefjes/plugins/kat_nuclei_cve/boefje.Dockerfile
@@ -1,5 +1,5 @@
 FROM golang:1.24-alpine AS build
-ARG NUCLEI_VERSION=v3.2.4
+ARG NUCLEI_VERSION=v3.9.0
 
 RUN go install -v github.com/projectdiscovery/nuclei/v3/cmd/nuclei@${NUCLEI_VERSION}
 
--- a/boefjes/boefjes/plugins/kat_nuclei_cve/normalize.py
+++ b/boefjes/boefjes/plugins/kat_nuclei_cve/normalize.py
@@ -1,30 +1,45 @@
 import json
+import logging
 from collections.abc import Iterable
 
 from boefjes.normalizer_models import NormalizerOutput
 from octopoes.models import Reference
 from octopoes.models.ooi.findings import CVEFindingType, Finding
 
+logger = logging.getLogger(__name__)
+
 
 def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
     url_reference = Reference.from_str(input_ooi["primary_key"])
-    if raw:
-        for line in raw.splitlines():
-            # Extract and parse values
+    if not raw:
+        return
+
+    for line in raw.splitlines():
+        line = line.strip()
+        if not line:
+            continue
+
+        try:
             data = json.loads(line)
-            id_ = data["info"]["classification"]["cve-id"][0].upper()
-            description = data["info"]["description"]
-            curl_command = data["curl-command"]
+        except json.JSONDecodeError:
+            # A single malformed line must not abort the whole task.
+            logger.warning("Skipping non-JSON line in nuclei output")
+            continue
 
-            # Create instances of CVEFindingType and Finding classes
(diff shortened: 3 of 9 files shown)


hasecon:
## End-to-end verified in a running stack

Rebuilt the `openkat-nuclei` boefje image from this branch and ran the **real `nuclei-cve` boefje task** against a genuinely vulnerable, header-based target (`log4shell.dvwa.cloud`, CVE-2021-44228) through a production Traefik. Full chain confirmed: boefje → Bytes → normalizer → finding.

### Build note (extra commit on this branch)
Nuclei v3.9.0 requires **Go ≥ 1.25.7**; the v3.2.4 build used `golang:1.24` and fails to compile v3.9.0. Bumped the build stage to `golang:1.25` (commit `8db2ba0a9`). After that the image builds and reports `Nuclei Engine Version: v3.9.0`.

### Results

| Stage | v3.2.4 (old) | v3.9.0 (this branch) |
|---|---|---|
| Bytes raw | **0 bytes** | **15 982 bytes** |
| Normalizer | crash / nothing persisted | parsed cleanly, finding saved |
| Octopoes | no CVE | `Finding\|Hostname\|internet\|log4shell.dvwa.cloud\|CVE-2021-44228` |
| Severity | — | **critical (risk 10.0)** via the CVE-finding-types path |
| `EXPOSED-PANELS` type | unresolved (absent from catalogue) | resolves correctly |

Normalizer log (trimmed):
```
[info] kat_nuclei_cve_normalize  source=Hostname|internet|log4shell.dvwa.cloud  source_method=nuclei-cve
       result=[CVEFindingType|CVE-2021-44228, Finding|...|log4shell.dvwa.cloud|CVE-2021-44228]
[info] Saved affirmation  CVEFindingType|CVE-2021-44228  risk_score=10.0  risk_severity=critical
```

### MIME cross-wiring fix also confirmed
Pre-fix logs on the same host showed `kat_nuclei_cve_normalize ... source_method='nuclei-takeover'` and `source_method='nuclei-cve'` running on the same raw (the cross-wiring described in the PR). After the fix only `kat_nuclei_cve_normalize` runs on the CVE output.

This is the exact "I don't see the CVE" symptom resolved end-to-end: the version bump makes the boefje produce output, and the normalizer hardening turns that output into a persisted finding.

boefjes/boefjes/plugins/kat_kat_finding_types/kat_finding_types.json
```diff
@@ -503,6 +503,14 @@
     "impact": "An attacker using your hosting provider may setup a virtual host for your domain and thus intercept and trick users. Since they control content available under your (sub-domain) they could reach user cookies/sessions and perform various other attacks without breaking the browsers same-site principles.",
     "recommendation": "To prevent subdomain takeover, organizations should regularly monitor their DNS records to identify and remove any unused subdomains. Additionally, they should ensure that all subdomains are properly configured and point to valid services that you control."
   },
+  "EXPOSED-PANELS": {
```
underdarknl: This seems like at least a partial duplicate of the next listed 'issue' in this file:   "EXPOSED-ADMIN-PANELS"

underdarknl (changes requested):

boefjes/boefjes/plugins/kat_kat_finding_types/kat_finding_types.json
hasecon: Good catch — you're right, it's a duplicate. Fixed in `243111e22`.

Instead of keeping the new `EXPOSED-PANELS`, I consolidated onto the existing **`EXPOSED-ADMIN-PANELS`**:
- removed the duplicate `EXPOSED-PANELS` entry;
- pointed the nuclei exposed-panels normalizer at `EXPOSED-ADMIN-PANELS` (it was the only code reference, so no other callers affected);
- refreshed the `EXPOSED-ADMIN-PANELS` description/impact/recommendation and swapped the infosecinstitute link for the OWASP WSTG admin-interface reference.

On risk: the old entry was `recommendation` and my new one was `medium`. Since the boefje only detects the *presence/reachability* of an admin panel (not an exploitable vuln), I settled on **`low`** as the honest middle. Happy to move it back to `recommendation` if you'd rather keep these purely informational.

underdarknl (approved):

Issue #5202, opened 2026-06-20
Findings Report misses all findings on descendant OOIs (regression from #5088 tree search_types pruning)


hasecon:
## Summary

The **Findings Report** (and, by reuse, the **DNS Report** findings section and the **Aggregate Organisation Report**) only shows findings that are attached **directly to the selected input OOI**. Any finding on a descendant OOI — e.g. an `HTTPHeader`, `HTTPResource`, `Website`, `SoftwareInstance`, `X509Certificate` — is silently dropped.

This is a **regression introduced by #5088** ("Add optimizations to various Octopoes queries", commit `931a1efed`), specifically the change *"Add search_types filtering to tree builder for query-level scope reduction"*.

Real-world impact: a **critical CVE-2021-41773** (CVSS 9.8) detected on an `Apache/2.4.49` `Server` header does **not** appear in the Findings Report, even though the `Finding` and the enriched `CVEFindingType` exist correctly in Octopoes.

## Reproduction

1. Have a hostname whose webserver returns `Server: Apache/2.4.49`. The bit `check_cve_2021_41773` creates a `Finding` attached to the `HTTPHeader` OOI (`Finding|HTTPHeader|...|Server|CVE-2021-41773`).
2. Generate a Findings Report (or Aggregate report) for that hostname.
3. The CVE — and every other web/port/cert finding — is missing. Only findings directly on the hostname (e.g. `KAT-WEBSERVER-NO-IPV6`) show up.

### Live evidence (against a running stack, input = `Hostname|internet|apache.dvwa.cloud`)

```
FindingsReport.generate_data(...) -> finding types: ['KAT-WEBSERVER-NO-IPV6']   # CVE absent
```

Octopoes tree API from the same hostname:

| query | findings reachable |
|-------|--------------------|
| `get_tree(types={Finding})`, depth 3..12 | **1** |
| `get_tree()` (no type filter), depth 12 | **41** (incl. `CVE-2021-41773`) |

The `Finding`, the `HTTPHeader` (`Apache/2.4.49 (Unix)`) and the `CVEFindingType` (`risk_score 9.8`, `risk_severity critical`) are all present and correct in the graph — this is purely a report/query-layer problem.

## Root cause

`rocky/reports/report_types/findings_report/report.py` collects findings with:

```python
tree = self.octopoes_api_connector.get_tree(reference, depth=9, types={Finding}, valid_time=valid_time).store
[...]

PR #5214, opened 2026-06-23
Add findings_by_ooi endpoint for descendant-aware findings collection

hasecon:
### Changes

**Part C** of #5202 (the dedicated findings endpoint, per the split agreed with @underdarknl). Adds `GET /{client}/findings_by_ooi`, returning **all Findings reachable from an OOI** — including findings on descendant OOIs (e.g. a CVE on an `HTTPHeader`) that `get_tree(search_types={Finding})` dropped after #5088.

**Part B (PR #5213)** fixes the user-facing regression in the `get_tree(search_types=...)` path itself. This PR adds the clean endpoint the report callers switch to (the Rocky swap is #5215, stacked on this).

- `XTDBOOIRepository.list_findings_by_ooi` builds the subtree with `get_tree` **without** a `search_types` filter — so it's unaffected by the #5088 pruning regression — and returns the finding leaves. This deliberately reuses the bounded, indexed **per-level tree query**, which stays fast on real graphs (~0.1s for a real asset's subtree). 

  > Note: an earlier revision used a single recursive XTDB datalog reachability query. Functional testing on a populated graph showed it does **not** scale (XTDB disconnects after ~12s even at depth 2 — the recursive `link` rule materialises connectivity globally), whereas the full-tree traversal is ~0.1s. The recursive approach was dropped.

- service + connector methods + the endpoint. The response bundles the findings **and their associated finding types** (`FindingsByOOIResponse`) so callers don't re-query them; no scan profiles.
- `depth` is clamped to `MAX_FINDINGS_BY_OOI_DEPTH = 15` (matching Rocky's `DEPTH_MAX`).
- Integration tests at the repository and connector/API layers: a finding three hops below the input OOI is returned with its type, and a sibling's finding reachable only via a shared `Network`/`FindingType` does not leak.

### Functional verification (real stack, populated DVWA graph)

`GET /dvwa/findings_by_ooi?reference=Hostname|internet|demo.dvwa.cloud&depth=9` → **200 in ~0.08s, 63 findings + 18 finding types** (incl. descendant `Finding|Website|…` ones). `FindingsReport`/`DNSReport` generate successfully against it.

### Issue link

Refs #5202 (B/#5213 closes the regression; the Rocky swap #5215 closes the issue).

---

### Code Checklist


octopoes/octopoes/repositories/ooi_repository.py
```diff
@@ -585,6 +592,20 @@ def get_tree(
         store = self.load_bulk(reference_node.collect_references(), valid_time, include_scan_levels=include_scan_levels)
         return ReferenceTree(root=reference_node, store=store)
 
+    def list_findings_by_ooi(self, reference: Reference, valid_time: datetime, depth: int = 9) -> list[Finding]:
```
underdarknl, line 604: I feel this is not the correct or optimal route.
* The endpoint should not be 'Findings' It should require a list of requested object (types), and possibly (direct decendents, eg, findings + findigntypes).
* We should construct an XTDB query using 'rules' to skip the repeated querying per tree layer, as this specific endpoint should not return the intermediate levels, or paths, but just return the objects.

hasecon: Thanks @underdarknl.

**On the API shape — fully agree.** I'll generalize it to take a `types` list and return the matching objects (with the associated objects, e.g. their finding types, as an opt-in flag) instead of hard-coding `Finding`.

**On the rules query — that was actually the first revision of this PR, and I replaced it after functional-testing showed it doesn't scale.** I built exactly that: recursive `reach`/`link` rules returning only the leaf objects, no intermediate levels/paths. On a populated DVWA org (`Hostname|internet|demo.dvwa.cloud`, a few hundred OOIs):

- The rules query **500s — XTDB disconnects after ~12s, even at depth 2**. The per-layer `get_tree` pull returns the same findings in **~0.08s** at depth 9.
- Root cause: XTDB 1.x doesn't allow a logic variable in the attribute position, so the connectivity rule has to enumerate every reference attribute as an `or-join` (two branches per foreign key, both directions). Inside the recursive `reach`, XTDB materialises that `link` relation *globally* (every connected pair in the node) before constraining to the target type, which explodes. I tried both a depth-counted `reach` (3-ary — defeats set-dedup) and an unbounded transitive closure (2-ary); both disconnect at ~12–14s.

So the per-layer pull wasn't a shortcut — it was a deliberate pivot for performance (bounded + indexed per level).

If you have a rules formulation that seeds from `?root` and avoids materialising the global `link` relation, I'd genuinely like to build it — you know the query layer far better than I do. Otherwise my proposal is to keep the per-layer pull internally but generalize the API surface exactly as you describe. Which direction do you prefer? Happy to pair on the datalog if you want to attempt a rules version that scales.

hasecon: Followed up on the 1.24.3 rules docs and applied both performance levers they recommend for recursive rules, tested against a populated DVWA org:

- **Bound rule-head args** (`[(reach [?a] ?b) ...]` — the "improves evaluation time by many orders of magnitude for large data sets" feature): still disconnects at **~12.8s**.
- **Explicit `link` rule bodies instead of the `or-join`** ("for maximum performance it is often best to use explicit rules instead of or/or-join"): **~15.3s**.
- Plus the earlier depth-counted (3-ary) and unbounded transitive-closure (2-ary) forms: ~12–14s.

All four hit the same "server disconnected" wall; `get_tree`'s per-layer pull returns the same findings in **~0.08s**.

The blocker looks fundamental to recursive reachability over our subgraph, which is cyclic: `Hostname ↔ ResolvedHostname ↔ IPAddress ↔ IPPort ↔ Website ↔ Hostname …`. The result set is tiny (~63 findings) but the recursion enumerates path-distinct intermediates, and XTDB 1.x's rule evaluation doesn't prune revisits the way a per-layer BFS does. (It also can't take a logic var in attribute position, so `link` has to enumerate every FK regardless.)

So my proposal: implement your API point — which I fully agree with — i.e. generalize the endpoint to take a `types` list and return the matching objects, with associated objects (e.g. finding types) as an opt-in — but keep the per-layer pull underneath, since it's a bounded, indexed BFS (~0.1s) while the rules-only "single query" doesn't hold up on real data.

If you have a rules formulation that prunes revisits / seeds differently and I just haven't found it, I'm very happy to test it — I've got a probe wired up against a populated org and can turn results around quickly.

underdarknl: Ok, then I tihnk we can conclude that because of the combination of our cyclic graph, and the absense of trimming already seen paths in the rule based queries, a tree view (recurrent calls per level), is the only way forward using xtdb1, we could revisit once we move to xtdb2.

hasecon:
@underdarknl rebased on latest `main` (now at `13322ba2`) and resolved the merge conflict.

The only conflict was in `octopoes/tests/integration/test_ooi_repository.py`: a "both-added" collision where this PR's `test_list_findings_by_ooi` and #5213's `test_get_tree_search_types_returns_descendant_findings` (since merged to main) were inserted at the same spot. Resolved by keeping **both** tests — they cover different paths. The production `ooi_repository.py` auto-merged cleanly: `list_findings_by_ooi` (builds the tree without a `search_types` filter) and #5213's path-preserving `get_tree` fix coexist.

Verified locally before pushing:
- `cd octopoes && make itest` → **39 passed, 2 xfailed, 0 failures** on the merged code
- Net diff vs `main` is exactly the intended 7 octopoes files (126 insertions), no stray artifacts

CI is re-running and green so far. Only remaining gate is the required code-owner review — ready when you are. 🙏

brennodewinter:
## Review #5214 — `findings_by_ooi` endpoint

Overall this is in good shape: CI is fully green, the API surface is clean, the response bundles finding types so callers avoid a second round-trip, and the integration tests cover both the positive case (finding 3 hops down is returned) and the negative case (sibling finding reachable only via a shared `Network`/`FindingType` does not leak). @underdarknl already signed off on the per-layer tree approach over a recursive datalog rule (the rules path disconnects at ~12s on a real graph — good that this was measured rather than assumed).

One substantive observation and a couple of nits below; none blocking.

### Substantive: the docstring rationale is stale post-#5213

`XTDBOOIRepository.list_findings_by_ooi` builds the tree **without** `search_types` and the docstring justifies it thus:

> Building the full tree (rather than `get_tree(search_types={Finding})`) sidesteps the #5088 pruning that dropped findings on descendant OOIs … see #5202

That was true when this PR was first opened, but #5213 ("Fix get_tree(search_types) dropping findings on descendant OOIs") has since landed on `main` and this branch was rebased onto it. `get_tree(search_types={Finding})` now preserves paths to descendant findings (the path-preserving `filter_children` re-added by #5213, plus the `test_get_tree_search_types_returns_descendant_findings` regression test that sits right below this code). So the pruning the docstring says it is "sidestepping" no longer exists.

Two options, either is fine:

1. **Switch to `search_types={Finding}`** — same findings set (intermediate levels stay unfiltered in `_get_tree_level`, so findings at any depth ≤ `depth` are still reached; `filter_children` keeps them), but `collect_references()` / `load_bulk` only touch OOIs on paths that lead to a finding instead of the entire subtree. On a real asset (~hundreds of OOIs, ~63 findings) this is a strictly smaller store and less `load_bulk` work for the same result.
2. **Keep the no-filter approach but update the docstring** to reflect the actual current reason (e.g. "deliberately loads the full subtree so the implementation does not depend on the path-preserving filter"), so future readers aren't left thinking `get_tree(search_types={Finding})` is still broken.

I'd lean toward (1) since the finding set is identical and the store is smaller, but (2) is acceptable if there's a reason to avoid relying on `filter_children` here.

### Nits (non-blocking)

- **`depth` default vs clamp.** The router/service/repository default `depth` to `9` while `MAX_FINDINGS_BY_OOI_DEPTH = 15`. The clamp (`max(1, min(depth, 15))`) is correct and matches Rocky's `DEPTH_MAX = 15`; just noting the default of 9 matches `findings_report.TREE_DEPTH`, not the clamp ceiling — worth a one-line comment so the two numbers don't look like they're meant to be the same.
- **`MockOctopoesAPIConnector.list_findings_by_ooi`** (`rocky/tests/conftest.py`) uses `getattr(self, "oois", {}).values()` for finding types. Test-only, so low stakes, but the `getattr` fallback reads as defensive against a fixture that may or may not set `oois` — if `oois` isn't guaranteed on the mock, finding types silently come back empty and Rocky report tests that depend on bundled finding types would pass vacuously. Worth confirming the mock always has `oois` populated when this path is exercised, or dropping the `getattr` fallback so a missing attribute fails loudly.

### What I verified

- `FindingsByOOIResponse` discriminator (`ConcreteFindingTypeType` + `Field(discriminator="object_type")`) matches the existing union in `octopoes/models/types.py` — serialisation round-trips.
- `finding_types_by_id` lookup keys (`finding_type.id`) match the lookup key (`finding.finding_type.tokenized.id`) in both `dns_report` and `findings_report` — `FindingType.id` is the natural key and `Reference.tokenized.id` resolves to it (`findings.py` / `models/__init__.py`).
- `load_bulk({finding.finding_type ...})` is safe on an empty findings set (`load_bulk_as_list` returns `[]` early).
- DNS report behaviour preserved: security flags still computed from all findings, `findings` list still only the KAT-* subset, missing finding type still logged + skipped (was `ObjectNotFoundException`, now `None` check — equivalent).
- findings_report no longer calls `list_objects(types={FindingType})` (all finding types in the org) — replaced by the bundled subset. Same orphan-finding-type behaviour (skip), strictly less work.

Happy to test option (1) against a populated org if useful — I can wire up the probe and report timings.

hasecon:
Thanks for the thorough review — all three addressed (rebased on latest `main` first):

1. **Substantive — switched to option (1), `search_types={Finding}`.** You're right that the #5088-sidestep rationale is stale now that #5213 restored the path-preserving `filter_children`. `list_findings_by_ooi` now builds the tree with `search_types={Finding}`, so only the branches that lead to a finding are loaded instead of the whole subtree — the finding set is identical (intermediate levels stay unfiltered in `_get_tree_level`, at any depth ≤ `depth`). Docstring rewritten to the actual current reason.
2. **Nit — depth default vs clamp.** Folded a one-line note into the docstring: the default of 9 matches `findings_report.TREE_DEPTH` and is intentionally below the `MAX_FINDINGS_BY_OOI_DEPTH = 15` ceiling.
3. **Nit — mock `getattr`.** Dropped the `getattr(self, "oois", {})` fallback so it matches the rest of the mock (which reads `self.oois` directly) and fails loudly. That immediately surfaced exactly your concern: `test_dns_report_no_findings` was passing vacuously without ever setting `oois`. Fixed it to set `oois = {}` explicitly, so its `finding_types == []` assertion is now meaningful.

Verified: full octopoes suite green (221 passed, incl. the repository + connector `findings_by_ooi` integration tests), rocky unit suite green (476 passed). pre-commit green.

brennodewinter (dismissed):
## Review: Approve

Clean endpoint addition that fixes the #5088 regression (descendant findings dropped from `get_tree(types={Finding})`).

### What I verified

**Endpoint** — `GET /{client}/findings_by_ooi` with `reference`, `valid_time`, `depth` params. Uses `get_tree(search_types={Finding})` which preserves paths to findings at any depth ≤ `depth`. Depth clamped to `MAX_FINDINGS_BY_OOI_DEPTH = 15` — sensible ceiling, default of 9 matches `findings_report.TREE_DEPTH`.

**Response model** — `FindingsByOOIResponse` bundles findings + finding_types so callers don't need a separate lookup. The service layer loads finding types via `load_bulk({finding.finding_type for finding in findings})` — single batched query, efficient.

**Report callers** — `dns_report` and `findings_report` updated to use the new endpoint. The old pattern (`get_tree(types={Finding})` + `list_objects(types={FindingType})` + try/except `ObjectNotFoundException`/`StopIteration`) is replaced by a clean dict lookup. Fewer API calls, simpler error handling.

**Sibling isolation test** — `test_list_findings_by_ooi` in `test_ooi_repository.py` verifies that a sibling hostname in the same network with the same finding type does NOT leak into the results. This is the key correctness property — findings only come from the subtree rooted at the input OOI, not from shared ancestors.

**API integration test** — `test_list_findings_by_ooi` in `test_api_connector.py` verifies end-to-end that both direct and deep (3-hop) findings are returned, and finding types are bundled.

**Mock connector** — Rocky's `MockOctopoesAPIConnector` updated with `list_findings_by_ooi` implementation. `test_dns_report.py` gets `mock_octopoes_api_connector.oois = {}` for the no-findings case.

### Non-blocking notes
- The `depth=9` default is appropriate for reports. If this endpoint is later used for ad-hoc queries, callers should be aware that very deep OOI trees may miss findings beyond depth 9.
- `finding_types_by_id` uses `finding.finding_type.tokenized.id` as the lookup key — consistent with the old code pattern.

CI: 39/39 green. Approving — needs rebase before merge (BEHIND main).

brennodewinter:
@hasecon @underdarknl — deze PR is goedgekeurd, gerebased op main, en alle 39 CI checks zijn groen.

Omdat ik de rebase zelf heb gepusht, telt mijn eigen approval niet meer. Zou een van jullie kunnen approven zodat deze gemerged kan worden? De enige wijziging na de rebase is een conflict-oplossing in `test_ooi_repository.py` (beide tests behouden — de pre-existing regression test en de nieuwe `test_list_findings_by_ooi`).

brennodewinter:
@hasecon @cookiemonster @underdarknl — CI is groen, review staat klaar, maar de branch protection eist approval van iemand anders dan de last pusher (brennodewinter heeft de rebase gedaan). Kan een van jullie een approval plaatsen zodat deze gemerged kan worden? 🙏

brennodewinter:
@hasecon @underdarknl — korte herinnering: CI staat volledig groen op de huidige head (`eac9e4f`), geen code-wijzigingen sinds de review, en de PR is gerebased op main. Enige wat ontbreekt is een re-approval-klik omdat ik de rebase zelf heb gepusht. Zou een van jullie dit willen approven zodat we kunnen mergen?

Issue #3638, opened 2024-10-08
Subreports urls throw a KeyError


madelondohmen:
**Describe the bug**
The links to subreports don't work as expected.

**To Reproduce**

1. Go to 'Reports'
2. Expand the row of a concatenated report
3. Click on one of the report ID urls

Also:
1. Go to 'Reports'
2. Expand the row of a concatenated report
3. Click on 'View all subreports'
4. Click on one of the report ID urls

**Expected behavior**
This should link to the correct subreport

**Screenshots**
![Image](https://github.com/user-attachments/assets/846d0882-a7ac-470e-8ee0-4e7d16d62304)
![Image](https://github.com/user-attachments/assets/4927c907-6fdc-4b64-9a51-db60191ed276)

```
Environment:

Request Method: GET
Request URL: http://localhost:8000/en/madelon123/reports/view?report_id=Report%7C065e2af2-79f2-46ad-9d09-70c58a9410e7&observed_at=2024-10-08%2023:59:59:999999

Django Version: 5.0.8
Python Version: 3.11.10
Installed Applications:
['whitenoise.runserver_nostatic',
 'django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.humanize',
 'django.forms',
 'django_components',
 'django_components.safer_staticfiles',
 'django_otp',
 'django_otp.plugins.otp_static',
 'django_otp.plugins.otp_totp',
 'two_factor',
 'account',
 'tools',
 'fmea',
 'rocky',
 'crisis_room',
 'onboarding',
 'katalogus',
 'django_password_validators',
 'django_password_validators.password_history',
 'rest_framework',
 'tagulous',
 'compressor',
 'reports',
 'knox',
 'csp']
Installed Middleware:
['django.middleware.security.SecurityMiddleware',
 'whitenoise.middleware.WhiteNoiseMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'rocky.middleware.auth_token.AuthTokenMiddleware',
 'django_structlog.middlewares.RequestMiddleware',
 'django_otp.middleware.OTPMiddleware',
[...]

PR #5227, opened 2026-07-10
First work towards making the temporal context part of the URL, also …

underdarknl:
…move OOI id to the url and remove the need to insert organization.code into the url at many places.

### Changes

* Make the temporal context a main player on both the URL level, and the navigation level.
* Introduce a custom url builder for the templates that uses this temporal context, but also takes into account the organization.code available in the context.
* Moves the OOI attribute from an url query argument to the url path (url-encoded).
This work removes a lot of boilerplate, but also is required for the bitemporal katalogus.
The kat-alogus in this PR is not yet (bi)temporal, or working correctly.

### Issue link

Closes ...

### Demo
Temporal selection in the header:
image

Displaying the current time when looking at historic/future data:
image

### QA notes

---

### Code Checklist

---

---

## Checklist for QA:


hasecon:
## Conceptual pass (as requested — draft, so no verdict, just the shape of the work)

**Concept: 👍.** Temporal context as a first-class path segment, an `app_url` tag that pulls org + temporal context from the render context, and the OOI id moving out of the query string — that's the right direction, and the amount of boilerplate it deletes is visible in the diff. Three design decisions are worth taking *before* the mechanical sweep, because they change what the sweep looks like. Then the "same mistake ×100": it's really five mechanical classes, each fixable in one pass. Everything below is verified on head `ad8435291` (branch checked out, `make utest` run in Docker: **185 failed / 285 passed** — almost all of it falls into the classes below).

### Three design decisions to take first

**A. OOI primary keys containing `/` cannot live in a `<ooi>` path segment.** Django resolves against the *percent-decoded* path (WSGI `PATH_INFO`), and the default `str` converter is `[^/]+`. So `quote(pk, safe="")` produces `%2F` in the link, but by the time the resolver sees it, it's a `/` again and the pattern stops matching. Every `URL`, `HTTPResource`, `HTTPHeader`, `Website`-path OOI (any pk with a `/` in it) 404s on its detail/tree/graph/edit pages. Options: put `<path:ooi>` as the **last** segment (`objects/detail/<path:ooi>` instead of `objects/<ooi>/detail`), or encode the pk into a slash-free token (urlsafe-b64), or keep the pk in the query string after all. This one determines the route layout, so decide it before touching 70 templates again.

**B. The temporal segment format.** `at-%Y-%m-%d %H:%M:%S.%f` puts a space and microseconds into every historic URL (`%20` everywhere), it's spelled out in four places (`temporal_context`, `get_temporal_url`, `app_url`, and the `removeprefix` parser in `setup`), and it has already drifted: `ViewReportView` stuffs the raw `datetime` into the path → `/org/2026-08-28 10:00:00+00:00/reports/…` (no `at-`, `+00:00` suffix) which only works because the parser tolerates it. Suggest `at-<ISO-8601 basic, UTC>` — e.g. `at-20260828T101233Z` (seconds unless microseconds are truly needed) — with one `format_temporal_context(dt)` / `parse_temporal_context(segment)` pair in `tools/view_helpers.py` that the mixin, the tag and the reports views all call.

**C. Which routes carry the segment?** Currently objects/findings/reports do, katalogus doesn't (but `SinglePluginView` gained `ObservedAtMixin`, whose `temporal_navigation` then uses `NoReverseMatch` as control flow to build a one-entry "Now" list), and `task_list` is registered **twice under the same name** (with and without the segment). Pick a rule — simplest is "every org-scoped route carries `<temporal_context>`" so `app_url` needs no per-route exceptions and the selector is always meaningful; otherwise "only temporal views" plus hiding the selector where the route has no segment (`if "temporal_context" not in resolver_match.kwargs: return []`, no try/except).

### The ×100 — five mechanical classes

1. **`app_url` uses the organization *name*, not the code.** `ooi_extra.py:32` does `kwargs.setdefault("organization_code", str(organization))`, and `Organization.__str__` returns `self.name`. So all **84** `{% app_url %}` sites render `/en/DVWA/now/…` (or `/en/Test%20Organization/now/…`) and `OrganizationView.setup` 404s on `get(code=…)`. Works only where name == code. One-line fix: `organization.code`. This is probably the ×100 you had in mind.

2. **Callers never migrated to the new route signature → `NoReverseMatch`.** Three sub-families, all greppable:
   - `get_ooi_url()` + the `{% ooi_url %}` tag still reverse with `organization_code` only and append `?ooi_id=` — callers: `finding_add.py:143`, `finding_type_add.py:64`, `scan_profile.py:59` (all **POST success redirects: the write happens, then the 500**), plus `crisis_room/partials/dashboard_finding_list_table.html` (×12), `oois/ooi_summary.html`, `report_overview/subreports_table.html`, `new_dashboard_item_modal_report_section.html`. Fix once: rewrite `get_ooi_url` to inject `temporal_context` + `quote(pk, safe="")` (or make the tag delegate to `app_url`) and everything heals.
   - ~25 bare `reverse()`/`redirect()`/`reverse_lazy()` in Python: `ooi_mute.py:53/59/70`, `ooi_add.py` (×8), `upload_raw.py` (×4), `upload_csv.py` (×3), `finding_type_add.py:27/31`, `report_overview.py:133`, `reports/views/base.py:539`, `reports/views/view_helpers.py:17-99`, `reports/viewsets.py:49`, `onboarding/views.py:439`, `tools/view_helpers.py:154` (`ObjectsBreadcrumbsMixin` → scan list), `ooi_list.py:39`. Recipe: add `ObservedAtMixin.app_reverse(name, **kwargs)` — the Python twin of `app_url` (injects org code + temporal context, quotes `ooi`) — and sweep `reverse(` → `self.app_reverse(` in the views. Also `ooi_detail_related_object.py:36` passes `ooi=` to `ooi_add`, whose route has no `<ooi>` → NoReverseMatch on every "Add related object" except Finding.
   - ~20 `{% url %}` sites in templates *outside* the diff: `crisis_room/` (dashboard items ×4, filter, action buttons ×4, explanation modal), `dashboard_redteam.html`/`dashboard_client.html`, `indemnification_present.html`, `finding_type_add.html`, `katalogus/normalizer_detail.html:51/70`, `katalogus/partials/objects_to_scan.html:53`, `reports/…/new_report_action_button.html:18` (→ **Reports tab 500s for everyone**), `subreports_header.html`. Plus `assets/js/taskDetails.js:148` still builds `/objects/detail/?observed_at=…&ooi_id=…` client-side.

3. **`SingleOOIMixin.setup` does `kwargs["ooi"]` unconditionally** — `KeyError` → 500 at setup time on routes without `<ooi>`: `objects/add/<ooi_type>/` (OOIAddView — adding *any* object), `findings/add/` (the no-OOI variant), `findings/mute/bulk/`, and four onboarding views (`OnboardingSetupScanOOIAddView`, `SetClearanceLevel`, `CreateReportRecipe`, `ReportView` → onboarding steps 5/6/9a/10). Old `get_ooi_id()` was lazy; either `kwargs.get("ooi")` + lazy raise, or split the mixin into "has OOI in route" vs not.

4. **`temporal_context=observed_at` passed explicitly (a `datetime`) in `finding_list.html` ×7, `tasks/*.html`, `report_*` templates.** In "now" mode `observed_at` is `datetime.now()`, which `app_url` treats as `!= "now"` → every link from the default Findings page lands on `/at-<render time>/…` → `historic_view=True` → Mute/Edit/Delete/Add-related toolbar hidden, selector says "3 seconds ago". Users can no longer mute a finding via the normal path. Fix: drop the explicit kwarg (the context default already carries the canonical string), and give `parse_observed_at` an explicit `isinstance(value, datetime)` branch instead of the `TypeError` trick (it currently also lets a bad string raise an uncaught `ValueError` during render).

5. **Tests on the old contract** — the 185 failures have three mechanical causes: `?ooi_id=` query/redirect assertions and `reverse()` without the new kwargs; `RequestFactory` requests without `resolver_match` → `AttributeError` in `temporal_navigation` (fix once in `tests/conftest.py::setup_request` by setting `request.resolver_match`); and `test_observed_at.py` asserting the removed GET-based parsing + the deleted "date is in the future" warning. Plus `robot/complete_onboarding/01_onboard_with_all_users.robot:126` uses the old detail URL.

### Smaller, same sweep
`now_url` is unused; `finding_add.py` adds `"date"` to form kwargs and deletes it six lines later; `app_url`'s `organization=None` branch deletes the key `organization` (not `organization_code`) so it can't work; `ooi` is now a `cached_property` *and* `OOIEditView`/`OOIDeleteView.setup` still do `self.ooi = self.get_ooi()` (shadowing), and `context["ooi"]` is set twice; `hasattr(self, "ooi")` in `get_ooi_class`/`get_readonly_fields` now silently triggers an Octopoes fetch (it's a property); three divergent date parsers (mixin / `valid_time` / tag); `hydrate_branch` reverses once per graph node (pre-existing shape, but now with three varying kwargs — reverse once with a sentinel and substitute).

### Tests worth adding with the sweep
`ObservedAtMixin` parsing (`now`, `at-<date>` → 23:59:59.999999 UTC, `at-<datetime>`, garbage → message + fallback), `temporal_navigation` on a route with/without the segment, and `app_url` (context defaults, `temporal_context=None`, quoting of `|` and `/` in `ooi`). Right now nothing touches the new code paths.

Happy to do sweeps 1, 2 and 4 as a stacked commit on this branch once A–C are decided — they're mechanical and I've got the grep lists.

hasecon:
### Addendum — one design move that solves A, B and three more at once: registered path converters

Verified on the branch in the rocky container, these all stem from `<ooi>` and `<temporal_context>` being bare `str` segments:

- **Route shadowing**: `resolve("/o/now/findings/finding_type/add/")` → `finding_add` with `ooi="finding_type"`. `findings/<ooi>/add/` is listed before `findings/finding_type/add/`, and `[^/]+` happily eats the literal — `FindingTypeAddView` is unreachable. Same trap for `findings/<ooi>` vs `findings/add/`, and for any future literal under `objects/`.
- **Double encoding**: `reverse("ooi_detail", ooi=quote(pk, safe=""))` emits `%257C` (Django re-encodes the `%`) — visible `%25…` in every OOI link — and a pk containing a literal `%` decodes to the wrong reference; the `quote`/`unquote` pair is hand-rolled at ~15 sites plus the tag, with the invariant enforced nowhere (`finding_add.py:53` already seeds the form with the *still-encoded* `kwargs["ooi"]` instead of `self.ooi_id`).
- **Garbage temporal segment locks the user in**: `/dvwa/foo/objects/` → `raw_temporal_context="foo"` → `is_historic_view=True`, `observed_at` falls back to now with a flash error, `temporal_context` becomes `at-<now>` and propagates into every link on the page — a typo or stale bookmark becomes a permanent "historic" session.

Two converters (`register_converter`) own all of this in one place:
```python
class TemporalContextConverter:   # now | at-<ISO-8601 basic UTC>
    regex = r"now|at-\d{8}T\d{6}Z"
    def to_python(self, value): ...  # -> None for "now", else aware datetime; anything else = 404 by regex
    def to_url(self, value): ...     # datetime | None -> segment (the single formatter)

class OOIReferenceConverter:      # requires the Type|… shape, owns quote/unquote
    regex = r"[A-Za-z0-9]+%7C[^/]+"  # or a urlsafe-b64 token if you go that way for the '/' problem (A)
```
Then `<temporal_context>`/`<ooi>` can't shadow literals, `reverse()` callers pass plain datetimes/pks with no `quote()`, `setup()` receives typed values, and `app_url`/`get_temporal_url`/`ViewReportView` stop each spelling their own format.

Two more verified items for the sweep list: the header temporal selector rebuilds the URL from `resolver_match.kwargs` only and drops `request.GET` — switching time on `/reports/view?report_id=…` loses `report_id` (→ 500 in `ViewReportView.setup`) and silently drops list filters; and `tasks/reports.html:55` passes a stray `query=ooi.mandatory_fields` kwarg into `app_url` → `NoReverseMatch` on the Tasks → Reports tab.

hasecon:
## Round 2 — delta since `ad84352` (4 commits, verified on head `f34ac68`)

Still a draft, so still no verdict — this is the round-over-round status. Everything below is verified on a checkout of the head (code read + live checks in the rocky container, `make utest` re-run: **186 failed / 284 passed**, vs 185/284 last round — the failures are the class-5 old-contract ones, dominated by `KeyError: 'ooi'`).

### Fixed since last round ✅

| Item | Status |
|---|---|
| ×100 / sweep class 1: `app_url` used the org **name** | ✅ `ooi_extra.py:32` now uses `organization.code` |
| `now_url` unused | ✅ now the single source (`get_temporal_url(None)`), exposed in context, and used by `ooi_past_due_warning.html` — nice |
| `self.ooi = self.get_ooi()` shadowing in Edit/Delete views | ✅ removed; the `cached_property` is now authoritative |
| Boefje-detail OOI links (SelectOOIForm → widget → cell templates) | ✅ the Organization object now flows through `widget.attrs` into `{% app_url %}` — chain verified |
| Ambiguous no-OOI `finding_add` route | ✅ removed; "select OOI first" is a sound design call |
| `field_input_radio.html` dropping the manual `checked` | ✅ safe — Django's `ChoiceWidget.create_option` already emits `checked` via `choice.data.attrs`, the removed line was redundant (and wrong on re-render after POST) |

### New in these 4 commits 🔴

1. **The new finding-add flow 500s on every GET — the form half of the migration isn't committed.** `FindingAddView.get_form_kwargs` (f34ac68) now sends `{"connector", "observed_at", "ooi_id"}`, but `FindingAddForm.__init__` still has signature `(connector, ooi_list, *args, **kwargs)`. Proven live in the rocky container:
   ```
   TypeError: FindingAddForm.__init__() missing 1 required positional argument: 'ooi_list'
   ```
   (and once `ooi_list` is gone, `observed_at`/`ooi_id` still land in `forms.Form.__init__` → second TypeError). The form needs the matching update: drop `ooi_list`/`set_choices_for_widget`, accept `observed_at`/`ooi_id`, and the `ooi_id` DataList field can become a hidden/readonly field seeded from the view.

2. **Duplicate "Add finding" button** — `oois/ooi_detail_findings_list.html` now renders the same `<a>` twice (copy-paste in f34ac68).

3. **`FindingAddView.setup` seeds the form with the still-URL-encoded pk**: `self.initial = {"ooi_id": kwargs["ooi"]}` — `kwargs["ooi"]` is percent-encoded (`Hostname%7Cinternet%7C…`), so the created Finding would reference a corrupted OOI. Use `self.ooi_id` (`SingleOOIMixin.setup` already unquotes it). Same bug family as the old `finding_add.py:53` note from round 1.

4. **`FindingTypeAddView` is now unreachable by any path.** The removed `findings_list_toolbar.html` include held the only "Add finding type" link (the partial is now orphaned — delete it or re-add a slimmed toolbar), *and* the route `findings/finding_type/add/` is still shadowed by `findings/<ooi>/add/` (round-1 addendum: `resolve()` hands it to `FindingAddView` with `ooi="finding_type"`). Both halves need an answer.

5. Small: `SelectOOIForm.__init__` type hint says `organization: str` but receives an `Organization`; `CheckboxTable.get_context` still carries the now-dead `organization_code` branch.

### Still open from round 1 (unchanged — the sweep)

* **B (format):** `%Y-%m-%d %H:%M:%S.%f` is now spelled in 3 places (`mixins.py:150,164`, `ooi_extra.py:40`) and `reports/views/base.py:746,754` still puts a **raw datetime** in the path.
* **Class 2:** `get_ooi_url`/`{% ooi_url %}` legacy (`finding_type_add.py:64`, `scan_profile.py:59` + 4 templates), the bare `reverse()` list, the unmigrated `{% url %}` templates, `taskDetails.js:148`.
* **Class 3:** `SingleOOIMixin.setup` still does `kwargs["ooi"]` unconditionally → `objects/add/<ooi_type>/` (OOIAddView), `MuteFindingsBulkView` and the onboarding views 500.
* **Class 4:** `finding_list.html` ×7 `temporal_context=observed_at`; `tasks/reports.html:55` stray `query=` kwarg.
* **Addendum:** route shadowing, `%257C` double-encoding, garbage-segment lock-in, selector dropping `request.GET` — the registered-path-converters proposal still covers all four at once.

### Housekeeping

* Branch conflicts with main (one file: `rocky/rocky/views/ooi_tree.py`). When merging main in, mind **Django 5.2** (#5319 landed): main changed `OrganizationAdmin.list_display` to `display_tags` because 5.2's `admin.E109` rejects the tagulous M2M — this branch still has `"tags"` and rocky **crash-loops at startup under a 5.2 image** (seen live). The merge brings the fix; just don't resolve it away.
* Only CodeQL ran on this head — there's no rocky CI signal on the branch right now.

Offer from round 1 stands: once A–C are decided I can do sweep classes 1/2/4 as a stacked commit — grep lists are ready.

brennodewinter:
## Review — temporal context in URL

hasecon's two rounds of review are thorough and I agree with the analysis. Posting to add weight to the key decision that's blocking progress and to +1 the proposed direction.

### The decision that unblocks everything: registered path converters

hasecon proposed two `register_converter` classes (`TemporalContextConverter` + `OOIReferenceConverter`) and I think that's the right call. It solves four problems at once:

1. **Route shadowing** — `findings/<ooi>/add/` currently eats `findings/finding_type/add/` because `<ooi>` is `[^/]+`. A converter with a stricter regex makes the literal route reachable.
2. **OOI pks containing `/`** — the single biggest correctness issue. Every `URL`, `HTTPResource`, `HTTPHeader` OOI 404s on detail/tree/graph/edit because Django percent-decodes `%2F` back to `/` in `PATH_INFO` before the resolver sees it. A converter that owns the quote/unquote cycle (or a urlsafe-b64 token) fixes this in one place.
3. **Temporal format sprawl** — `at-%Y-%m-%d %H:%M:%S.%f` with spaces and microseconds is spelled in 3+ places and already drifted (`ViewReportView` puts a raw datetime in the path). A converter with `to_python`/`to_url` centralises this.
4. **Double-encoding** — `reverse("ooi_detail", ooi=quote(pk, safe=""))` emits `%257C` because Django re-encodes the `%`. With a converter, callers pass plain pks and the converter handles encoding.

### What I'd ask @underdarknl to decide before the sweep

**A.** How to handle `/` in OOI pks: `<path:ooi>` as the last segment, or a slash-free token (urlsafe-b64)? The converter approach works either way, but the route layout differs.

**B.** Temporal format: `at-20260828T101233Z` (ISO-8601 basic, UTC, seconds) as hasecon suggested is clean and URL-safe. Confirm or propose alternative.

**C.** Which routes carry the segment: "every org-scoped route" (simplest, no per-route exceptions in `app_url`) or "only temporal views" (needs selector hiding logic)?

Once A–C are decided, the ~100 template/Python migrations hasecon enumerated are mechanical and can be done in a focused sweep.

### Current state

186 tests failing, `FindingAddForm.__init__` signature mismatch causes a 500 on every GET, `SingleOOIMixin.setup` does `kwargs["ooi"]` unconditionally causing KeyErrors on routes without `<ooi>`. These are all known and listed in hasecon's round 2 — flagging that they're blocking, not just nits.

Happy to help with the sweep once the design calls are made.

brennodewinter (changes requested):
## Round 3 — delta since `f34ac68` (1 commit, verified on head `cb567dc5`)

The path converters are a big improvement — they address design decisions A and B from hasecon's rounds and fix the route shadowing, double-encoding, and temporal format sprawl in one move. The `OOIReferenceConverter` regex `[A-Za-z0-9]+\|.+` correctly allows `/` in pks (greedy `.+` with backtracking against the trailing route suffix), and `TemporalContextConverter` uses the clean `at-YYYYMMDDTHHMMSSZ` format. `SinglePluginView` dropping `ObservedAtMixin` and the duplicate `task_list` route removal resolve decision C. Good progress.

Still request-changes — the converters are in but the sweep around them isn't done. Below: what's fixed, what's newly broken, and what's still open.

### Fixed since round 2 ✅

| Item | Status |
|---|---|
| Path converters (`TemporalContextConverter` + `OOIReferenceConverter`) | ✅ registered in `urls.py`, used on all temporal and OOI routes |
| Route shadowing (`findings/<ooi>/add/` eating `finding_type/add/`) | ✅ `finding_type/add/` moved to its own route, no longer shadowed |
| `app_url` using org name instead of code (class 1) | ✅ confirmed in `ooi_extra.py:32` |
| `SinglePluginView` inheriting `ObservedAtMixin` unnecessarily | ✅ now `OrganizationView` only — katalogus is non-temporal |
| Duplicate `task_list` route (with/without temporal segment) | ✅ removed the temporal variant |
| `now_url` / `self.ooi` shadowing in edit/delete | ✅ carried over from round 2 |
| Temporal format sprawl (3+ places spelling `at-%Y-%m-%d %H:%M:%S.%f`) | ✅ now centralised in the converter's `to_url` |

### Newly broken in `cb567dc5` 🔴

**1. `SingleOOIMixin.setup` double-decodes OOI pks.** `OOIReferenceConverter.to_python` returns the percent-decoded value (Django decodes `PATH_INFO` before matching). Then `setup` does `urllib.parse.unquote(kwargs["ooi"])` — a second decode. For pks containing `%` (e.g. `URL|internet|https://example.com/50%2520off`), the first decode gives `…50%20off`, the second turns `%20` into a space → `…50 off`. Corrupted pk, wrong OOI fetched or 404. **Fix:** remove the `unquote`, use `self.ooi_id = kwargs["ooi"]` directly. The converter owns the decode cycle now.

**2. `FindingAddForm.__init__` is broken — the form half of the migration isn't right.**
- `self.set_choices_for_widget("ooi_id", self.ooi_id)` — `self.ooi_id` is a string (a pk), not a list of `(value, label)` tuples. `set_choices_for_widget` sets `widget.choices`, which expects an iterable of pairs. The `ooi_id` field is now a plain `CharField` (no `DataListInput` widget), so `widget.choices` is meaningless on a `TextInput` anyway.
- `self.set_choices_for_widget("date", self.observed_at)` — `self.observed_at` is a datetime, not choices. Same problem on a `DateTimeInput` widget.
- `clean_date` and `clean_ooi_id` were removed — no validation that the OOI exists or that the date isn't in the future.
- `date` field `initial=None` — no default date, the user gets an empty field.

**3. `FindingAddView.setup` seeds initial from the wrong value.** `self.initial = {"ooi_id": kwargs["ooi"]}` uses the converter-decoded pk, but `get_form_kwargs` passes `self.ooi_id` (double-decoded via `SingleOOIMixin`). For pks with `%`, the form initial and the form kwarg diverge. Use `self.ooi_id` consistently (after fixing #1).

**4. Duplicate `urlconverters.py` files.** `rocky/rocky/urlconverters.py` and `rocky/tools/urlconverters.py` are identical. `urls.py` imports from `tools.urlconverters`. The `rocky/rocky/urlconverters.py` copy is dead code — delete it.

**5. `OOIReferenceConverter.to_url` doesn't convert OOI objects to strings.** Python code passes `self.ooi` (an OOI object) directly to `reverse()` in ~8 places (`ooi_view.py:183`, `ooi_tree.py:53/79/99`, `ooi_delete.py:39`, `ooi_edit.py:65`, `mixins.py:575`, `finding_add.py:96`). `to_url` returns the value as-is. `OOI.__str__` returns `primary_key` (verified in `octopoes/models/__init__.py:128`), so Django's `str()` coercion happens to work — but it's fragile and the contract is "to_url returns a string". **Fix:** `def to_url(self, value): return str(value)`.

**6. `OOIDeleteView.delete` and `BaseOOIFormView.form_valid` use `self.observed_at` as valid_time.** In historic view, this uses the historic datetime for create/delete operations. You can't create or delete in the past — these should use `datetime.now(timezone.utc)` (as the original code did). `self.observed_at` is for *reading*, not *writing*.

### Still open from rounds 1–2 (unchanged)

**Class 3 — `SingleOOIMixin.setup` does `kwargs["ooi"]` unconditionally.** `KeyError` → 500 on routes without `<ooi>`: `OOIAddView` (`objects/add/<ooi_type>`), `MuteFindingsBulkView` (`findings/mute/bulk/`), and onboarding views. Use `kwargs.get("ooi")` and handle `None`.

**Class 4 — `app_url` uses `context.get("observed_at", "now")` instead of `context.get("temporal_context")`.** In "now" mode, `observed_at` is `datetime.now()`, so every `{% app_url %}` link produces `at-<timestamp>` instead of `now`. The destination page thinks it's historic → mute/edit/delete toolbar hidden, selector says "3 seconds ago". **Fix:** change `ooi_extra.py:29` to `context.get("temporal_context")` (which is `None` in "now" mode — the converter maps `None` → `"now"`). Also remove the explicit `temporal_context=observed_at` from `finding_list.html` (×7) and other templates — pass `temporal_context=temporal_context` (None in "now" mode) or just drop the kwarg and let the context default flow.

**Class 2 — unmigrated callers (partial list, still greppable):**
- `tasks/reports.html:55` — stray `query=ooi.mandatory_fields` kwarg → `NoReverseMatch` on Tasks → Reports tab.
- `get_ooi_url()` / `{% ooi_url %}` legacy callers still using the old `?ooi_id=` query contract.
- `OOIListView.breadcrumbs` class attribute: `reverse_lazy("ooi_list")` without `organization_code`/`temporal_context` kwargs — would `NoReverseMatch` if evaluated. Overridden in `get_context_data` but fragile.

**`hasattr(self, "ooi")` in `get_ooi_class`/`get_readonly_fields` triggers the `cached_property`.** `hasattr` now makes an Octopoes API call just to check existence. On routes without `<ooi>` (after the class-3 fix), `self.ooi_id` would be `None` → `Reference.from_str(None)` raises `TypeError`, which `hasattr` doesn't catch → 500. The old code had `self.ooi` as a plain attribute set in `setup`; the `cached_property` changes the semantics of `hasattr`.

**No tests for the new code paths.** Converters (`to_python`/`to_url` round-trip, regex matching), `app_url` with the new context defaults, `temporal_navigation` on routes with/without the segment, `ObservedAtMixin` parsing. hasecon flagged this in round 1 — still nothing. The 185+ test failures from round 1 are mostly the old-contract ones (class 5), but without tests on the new paths, regressions in the sweep will be invisible.

**Branch conflicts with main** (`mergeable: CONFLICTING`). Only CodeQL ran — no rocky CI signal. When merging main, mind Django 5.2 (#5319): `OrganizationAdmin.list_display` changed to `display_tags` because 5.2's `admin.E109` rejects the tagulous M2M — this branch still has `"tags"` and rocky crash-loops at startup under a 5.2 image.

### Summary

The converter approach was the right call and it's implemented correctly at the converter level. The remaining work is the sweep around it: fix the double-decode, fix `FindingAddForm`, use `self.observed_at` only for reads, use `context.get("temporal_context")` in `app_url`, guard `kwargs["ooi"]` in `SingleOOIMixin.setup`, and migrate the remaining callers. The `hasattr(self, "ooi")` issue needs a design call — either split `SingleOOIMixin` into "has OOI in route" vs not, or use a separate flag set in `setup`.

Happy to help with the sweep once these blocking items are addressed.

hasecon (changes requested):
## Round 4 — verified on head `be6e14de` (`make utest` in Docker, rocky)

Real progress since round 3 — several converter-era items are genuinely fixed. But the caller/test migration sweep still isn't finished: **`make utest` = 231 failed / 249 passed**, and rocky `test (3.10–3.13)` is red in CI on this exact head. So this stays REQUEST_CHANGES; the remaining work is the mechanical sweep + the test harness, not new logic.

### Fixed since round 3 ✅ (verified on head)
| Item | Status |
|---|---|
| Duplicate `rocky/rocky/urlconverters.py` (r3 #4) | ✅ gone; only `tools/urlconverters.py` remains |
| `app_url` used `observed_at` instead of `temporal_context` (r2 class 4 / r3) | ✅ `ooi_extra.py` now reads `context.get("temporal_context")`, maps `None`→"now", coerces `ooi`→`str()` |
| Double-decode of the OOI pk in `SingleOOIMixin.setup` (r3 #1) | ✅ the extra `unquote` is gone; `self.ooi_id` is used |
| `FindingAddForm` signature mismatch / 500 on GET (r2 #1, r3 #2) | ✅ reworked to `ooi_id`/`finding_type_ids` fields |

### Still blocking 🔴

**1. 231 unit tests fail — the migration sweep is incomplete.** Distribution:
- **274 `NoReverseMatch`** — callers still reverse the old contract without `temporal_context`: `finding_list` ×118, `ooi_list` ×52, `scheduled_reports` ×18, `upload_csv` ×16, `ooi_list_export`, `upload_raw`, `report_history`, `ooi_graph`, `aggregate_report_select_oois`, `view_report`. `app_url` itself is now correct, so these are the direct `reverse()`/`{% url %}` sites (production + tests) that still bypass it — the class-2 sweep from rounds 1–2.
- **86 `AttributeError`** — `tests/conftest.py::setup_request` still does not set `request.resolver_match`, so `temporal_navigation`/`app_url` blow up during render in every view/template test. This is the one-spot test-harness fix flagged in round 1 (set `request.resolver_match` with `organization_code` + `temporal_context` kwargs).
- **96 `KeyError` / 137 `ValueError`** — the `kwargs["ooi"]` / temporal-parse paths on routes without those segments.

**2. Writes still use `self.observed_at` as valid_time (r3 #6, still open).** `rocky/views/ooi_delete.py:15` → `octopoes_api_connector.delete(self.ooi.reference, self.observed_at, sync=True)`. In a historic view this deletes at the historic timestamp; create/delete must use `datetime.now(timezone.utc)`. `observed_at` is for reads only.

### Smaller
- `parse_observed_at` in `ooi_extra.py` only catches `TypeError`; a malformed `at-…` string still raises an uncaught `ValueError` during render.
- `makelang` is red and this PR changes translatable strings, so the `.po` regen belongs with this PR, not a follow-up.
- Branch is `blocked`/behind main — mind the Django 5.2 `OrganizationAdmin` `display_tags` fix when merging main in (rocky crash-loops under a 5.2 image otherwise).

### Path to green (and to the tests)
The converters landed and `app_url` is correct, so design decisions A–C are effectively settled. What's left is exactly the sweep we offered in rounds 1–2: (a) the `setup_request` `resolver_match` fix (clears the AttributeError class at once), (b) migrate the remaining `reverse()`/`{% url %}` callers to pass `temporal_context` (or route them through `app_url`/a Python `app_reverse` twin), (c) the write-valid_time fix, then (d) the new-path tests (converter round-trip, `app_url` context defaults, `temporal_navigation` with/without the segment, `ObservedAtMixin` parsing). Adding tests before (a)–(c) can't go green — the suite is red from unmigrated callers, not missing tests. Happy to take the sweep + tests on as commits on this branch (grep lists ready from the earlier rounds) so it lands green for approval.

### Verdict
**CHANGES REQUESTED** — rocky `test` red on head (231 failures), and the historic-delete write-semantics bug is a correctness issue. Both are addressable with the mechanical sweep + the harness fix; no design change needed.

underdarknl:
2. Writes still use self.observed_at as valid_time (r3 https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/pull/6, still open). rocky/views/ooi_delete.py:15 → octopoes_api_connector.delete(self.ooi.reference, self.observed_at, sync=True). In a historic view this deletes at the historic timestamp; create/delete must use datetime.now(timezone.utc). observed_at is for reads only.

I'd argue that deleting an object on any time, needs to reflect that time as its deletion time in the database. Now is the default when no time is specified (and will be filled in as the current datetime in the calls you mention), but any other time is a specific command from the user. Perhaps a warning about this behaviour in the rocky edit/delete flow is a good step though.

rocky/tests/objects/test_objects_delete.py
```diff
@@ -9,7 +9,9 @@
 def test_ooi_delete(rf, redteam_member, mock_organization_view_octopoes, network):
     mock_organization_view_octopoes().get.return_value = network
     request = setup_request(rf.get("ooi_delete", {"ooi_id": "Network|testnetwork"}), redteam_member.user)
```
underdarknl: Shouldn't the dict containing ooi_id be removed from these as this is now aprt of the url?

underdarknl (commented):
see remark

rocky/rocky/views/mixins.py
```diff
@@ -515,11 +520,14 @@ def report_type_summary(reports: list[AssetReport]) -> dict[str, int]:
 
 
 class SingleOOIMixin(OctopoesView):
-    ooi_id: str
+    ooi_id: str | None
 
     def setup(self, request, *args, **kwargs):
         super().setup(request, *args, **kwargs)
-        self.ooi_id = kwargs["ooi"]
+        # Not every route that uses this mixin carries an <ooi> segment (e.g. OOIAddView,
```
underdarknl: I think (ideally) those routes should be fixed to not include this mixin.

underdarknl (commented):
and one more

rocky/rocky/views/ooi_delete.py
```diff
@@ -12,7 +14,9 @@ class OOIDeleteView(OrganizationPermissionRequiredMixin, SingleOOIMixin, Templat
     permission_required = "tools.can_delete_oois"
 
     def delete(self, request):
-        self.octopoes_api_connector.delete(self.ooi.reference, self.observed_at, sync=True)
+        # Deleting is a write, so it always applies to the present, never to a historic
```
underdarknl: No, this was a feature I wanted to add, see my comment about this earlier.

underdarknl (commented):
see remark

hasecon:
## Sweep done — branch is green

Picked up the offer from the earlier rounds and ran the caller/test-migration sweep on this branch. On head `1175a31ad`, `make utest` = **480 passed / 1 skipped**, and CI is green on every blocking check (`test (3.10–3.13)`, `pre-commit`, `build (rocky)` ×4, CodeQL, container image). Only `makelang` is red — translation regen, non-blocking; a `make lang` commit clears it.

### Blocking items from round 4 — fixed
- **`app_url` dropped `temporal_context` on "now" pages**, so every object/finding link raised `NoReverseMatch`. It now defaults to the context value (`None` → `"now"`), with a fallback for the non-temporal routes that also use the tag.
- **`SingleOOIMixin.setup` `KeyError`** on routes without `<ooi>` → `kwargs.get("ooi")`, and it is set *before* `super().setup()` so `BaseReportView.setup` can read `self.ooi`.
- **`resolver_match` is `None`** under `RequestFactory` → guarded in `get_temporal_url` (degrades to an empty temporal link instead of `AttributeError`).
- **The remaining `reverse()`/`redirect()`/`{% url %}` callers** migrated: `ooi_mute`, `upload_csv`/`upload_raw`, `ooi_add`, the report step navigations, `report_history`/`scheduled_reports`, onboarding, `katalogus/objects_to_scan`, and the stray `query=` kwarg in `tasks/reports.html`.
- **Old-contract tests** ported across ~35 files; `test_observed_at` rewritten for the URL-segment contract, plus converter round-trip tests.

### Also fixed while migrating
- `OOIEditView` save used an unset `self.ooi_class` → use `get_ooi_class()`.
- `get_ooi_class`/`get_readonly_fields` used `hasattr(self, "ooi")`, which triggered an Octopoes fetch on ooi-less views → guard on `ooi_id`.
- `OOIListView` bulk POST redirected through a `self.self.breadcrumbs` typo.
- Double-encoded pks (`%257C`) from hand-rolled `quote()` → the converter owns encoding now.
- `context["ooi"]` fetched OOI "None" on ooi-less views → only set when an ooi is present.
- `step_10_report` reverted to no `<ooi>` (the view never uses one and the steps-nav can't supply it); `step_9a` threads the ooi via `request.GET.ooi`.

### On your review comments, @underdarknl
- **Historic delete is intentional** — agreed, reverted. `OOIDeleteView.delete` uses `self.observed_at` again; `now` stays the default when no temporal context is set. A warning in the delete/edit flow is a good follow-up.
- **`ooi_id` dicts in the delete tests** — removed the redundant query dicts across the object tests, since the ooi is part of the URL now.
- **Those routes should not include the mixin** — agreed that is cleaner; left as a follow-up since the `kwargs.get` guard keeps everything green and unblocks review. Happy to do that refactor separately.

Functional test (`make kat` + browser E2E) has not been re-run here since it touches ~every page's links — worth a smoke before merge.

rocky/onboarding/templates/step_9_choose_report_type.html
```diff
@@ -23,7 +23,7 @@ <h2>{% translate "Generate a report" %}</h2>
                     {% endblocktranslate %}
                 </p>
                 <form method="post"
-                      action="{% url "step_9a_setup_scan_ooi_detail" organization.code %}?{{ request.GET.urlencode }}"
+                      action="{% url "step_9a_setup_scan_ooi_detail" organization.code request.GET.ooi %}?{{ request.GET.urlencode }}"
```
underdarknl, line 26: Should this ooi possibly live on the views url /  kwargs instead of the request params too?

rocky/reports/templates/partials/report_findings_table.html
```diff
@@ -41,7 +41,7 @@
                         <td>{{ info.occurrences|get_first_seen }}</td>
                         {% if is_crisis_room %}
                             <td class="number">
-                                <a href="{% url "view_report" dashboard_item.item.dashboard.organization.code %}?report_id={{ dashboard_item.data.report }}#{{ info.finding_type.id }}"><span class="icon ti-arrow-right"></span></a>
+                                <a href="{% app_url "view_report" organization_code=dashboard_item.item.dashboard.organization.code %}?report_id={{ dashboard_item.data.report }}#{{ info.finding_type.id }}"><span class="icon ti-arrow-right"></span></a>
```
underdarknl, line 44: as reports themselves are just ooi's. I feel like we should move away from report_id as an url param, and just let the ooi mixin and <ooi:ooi> style routing handle this.

rocky/rocky/views/ooi_edit.py
```diff
@@ -64,10 +59,20 @@ def get_context_data(self, **kwargs):
...
+                        "temporal_context": self.temporal_context,
+                        "ooi": self.ooi,
+                    },
+                ),
+                "text": _("Edit"),
+            }
         )
 
-        context["type"] = self.ooi_class.get_ooi_type()
+        context["type"] = self.get_ooi_class().get_ooi_type()
```
underdarknl: Maybe we should promote both of these (ooi_type) and (ooi_class) to properties made available by the SingleOOIMixin. I bet there's more placed that use these values for the current loaded OOI.

rocky/rocky/views/ooi_edit.py
```diff
@@ -64,10 +59,20 @@ def get_context_data(self, **kwargs):
...
         context["ooi_human_readable"] = self.ooi.human_readable
```
underdarknl: Isn't there a human readable template tag that we could use?

rocky/rocky/views/ooi_edit.py
```diff
@@ -64,10 +59,20 @@ def get_context_data(self, **kwargs):
         # Construct breadcrumbs
         breadcrumb_list = self.get_breadcrumb_list()
         breadcrumb_list.append(
```
underdarknl: Should we instead overload the build_breadcrumbs method, and have it call its parent?

rocky/rocky/views/ooi_list.py
```diff
@@ -144,7 +150,12 @@ def _set_scan_profiles(
             messages.SUCCESS,
             _("Successfully set scan profile to %s for %d OOIs.") % (level.name, len(selected_oois)),
         )
-        return redirect(reverse("ooi_list", kwargs={"organization_code": self.organization.code}))
+        return redirect(
```
underdarknl: Can we pick this specific route from the availlable breadcrums?

rocky/rocky/views/ooi_list.py
```diff
@@ -169,7 +180,12 @@ def _set_oois_to_inherit(
         messages.add_message(
             request, messages.SUCCESS, _("Successfully set %d OOI(s) clearance level to inherit.") % len(selected_oois)
         )
-        return redirect(reverse("ooi_list", kwargs={"organization_code": self.organization.code}))
+        return redirect(
```
underdarknl: Can we pick this specific route from the availlable breadcrums?

rocky/rocky/views/ooi_list.py
```diff
@@ -192,7 +208,12 @@ def _delete_oois(self, selected_oois: list[str], request: HttpRequest, *args: An
             _("Successfully deleted %d ooi(s). Note: Bits can recreate objects automatically.") % len(selected_oois),
         )
 
-        return redirect(reverse("ooi_list", kwargs={"organization_code": self.organization.code}))
+        return redirect(
```
underdarknl: Can we pick this specific route from the availlable breadcrums?

rocky/rocky/views/ooi_mute.py
```diff
@@ -50,13 +50,23 @@ def post(self, request, *args, **kwargs):
 
         if not selected_findings:
             messages.add_message(self.request, messages.WARNING, _("Please select at least one finding."))
-            return redirect(reverse("finding_list", kwargs={"organization_code": self.organization.code}))
+            return redirect(
```
underdarknl: Lets deduplicate all three possible return redirect urls in this method by pre-creating the url once, and using it where needed. 
creating a more complex if/elif/else tree could also work, but that increases cyclomatic complexity

rocky/rocky/views/ooi_mute.py
```diff
@@ -50,13 +50,23 @@ def post(self, request, *args, **kwargs):
...
 
             messages.add_message(self.request, messages.SUCCESS, _("Finding(s) successfully unmuted."))
-            return redirect(reverse("finding_list", kwargs={"organization_code": self.organization.code}))
+            return redirect(
+                reverse(
+                    "finding_list",
+                    kwargs={"organization_code": self.organization.code, "temporal_context": self.temporal_context},
+                )
+            )
         else:
```
underdarknl: else is not needed, is either unmute, and returned, or we get to here.

rocky/rocky/views/ooi_view.py
```diff
@@ -141,42 +140,55 @@ def setup(self, request, *args, **kwargs):
...
 
         context["ooi"] = self.ooi
-        context["ooi_current"] = self.get_current_ooi
-        context["mandatory_fields"] = get_mandatory_fields(self.request)
+        context["ooi_current"] = self.get_now_ooi
         return context
 
     def build_breadcrumbs(self) -> list[Breadcrumb]:
         start: Breadcrumb
         if isinstance(self.ooi, Finding):
```
underdarknl, line 165: Maybe we should toggle the path_name (finding_list / ooi_list) based on the class, and have the rest of the start dict and url construction occur only once

rocky/rocky/views/upload_csv.py
```diff
@@ -65,14 +65,30 @@ def setup(self, request, *args, **kwargs):
             self.add_error_notification(CSV_ERRORS["no_org"])
 
     def get_success_url(self):
-        return reverse_lazy("ooi_list", kwargs={"organization_code": self.organization.code})
+        return reverse_lazy(
```
underdarknl: Lets pull this url from the breadcrumbs available on the view.

rocky/rocky/views/upload_raw.py
```diff
@@ -31,22 +31,42 @@ def get_initial(self):
         return initial
 
     def get_success_url(self):
-        return reverse_lazy("ooi_list", kwargs={"organization_code": self.organization.code})
+        return reverse_lazy(
```
underdarknl: same here, already available on the breadcrumbs.

rocky/rocky/views/upload_raw.py
```diff
@@ -31,22 +31,42 @@ def get_initial(self):
...
+                ),
                 "text": _("Upload raw"),
             },
         ]
         return context
 
     def form_valid(self, form):
         if not self.process_raw(form):
-            return redirect("upload_raw", organization_code=self.organization.code)
+            return redirect(
```
underdarknl: and here, pull from the breadcrumbs.

rocky/tests/objects/test_objects.py
underdarknl, line 1: Should we make the kwargs available as a fixture? It seems like we can avoid a lot of duplication that way.

rocky/tests/test_members.py
```diff
@@ -182,7 +182,7 @@ def test_check_add_redteamer_form(rf, admin_member):
     assertContains(
         response,
         '<input type="radio" name="trusted_clearance_level" value="-1" radio_paws="True" '
-        'id="id_trusted_clearance_level_0" required="True" checked="True" checked="True">',
+        'id="id_trusted_clearance_level_0" required="True" checked="True">',
```
underdarknl, line 185: Its good to see that whomever wrote the html, also did not spot the double checked -= true and just included it in the test :P

underdarknl (commented):
small nits that should above all remove lots of lines, but, I think but overal we are looking good.

hasecon:
Picked up your review round, @underdarknl — thanks, the "remove lots of lines" framing was the right lens. Pushed as `382d43192`; `make utest` = **480 passed / 1 skipped**, pre-commit clean.

### Done

| Your comment | What changed |
|---|---|
| `ooi_mute.py` — dedupe the three redirect urls | `finding_list_url` built once and reused |
| `ooi_mute.py` — `else` not needed | dropped, the unmute branch returns |
| `ooi_list.py` ×3 — pick the route from the breadcrumbs | `redirect(self.build_breadcrumbs()[0]["url"])` |
| `upload_csv.py` / `upload_raw.py` ×3 — already on the breadcrumbs | both now use `BreadcrumbsMixin` + `build_breadcrumbs`; `get_success_url` and the failure redirect read from it. `UploadRaw.get_context_data` disappeared entirely |
| `ooi_edit.py` — overload `build_breadcrumbs`, call the parent | `ooi_edit` and `ooi_delete` now do `[*super().build_breadcrumbs(), {leaf}]` |
| `ooi_view.py` — toggle only the path_name (you resolved it, but it fitted) | the shared builder toggles `list_route`/`list_text` and constructs the dicts once |
| `ooi_edit.py` — `ooi_type` as a mixin property | `SingleOOIMixin.ooi_type` added and used |
| `ooi_edit.py` — human readable via the template | context var dropped; the template renders `ooi.human_readable` |
| `test_objects.py` — kwargs as a fixture | `url_kwargs` fixture, 16 duplicated dicts gone |

`get_breadcrumb_list` is gone: it was a near-duplicate of `BaseOOIDetailView.build_breadcrumbs`, so both are now one `build_breadcrumbs` on `SingleOOIMixin`.

### One thing that refactor exposed — it backs up your mixin point

Renaming it to `build_breadcrumbs` made `SingleOOIMixin` **shadow the report breadcrumbs** in `OnboardingReportView`, which mixes both `SingleOOIMixin` and `BaseReportView` while its route has no `<ooi>` at all. Four onboarding tests failed on it. I fixed it with an explicit fallback (no ooi in the route → defer to the next builder in the MRO), but that is exactly the mess your *"those routes should be fixed to not include this mixin"* comment is about — so that follow-up is more than cosmetic.

### Deliberately not in this commit

- **`ooi_class` as a property.** `ooi_type` was easy, but `OOIAddView.setup` assigns `self.ooi_class = self.get_ooi_class()` (`ooi_add.py:78`), which a read-only property would break. Doable with a setter or by refactoring `OOIAddView`, but that belongs in its own change rather than hidden in this sweep.
- **`step_9` ooi on the url/kwargs** and **`report_id` → `<ooi:ooi>` routing.** Both right, but the onboarding wizard carries the ooi through `get_selection()` across steps 5→10, and `report_id` runs through the whole reports app. Migrating one step or one template makes it *less* consistent; happy to do either as a follow-up PR.

rocky/rocky/views/ooi_mute.py
```diff
@@ -48,38 +48,25 @@ def post(self, request, *args, **kwargs):
...
-                reverse(
-                    "finding_list",
-                    kwargs={"organization_code": self.organization.code, "temporal_context": self.temporal_context},
-                )
-            )
+            return redirect(finding_list_url)
+
         if unmute:
             mutes_finding_refs = [MutedFinding(finding=finding).reference for finding in selected_findings]
             self.octopoes_api_connector.delete_many(mutes_finding_refs, datetime.now(timezone.utc))
```
underdarknl: This delete_many should adhere to the temporal_context, unmuting at a given time should do just that.

self.octopoes_api_connector.delete_many(mutes_finding_refs, self.observed_at)

rocky/rocky/views/ooi_mute.py
```diff
@@ -48,38 +48,25 @@ def post(self, request, *args, **kwargs):
...
-                reverse(
-                    "finding_list",
-                    kwargs={"organization_code": self.organization.code, "temporal_context": self.temporal_context},
-                )
-            )
+            return redirect(finding_list_url)
+
+        oois = [self.ooi_class.model_validate({"finding": finding, "reason": reason}) for finding in selected_findings]
+
+        create_oois(self.octopoes_api_connector, self.bytes_client, oois, datetime.now(timezone.utc), end_valid_time)
```
underdarknl: This create should also use the temporal_context for this request.

create_oois(self.octopoes_api_connector, self.bytes_client, oois, self.observed_at, end_valid_time)

underdarknl (commented):
We should use the given temporal_context for ooi writes/deletes just like how we use it for reads. I suspect there are more locations where this is happening and now becoming more clear.

The user is navigating on a given timeline, making changes on that timeline should happen then, not in the 'now'.
We might want to create a template partial that we can sprinkle near the various edit/delete pages / confirm pages that allows the user to either continue with editing then 'then', or refresh the page/form to the 'now'.

rocky/rocky/views/ooi_edit.py
underdarknl: moved to a new issue.

underdarknl:
Lets spin out the restructuring of the report url's to include the report_id in the route itself into a new issue. Im guessing this will remove a bunch of duplicate code once we start just using the singleOOIMixin etc on those routes.

hasecon:
Picked up this morning's comments, @underdarknl. Pushed as `b67d3a44d` (on top of your merge of main); `make utest` = **482 passed / 1 skipped**, pre-commit clean.

### Both mute writes now follow the temporal context

| Your comment | What changed |
|---|---|
| `ooi_mute.py:62` — unmuting at a given time should do just that | `delete_many(mutes_finding_refs, self.observed_at)` |
| `ooi_mute.py:69` — this create should also use the temporal context | `create_oois(..., oois, self.observed_at, end_valid_time)` |

`observed_at` falls back to `datetime.now(timezone.utc)` when the route carries no temporal context, so "now" behaviour is unchanged and only an explicit historic view writes at that moment — the same rule you set for delete.

Two regression tests pin it (`test_mute_findings_bulk_writes_at_the_viewed_time`, `test_unmute_findings_bulk_deletes_at_the_viewed_time`): both assert the valid_time that actually reaches the connector, and both go red if the `datetime.now()` calls come back.

### One thing worth your call

The same "writes adhere to the temporal context" rule arguably applies to more write sites that still hardcode `datetime.now(timezone.utc)`. I did **not** touch them, because I would rather you say which ones are deliberate than have me guess — the delete case earlier this round is exactly why:

- `finding_type_add.py:64` — `Declaration(valid_time=...)` for a manually added finding type
- `upload_csv.py:119,167` — declarations for uploaded objects
- `ooi_list.py:161,182` — bulk scan-profile saves
- `scan_profile.py:53` — resetting a scan profile to `EmptyScanProfile`
- `report_overview.py:252,335` — deleting reports and creating a report OOI

My reading is that the declaration-style ones (finding type, CSV upload) follow the same logic as mute and delete, while scan-profile changes are arguably always "now" because clearance is a statement about the present. Happy to sweep whichever set you name.

### Follow-ups spun out

- #5385 — report URLs carrying the report in the route instead of `?report_id=`, as you asked. It also closes a concrete bug: `get_temporal_url` rebuilds from `resolver_match.kwargs` only, so switching the temporal selector on a report view drops `report_id` and `get_report_ooi()` ends up calling `get_report()` with `Reference.from_str(None)`. With the report in the route that cannot happen.
- #5384 (yours) covers `ooi_type`/`ooi_class` as `SingleOOIMixin` properties — noting there that `OOIAddView.setup` assigns `self.ooi_class` (`ooi_add.py:78`), so that one needs a setter or a small `OOIAddView` refactor.

Still not re-run here: the functional test (`make kat` + browser E2E). It touches nearly every link on the branch, so it is worth a smoke before merge — say the word and I will run it.

underdarknl:
I think the intention of a user modifying any OOI (or clearance level) is to modify it in the given time.
Since we do have both the transaction time and the valid time of each 'change', we can still explain why for example boefje jobs have ran `after` the newly set valid time on a clearance level. 
Changing the valid time does however possibly have a ripple effect on the inherited clearance levels which I'm not sure on which time they would be registered. My best understanding is that they should probably be using the same timeframe as the newly changed clearance level and I guess they wont at this moment.
Changing these probably also means the scheduler will consume these changes and create new jobs (in the now) for Ooi's and clearance levels that changes in the `then`. Maybe we simply guard against this for now by ignoring those events.

hasecon:
Clear — swept the rest on that rule. Pushed as `5e8ba666d`; `make utest` = **485 passed / 1 skipped**, `make itest` = 5 passed, pre-commit clean.

### Swept

| Site | Was |
|---|---|
| `account/mixins.py:173,187` — `raise_clearance_level` / `raise_clearance_levels` | declared clearance written at `now()` |
| `ooi_list.py:161` — set clearance to inherit | `now()` |
| `ooi_list.py:182` — bulk object delete | `now()` |
| `scan_profile.py:53` — reset to `EmptyScanProfile` | `now()` |
| `finding_type_add.py:64` — manually added finding type | `now()` |
| `upload_csv.py:119,167` — declarations for uploaded objects | `now()` |
| `report_overview.py:121,252,335` — recipe delete, report delete, report rename | `now()` |
| `reports/views/base.py:307` — `ReportRecipe` create | `now()` |

`FindingTypeAddView` and `UploadCSV` needed `ObservedAtMixin` — both already sat on a route carrying `<tc:temporal_context>` and simply never read it.

Three regression tests pin the clearance and bulk-delete paths (`test_set_clearance_level_at_the_viewed_time`, `test_set_clearance_level_to_inherit_at_the_viewed_time`, `test_delete_oois_at_the_viewed_time`); all three assert the valid_time that reaches the connector and go red on the old code.

**Left as `now()` on purpose** — none of these is a user modifying an OOI: the API viewset `valid_time` fallback (`account/mixins.py:297`), the scheduler `deadline_at` values (`ooi_edit.py:44`, `report_overview.py:310`), the `observed_at`-vs-now clamps in the report views (`base.py:267,592`), and `get_now_ooi` (`ooi_view.py:152`), which exists precisely to fetch today's version.

### Your two open points — both check out, and the second is worse than "probably won't"

**Inherited clearance levels do not follow the changed level's timeframe.** Confirmed. `recalculate_scan_profiles(valid_time)` writes each `InheritedScanProfile` at the `valid_time` it is handed (`octopoes/core/service.py:350`), but its only caller is the periodic worker, which always hands it the present:

```python
# octopoes/tasks/scanprofiles.py:89
recalc_org["octopoes"].recalculate_scan_profiles(datetime.now(timezone.utc))
```

So a declaration written at `T` propagates — a fact valid from `T` is in effect at now — but every inherited profile it produces is stamped `now`, not `T`. Nothing in the save path triggers propagation at `T`; `_on_create_scan_profile` / `_on_update_scan_profile` only run inferences (`service.py:548-554`). Making inheritance follow the declaration would mean recalculating at the event's `valid_time` rather than on a fixed interval — an octopoes change, not a rocky one.

**The scheduler cannot currently be guarded, because the mutation carries no time at all.** `ScanProfileMutation` is only `operation`, `primary_key`, `value`, `client_id` (`octopoes/events/manager.py:25-29`, mirrored in `mula/scheduler/models/ooi.py:31`). The publisher has `event.valid_time` in hand and does not include it. So mula has no field to filter on — "ignore historic events" needs `valid_time` added to the payload on the octopoes side first, then a check in `BoefjeScheduler.process_mutations`. That is an octopoes + mula change and I would not put it in this PR.

Worth being explicit about the consequence in the meantime: with this commit a user can set a clearance level in the past, and the scheduler will queue boefjes for it in the now. That is the behaviour you described; it just isn't guarded yet. Happy to open the payload change as its own issue, or to hold the clearance half of this commit back until the guard exists — your call, and easy to split off since it is two lines in `account/mixins.py`.

brennodewinter:
Coordination note re #5409 / #3638 (the `Y-m-d H:i:s:u` → `s.u` fix for `observed_at` in report links) — flagged on #5409 that it clashes here, so noting what I found on this branch's head:

**Leftover dead param in `scheduled_reports_table.html`** (lines ~94/97): the "Recent reports" links still append `&observed_at={{ ...|date:"Y-m-d H:i:s:u" }}`. Two issues: the format is still the buggy colon variant from #3638, and more importantly `ObservedAtMixin` on this branch no longer reads `request.GET["observed_at"]` at all — `observed_at` comes from `temporal_context` in the path only. So these params are dead weight, and the links silently use the page's temporal context instead of the report's own `observed_at`. `report_history_table.html` already does this correctly with `{% app_url "view_report" temporal_context=report.report.observed_at %}` — suggest the same here (or drop the params if report-URL restructuring is deferred anyway).

**Four templates outside this PR's diff** still build `view_report` links with `?observed_at=`: `organization_crisis_room_dashboard_items.html`, `partials/report_dashboard_item_filter.html`, `subreports_header.html`, `subreports_table.html`. Once this lands and the GET param is no longer read, those links lose their timestamp and fall back to "now" — worth tracking in the report-URL follow-up issue you proposed spinning out.

Meanwhile I'd suggest landing #5409 first (small, fixes a live bug on main); the conflict here is confined to the two files you rewrite anyway.

PR #5257, opened 2026-07-23
Fix vulnerability report HTTP header findings

### LINKED ISSUE #5203: Vulnerability Report returns no findings: missing HTTPHeader query path + inverted KAT filter + cross-IP accumulation
## Summary

The **Vulnerability Report** (`rocky/reports/report_types/vulnerability_report/report.py`) produces an **empty result** for a host that clearly has vulnerabilities, and structurally **cannot** show HTTP-header–based findings such as **CVE-2021-41773** (CVSS 9.8) on an `Apache/2.4.49` `Server` header.

These are pre-existing bugs (originating in #2154 / #2586), independent from the tree-traversal regression tracked in #5202. Both issues happen to hide the same CVE, which is how this surfaced.

## Live evidence (input = `Hostname|internet|apache.dvwa.cloud`)

```
VulnerabilityReport.collect_data(...) -> vulnerabilities: []   # completely empty
```

The host has, in Octopoes: `CVE-2021-41773` (on an `HTTPHeader`), `KAT-OUTDATED-SOFTWARE` (Apache 2.4.49, on a `SoftwareInstance`), `KAT-UNCOMMON-OPEN-PORT` (on an `IPPort`), and more. None reach the report.

### 1. No query path covers `HTTPHeader`-attached findings
Only two finding-bearing paths are queried:

```python
"IPAddress.<address[is IPPort].<ooi[is Finding]"
"IPAddress.<address [is ResolvedHostname].hostname.<netloc [is HostnameHTTPURL].<ooi [is SoftwareInstance].<ooi [is Finding]"
```

CVE-2021-41773 is attached to an `HTTPHeader` (`Finding|HTTPHeader|...|Server|CVE-2021-41773`), reachable via `Website → HTTPResource → HTTPHeader`. That path is never queried, so the CVE is structurally unreachable.

### 2. Inverted `KATFindingType` filter drops every KAT finding

```python
for finding in findings:
    if finding.finding_type.class_type != KATFindingType:   # keeps ONLY non-KAT
        aggregated_findings.append(finding)
```

(and the same `!=` on `finding_type` a few lines up). This keeps only non-KAT finding types and **discards all KAT findings** — including the ones the queries *do* return (`KAT-UNCOMMON-OPEN-PORT`, `KAT-OUTDATED-SOFTWARE`). For a report titled *"Vulnerabilities found are grouped for each system"* this looks inverted; it should almost certainly be `==` (or the filter removed). This is why the result is fully empty: query #1's KAT findings are dropped, and the CVE never arrives via the paths in defect #1.

### 3. Cross-IP accumulation
`all_finding_types` and `aggregated_findings` are declared once per input OOI but appended to across the inner `for ip in ip_references` loop, then assigned to `findings_data[ip]` every iteration. Findings from one IP leak into another IP's data (and get duplicated).

## Net effect for the reported case

- `KAT-OUTDATED-SOFTWARE` (Apache 2.4.49) **is** retrieved via the SoftwareInstance path → dropped by defect #2.
- The real `CVE-2021-41773` is **never** retrieved → defect #1.

Result: the user sees neither "outdated software" nor the critical CVE → "OpenKAT detects it but the Vulnerability Report is empty."

## Suggested fix (one PR)

1. Add a finding query path that traverses HTTP headers/resources (e.g. `... HostnameHTTPURL.<resource [is HTTPResource].<resource [is HTTPHeader].<ooi [is Finding]`), or — preferably — collect findings by walking the asset subtree once (and reuse the corrected helper from #5202 so both reports share one correct collection path).
2. Fix the inverted `!= KATFindingType` filter (confirm intent: KAT findings belong in a vulnerability report).
3. Move/reset the `all_finding_types` / `aggregated_findings` accumulators inside the per-IP loop.

Related regression: #5202 (Findings/DNS/Aggregate reports miss the same CVE via a different mechanism).

brennodewinter:
## Summary

- include vulnerability findings attached to `HTTPHeader` objects in the Vulnerability Report
- deduplicate findings that are reachable through multiple relation paths
- keep findings isolated per IP address so counts cannot leak between systems
- bulk-load finding types while preserving the existing exclusion of `KATFindingType`
- add unit and integration regression coverage for HTTP-header CVEs, deduplication, KAT filtering, and IPv4/IPv6 isolation

## Root cause

The report only queried findings attached directly below an IP port or below a software instance reached through a hostname. Findings attached to an HTTP header therefore had no relation path into the report. The existing aggregation also reused mutable lists between IP addresses, which could duplicate counts.

## Impact

CVE and other non-KAT vulnerability findings attached to HTTP headers are now shown under the IP address where the header was observed, with stable occurrence counts.

## Validation

- `7 passed` in the targeted Vulnerability Report test module on PostgreSQL
- `44 passed` across the relevant Rocky report-type tests
- all three Octopoes finding paths parsed and compiled to XTDB queries
- Ruff lint and format checks passed
- Python compile check and `git diff --check` passed

The full local UI/PDF test collection requires native WeasyPrint/GLib libraries that are not installed in this environment; the relevant report logic and regression suite passed.

Closes #5203

### DIFF
--- a/rocky/reports/report_types/vulnerability_report/report.py
+++ b/rocky/reports/report_types/vulnerability_report/report.py
@@ -4,12 +4,23 @@ from datetime import datetime
 from typing import Any, TypedDict
 
 from django.utils.translation import gettext_lazy as _
-
 from octopoes.models import Reference
 from octopoes.models.ooi.dns.zone import Hostname
 from octopoes.models.ooi.findings import Finding, FindingType, KATFindingType, RiskLevelSeverity
 from octopoes.models.ooi.network import IPAddressV4, IPAddressV6
+
 from reports.report_types.definitions import Report
+
+PORT_FINDINGS_PATH = "IPAddress.<address[is IPPort].<ooi[is Finding]"
+SOFTWARE_INSTANCE_FINDINGS_PATH = (
+    "IPAddress.<address [is ResolvedHostname].hostname.<netloc [is HostnameHTTPURL]"
+    ".<ooi [is SoftwareInstance].<ooi [is Finding]"
+)
+HTTP_HEADER_FINDINGS_PATH = (
+    "IPAddress.<address[is IPPort].<ip_port [is IPService].<ip_service [is Website]"
+    ".<website [is HTTPResource].<resource [is HTTPHeader].<ooi [is Finding]"
+)
+FINDING_PATHS = (PORT_FINDINGS_PATH, SOFTWARE_INSTANCE_FINDINGS_PATH, HTTP_HEADER_FINDINGS_PATH)
 
 
 class FindingsData(TypedDict):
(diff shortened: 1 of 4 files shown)


underdarknl:
It looks like there's a bunch of changed lines that are not in line with our precommit / prek rules.
These lines functionally dont realy change anything anyway (they have just moved), and as such add noise to the PR.
Please make sure to run prek (and our other linters as stipulated in the workflows) to minimize drift between coding styles and PR noise.

brennodewinter:
Superseded by #5258. The replacement includes the CI fixes; this branch could not be updated because the repository ruleset requires verified signatures for branch updates.

PR #5271, opened 2026-07-31
Add get_normalizer_metas to BytesAPIClient and use it in the integration tests

### LINKED ISSUE #5270: add missing client functionality in BytesAPIClient, and rewrite the test to use that after.
Lets do a follow up PR which adds the missing client functionality in BytesAPIClient, and rewrite the test to use that after.

_Originally posted by @underdarknl in https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/issues/5114#issuecomment-5141474173_

hasecon:
### Changes

Follow-up to #5114, as requested in #5270:

- Adds a `get_normalizer_metas` method to the test `BytesAPIClient` for `GET /bytes/normalizer_metas` (list of UUIDs → `dict[str, NormalizerMeta]`). It follows the conventions of the sibling methods: `@retry_with_login`, `_verify_response`, and pydantic model validation of the response.
- Rewrites the two `/bytes/normalizer_metas` integration tests from #5114 to go through the client method instead of raw `bytes_api_client.client.get(...)` calls. Since the client now returns validated `NormalizerMeta` models, the assertions also check the returned models (`id`, `normalizer.id`), not just the response keys.

### Issue link

Closes #5270

### QA notes

Test-only change (the client lives in `bytes/tests/`); no runtime code touched.

Verified locally in Docker:
- `bytes make utest` — 14 passed, 1 skipped
- `bytes make itest` — 31 passed, 1 skipped (both rewritten tests pass through the new client method)
- `pre-commit` on the changed files — passed

---

### Code Checklist

### DIFF
--- a/bytes/tests/client.py
+++ b/bytes/tests/client.py
@@ -119,6 +119,25 @@ class BytesAPIClient:
         return [NormalizerMeta.model_validate(normalizer_meta) for normalizer_meta in normalizer_meta_json]
 
     @retry_with_login
+    def get_normalizer_metas(
+        self, normalizer_meta_ids: list[UUID], limit: int = 100, offset: int = 0
+    ) -> dict[str, NormalizerMeta]:
+        response = self.client.get(
+            "/bytes/normalizer_metas",
+            params={
+                "normalizer_metas": [str(normalizer_meta_id) for normalizer_meta_id in normalizer_meta_ids],
+                "limit": limit,
+                "offset": offset,
+            },
+        )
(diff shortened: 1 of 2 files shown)


underdarknl (approved):

underdarknl:
Maybe a good followup PR would be to move this bytes client from the test folder and have it present as a utility in the bytes folder itself.
We could then also use that client from within Rocky instead of defining mostly the same code there.

PR #5267, opened 2026-07-24
Allow superusers to promote accounts in Rocky

brennodewinter:
### Changes

Adds a separate installation-wide **Grant superuser access** action to Rocky's member edit flow.

- only an existing superuser can see or invoke the action;
- the target is identified on a dedicated confirmation page that explains the global impact;
- POST + CSRF sets both `is_superuser` and `is_staff` inside a locked transaction;
- inactive accounts fail closed and already-promoted accounts remain an idempotent no-op;
- event `900113` records actor, target, previous/new state and timestamp after the database commit;
- unauthorized requests resolve before the target member lookup, avoiding a membership-ID oracle;
- the interface and documentation give the exact Django Admin recovery path because Rocky cannot revoke the permission yet.

The root cause was that Rocky modeled organization roles separately from global account flags and exposed no controlled UI action for `KATUser.is_superuser`; operators therefore needed shell or Django Admin access.

The generated `django.pot` refresh is isolated in its own commit. It includes the feature strings plus reproducible pre-existing catalog drift from current scheduler/task/bytes sources because the repository-native generator refreshes the complete catalog.

No migration, new dependency, configuration change, organization-role change, or breaking change is required.

### Issue link

Closes #5266

### Demo

The action appears under **Global account access** on an active non-superuser's member edit page. It opens a dedicated warning page showing the target email, installation-wide scope, recovery path, destructive confirmation, and cancel action.

### QA notes

1. Sign in as a superuser and edit an active non-superuser member.
2. Select **Grant superuser access** under **Global account access**.
3. Verify the target email, global warning, and Django Admin recovery instructions.
4. Confirm and verify the account now has both superuser and staff status and event `900113` is logged.
5. Repeat the POST and verify it is a no-op with clear feedback.
6. Verify admin, redteam, and client accounts neither see the action nor can invoke its URL.
7. Verify an inactive account does not show the action and cannot be promoted by direct POST.

Local validation:

- `prek run --all-files --show-diff-on-failure --color never`
- full Rocky unit test suite on PostgreSQL 15
- `pytest tests/test_members.py -q --no-cov` — 28 passed

Docker-based integration tests were not available locally; the repository CI workflow covers them.

### LibreKAT core values assessment

[...]

PR #5279, opened 2026-08-03
Allow superusers to grant and revoke superuser access in Rocky

brennodewinter:
## Summary

Supersedes #5267, addressing all three review points from @underdarkNL:

- **Unreachable test code**: rewrote the rolled-back-grant test to use `transaction.set_rollback(True)` instead of an unreachable `raise RuntimeError`, with a docstring explaining the intent.
- **Revoke superuser access**: adds `RevokeSuperuserAccessView` mirroring the grant flow, with a last-active-superuser guard to prevent installation lockout, audit event `900114`, a dedicated confirmation page, and a revoke link on the member edit page. The grant page and docs now reflect that Rocky can revoke the permission itself.
- **Confirmation dialog**: both grant and revoke open a dedicated confirmation page with `role="alert"` warning, target email, installation-wide impact, destructive confirm button, and cancel link — a11y-friendly without requiring JS.

### Changes from #5267

- `rocky/views/organization_member_superuser.py`: shared `SuperuserAccessView` base, new `RevokeSuperuserAccessView` with last-superuser guard
- `rocky/templates/organizations/organization_member_revoke_superuser.html`: new revoke confirmation page
- `rocky/templates/organizations/organization_member_edit.html`: show grant OR revoke link based on target status
- `rocky/templates/organizations/organization_member_grant_superuser.html`: updated text (Rocky can now revoke)
- `rocky/urls.py`: revoke URL
- `rocky/tests/test_members.py`: revoke tests (confirmation, success, rollback, idempotent, last-superuser guard, non-superuser denied, CSRF), fixed rollback test
- `docs/`: event 900114, updated users-and-organizations page
- `rocky/rocky/locale/django.pot`: revoke strings

#### Test plan

- [x] 38/38 `test_members.py` tests pass (PostgreSQL 15, Django 5.1)
- [x] ruff: 0 errors
- [x] djlint: 0 errors on all templates
- [ ] CI on this PR

### DIFF
--- a/docs/source/installation-and-deployment/events-and-logging.rst
+++ b/docs/source/installation-and-deployment/events-and-logging.rst
@@ -61,6 +61,7 @@ Event code Model              Routing key            Description                
 900111     TOTPDevice         account_change         2FA is removed.                             D
 900112     TOTPDevice         account_change         2FA is updated.                             U
 900113     KATUser            account_change         Superuser access is granted.                U
+900114     KATUser            account_change         Superuser access is revoked.                U
 900201     Organization       organization_change    A new organization is created.              C
 900202     Organization       organization_change    Organization information changed.           U
 900203     Organization       organization_change    Organization is removed.                    D
--- a/docs/source/installation-and-deployment/events-and-logging.rst
+++ b/docs/source/installation-and-deployment/events-and-logging.rst
@@ -60,6 +60,7 @@ Event code Model              Routing key            Description                
 900110     KATUser            account_change         A user is deleted.                          D
 900111     TOTPDevice         account_change         2FA is removed.                             D
 900112     TOTPDevice         account_change         2FA is updated.                             U
+900113     KATUser            account_change         Superuser access is granted.                U
 900201     Organization       organization_change    A new organization is created.              C
 900202     Organization       organization_change    Organization information changed.           U
 900203     Organization       organization_change    Organization is removed.                    D
--- a/docs/source/installation-and-deployment/users-and-organizations.rst
+++ b/docs/source/installation-and-deployment/users-and-organizations.rst
@@ -17,6 +17,10 @@ Users
 
 OpenKAT knows four types of users: the client, the red team user, the admin and the superuser. In OpenKAT, permissions utilise a stacked model. This means that a higher permission level includes all lower permissions of the lower levels. The client is a 'read only' type of user, the red teamer is a researcher who can start scans. The admin is an administrative user who can do user management etc, the superuser has the ability to do everything.
 
+Superuser access is global and applies to every organization in an OpenKAT installation. An existing superuser can grant this access to another active account by opening the organization's member overview, editing the member, and selecting ``Grant superuser access`` under ``Global account access``. Rocky shows the installation-wide impact for confirmation before applying the change. Organization admins and other account types cannot grant superuser access.
+
+Rocky cannot revoke superuser access. To revoke it, open the account under :guilabel:`Accounts` > :guilabel:`Users` in Django administration, clear both :guilabel:`Superuser status` and :guilabel:`Staff status`, and save the account.
+
 Rights and functions per user type
 ----------------------------------
 
--- a/docs/source/installation-and-deployment/users-and-organizations.rst
+++ b/docs/source/installation-and-deployment/users-and-organizations.rst
@@ -19,7 +19,7 @@ OpenKAT knows four types of users: the client, the red team user, the admin and 
 
 Superuser access is global and applies to every organization in an OpenKAT installation. An existing superuser can grant this access to another active account by opening the organization's member overview, editing the member, and selecting ``Grant superuser access`` under ``Global account access``. Rocky shows the installation-wide impact for confirmation before applying the change. Organization admins and other account types cannot grant superuser access.
 
-Rocky cannot revoke superuser access. To revoke it, open the account under :guilabel:`Accounts` > :guilabel:`Users` in Django administration, clear both :guilabel:`Superuser status` and :guilabel:`Staff status`, and save the account.
+An existing superuser can revoke this access again from the same member edit page by selecting ``Revoke superuser access``. Rocky refuses to revoke the last active superuser in an installation to prevent lockout; promote another account first, or revoke via Django administration. To revoke superuser access outside Rocky, open the account under :guilabel:`Accounts` > :guilabel:`Users` in Django administration, clear both :guilabel:`Superuser status` and :guilabel:`Staff status`, and save the account.
 
 Rights and functions per user type
 ----------------------------------
--- a/rocky/rocky/locale/django.pot
+++ b/rocky/rocky/locale/django.pot
@@ -9,7 +9,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-07-24 20:14+0000\n"
+"POT-Creation-Date: 2026-08-19 13:43+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
@@ -590,6 +590,7 @@ msgstr ""
 #: rocky/templates/organizations/organization_edit.html
 #: rocky/templates/organizations/organization_member_edit.html
 #: rocky/templates/organizations/organization_member_grant_superuser.html
+#: rocky/templates/organizations/organization_member_revoke_superuser.html
 #: rocky/templates/partials/delete_ooi_modal.html
 #: rocky/templates/partials/edit_ooi_clearance_level_modal.html
 #: rocky/templates/partials/mute_findings_modal.html
@@ -931,6 +932,7 @@ msgstr ""
 #: reports/templates/report_overview/scheduled_reports_table.html
 #: reports/templates/summary/selected_plugins.html
 #: rocky/templates/findings/finding_list.html
+#: rocky/templates/oois/ooi_detail_origins_observations.html
 #: rocky/templates/tasks/boefjes.html rocky/templates/tasks/normalizers.html
 #: rocky/templates/tasks/ooi_detail_task_list.html
 #: rocky/templates/tasks/plugin_detail_task_list.html
@@ -946,6 +948,7 @@ msgstr ""
 #: reports/templates/report_overview/scheduled_reports_table.html
 #: reports/templates/summary/selected_plugins.html
 #: rocky/templates/findings/finding_list.html
+#: rocky/templates/oois/ooi_detail_origins_observations.html
 #: rocky/templates/tasks/boefjes.html rocky/templates/tasks/normalizers.html
 #: rocky/templates/tasks/ooi_detail_task_list.html
 #: rocky/templates/tasks/plugin_detail_task_list.html
@@ -6058,6 +6061,47 @@ msgstr ""
 msgid "Warning"
 msgstr ""
 
+#: rocky/templates/oois/ooi_detail_origins_observations.html
+#: rocky/templates/tasks/partials/task_actions.html
+msgid "Loading..."
+msgstr ""
+
+#: rocky/templates/oois/ooi_detail_origins_observations.html
+#: rocky/templates/tasks/partials/task_actions.html
+msgid "Normalizer task"
+msgstr ""
+
+#: rocky/templates/oois/ooi_detail_origins_observations.html
+#: rocky/templates/tasks/partials/task_actions.html
+msgid "Consuming Raw file"
+msgstr ""
+
+#: rocky/templates/oois/ooi_detail_origins_observations.html
+#: rocky/templates/tasks/partials/task_actions.html
+msgid "Yielded objects"
+msgstr ""
+
+#: rocky/templates/oois/ooi_detail_origins_observations.html
+#: rocky/templates/tasks/partials/task_actions.html
+msgid "Yielded raw files for task"
+msgstr ""
+
+#: rocky/templates/oois/ooi_detail_origins_observations.html
+#: rocky/templates/tasks/boefje_task_detail.html
+#: rocky/templates/tasks/partials/task_actions.html
+msgid "Download meta and raw data"
+msgstr ""
+
+#: rocky/templates/oois/ooi_detail_origins_observations.html
+#: rocky/templates/tasks/partials/task_actions.html
+msgid "Origin boefje task"
+msgstr ""
+
+#: rocky/templates/oois/ooi_detail_origins_observations.html
+#: rocky/templates/tasks/partials/task_actions.html
+msgid "See all similar tasks"
+msgstr ""
+
 #: rocky/templates/oois/ooi_edit.html
 #, python-format
 msgid "Edit %(type)s: %(ooi_human_readable)s"
@@ -6112,21 +6156,13 @@ msgid "List of views for OOI"
 msgstr ""
 
 #: rocky/templates/oois/ooi_past_due_warning.html
-msgid "This object is past due"
-msgstr ""
-
-#: rocky/templates/oois/ooi_past_due_warning.html
-msgid "This object is past due and has been deleted"
+msgid "You are looking at a historic view of this object."
 msgstr ""
 
 #: rocky/templates/oois/ooi_past_due_warning.html
 msgid ""
-"This object is past due. You are viewing the object state in a past state."
-msgstr ""
-
-#: rocky/templates/oois/ooi_past_due_warning.html
-msgid ""
-"You will not be able to add Findings or other OOI's to past due objects."
+"You are looking at a historic view of this object which has been deleted in "
+"the current timeline"
 msgstr ""
 
 #: rocky/templates/oois/ooi_past_due_warning.html
@@ -6264,6 +6300,12 @@ msgid ""
 msgstr ""
 
 #: rocky/templates/organizations/organization_member_edit.html
+#: rocky/templates/organizations/organization_member_revoke_superuser.html
+#: rocky/views/organization_member_superuser.py
+msgid "Revoke superuser access"
+msgstr ""
+
+#: rocky/templates/organizations/organization_member_edit.html
 #: rocky/templates/organizations/organization_member_grant_superuser.html
 #: rocky/views/organization_member_superuser.py
 msgid "Grant superuser access"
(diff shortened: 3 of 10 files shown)


hasecon (changes requested):
## QA Review

### Setup
- Branch: `pr-5267` @ `30464b3`
- Stack: `make clean && make kat` — ✅ 13/13 services up

Reviewed as a security-sensitive change (privilege escalation path), so the functional test focused on the authorization boundary and the audit trail rather than only the happy path.

### CI/CD Status

| Check | Status | Blocking? |
|---|---|---|
| `test (3.10–3.13)` Rocky | ✅ green | — |
| `CodeQL`, `Analyze (python/js/actions)` | ✅ green | — |
| `build (debian12/13, ubuntu22.04/24.04, rocky)` | ✅ green | — |
| **`pre-commit`** | ❌ **failure** | **yes — Rule 1 (required on any PR)** |
| `makelang` | ❌ failure | no — documented exception, and red repo-wide for weeks |

**`pre-commit` is the one blocker, and it is trivial.** Reproduced locally on the changed files; ruff fixes both automatically:

1. `rocky/tests/test_members.py` — import grouping. The blank line separating third-party from first-party imports was moved: it needs to be after `pytest_django.asserts`, not before `tests.conftest`.
2. `rocky/rocky/views/organization_member_superuser.py` — ruff-format collapses the last breadcrumb dict to one line.

`pre-commit run --all-files` and a commit clears it.

### Code Review Findings

| 1 | **Blocking (trivial)** | `test_members.py`, `organization_member_superuser.py` | `pre-commit` red — formatting only, see above |
| 2 | Low | `organization_member_superuser.py:137` | The comment starts with `# ponytail:` — that prefix appears exactly once in the whole repo and means nothing to a future reader. The comment *content* is valuable (it documents the TOCTOU limit and the recovery path); please keep the text and drop the marker. |
| 3 | Low | `organization_member_superuser.py:138-145` | The last-superuser guard's TOCTOU, which the comment honestly acknowledges: `select_for_update()` locks only the target row, so two superusers concurrently revoking two *different* superusers can each count 1 remaining and both proceed. Reachable with exactly two superusers revoking each other. If you want to close it now rather than "later", locking the set works: `list(User.objects.select_for_update().filter(is_superuser=True, is_active=True))` and count from that — it serialises concurrent revokes without a new constraint. |
| 4 | Low | `organization_member_superuser.py:133` | Revoke asymmetry: grant sets `is_superuser` **and** `is_staff`, but revoke returns early on `if not previous_is_superuser` and leaves `is_staff` untouched. An account with `is_staff=True, is_superuser=False` (reachable via Django admin) is told "does not have superuser access" while retaining Django-admin access. Clearing `is_staff` in that branch — or saying so in the message — would close the loop. |
| 5 | Nit | `test_members.py` | `test_superuser_can_view_revoke_superuser_confirmation` asserts `assertContains(response, "revoke")` — that substring is in the URL and the page title regardless, so it can't fail. Asserting on the warning text (as the grant test does) would make it meaningful. |

### What I verified rather than assumed

- **Authorization boundary holds.** A non-superuser org member POSTing the grant URL for their *own* membership gets **403** and is not promoted; GET is also 403; anonymous is redirected. Target resolution genuinely happens after the permission check, so the 404-vs-403 distinction can't be used to probe membership IDs — the docstring's claim checks out.
- **Audit trail lands in the real stack**, not just in the structlog test fixture:
  `Superuser access granted  event_code=900113 actor_email=… actor_id=1 target_email=… target_id=2 previous_is_superuser=False previous_is_staff=False is_superuser=True is_staff=True changed_at=… ip=… organization_code=dvwa request_id=…`
  Using `transaction.on_commit` so a rolled-back change logs nothing is the right call, and the rollback tests cover it.
- **Event codes 900113/900114** are new (no collision) and documented in `events-and-logging.rst`.
- **The Django admin link** in the grant page resolves — `admin.site.urls` is unconditionally mounted at `rocky/urls.py:72`, so no `NoReverseMatch` risk.
- **`destructive` is an existing Manon class** (used by `ooi_delete.html`), so styling is consistent rather than invented.
- **Emails are lowercased** in `account/models.py:60`, which is why the `.lower()` in the revoke assertion is correct and not accidental.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| rocky | `make utest` (Docker) | ✅ 492 passed, 1 skipped (470 on main + 22 new) |
| rocky | `make itest` (Docker) | ✅ 5 passed |
| rocky | CI `test (3.10–3.13)` | ✅ green |

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Grant confirmation page | ✅ 200 | installation-wide warning, CSRF token, Django-admin link all present |
| Grant | ✅ 302 | target becomes `is_superuser=True, is_staff=True` |
| Grant idempotency | ✅ 302 | second grant warns, state unchanged |
| Revoke | ✅ 302 | target back to `False, False` |
| Last-superuser guard | ✅ | revoking the only superuser refused, account unchanged |
| **Non-superuser self-escalation** | ✅ **403** | not promoted |
| Anonymous access | ✅ 302 | redirected to login |
| Audit log | ✅ | `event_code=900113` with full actor/target/before/after context |
| Log analysis | ✅ | only known startup noise |

### Verdict

**CHANGES REQUESTED** — solely because `pre-commit` is red, which Rule 1 makes blocking on any PR. Run `pre-commit run --all-files`, commit, and this is done; findings 2–5 are polish you can take or leave in the same push.

Substantively this is a careful piece of work. The permission model, the deliberate ordering of permission-check-before-target-resolution, the on-commit audit logging, the last-superuser guard and the honest TOCTOU comment are all better than I usually see on an authorization change, and the 22 tests cover the negative paths as well as the happy one. Nice to review.

One process note, not a finding: the description says the PR was generated with Devin. That's fine by me, but for a privilege-escalation path it's worth having a human walk the grant and revoke flows in a browser once before merge — I exercised them through Django's test client, which skips the real 2FA middleware chain and the actual rendered UI.

underdarknl:
@brennodewinter could you run prek / precommit again?

underdarknl:
Looks like this Pr still needs a pre-commit run to be eligible for merging.

brennodewinter:
Thanks for the review — all five findings addressed in the last push:

- **#1 (blocker):** `pre-commit` now green (ruff-format collapses the last breadcrumb dict; import separator moved to after `pytest_django.asserts`).
- **#2:** `# ponytail:` marker dropped, the TOCTOU explanation kept (and updated, see #3).
- **#3:** TOCTOU closed: the revoke path now locks the full active-superuser set (`select_for_update().order_by("pk")`) *before* the target row, so two concurrent revokes serialise on the set in pk order instead of racing the count (and without deadlocking, since both lock the same rows in the same order).
- **#4:** Revoke no longer stays silent when the target has `is_staff` without `is_superuser` — it now warns that Django administration access is retained. `is_staff` set independently is left untouched (this flow only manages the superuser grant).
- **#5:** Replaced the always-true `assertContains(response, "revoke")` with an assertion on the confirmation warning text; added a test for the staff-without-superuser path.

Verified: `pre-commit run --files …` green (ruff, ruff-format, mypy, vulture, pyupgrade, django-upgrade, codespell, prettier) and `pytest tests/test_members.py` 39 passed.

brennodewinter:
Also fixed the translation template: the new `%(email)s does not have superuser access but still has Django administration access` message was missing from `django.pot`, which would keep the `makelang` check red. Verified the updated `django.pot` is now byte-identical to a fresh `make lang` run (apart from the POT-Creation-Date).

brennodewinter:
Rebased on `main` to resolve the merge conflict (the `django.pot` conflict was just a stale POT-Creation-Date and new translation strings from main). All five findings from the original review are still addressed:

- **#1 (blocker):** `pre-commit` green
- **#2:** `# ponytail:` marker dropped, TOCTOU explanation kept
- **#3:** TOCTOU closed — revoke path locks the full active-superuser set before the target row
- **#4:** Revoke now warns when `is_staff` is retained without `is_superuser`
- **#5:** Replaced the always-true assertion with one on the confirmation warning text

@hasecon — ready for re-review when you have a moment.

brennodewinter:
## Review findings addressed — branch rebased on main

All five findings from @hasecon's review are resolved:

| # | Finding | Status |
|---|---------|--------|
| 1 | `pre-commit` red (formatting) | ✅ Fixed — `pre-commit run --all-files` green |
| 2 | `ponytail:` marker | ✅ Dropped — comment text kept, marker removed |
| 3 | TOCTOU in last-superuser guard | ✅ Fixed — `select_for_update()` now locks the full active-superuser set in pk order, serialising concurrent revokes |
| 4 | Revoke asymmetry (`is_staff` left untouched) | ✅ Fixed — revoke now warns when `is_staff=True` without `is_superuser`, and clears both flags on actual revoke |
| 5 | Meaningless test assert | ✅ Fixed — asserts on warning text, not just "revoke" substring |

### Rebase & django.pot
- Rebased on latest main (was CONFLICTING — `django.pot` conflict resolved).
- `django.pot` rebuilt from main's version + this PR's new strings, avoiding environment-specific `#, python-brace-format` differences that would fail the `makelang` CI check.

### Verification
- `pre-commit run --all-files` — ✅ green
- `make utest` (Docker, Python 3.13) — ✅ 499 passed, 1 skipped (incl. 39 `test_members.py` tests)
- `make lang` — ✅ only new superuser msgids added, no environment-specific drift

Ready for re-review.

hasecon (changes requested):
## QA Review — round 2

### Setup
- Branch: `pr-5267` @ `a652c77`
- Stack: `make clean && make kat` — 13/13 containers up, Rocky HTTP 200
- Test org: `setup_test_org --code dvwa` — OK, plus a second org and four purpose-built accounts (a blocked member, a second superuser, and an account with `is_staff` set independently)

### Round 1 findings — all five verified fixed

Checked against the head, not the summary:

| # | Round-1 finding | Status | How I verified |
|---|---|---|---|
| 1 | `pre-commit` red (blocker) | ✅ Fixed | CI `pre-commit` green on `a652c77` |
| 2 | `# ponytail:` marker | ✅ Fixed | `grep -rn ponytail rocky/` → no matches; the TOCTOU explanation is kept |
| 3 | TOCTOU in last-superuser guard | ✅ Fixed | Revoke now locks the full active-superuser set via `select_for_update().order_by("pk")` before the target row. I re-checked the lock ordering for a new deadlock risk between the grant and revoke paths and found none — grant takes member → single user, revoke takes member → superuser set (pk order) → target, with no inversion. |
| 4 | Revoke asymmetry on `is_staff` | ✅ Fixed as asked | The staff-without-superuser branch now warns instead of staying silent. See new finding 2 for the other half of this. |
| 5 | Always-true `assertContains(response, "revoke")` | ✅ Fixed | Now asserts on the confirmation warning text |

Thanks for the thorough turnaround — every point was addressed, and the concurrency fix in particular is done properly rather than papered over.

### CI/CD Status

All 20 checks green (`test` 3.10–3.13, `pre-commit`, `CodeQL`, `Analyze` ×3, all four `build` legs, `makelang`, `create_container_image`, `build-artifact`). No blocking check is red.

Note on `makelang`: it is green here even though only `django.pot` changed and no `.po` files did, so there is nothing to follow up on that front.

### Code Review Findings

Two new substantive issues, both reproduced end to end. Neither is a regression from round 1 — they were present before and I did not catch them, so they are on me as much as anyone.

| # | Sev | File | Issue |
|---|-----|------|-------|
| 1 | Medium | `views/organization_member_superuser.py:82` | Grant has no `blocked` guard — a blocked member can be promoted to superuser, creating a state the members list is built to prevent |
| 2 | Medium | `views/organization_member_superuser.py:165` | Revoke unconditionally clears `is_staff`, contradicting the view's own stated policy 15 lines above; the existing test cannot catch it |
| 3 | Low | `views/organization_member_superuser.py:63` | Self-revoke redirects the actor to a 403 page |
| 4 | Low | `templates/…/organization_member_edit.html:33` | Revoke link is shown for the last active superuser, where the POST always refuses |

---

#### 1. (Medium) Grant has no `blocked` guard

`post()` guards `is_active` but never checks `OrganizationMember.blocked`, and the edit page offers the link for blocked members. Reproduced on the running stack:

```
before: superuser=False staff=False blocked=True
POST /en/dvwa/members/3/grant-superuser/ -> 302
after : superuser=True  staff=True  blocked=True
```

What that account can then reach:

```
/en/dvwa/members      (the org that blocked them)  -> 403   block still holds here
/en/otherorg/members  (an org they never joined)   -> 200
/en/admin/tools/organizationmember/                -> 200   can clear their own block
```

So the block survives only in the blocking organization, while the account gains full access to every other organization plus Django admin, where the `blocked` flag itself can be edited.

The more concrete problem is the state it leaves behind. `organization_member_list.html` deliberately makes superusers unblockable — line 117 passes `input_disabled=member.user.is_superuser` to the unblock checkbox, and line 107 disables the Edit button for superusers. After the grant, the members list shows that row as **"Super user"** with the Blocked checkbox checked *and disabled*, and no working Edit button. I confirmed this in the browser: the org admin can no longer unblock or edit that member from the members list at all. The grant flow produces exactly the combination the rest of the UI is written to make impossible.

Suggested fix: guard it the same way `is_active` is guarded, e.g.

```python
if target_member.blocked:
    messages.warning(
        request,
        _("%(email)s is blocked in this organization and cannot be granted superuser access.")
        % {"email": target_user.email},
    )
    return HttpResponseRedirect(self.get_success_url())
```

and hide the link in `organization_member_edit.html` when `object.blocked`. If you would rather allow it deliberately, then the confirmation page should say the target is currently blocked — right now it says nothing, and the blocked checkbox sits on the page the admin just came from.

#### 2. (Medium) Revoke clears `is_staff` unconditionally, contradicting its own policy

Lines 137-150 explicitly decline to touch `is_staff`, telling the actor to "revoke that via Django administration", and the test for that path comments that "is_staff is untouched: this flow only manages the superuser grant, not staff access set independently". But line 165 clears `is_staff` unconditionally on the actual revoke. Reproduced:

```
before: superuser=True  staff=True   (staff set independently, before any grant)
POST revoke -> 302
after : superuser=False staff=False  <- Django-admin access silently removed
```

So an account that had Django admin access on its own merits, was later granted superuser, and is later revoked, silently loses the admin access it had beforehand. Nothing in the confirmation page or the success message mentions it.

The existing test cannot catch this: `create_user` in `tests/conftest.py:100-108` never sets `is_staff`, so `superuser_b.is_staff` is already `False` and `assert superuser_member_b.user.is_staff is False` passes vacuously.

Either behaviour is defensible — what is not is having the two branches disagree. Pick one: leave `is_staff` alone on revoke (and say so in the success message), or clear it deliberately and fix the contradictory warning text and the test above.

#### 3. (Low) Self-revoke lands the actor on a 403

`get_success_url()` always redirects to `organization_member_list`, which requires `tools.view_organization`. A superuser who revokes their own access no longer has it:

```
POST revoke (self, another superuser exists) -> 302 -> /en/dvwa/members -> 403
```

This is the default case rather than an edge case: superuser org-memberships carry no group, since superuser status already covers everything. I checked every membership on the stack, including the one `setup_test_org` creates — all `groups=NONE`. So any superuser who demotes themselves through this flow lands on an error page.

Redirecting to the crisis room or landing page when `target_user.pk == request.user.pk` would fix it.

#### 4. (Low) Revoke link offered for the last active superuser

The template shows the revoke link for any superuser, so on a single-superuser installation the admin can click through to a destructive-styled confirmation page for an action the POST always refuses. The guard is correct; it just fires late. Hiding the link (or disabling it with the reason) when the target is the only active superuser would save the round trip.

Both findings above ship without coverage, and finding 2 specifically survived because its test asserts on a value that was already `False`. These two drop-ins pin them:

```python
def test_blocked_member_cannot_be_granted_superuser_access(rf, superuser_member, admin_member):
    admin_member.blocked = True
    admin_member.save(update_fields=["blocked"])
    request = setup_request(rf.post("organization_member_grant_superuser"), superuser_member.user)

    response = GrantSuperuserAccessView.as_view()(
        request, organization_code=admin_member.organization.code, pk=admin_member.id
    )

    assert response.status_code == 302
    admin_member.user.refresh_from_db()
    assert admin_member.user.is_superuser is False
    assert admin_member.user.is_staff is False


def test_revoke_preserves_independently_set_staff_access(
    rf, superuser_member, superuser_member_b, django_capture_on_commit_callbacks
):
    superuser_member_b.user.is_staff = True   # set independently, not by this flow
    superuser_member_b.user.save(update_fields=["is_staff"])
    request = setup_request(rf.post("organization_member_revoke_superuser"), superuser_member.user)

    with django_capture_on_commit_callbacks(execute=True):
        RevokeSuperuserAccessView.as_view()(
            request, organization_code=superuser_member_b.organization.code, pk=superuser_member_b.id
        )

    superuser_member_b.user.refresh_from_db()
    assert superuser_member_b.user.is_superuser is False
    assert superuser_member_b.user.is_staff is True
```

I ran both in the Rocky test container on this branch and they fail as intended:

```
FAILED test_blocked_member_cannot_be_granted_superuser_access - assert True is False
FAILED test_revoke_preserves_independently_set_staff_access  - assert False is True
2 failed, 10 warnings in 2.24s
```

The second encodes the "leave `is_staff` alone" option; if you go the other way, invert it to assert `is_staff is False` and give `superuser_member_b` staff access in the setup so it stops being vacuous.

### Test Results

| Component | Test | Result |
|-----------|------|--------|
| rocky | `make utest` (Docker) | ✅ 503 passed, 1 skipped (123s) |
| rocky | 34 tests in `test_members.py` | ✅ all pass |
| rocky | CI `test` 3.10–3.13 | ✅ green |
| rocky | proposed regression tests | ❌ 2 failed (as intended — they pin findings 1 and 2) |

### Functional Test

| Check | Status | Details |
|-------|--------|---------|
| Stack build | ✅ | 13/13 containers, no restart loops |
| Log analysis | ✅ | 5 tracebacks, all startup noise (rocky_worker scheduler race, bytes deque, RabbitMQ EOL notice) |
| **Non-superuser self-escalation** | ✅ 403 | GET and POST both 403, `is_superuser` unchanged |
| Anonymous access | ✅ 302 | redirected to `/en/login/?next=…` |
| Audit trail | ✅ | `event_code=900113 actor_email=superuser@localhost target_email=… previous_is_superuser=False is_superuser=True is_staff=True` |
| Browser — grant flow | ✅ | login → member edit → confirmation → success message → member list, all correct |
| Confirmation page a11y | ✅ | `role="alert"` live region, target email, admin link, cancel, destructive confirm |
| Grant on blocked member | ❌ | succeeds (finding 1) |
| Revoke with independent `is_staff` | ❌ | staff silently cleared (finding 2) |
| Self-revoke | ⚠️ | 302 → 403 (finding 3) |

I walked the grant flow in a real browser this round, which closes the process note I left last time about only having exercised it through Django's test client. The UI itself is good: the confirmation page is a proper live region, the warning is clear, cancel works, and the success message reads well.

### Verdict

**CHANGES REQUESTED**

All five round-1 findings are genuinely fixed, and the concurrency work is solid. The two new Mediums are what hold it: granting superuser to a blocked member produces a member the org admin can no longer unblock or edit from the members list, and revoke silently strips Django-admin access that the same view explicitly promises not to touch. Both are small, contained changes, and both come with a test above that is red today.

Findings 3 and 4 are polish you can fold into the same push or leave.

One note for the NIS2 logging story rather than a finding: successful grants and revokes log 900113/900114 with full actor/target context, but a denied attempt raises a bare `PermissionDenied` and logs nothing. On a privilege-escalation endpoint, failed attempts are usually the ones worth alerting on. Worth considering separately — not something to hold this PR for.

brennodewinter:
Round-2 findings status — branch merged with current `main` (was CONFLICTING):

| # | Finding | Status |
|---|---------|--------|
| 1 | Grant has no `blocked` guard | ✅ `target_member.blocked` guard in `GrantSuperuserAccessView.post()`; edit page hides the grant link for blocked members; regression test `test_blocked_member_cannot_be_granted_superuser_access` |
| 2 | Revoke clears `is_staff` unconditionally | ✅ Revoke now only writes `is_superuser`; independently-set staff access survives. Also folded in the parenthetical: the success message now says the account retains Django administration access when it does (`44e209a`) |
| 3 | Self-revoke lands on a 403 | ✅ Self-revoke redirects to the crisis room |
| 4 | Revoke link shown for last active superuser | ✅ Link hidden via `can_revoke_superuser` context flag |

Verification:
- `pre-commit run` on all changed files — ✅ green (ruff, ruff-format, mypy, vulture, codespell, djLint, prettier, rstcheck)
- `pytest tests/test_members.py` (Docker, py3.13) — ✅ 41 passed
- `django.pot` regenerated via `makemessages` in the rocky container after the merge — only the PR's new msgids added

@hasecon — ready for re-review.

PR #5284, opened 2026-08-05
Fix scan level inheritance chain lookup, also pre-filter eligble neighbours based on known model/path inheritance

### LINKED ISSUE #5283: Scan level inheritance sometimes showed an incorrect inheritance path
**Describe the bug**
Sometimes an objects scan level inheritance chain showed a path going *down* instead of going up.

**To Reproduce**

Add both innerheight.com and underdark.nl hostnames and declare l2.
wait for scans to complete.
Navigate to the object: css.underdark.nl AAAA 2001:1af8:2100:b080:1::80 or DNSAAAARecord%257Cinternet%257Ccss.underdark.nl%257C2001%253A1af8%253A2100%253Ab080%253A1%253A%253A80
See the clearance level path go from 2, down to 0, back to two .

**Expected behavior**
Show an actual logical chain.

**OpenKAT version**
1.22

**Additional context**
The scan level itself was correctly inherited, the chain was just collected incorrectly, as there might be many chains

underdarknl:
### Changes

This fixes a few bugs in the scan level chain, and optimized the code by only fetching neighbours that can (as stated by the model) give us a scan level equal or higher than we currently have, and as such can be a path towards the declaration source.

### Issue link

Closes #5283 

### Demo

image

### QA notes
See the issue for reproducing steps, however, we might need unit tests to cover this.

### `Website.ip_service` `max_inherit_scan_level` 4 → 1 — rationale

This is a conscious scan-coverage change, not a display fix. The old L4 meant a Website inherited full clearance from its `IPService`, so a Website on a high-clearance IP but low-clearance hostname would reach L4 purely through the IP path. Lowering this to L1 means co-hosted websites on a given IP are still included (L1), but the IP path no longer carries high clearance to the Website.

In the common case (hostname declared high) the Website still reaches L4 via the unchanged `hostname` path (`max_inherit_scan_level=4`), so most sites are unaffected. The edge case (IP high, hostname low) now gets L1 instead of L4 — more conservative, and consistent with the principle that a Website's clearance should follow its hostname, not whatever else happens to share the IP.

### Follow-up items (non-blocking, tracked for future work)

1. **`_path_can_inherit_level` naming** — the leading underscore says "private" but it's called from `service.py` outside the `Path` class. Drop the underscore or inline the check.
2. **`get_max_scan_level_issuance` docstring** — still references caching behaviour while its sibling `get_max_scan_level_inheritance` was updated to describe semantics. Minor inconsistency.
3. **Unit test coverage** — the test pins the duplicate-neighbour regression (round-1 400) but doesn't exercise the `reversed()`/highest-edge-wins bug, which would need a two-different-level-edges topology.

---

### Code Checklist

---

---

## Checklist for QA:


hasecon (changes requested):
## QA Review

### Setup
- Branch: `fix/octopoes-scanlevel-chain` @ `338d102`
- Stack: `make clean && make kat` — ✅ 13/13 services up
- Test org: `setup_test_org --code dvwa` + `dvwa.cloud` / `juice.dvwa.cloud` for a DNS-zone topology — ✅

### CI/CD Status
**All 38 checks green**, including `rtest`, `Tests (3.10–3.13)`, `pre-commit` and `CodeQL`. Nothing red.

That is worth pausing on, because the blocker below makes the inheritance endpoint return HTTP 400 for an ordinary DNS zone. Green CI here means the code path has no coverage at all, not that it works.

### The fixes in this PR are correct — I verified each one

Before the bad news, the good: the diagnosis in #5283 is right and the fixes are real.

**The `reversed()` removal is the core fix and it is subtle.** The old dict comprehension was named `highest_inheritance_per_neighbour` but kept the **lowest** level per neighbour, because iterating a descending list means the last write per key wins:

```
sorted ascending: [('A',1), ('B',2), ('A',3)]
old (reversed):   {'A': 1, 'B': 2}   <- lowest
new:              {'A': 3, 'B': 2}   <- highest
```

That is exactly the "level 2, down to 0, back to 2" symptom from the issue.

**`(field.json_schema_extra or {})`** correctly guards against fields without `json_schema_extra`, where the old `.get()` would have raised `AttributeError` on `None`.

**Filtering on `inherited_level < required_level`** instead of the neighbour's raw level is more correct: the old code filtered on the neighbour's own level but then appended the possibly-lower `min(...)` value, so a step that cannot carry the required level could still enter the chain.

**Pre-filtering paths before querying** is a sound optimization, and `get_neighbours(..., paths=...)` already exists in the repository, so no interface change is needed.

### Code Review Findings

| 1 | **Critical** | `core/service.py` (dedupe before `_populate_scan_profiles`) | Same neighbour reached via two paths raises `ValueError: neighbour scan_profile is None`. **Reproduced against real data** — see below. |
| 2 | Medium | — | No tests. `get_scan_profile_inheritance` and `InheritanceSection` have **zero** references anywhere in `octopoes/tests/`, which is why a 400-on-ordinary-data change passes CI. |
| 3 | Nit | `models/path.py` | `_path_can_inherit_level` is underscore-private but called from `core/service.py`. Either drop the underscore or make it a module-level helper next to `get_max_scan_level_inheritance`. |

### Finding 1 in detail

`get_neighbours` deserializes **a fresh object per path**:

```python
ret[path] = [self.deserialize(serialized) for serialized in value]
```

so the same OOI reachable via two paths yields two distinct Python objects with equal references. The new code then dedupes by reference before populating:

```python
neighbours_by_reference = {neighbour.reference: neighbour for _, neighbour, _ in eligible_edges}
self._populate_scan_profiles(list(neighbours_by_reference.values()), valid_time)
```

`_populate_scan_profiles` mutates only the objects handed to it, so the *other* object keeps `scan_profile is None` and the loop below raises. The old code passed every neighbour object, which is why it never hit this.

**Reproduced end-to-end on the running stack**, with nothing exotic in the data — plain `setup_test_org` output:

```
GET /dvwa/scan_profiles/inheritance?reference=DNSZone|internet|dvwa.cloud

this PR : 400  {"detail":"neighbour scan_profile is None"}
main    : 200  [{"reference":"DNSZone|internet|dvwa.cloud","level":1,"scan_profile_type":"inherited"},
                {"reference":"Hostname|internet|dvwa.cloud","level":1,"segment":"<dns_zone[is Hostname]",
                 "scan_profile_type":"declared"}]
```

Same request, same data, only `core/service.py` swapped. Also confirmed in isolation: feeding the real method two paths whose lists hold distinct objects with one reference logs `Populating scan profiles for 1 oois` and raises, where main logs `2 oois` and returns a chain.

This is not a rare shape. Twelve model combinations have two *inheritable* segments to the same neighbour type, and two of them are everyday structures:

```
Hostname    <- DNSZone        [('hostname', 2), ('dns_zone', 2)]     triggers at required_level <= 2
IPAddressV4 <- IPV4NetBlock   [('start_ip', 4), ('netblock', 4)]     triggers at required_level <= 4
```

The first is simply a zone apex — every zone in the graph has one, which is precisely what `dvwa.cloud` hit above.

**Suggested fix**, keeping the dedupe benefit of one profile fetch per reference — canonicalise the edges onto the populated objects:

```python
neighbours_by_reference = {neighbour.reference: neighbour for _, neighbour, _ in eligible_edges}
self._populate_scan_profiles(list(neighbours_by_reference.values()), valid_time)
eligible_edges = [
    (segment, neighbours_by_reference[neighbour.reference], level)
    for segment, neighbour, level in eligible_edges
]
```

### Finding 2 — suggested test scenarios

Since the recursion is subtle and now demonstrably fragile, unit tests on `OctopoesService.get_scan_profile_inheritance` with a stubbed `ooi_repository` would be cheap and high value:

1. **Duplicate neighbour across two paths** — the regression above; asserts no exception and one profile fetch per reference.
2. **Highest edge wins** — two edges to the same neighbour at levels 1 and 3; the chain must continue via 3. This is the `reversed()` bug and would have caught it.
3. **Chain never steps down** — assert levels are non-decreasing toward the declaration, which is the user-visible complaint in #5283.
4. **Declared short-circuit** — a declared neighbour ends the chain immediately at the highest declared level.
5. **No eligible paths** — returns the chain unchanged rather than recursing.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| octopoes | CI `rtest`, `Tests (3.10–3.13)` | ✅ green (no coverage of this path) |
| octopoes | inheritance endpoint, 20 inherited OOIs | ❌ **19 OK, 1 × HTTP 400** (`DNSZone\|internet\|dvwa.cloud`) |
| octopoes | same OOI on `main` | ✅ 200 with a correct 2-step chain |
| any | `pre-commit`, `CodeQL` | ✅ green |

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Stack build | ✅ | 13/13 services up |
| Inheritance for most OOIs | ✅ | 19 of 20 return a chain |
| Inheritance for a zone apex | ❌ | 400 `neighbour scan_profile is None` |
| Regression vs main | ❌ | main returns 200 for the same reference |

### Verdict

**CHANGES REQUESTED**

The three bugs this PR set out to fix are genuinely fixed, and the `reversed()` one is a nice catch that explains the reported symptom precisely. The blocker is the new dedupe: it turns a working request into a 400 for any OOI whose neighbour is reachable by two inheritable paths, which includes every DNS zone apex. The four-line canonicalisation above should settle it.

Given how easily a subtle change here breaks a path with no test coverage, I would really like scenarios 1–3 to land with the fix rather than as a follow-up.

underdarknl:
Im puzzled why the unit-test is failing, more investigating needed.

hasecon (changes requested):
## QA Review — round 2

### Setup
- Branch: `fix/octopoes-scanlevel-chain` @ `e872423`
- Stack: `make clean && make kat` — ✅ 13/13 services up
- Test org: `setup_test_org --code dvwa` + `dvwa.cloud`/`juice`/`apache` DNS topology — ✅

### Round-1 findings — status (verified against this head, not comments)

| 1 | **Critical** — dedupe raised `neighbour scan_profile is None` → HTTP 400 on every zone apex | ✅ **Fixed & verified live.** The canonicalisation (`neighbours_by_reference` + remapped `eligible_edges`) is in place. `GET /dvwa/scan_profiles/inheritance?reference=DNSZone\|internet\|dvwa.cloud` now returns **200** (was 400); a sweep of all 50 OOIs returns **0 non-200**. Chains trace correctly to a DECLARED source (e.g. `IP → ResolvedHostname → Hostname(L4 declared)`). |
| 2 | Medium — no unit tests for `get_scan_profile_inheritance` | ⚠️ **Still open.** No unit tests added; only the robot test was touched — and it is now the blocker (below). |
| 3 | Nit — `_path_can_inherit_level` visibility | ⚠️ Evolved into **dead code** — the method is defined in `path.py:160` but never called; `service.py:573-580` inlines the same check. |

### CI/CD status

| Check | Result | Blocking? |
|-------|--------|-----------|
| `rtest` (octopoes robot) | ❌ **failure** | **Yes (Rule 1)** — diagnosed below |
| `test` (octopoes unit+integration, 3.10–3.13) | ✅ | — |
| `pre-commit`, `CodeQL`, all `build`/container | ✅ | — |
| `makelang` | ❌ failure | No — this PR changes no translatable strings (no `.po`/`.html`), pre-existing drift |

Per Rule 1 (and the #5088 lesson), a red `rtest` on an octopoes PR is an automatic blocker. But I investigated it specifically rather than hand-waving, and the news is good — see below.

### 🔎 `rtest` root cause (this is the answer to "puzzled why the unit-test is failing")

The failing test is `06_scan_profile_inheritance :: Simple Scan Profile Inheritance`. **It is not an algorithm bug — it's a propagation-timing bug in the test.** I reproduced the exact fixture on the running stack (same `normalizer_output.json` observation, declare L4 on `Hostname|internet|example.com`, then poll the inheritance for `IPAddressV4|internet|1.1.1.1`):

```
t≈1s … t≈12s : length=1 | IPAddressV4(L0)        <- IP has not received the level yet
t≈20s        : length=3 | IP(L4) → ResolvedHostname(L4) → Hostname(L4)   ✅ exactly what the test asserts
```

Why it regressed even though the topology is unchanged: the **old** algorithm traced *potential* inheritance and returned the length-3 chain immediately, independent of propagation. The **new** stricter algorithm (`inherited_level == required_level`) only explains the target's *actual current* level, so it correctly returns `length=1` until the declared L4 has propagated `Hostname → ResolvedHostname → IP` through the celery worker — which takes ~15–20s here. The `Sleep 1s` added in this PR is ~20× too short, so `Length Should Be 3` fires while the IP is still L0.

The `.ci` rtest env *does* run the propagation worker (`ci_octopoes_api_worker`), so propagation happens — the test just doesn't wait for it.

**Suggested fix (pick one):**
- **Preferred — deterministic, race-immune:** replace `Sleep 1s` with a poll:
  ```robotframework
  Wait Until Keyword Succeeds    30s    1s
  ...    Scan Profile Inheritance Length Should Be    ${REF_IPADDR}    3
  ```
- **Simpler, codebase-consistent:** replace `Sleep 1s` with `Await Sync` (as `01_scan_profiles.robot` does after every declaration). Note: if scan-profile propagation is a multi-step cascade, `Await Tasks Done` can return *between* steps, so the `Wait Until Keyword Succeeds` form is the safer of the two.

Either way this also resolves the fragile-`Sleep` concern and would make the test actually exercise the fixed path.

### New code-review findings (round 2)

| A | **Medium** | `rocky/views/ooi_detail.py:168` + `models/explanation.py` | The Rocky half is only wired halfway. `InheritanceSection.level` changed meaning (now the neighbour's **own** level, was the **transferred** level) and the new `inherited_level` is added to context as `given` — but `partials/explanations.html` still renders only `L{{ section.level }}` and never uses `given`. On a capped edge this **misreports clearance**: verified live, the `DNSZone\|internet\|dvwa.cloud` row shows `level=4 given=1` (only L1 was actually transferred), so the UI would display "inherited … L4" for an L1 inheritance. Either render `given` in the template or split the Rocky change out. |
| B | Low | `models/path.py:215` | `get_max_scan_level_issuance` still uses the unguarded `.json_schema_extra.get(...)` that its dual `get_max_scan_level_inheritance` was just hardened against with `(field.json_schema_extra or {})`. Same latent `AttributeError` left in the sibling. |
| C | Note | `models/ooi/web.py` | `Website.ip_service` `max_inherit_scan_level` **4 → 1** is a real behavioural change (co-hosted sites on a shared IP now inherit only L1). Looks intentional and sensible, but it's beyond "fix the chain lookup" — worth calling out in the PR description and ideally a test. |
| D | Nit | `core/service.py:620` | Redundant `neighbour = neighbours_by_reference[neighbour.reference]` inside the loop — `eligible_edges` was already remapped to canonical instances at 611-614. |
| E | Nit | `core/service.py:622` | `neighbour.scan_profile is None` check is now unreachable (`_populate_scan_profiles` assigns `EmptyScanProfile` to all) — harmless dead defence. |

### Test results

| Component | Test | Result |
|-----------|------|--------|
| octopoes | CI `test` (unit+integration 3.10–3.13) | ✅ green |
| octopoes | CI `rtest` | ❌ red — test-timing (`Sleep 1s`), diagnosed above |
| octopoes | inheritance endpoint, all 50 dvwa OOIs | ✅ 50/50 HTTP 200 (round-1 400 gone) |
| octopoes | live chain correctness (IP / zone apex / HTTP URL) | ✅ correct, non-decreasing, ends DECLARED |

*(An in-container `pytest tests/test_api.py::test_get_scan_profiles` 500 is an environment artifact — the test mocks XTDB at `http://xtdb:3000` but the live container talks to `http://crux:3000`; unrelated to this PR, and the CI `test` job is green.)*

### Functional test

| Check | Status | Details |
|-------|--------|---------|
| Stack build | ✅ | 13/13 up |
| Round-1 400 regression | ✅ Fixed | zone apex 200; 0/50 non-200 |
| Chain correctness | ✅ | verified live incl. capped-edge (`given`) values |
| Logs | ✅ | no `scan_profile is None`; only unrelated VPN-off normalizer `JSONDecodeError` + RabbitMQ reconnect noise |

### Verdict

**CHANGES REQUESTED**

The substance is in good shape: the round-1 critical is genuinely fixed and verified live, the algorithm produces correct chains, and the `reversed()`/dedupe rework is sound. The blocker is not the code logic — it's that `rtest` is red because `06_scan_profile_inheritance` waits `Sleep 1s` for a propagation that takes ~15–20s. Swap that for `Await Sync` / `Wait Until Keyword Succeeds` and the suite should go green. I'd also like finding A (the half-wired `given`/`level` display) fixed before merge, since it ships a misleading clearance number, and the round-1 request for at least one unit test on the duplicate-neighbour path still stands.

hasecon (changes requested):
## QA Review — round 3 (head `ec0a936`)

Good progress — the round-2 blocker is gone. One user-facing item and the standing test gap keep it from a green light.

### Round-2 findings — status (verified against head, not comments)

| # | Round-2 finding | Status |
| rtest | **Blocker** — `06_scan_profile_inheritance` asserted before propagation (`Sleep 1s` ≪ ~15–20s) | ✅ **Fixed & green.** You replaced the sleep with `Recalculate Scan Profile Propagation` → `GET /scan_profiles/recalculate?sync=true`, which runs `recalculate_scan_profiles()` then `commit(sync=True)` — it blocks until propagation is committed, so the length-3 assertion is now deterministic. `rtest` is **success** on this head (job 94125368843). Cleaner than the poll I suggested. |
| A | **Medium** — `InheritanceSection.level` now = neighbour's *own* level; `inherited_level`/`given` added to the Rocky context but the template never renders it | ❌ **Still open.** `partials/explanations.html` is byte-identical to `main` — it renders `L{{ section.level }}` and ignores `given`. So on a **capped edge** the clearance-inheritance table shows the neighbour's own level, not what was transferred. Live evidence from round 2: the `DNSZone\|internet\|dvwa.cloud` row is `level=4 given=1` → the table displays "L4 inherited from ↓" for an inheritance that only carried L1. `given` is wired into the context (`ooi_detail.py:168`) but unused — clearly half-finished. |
| — | Medium — no unit tests on `get_scan_profile_inheritance` | ❌ **Still open.** Only the robot test was touched; `octopoes/tests/` still has zero references to this function — on a recursion that has now regressed twice, both times caught only by live testing, never by CI. |
| B | Low — `get_max_scan_level_issuance` still uses unguarded `.json_schema_extra.get(...)` | ❌ Still open (its sibling `get_max_scan_level_inheritance` was hardened; this one wasn't). |
| C | Note — `Website.ip_service` `max_inherit 4→1` | ✅ Now documented in-code with a comment. |
| D | Nit — redundant `neighbour = neighbours_by_reference[neighbour.reference]` in the loop (edges already remapped) | ❌ Still present (harmless). |
| E | Nit — `neighbour.scan_profile is None` check now unreachable | ❌ Still present (harmless). |

### CI/CD Status
**All 39 checks green** — `rtest`, `test (3.10–3.13)`, `Tests (3.10–3.13)`, `pre-commit`, `CodeQL`, `makelang` (green again), all builds. `mergeable_state` is **behind** main, so it needs a branch update before merge.

### Verified this round
- `rtest` is genuinely green on `ec0a936` (the actual CI result on this head, not a claim), and the sync-recalculate mechanism is sound.
- `core/service.py` is **byte-identical to `e872423`** — the head I verified live in round 2 (all 50 dvwa OOIs → HTTP 200, chains correct and ending DECLARED). So the algorithm is unchanged and already validated; I did not re-run `make kat` for it. Finding A I re-checked statically (`git diff main…head -- explanations.html` is empty).

### What's left before merge
1. **Finding A — render `given`.** It's a one-line template change and the value is already in context. Something like:
   ```django
   L{{ section.given|default:section.level }}
   {% if section.given and section.given != section.level %}<span class="muted">(source L{{ section.level }})</span>{% endif %}
   ```
   or whatever phrasing you prefer — the point is the table should show the *transferred* level, not the source's own level, so it stops overstating clearance on capped edges. (If showing the source's own level is actually intended, then dropping `given` from the context would at least make it consistent — but wiring it in and not using it reads as unfinished.)
2. **One or two unit tests** on `get_scan_profile_inheritance` (stubbed `ooi_repository`): the duplicate-neighbour-across-two-paths case (round-1 regression) and highest-edge-wins (the `reversed()` bug). Both are cheap and would have caught the two regressions this PR went through. Not a hard blocker on their own, but on a function this fragile I'd really like them to land here.

### Verdict
**CHANGES REQUESTED** — but a light one. The hard part is done: the algorithm is correct and verified, and `rtest` (the round-2 blocker) is fixed properly. What remains is wiring the `given` value you already pass into the template so the clearance table stops showing a misleading number on capped edges, plus ideally a unit test or two. Happy to re-review immediately once that's in; please also update the branch with main before merging.

hasecon:
Concrete proposal for the two remaining items, to save you a round.

## 1. Finding A — render the `given` value in the inheritance table

The value is already in the context (`ooi_detail.py:168` sets `"given": section.inherited_level`); `partials/explanations.html` just doesn't use it, so on a capped edge the cell shows the neighbour's **own** level instead of what was transferred. The recursion sets:

- `section.level` = the neighbour's **own** `scan_profile.level`
- `section.given` (= `inherited_level`) = the level actually **transferred** along that edge

On an uncapped edge those are equal; on a capped one (`Website.ip_service` L1, a DNSZone apex, …) `level > given`, and today the table prints `level`, overstating the inherited clearance.

Minimal template change that shows both, so nothing is hidden and nothing is overstated (`rocky/rocky/templates/partials/explanations.html`, the "Clearance level" cell):

```django
<td>
    L{{ section.level }}
    {% if section.given is not None and section.given != section.level %}
        (↓ L{{ section.given }})
    {% endif %}
    {% if forloop.last %}
        {% translate "declared" %}
    {% else %}
        {% translate "inherited from ↓" %}
    {% endif %}
</td>
```

A capped edge then reads `L4 (↓ L1) inherited from ↓` — the source's own clearance stays visible, but the number that actually flowed down is explicit. Wording/styling is yours to polish; the point is just to surface `given`. (If you'd rather show only the source's own level, then dropping `given` from the `ooi_detail.py` context would at least make it consistent — but since you wired it in, showing it seems to be the intent.)

## 2. A unit test for `get_scan_profile_inheritance`

The scaffolding already exists in `octopoes/tests/conftest.py`: the `network` / `hostname` / `ipaddressv4` / `resolved_hostname` fixtures build a real graph against `ooi_repository` + `scan_profile_repository`, and there's a service fixture over the same repos. An integration-style test there is the most faithful (it exercises the real `get_neighbours` per-path deserialization that produced both regressions):

```python
def test_scan_profile_inheritance_duplicate_neighbour_paths(octopoes_service, ...fixtures..., valid_time):
    # Declare L4 on the hostname; the zone apex reaches the same neighbour via two
    # inheritable paths (hostname <- DNSZone), which is the round-1 400 regression.
    declare_scan_profile(REF_HOSTNAME, 4)
    recalculate(sync=True)

    chain = octopoes_service.get_scan_profile_inheritance(REF_DNSZONE, valid_time, [start_section])

    assert chain[-1].scan_profile_type == ScanProfileType.DECLARED   # resolves, no ValueError
    # highest-edge-wins (the reversed() bug): levels toward the source are non-decreasing
    assert [s.level for s in chain] == sorted(s.level for s in chain)
```

Two cases worth pinning: **duplicate neighbour across two paths** (round-1 400) and **highest edge wins** (the `reversed()` bug) — both regressed here and neither was caught by CI. Happy to write these up properly and verify them run, as a small stacked PR onto this branch if that's easier for you — just say the word.

Everything else on the PR is in good shape: the algorithm is correct (verified live, all 50 dvwa OOIs → 200) and `rtest` is green with the sync-recalculate fix.

hasecon (dismissed):
## QA Review — round 4 (head `d5d2935`)

So close. Both round-3 items were picked up — the template fix is correct and complete, and the unit test you added is *exactly* the right test — but the test was committed without the mock-repository support it needs, so it can't run: it dies on `NotImplementedError` before reaching any assertion, and that is why **all four octopoes `test (3.10–3.13)` CI jobs are red on this head**. One small conftest addition (drop-in below, verified) turns everything green.

### Round-3 findings — status (verified against this head, not comments)

| # | Round-3 finding | Status |
| A | **Medium** — `given` wired into context but template never renders it → capped edges overstate clearance | ✅ **Fixed.** `partials/explanations.html` now renders `(↓ L{{ section.given }})` when `given` differs from `level`. Conditions handle the start row (`given=None`) and uncapped edges correctly, and both consumers of the partial (`OOIDetailView`, `ScanProfileDetailView` via inheritance) provide `given` — verified by tracing the context, per the both-views check. |
| — | Medium — no unit tests on `get_scan_profile_inheritance` | ⚠️ **Attempted, currently broken — now the blocker.** See below. |
| B | Low — `get_max_scan_level_issuance` still unguarded (`.json_schema_extra.get` on possibly-`None`) | ❌ Still open. |
| D/E | Nits — redundant remap line; unreachable `scan_profile is None` check | ❌ Still present (harmless). |

### The blocker: the new test cannot run

`tests/test_scan_profile_inheritance_path.py` calls `octopoes_service.get_scan_profile_inheritance` against the conftest mocks, but two methods it needs were never implemented on them:

1. `MockScanProfileRepository.get_bulk` → first crash (`scan_profile_repository.py:40`, via `_populate_scan_profiles`)
2. `MockOOIRepository.get_neighbours` → second crash (`ooi_repository.py:93`)

Reproduced in the octopoes container on this head: `1 failed in 0.74s`, `NotImplementedError`. This is exactly what CI's red `test (3.10)`–`test (3.13)` are.

### Drop-in fix (verified end-to-end locally)

```python
# conftest.py — MockScanProfileRepository
    def get_bulk(self, references: set[Reference], valid_time: datetime) -> list[ScanProfileBase]:
        return [profile for reference, profile in self.profiles.items() if reference in references]

# conftest.py — MockOOIRepository
    def get_neighbours(
        self, reference: Reference, valid_time: datetime, paths: set[Path] | None = None
    ) -> dict[Path, list[OOI]]:
        from octopoes.models.path import get_paths_to_neighbours

        if paths is None:
            paths = get_paths_to_neighbours(reference.class_type)

        ret: dict[Path, list[OOI]] = {}
        for path in paths:
            neighbours = self.list_neighbours({reference}, {path}, valid_time)
            if neighbours:
                # The real XTDB repository deserializes a fresh object per path, without
                # a scan profile; mimic that so per-path duplicate instances are faithful.
                copies = [neighbour.model_copy(deep=True) for neighbour in neighbours]
                for copy in copies:
                    copy.scan_profile = None
                ret[path] = copies
        return ret
```

**The per-path `model_copy` + `scan_profile = None` part is load-bearing, not cosmetic.** The real `XTDBOOIRepository.get_neighbours` deserializes a *fresh* object per path with no scan profile — that duplicate-instance behaviour is precisely what caused the round-1 400 regression. I verified the difference empirically:

| `core/service.py` version | naive stub (shared instances) | faithful stub (above) |
|---|---|---|
| this head `d5d2935` | ✅ passes | ✅ passes |
| `338d102` (round-1 dedupe bug) | ✅ passes (!) | ❌ `ValueError: neighbour scan_profile is None` |

With the naive stub the test silently pins nothing; with the faithful stub it fails on exactly the regression this PR went through and passes on the fix. (It doesn't exercise the `reversed()` bug — that would need a two-different-level-edges topology — but pinning the dedupe class is the important half.)

### Verified this round (round-4 explicit list)
- `core/service.py` is **byte-identical to `e872423`**, the head verified live in round 2 (50/50 OOIs → 200, chains correct) — algorithm unchanged, so no stack re-test needed.
- CI red cause reproduced locally in Docker (`NotImplementedError`), fix applied locally → `1 passed`; revert matrix above run against `d5d2935`, `main`, and `338d102`.
- Template fix checked against both consuming views; `rtest` green on this head; `makelang` green; boefjes `Tests (3.10–3.13)` green (the merge-in of current main rides along cleanly).
- Working tree restored to the PR head after testing — no changes pushed from my side.

### CI/CD Status
| Check | Result | Blocking? |
|-------|--------|-----------|
| octopoes `test (3.10–3.13)` | ❌ failure ×4 | **Yes (Rule 1)** — caused by this PR's test, diagnosis above |
| `rtest` | ✅ | — |
| `Tests (3.10–3.13)` (boefjes), `pre-commit`, `CodeQL`, `makelang`, builds | ✅ | — |

### Verdict
**CHANGES REQUESTED** — one commit away from approval: add the two mock methods (drop-in above, please keep the fresh-copy semantics). Everything else is done: the algorithm has been verified live since round 2, the `given` display is fixed, `rtest` is green, and once the conftest lands, the unit test genuinely pins the duplicate-neighbour regression. Finding B and the two nits remain optional cleanups. Happy to re-review the moment it's pushed.

hasecon:
Picked this up as discussed — pushed `0b47264` completing the mock support that the round-4 review identified, building on Jan's `get_bulk` from `7d910d6`:

- **`MockOOIRepository.get_neighbours`** — returns per-path fresh copies *without* a scan profile, matching how `XTDBOOIRepository` deserializes. That per-path-instance behaviour is load-bearing: it's what makes `test_scan_profile_inheritance_duplicate_neighbour_paths` actually pin the duplicate-neighbour regression instead of silently passing on the broken code.
- Also folded in the small round-4 "B" cleanup: `get_max_scan_level_issuance` now guards `json_schema_extra` the same way its inheritance sibling does.

Verified in the octopoes container before pushing:
- the chain test **passes at this head** (`1 passed`);
- it **still fails with the exact `ValueError`** when `core/service.py` is swapped to the pre-canonicalization code (`338d102`) — so the regression is genuinely pinned;
- full unit suite: **157 passed**; the remaining `tests/test_api.py` failures are the known local XTDB-env mock artifact (identical on the unmodified head, green in CI);
- `pre-commit` clean on the changed files.

On the user-test-vs-regression-test point: agreed the robot suite leans functional — the important thing is that `get_scan_profile_inheritance` now has *any* in-suite coverage, whatever we call it. 

Since hasecon now has commits on this branch, the approval is @brennodewinter's to give (Jan's earlier commits + this one). CI on `0b47264` should turn the four `test (3.x)` jobs green — I'll confirm once the run completes.

hasecon:
## Self-QA on the final head (`0b47264`) — full collin pass for @brennodewinter's review

Since we now have commits on this branch, here is the complete QA evidence on the final state (posted as a comment, not a review — the approval is Brenno's).

### Functional test (fresh `make clean && make kat` from this head)
| Check | Result |
|-------|--------|
| Stack | ✅ 13/13 services up, Rocky 200 |
| Inheritance endpoint sweep | ✅ **68/68 OOIs → HTTP 200** (dvwa org incl. `dvwa.cloud`/`juice`/`apache` DNS topology; the round-1 zone-apex 400 class stays gone on real data) |
| Chain correctness | ✅ zone apex: `DNSZone (L1, inherited)` ← `Hostname (level 4, inherited_level 1, declared)` — capped edge resolved correctly |
| **Live template render** (the round-3 finding A) | ✅ `/objects/scan-profile/?ooi_id=DNSZone\|internet\|dvwa.cloud&show_clearance_level_inheritance=True` renders the origin row as **`L4 (↓ L1) declared`** — the transferred level is now explicit; uncapped rows and the "This OOI" row (no `given`) render unchanged |
| Logs | ✅ octopoes api/worker/scanprofiles: 0 error lines; only live-scan noise in boefje workers |

### Tests on this head (Docker)
| Suite | Result |
|-------|--------|
| `tests/test_scan_profile_inheritance_path.py` | ✅ 1 passed — and still fails with the exact `ValueError` when `core/service.py` is swapped to the pre-canonicalization code (`338d102`), so the regression stays pinned |
| Octopoes unit suite | ✅ 157 passed (the `tests/test_api.py` failures are the known local XTDB-env mock artifact — identical on the head before our commit, green in CI) |
| Octopoes `make itest` | ✅ **37 passed, 2 xfailed** |

### CI on `0b47264`
37 green, 1 red: **`test (3.11)`**, failing in the *Run integration tests* step. Same code is green on 3.10/3.12/3.13, and the same integration suite passes locally (37 passed above) — this looks like a CI flake, not a code signal. It needs a **re-run of the failed job** (I can't trigger that from here); if it stays red after a re-run, treat it as real and ping me.

### Status of all findings across rounds 1–4
Everything is resolved on this head: round-1 dedupe-400 (fixed r2, re-verified live here), round-2 rtest timing (fixed r3, rtest green), round-3 finding A `given` display (fixed r4, **live-verified above**), round-4 missing mocks (Jan's `get_bulk` + our `get_neighbours`), and finding B (`get_max_scan_level_issuance` guard, in our commit — note its key mapping is intentionally mirrored vs the inheritance getter). Remaining nits D/E are harmless style points.

From our side this is ready: one CI re-run of `test (3.11)` away from green.

underdarknl:
All tests are green on that last commit now. Needs approval from @brennodewinter

brennodewinter (dismissed):
Approved. Algorithm is correct, well-tested, and all CI is green. Two items to address (neither a blocker for merge, but please follow up):

## 1. `Website.ip_service` `max_inherit_scan_level` 4 → 1 — please confirm intent

This is a real scan-coverage change, not a display fix. A Website on a high-clearance IP but low-clearance hostname now inherits only L1 instead of L4. Issue #5283 explicitly states *"The scan level itself was correctly inherited, the chain was just collected incorrectly"* — so this changes the thing the issue said was already correct.

In the common case (hostname declared high) the Website still reaches L4 via the unchanged `hostname` path, so most sites are unaffected. But the edge case is a coverage reduction, it has no test, and the in-code comment describes the L1 semantics without justifying *why* the old L4 was wrong.

Please add one line of rationale (which case this corrects) in the PR description or a code comment, or split it into its own PR. Not blocking given the prior sign-offs, but it should be a conscious decision.

## 2. Dead code: `_path_can_inherit_level`

`octopoes/octopoes/models/path.py:160` defines `_path_can_inherit_level` but it is never called — `service.py` inlines the same check directly. Flagged since round 1/2. Please delete it (deletion over addition).

## Nits (optional)

- `core/service.py:620` — redundant `neighbour = neighbours_by_reference[neighbour.reference]` inside the loop; `eligible_edges` was already remapped to canonical instances at 612.
- `core/service.py:622` — `neighbour.scan_profile is None` check is now unreachable (`_populate_scan_profiles` assigns `EmptyScanProfile` to all).

underdarknl:
All should be adressed

brennodewinter (approved):
## Re-review — round 5 (head `364e4a7`)

Re-approving after the "Fix per review" commit (`9b44b20`) and the latest main merge. My previous approval on `7ecac39` was dismissed due to the new commit; this review covers the delta.

### Delta since last approval (`7ecac39` → `364e4a7`)

| Previous finding | Status |
|---|---|
| Nit D — redundant `neighbour = neighbours_by_reference[...]` remap in loop | ✅ **Fixed.** Removed in `9b44b20`. |
| Nit E — unreachable `neighbour.scan_profile is None` check | ✅ **Fixed.** Removed in `9b44b20`. |
| Finding #2 — dead code `_path_can_inherit_level` | ✅ **Addressed.** Instead of deleting, Jan wired it in to replace the inline eligibility check. DRY — the logic in `service.py` and `path.py` is now single-sourced. Valid choice. |

### CI/CD Status
All checks green on this head: `test (3.10–3.13)` ×2 (Octopoes + Rocky), `Tests (3.10–3.13)` (Boefjes), `rtest`, `pre-commit`, `makelang`, `CodeQL`, all container images. Remaining `build (debian/ubuntu)` jobs are the slow package builds, unrelated to the code changes.

### Algorithm verification (static, against current head)

The core logic in `get_scan_profile_inheritance` is unchanged from `e872423` (the head hasecon verified live in round 2 — 50/50 dvwa OOIs → HTTP 200, correct chains ending DECLARED). The `9b44b20` commit only touched the eligibility pre-filter (replaced inline check with `_path_can_inherit_level` call) and removed two dead lines. No behavioural change.

### Items still open (non-blocking, follow-up)

1. **`_path_can_inherit_level` naming** — the leading underscore says "private" but it's called from `service.py` outside the `Path` class. Drop the underscore to make it `path_can_inherit_level`, or inline the two-line check back. Minor.

2. **`Website.ip_service` `max_inherit_scan_level` 4→1** — still a conscious scan-coverage change, not a display fix. The comment ("this means any co-hosted websites on a given IP are included") describes *what* but not *why* the old L4 was wrong. In the common case (hostname declared high) the Website still reaches L4 via the `hostname` path, so most sites are unaffected. The edge case (IP high, hostname low) now gets L1 instead of L4 — more conservative, defensible, but a policy decision that should be documented with one line of rationale. Not a blocker.

3. **`get_max_scan_level_issuance` docstring** — still says "This does not change during runtime as the models are static and as such can be cached." while its sibling `get_max_scan_level_inheritance` was updated to "Return the maximum level transferable while tracing inheritance." Minor inconsistency.

4. **Unit test coverage** — the test pins the duplicate-neighbour regression (round-1 400) but doesn't exercise the `reversed()`/highest-edge-wins bug, which would need a two-different-level-edges topology. Acknowledged by hasecon in round 4. The robot test covers the end-to-end path. Not a blocker.

### Verdict

**Approved.** The algorithm is correct and live-verified, the round-1 and round-2 regressions are fixed and pinned by a unit test, the template correctly surfaces `inherited_level` on capped edges, all CI is green, and the latest commit cleaned up the dead code and unreachable branches I flagged. The remaining items are nits and a documentation request — none are merge blockers.

PR #5289, opened 2026-08-05
Fix kat_crt_sh crash on wildcard and email certificate identities

brennodewinter:
## What

crt.sh's `name_value` and `common_name` routinely contain wildcard identities (`*.example.com`) and rfc822 (email) identities (`admin@example.com`). Neither is a valid `Hostname`: the `*` and `@` are rejected by the Hostname validator. The normalizer yielded these verbatim, so a single such identity raised a `ValidationError`. Because the normalizer runner materializes the result generator, that exception aborts the **whole** normalizer and drops **every hostname and certificate** found in the scan.

## Fix

Normalise each identity through a small `_clean_identity` helper before constructing a `Hostname`: strip the wildcard prefix (as `common_name` already did via `lstrip(".*")`) and skip email identities. A defensive `try/except ValueError` around the `Hostname` construction ensures any other malformed identity from this untrusted third-party source can no longer abort the run.

## Test

`boefjes/tests/plugins/test_crt_sh.py`: a `name_value` mixing wildcard, email and plain identities no longer crashes; the wildcard collapses onto the bare domain, plain names survive, email identities are skipped, and the certificate is still yielded.

### DIFF
--- a/boefjes/boefjes/plugins/kat_crt_sh/normalize.py
+++ b/boefjes/boefjes/plugins/kat_crt_sh/normalize.py
@@ -1,5 +1,6 @@
 import datetime
 import json
+import logging
 from collections.abc import Iterable
 
 from dateutil.parser import parse
@@ -8,6 +9,25 @@ from boefjes.normalizer_models import NormalizerOutput
 from octopoes.models.ooi.certificate import X509Certificate
 from octopoes.models.ooi.dns.zone import Hostname
 from octopoes.models.ooi.network import Network
+
+logger = logging.getLogger(__name__)
+
+
+def _clean_identity(identity: str) -> str | None:
+    """Normalise a crt.sh certificate identity into a valid hostname, or None.
+
+    crt.sh's ``common_name`` and ``name_value`` can contain wildcard identities
+    (``*.example.com``) and rfc822 (email) identities (``admin@example.com``).
+    Neither is a valid Hostname: ``*`` and ``@`` are rejected by the Hostname
+    validator. Left unchecked, a single such identity raises a ValidationError
+    that aborts the whole normalizer run and drops every hostname and
+    certificate found in the scan (the runner materializes the generator, so an
+    exception discards everything already yielded).
+    """
+    name = identity.strip().lstrip(".*")
+    if not name or "@" in name:
+        return None
+    return name
 
 
 def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
@@ -25,13 +45,18 @@ def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
         # walk over all name_value parts (possibly just one, possibly more)
         names = certificate["name_value"].lower().splitlines()
         for name in names:
-            if not name.endswith(current):
-                # todo: do we want to hint other unrelated hostnames using the same certificate / and this possibly
-                #  the same private keys for tls?
-                pass
-            if name not in unique_domains:
-                yield Hostname(name=name, network=network_reference)
-                unique_domains.add(name)
+            # todo: do we want to hint other unrelated hostnames using the same certificate / and this possibly
+            #  the same private keys for tls?
+            name = _clean_identity(name)
+            if name is None or name in unique_domains:
+                continue
+            try:
+                hostname = Hostname(name=name, network=network_reference)
+            except ValueError:
+                logger.debug("Skipping invalid crt.sh identity %r", name)
+                continue
+            yield hostname
+            unique_domains.add(name)
 
         # Yield only current certs.
         expires_in = parse(certificate["not_after"]).astimezone(datetime.timezone.utc) - datetime.datetime.now(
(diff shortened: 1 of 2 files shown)


hasecon (approved):
## QA Review

### Setup
- Branch: `fix/kat-crt-sh-invalid-identity-crash` @ `9298dca`
- Verified by A/B against main, exhaustive edge cases, and an end-to-end persistence run on the live stack

### CI/CD Status

The two `github-advanced-security` inline comments are both **resolved and outdated** — CodeQL flagged "Incomplete URL substring sanitization" on the `.endswith(current)` pattern **in the test file's assertions**, not in shipped code. That `endswith` is the pre-existing dedup heuristic (unchanged by this PR — same logic on main, just now guarded), used only to decide whether to emit an extra hostname, not as a security boundary. The CodeQL check itself is green.

### Code Review Findings

No issues. This is a well-scoped robustness fix, same class as the sibling normalizer PRs: an untrusted third-party field (crt.sh `name_value` / `common_name`) contains identities that aren't valid Hostnames, and one bad value aborted the whole run.

**Headline claim proven by A/B against main** — using the exact production input shape:

```
name_value = "*.example.com\nexample.com\nadmin@example.com\nwww.example.com"

main: CRASH ValidationError: Invalid hostname character      <- whole run aborts, ZERO output
this: hostnames=['example.com', 'www.example.com']  certs=1   <- full result
```

So on main a single `*` or `@` identity drops **every** hostname and the certificate of the scan (the runner materializes the generator, so the exception discards everything already yielded). The fix recovers all valid data.

**Verified end-to-end on the live stack** (Bytes → normalizer → Octopoes) with a mixed wildcard/email/plain `name_value`:

```
Hostname|internet|crtzone.example.org           (wildcard *.crtzone… collapsed onto the bare domain)
Hostname|internet|www.crtzone.example.org
X509Certificate|C=US, O=Example CA|0A1B2C        subject=crtzone.example.org
```

The email identity was skipped, the wildcard collapsed with no `*` leaking, and both the hostnames **and** the certificate persisted — exactly the data that is lost on main.

**The defensive `try/except ValueError` genuinely holds.** `Hostname(name=…)` raises pydantic `ValidationError` (a `ValueError` subclass, so `except ValueError` catches it), and the validator's `v.encode("idna")` can raise `UnicodeError` (also a `ValueError` subclass). I fed the normalizer a spread of malformed identities the two unit tests don't cover — mid-string wildcard (`foo.*.example.com`), underscore (`_dmarc…`), embedded space, whitespace-only line, bare `*`, `xn--`, an over-long (>63) label — and **none crashed**; each invalid identity was dropped and the valid `example.com` plus the cert always survived. The `_clean_identity` + `try/except` combination covers both the anticipated (wildcard/email) and the unanticipated malformed cases.

Minor, non-blocking: `_clean_identity` uses `lstrip(".*")`, which strips a *character set* rather than a prefix — harmless here (it correctly reduces `*.`, `.*.`, `**.` etc. to the bare domain, and `common_name` already did exactly this on main), just worth knowing it isn't a literal-prefix strip.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `test_crt_sh.py` | 2 passed |
| boefjes | A/B main vs PR (production input shape) | main crashes → 0 output; PR → full result |
| boefjes | 7 extra malformed-identity edge cases | no crashes, valid names + cert survive |
| boefjes | live raw → Octopoes | 2 Hostnames + 1 X509Certificate persisted |

The two new tests are consistent and assert the right behaviour: the mixed wildcard/email/plain case collapses to `{example.com, www.example.com}` with the cert still yielded, and the all-invalid case still yields the cert. Read the full functions — setup and assertions line up.

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Live raw → normalizer → Octopoes | OK | 2 Hostnames + 1 cert persisted (email skipped, wildcard collapsed) |
| Log analysis | OK | no crt_sh normalizer errors on correct input |

(One transparency note on my own process: my first E2E attempt fed the normalizer a `Hostname`, but this boefje consumes a **`DNSZone`** — the normalizer reads `input_ooi["hostname"]["name"]`, which is the DNSZone shape — so it logged a `KeyError: 'hostname'`. That was my harness using the wrong input OOI type, not a PR defect; the rerun above with the correct DNSZone-shaped input is the valid result. The normalizer/input contract is internally consistent: boefje consumes DNSZone, normalizer and unit tests both use the `{"hostname": {"name": …}}` shape.)

### Verdict
**APPROVED**

Real data-loss bug — one malformed crt.sh identity discarded every hostname and certificate of the scan — fixed narrowly, with the defensive layer verified to hold against inputs well beyond the two the tests cover. Proven by A/B against main and confirmed end-to-end on a live stack. Please rebase on main before merging.

underdarknl (approved):

underdarknl (approved):

boefjes/boefjes/plugins/kat_crt_sh/normalize.py
```diff
@@ -25,13 +45,18 @@ def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
...
+            name = _clean_identity(name)
+            if name is None or name in unique_domains:
+                continue
+            try:
+                hostname = Hostname(name=name, network=network_reference)
+            except ValueError:
+                logger.debug("Skipping invalid crt.sh identity %r", name)
+                continue
+            yield hostname
+            unique_domains.add(name)
```
underdarknl, line 59: Ideally, we'd want a lower() call around the string here, but as the normalizer runner and model will deduplicate double yielded objects anyway, there's no harm in keeping these.

underdarknl:
@brennodewinter  Please recommit with a valid signing key.

underdarknl (approved):

PR #5290, opened 2026-08-05
Fix kat_external_db discarding one section when the other key is missing

brennodewinter:
## What

`kat_external_db` reads the `ip_addresses` and `domains` lists with `follow_path_in_dict`, which raises `KeyError` when a key is absent. Because the normalizer runner materializes the result generator, a `KeyError` on the domains loop **discards every IP address already yielded**, and a response without `ip_addresses` yields nothing at all even when it does contain `domains`.

## Fix

Add `follow_list_path_in_dict`, which returns an empty list (and logs a warning) when a list path is absent, so the two sections import independently. A response containing only one of the two sections now imports what it has.

## Test

`boefjes/tests/plugins/test_external_db.py`: only-ip_addresses, only-domains, and both-sections inputs all import their content without crashing.

### DIFF
--- a/boefjes/boefjes/plugins/kat_external_db/normalize.py
+++ b/boefjes/boefjes/plugins/kat_external_db/normalize.py
@@ -31,6 +31,22 @@ def follow_path_in_dict(path, path_dict):
     return path_dict
 
 
+def follow_list_path_in_dict(path, path_dict):
+    """Return the list found at path, or an empty list when the path is absent.
+
+    The ip_addresses and domains sections must be imported independently: the
+    runner materializes the result generator, so a KeyError raised while
+    iterating one section discards every OOI already yielded by the other. A
+    response with only domains (or only ip_addresses) must still import what it
+    does contain.
+    """
+    try:
+        return follow_path_in_dict(path=path, path_dict=path_dict)
+    except KeyError:
+        logging.warning("Path %s not found in results, skipping that section.", path)
+        return []
+
+
 def get_indemnification_level(path_dict):
     """Return indemnification level from metadata or default."""
     try:
@@ -49,7 +65,7 @@ def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
     network = Network(name=input_ooi["name"])
     addresses_count, blocks_count, hostnames_count = 0, 0, 0
 
-    for address_item in follow_path_in_dict(path=IP_ADDRESS_LIST_PATH, path_dict=results):
+    for address_item in follow_list_path_in_dict(path=IP_ADDRESS_LIST_PATH, path_dict=results):
         interface = ip_interface(follow_path_in_dict(path=IP_ADDRESS_ITEM_PATH, path_dict=address_item))
         indemnification_level = get_indemnification_level(path_dict=address_item)
         address, mask_str = interface.with_prefixlen.split("/")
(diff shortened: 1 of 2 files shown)


hasecon (approved):
## QA Review

### Setup
- Branch: `fix/kat-external-db-partial-import` @ `c254f1f`
- Verified by A/B against main, edge-case probing, and an end-to-end import on the live stack

### CI/CD Status

### Code Review Findings

Correct, minimal fix, same failure class as the sibling normalizer PRs: `follow_path_in_dict` raises `KeyError` on a missing section, and because the runner materializes the generator, a `KeyError` on the `domains` loop discards every IP already yielded. `follow_list_path_in_dict` (try/except → `[]` + warning) lets the two sections import independently.

**Proven by A/B against main** on an only-`ip_addresses` response (no `domains`):

```
main: CRASH KeyError: "Key domains not in ['ip_addresses']"   <- discards the IP it should import
this: OK  IPAddressV4 + DeclaredScanProfile
```

Symmetric for only-`domains` (main crashes on the `ip_addresses` loop before reaching any domain → nothing at all). Confirmed all three shapes import cleanly: only-ip, only-domains, both.

**Verified end-to-end on the live stack** — pushed an only-`ip_addresses` raw (no `domains`) through Bytes → normalizer → Octopoes:

```
IPAddressV4|internet|198.51.100.23   persisted   (normalizer logged "Done", no error)
```

On main this response imports nothing; here the IP survives. The section fix does exactly what it claims.

**One substantive observation — same failure class remains at item level (non-blocking, pre-existing).**

The PR's own rationale — "the runner materializes the generator, so one exception discards everything already yielded" — applies equally to the *item-level* accesses inside the loops, which this PR leaves unguarded. A single malformed item still aborts the whole run and discards both sections. Reproduced (each with a valid item present, to show the valid data is lost too):

| Input | Result |
|-------|--------|
| `ip_addresses: [{valid}, {"foo":"bar"}]` (item missing `address`) | `KeyError` — whole run aborts |
| `ip_addresses: [{valid}, {"address":"not-an-ip"}]` | `ValueError` from `ip_interface` — aborts |
| `domains: [{"x":"y"}]` (item missing `name`) | `KeyError` — aborts |
| `ip_addresses: [{"address":..,"indemnification_level":9}]` | `ValueError: Invalid indemnificationlevel` — aborts |
| `…"indemnification_level":"abc"` | `ValueError` from `int()` (the except only catches `KeyError`) — aborts |

All of these are pre-existing (the item-level `follow_path_in_dict`, `ip_interface` and `get_indemnification_level` calls are unchanged by this diff), so none of them block this PR — it fixes precisely the section-level case it set out to. But since the motivating principle is "don't let one bad value discard everything," and this is an **untrusted external-DB** source where a single bad row is exactly the likely failure, the item level is the natural next step: wrap the per-item body in a skip-and-warn so one malformed entry drops only itself. Worth considering here or as a follow-up. (The `indemnification_level` `except KeyError` not covering the `int()`/range `ValueError` is the sharpest of these — a non-numeric level aborts the import.)

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `test_external_db.py` | 3 passed |
| boefjes | A/B main vs PR (only-ip) | main `KeyError` → 0 output; PR imports the IP |
| boefjes | live raw → Octopoes | IPAddressV4 persisted |

The three new tests are consistent and cover the fix precisely: only-ip, only-domains, both — each asserting the expected OOI type is present. Read the full file; setup and assertions line up.

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Live only-ip raw → Octopoes | OK | IPAddressV4 persisted (would be 0 on main) |
| Log analysis | OK | `kat_external_db_normalize` "Done", no errors |

(Process note: the `DeclaredScanProfile(level=…)` yielding is byte-for-byte unchanged by this PR, so the scan-level applied to imported OOIs is out of scope here — I didn't chase the level my manual `source_method="manual"` upload ended up with, since it's orthogonal to the section fix.)

### Verdict
**APPROVED**

Real partial-import data-loss bug, fixed narrowly with tests, and proven end-to-end that a one-section response now imports instead of discarding everything. The item-level residual is the same failure class one level deeper and pre-existing, so it doesn't block — but given this is an untrusted external source, extending the same skip-and-warn to individual items (especially the non-numeric `indemnification_level` case) would be a worthwhile follow-up. Please rebase on main before merging.

underdarknl:
@brennodewinter please recommit with a valid signing key

underdarknl (approved):

PR #5291, opened 2026-08-05
Fix kat_cve_finding_types crash on CVSS v4.0-only CVEs

brennodewinter:
## What

When a CVE's `metrics` contained neither `cvssMetricV31` nor `cvssMetricV30`, the normalizer fell through to `metrics["cvssMetricV2"]`. NVD increasingly returns entries scored **only** with CVSS v4.0 (`cvssMetricV40`), for which that access raises `KeyError`, aborting the normalizer and dropping the finding type.

## Fix

Select the CVSS metric from a known-version preference list (`cvssMetricV31`, `cvssMetricV30`, `cvssMetricV40`, `cvssMetricV2`) and fall back to `RiskLevelSeverity.UNKNOWN` when none is present, instead of crashing. v4.0 shares the `type` / `cvssData.baseScore` shape the extraction loop already uses.

## Test

`boefjes/tests/plugins/test_cve_finding_types.py::test_cve_with_only_cvss_v4` with a new `cve-result-with-cvss4.json` fixture; the three existing tests still pass.


hasecon (dismissed):
## QA Review

### Setup
- Branch: `fix/kat-cve-finding-types-cvss-v4` @ `d488639`
- Scope: one normalizer + one test + one fixture → verified by running the normalizer against real inputs rather than a full `make kat` pass.

### CI/CD Status
**All 18 checks green**, including `pre-commit` this time, `Tests (3.10–3.13)`, `CodeQL`, `Analyze` and all four boefjes dist builds. Nothing red. The branch is *behind main*, so it will want an update before merging.

### The premise checks out — verified, not taken on trust

**The crash reproduces on `main`** with the new fixture:

```
main : KeyError 'cvssMetricV2'
this : severity=critical  score=9.3
```

So the `else: cvss = metrics["cvssMetricV2"]` fall-through really does abort the normalizer for a v4.0-only CVE, and the finding type is dropped for every finding referencing it. The `for/else` replacement is correct Python and the preference order is sound.

I also pulled a real CVE from the live NVD API to sanity-check the shape assumption. `CVE-2024-3400` returns:

```
metrics: ['cvssMetricV31', 'ssvcV203']
cvssMetricV31[0]: fields=['cvssData','exploitabilityScore','impactScore','source','type']  type=Secondary  baseScore=10.0
```

Two things worth noting from that. First, the `cvssData.baseScore` / `type` shape the extraction loop relies on is confirmed against production data. Second — and this supports the fix — NVD does return metric keys outside the CVSS family (`ssvcV203` here), which is exactly the kind of surprise the old code could not survive.

One honest gap: I could **not** find a live v4.0-scored CVE through the NVD queries I tried, so the `cvssMetricV40` shape rests on the hand-written fixture rather than a real payload. It matches the v3.1 shape minus `exploitabilityScore`/`impactScore`, which is what I would expect, but if you have a real v4.0 response to hand it would be worth diffing the fixture against it — the whole PR hinges on that shape.

### Code Review Findings

| 1 | Medium | `normalize.py` | An **empty** metric list still crashes. `{"metrics": {"cvssMetricV31": []}}` passes the `in metrics` check, so `cvss = []`, the `for` finds no `Primary`, and the `else` branch hits `cvss[0]` → **`IndexError: list index out of range`** (reproduced). Pre-existing rather than introduced here, but it is the same failure mode one layer deeper, and this PR is the natural place to close it — `if not cvss:` instead of `if cvss is None:` covers both. |
| 2 | Low | `normalize.py` | `item["cvssData"]["baseScore"]` is still unguarded. If NVD ever emits a metric entry without `baseScore` the normalizer aborts again. Same class; worth a `.get()` while you are in here, or explicitly out of scope. |
| 3 | Question | `normalize.py` | The order places `cvssMetricV40` *after* v3.0 but before v2. That reads deliberate — keeping scores comparable with historical v3.x data and with `get_risk_level`'s bands, which were drawn for v3.x — but the reasoning is not written down. One sentence in the comment would stop a future reader "fixing" the order to prefer the newest version. |

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `test_cve_finding_types.py` (Docker) | ✅ 4 passed, incl. the new v4.0 case |
| boefjes | CI `Tests (3.10–3.13)` | ✅ green |
| any | `pre-commit`, `CodeQL`, `Analyze` | ✅ green |

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Crash reproduced on `main` | ✅ | `KeyError: 'cvssMetricV2'` on the v4.0-only fixture |
| Same input on this branch | ✅ | CRITICAL / 9.3 |
| Existing v3.1, v2 and no-metrics paths | ✅ | the three original tests still pass |
| Real NVD payload shape | ✅ | `cvssData.baseScore` + `type` confirmed against `CVE-2024-3400` |
| Empty metric list | ❌ | `IndexError` — finding 1 |

### Verdict

**APPROVED**

Real bug, verified reproducible on main, fixed narrowly and with a test that pins the behaviour. Nothing here blocks: CI is fully green including `pre-commit`, and the three findings are hardening opportunities rather than defects in what you changed.

That said, finding 1 is close enough to the bug you are fixing that I would rather see it in this PR than as a follow-up — changing `if cvss is None:` to `if not cvss:` is a one-word edit that closes an identical crash. Please also update the branch with main before merging.

brennodewinter:
Thanks for the thorough QA review — all three findings are addressed in f1d7ccf0b, and the branch has been updated with main.

**Finding 1 (empty metric list → IndexError):** Fixed, slightly broader than the suggested one-word edit. The selection loop now skips empty lists (`metrics.get(metric)` instead of `metric in metrics`), so an empty `cvssMetricV31` no longer shadows a populated lower-priority version, and the fallback check is now `if not cvss:`. Pinned by two new tests: empty list → UNKNOWN, and empty v3.1 + populated v2 → falls through to v2.

**Finding 2 (unguarded `baseScore`):** Taken in scope. Access now goes through `.get("cvssData", {}).get("baseScore")`; a metric entry without a base score yields UNKNOWN instead of aborting the normalizer. Also covered by a new test.

**Finding 3 (ordering rationale):** The comment now spells out why `cvssMetricV40` comes after v3.x: the severity bands in `get_risk_level` were drawn for v3.x scores, and preferring v3.x keeps scores comparable with historical data.

On the honest gap you flagged: agreed that the v4.0 fixture is hand-written. If a real v4.0-scored NVD payload turns up we should diff it against the fixture, but the guards added here mean a shape surprise now degrades to UNKNOWN rather than dropping the finding type.

hasecon (approved):
## QA Review — round 2 (re-review of `f1d7ccf`)

### Setup
- Branch: `fix/kat-cve-finding-types-cvss-v4` @ `f1d7ccf` (was `d488639`)
- Branch is now genuinely current: `origin/main` is an ancestor of HEAD, verified rather than read off the mergeable state.

### Round-1 findings — all three verified fixed by running them, not by reading the diff

| 1 | Medium: empty metric list → `IndexError` on `cvss[0]` | ✅ **Fixed, and better than I suggested** |
| 2 | Low: unguarded `item["cvssData"]["baseScore"]` | ✅ **Fixed** |
| 3 | Question: ordering rationale undocumented | ✅ **Documented** |

Exercised against the live normalizer:

```
cvssMetricV31: []                          -> unknown   score=None
v31 empty + v2 populated (5.0)             -> medium    score=5.0     <- finding 1b
v31 entry without baseScore                -> unknown   score=None
v31 entry without cvssData                 -> unknown   score=None
only cvssMetricV40 (9.3)                   -> critical  score=9.3     <- original case, no regression
```

On finding 1 — you went wider than the one-word edit I proposed, and you were right to. Moving the emptiness check into the *selection* loop (`metrics.get(metric)` rather than `metric in metrics`) fixes a second, quieter bug I had not separated out: with `if not cvss:` alone, an empty `cvssMetricV31` would still have shadowed a populated `cvssMetricV2` and silently produced UNKNOWN for a CVE that does have a score. The `v31 empty + v2 populated` line above is that case, and it now resolves to MEDIUM/5.0. Good catch on your side.

The `next((item for item in cvss if item["type"] == "Primary"), cvss[0])` rewrite also reads better than the `for/else` it replaces.

### CI/CD Status
**All 18 checks green** — `pre-commit`, `Tests (3.10–3.13)`, `CodeQL`, `Analyze` and all four boefjes dist builds. I waited for the run to finish rather than reading a partial state; at first look several jobs were still in progress and `CodeQL` was `neutral`, both of which settled to success.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `test_cve_finding_types.py` (Docker) | ✅ **7 passed** (was 4) |
| boefjes | `pre-commit` (local + CI) | ✅ green |
| boefjes | CI `Tests (3.10–3.13)` | ✅ green |

The three added tests pin exactly the three behaviours: empty list, empty-preferred-falls-through, and missing base score.

### One observation, explicitly not a request

`item["type"]` in the `next(...)` generator is still a bare subscript, so a metric entry lacking `type` would raise `KeyError` — the same shape of hole as the two you just closed. NVD marks `type` as required and I have no evidence of it ever being absent, so I am flagging it for the record rather than asking for a third round. This PR is now well past the bar.

### Verdict

**APPROVED**

All three findings addressed, the fix for the first one is broader and better than what I asked for, and the behaviour is verified by execution rather than inspection. Ready to merge as far as I am concerned.

Carried over from round 1 and still true: the crash reproduces on `main` (`KeyError: 'cvssMetricV2'`), the real-NVD shape check against `CVE-2024-3400` confirms `cvssData.baseScore` / `type`, and the `cvssMetricV40` fixture remains hand-written — worth diffing against a real v4.0 payload if one turns up, though the guards you added now mean a shape surprise degrades to UNKNOWN instead of dropping the finding type.

hasecon:
## Post-approval note: an adjacent crash of the same class, one guard higher

My round-2 **APPROVED** still stands for `b0822b9` — CI is 18/18 green, the branch is current with main, and the three round-1 findings remain fixed. This is not a request to change that.

But while reviewing #5296 I reproduced an adjacent crash in **this same normalizer**, and it's the *exact* failure mode this PR's description decries ("aborting the normalizer and drops the finding type") — just one guard higher up, on lines the diff doesn't touch:

```python
# normalize.py:31-34, unchanged by this PR
descriptions = data["cve"]["descriptions"]          # <- KeyError if the envelope has no "cve"
...
if not data["cve"]["metrics"]:                       # <- same
```

The CVE lookup source (`cve.circl.lu`) does not always return a `{"cve": {...}}` envelope — an unknown, reserved, disputed or very-new CVE comes back without it. Reproduced against the live normalizer:

```
response {}                    -> KeyError: 'cve'
response {"message":"..."}     -> KeyError: 'cve'
response null                  -> TypeError: 'NoneType' object is not subscriptable
{"cve":{"metrics":{}}}  (no descriptions) -> KeyError: 'descriptions'
```

This is arguably higher-frequency than the v4.0-only case you're fixing: a scan reports a CVE id that circl.lu doesn't know (fresh CVEs especially), the boefje stores the "not found" envelope, and the normalizer aborts here — dropping the finding type for every finding referencing that CVE. I hit it end-to-end during #5296 with a scan-reported CVE that circl.lu had no record of.

Since this PR is already hardening exactly this normalizer against exactly this "unexpected response shape" failure, it's the natural home for the outer-envelope guard too. Roughly:

```python
cve = data.get("cve") or {}
descriptions = cve.get("descriptions") or []
english = next((d for d in descriptions if d.get("lang") == "en"), None)
english_description = english["value"] if english else ""
metrics = cve.get("metrics") or {}
if not metrics:
    ...
```

Your call whether to fold it in here (one small block, same theme) or take it as a follow-up — flagging it because I have a live reproduction and it's the same bug class you're closing. The v4.0 fix itself is complete and correct; I re-verified it still returns CRITICAL/9.3 on this branch.

boefjes/boefjes/plugins/kat_cve_finding_types/normalize.py
```diff
@@ -36,20 +36,28 @@ def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
...
+        # The old code fell through to metrics["cvssMetricV2"], raising KeyError
+        # in that case, which aborts the normalizer and drops the finding type.
+        # Prefer the known metric versions and fall back to UNKNOWN instead.
+        # v4.0 deliberately comes after v3.x: the severity bands in
+        # get_risk_level were drawn for v3.x scores, so preferring v3.x keeps
+        # scores comparable with historical data. Empty metric lists are
+        # skipped so a populated later version wins over an empty earlier one.
+        for metric in ("cvssMetricV31", "cvssMetricV30", "cvssMetricV40", "cvssMetricV2"):
+            if metrics.get(metric):
+                cvss = metrics[metric]
```
underdarknl, line 49: Should we add the metric-type to the CSV octopoes model? (additions are pretty cheap). That way we can possibly convert from one into the other, but at minimum at least show what the original metric was when we did discovery/hydration.

hasecon (approved):
## QA Review — round 3 (re-run of `/collin 5291`)

### Setup
- Branch: `fix/kat-cve-finding-types-cvss-v4` @ `952f71ab`
- Every commit since my round-2 approval (`3580308…` onward) is a **pure `Merge branch 'main'`** — I diffed the head `normalize.py` and it is byte-identical to the code I approved. So this re-review is about (a) re-confirming that, (b) answering @underdarknl's new comment, and (c) the still-open adjacent crash. No re-run of the suite: the normalizer is unchanged and CI is green on this head.

### Round-1/2 findings — still fixed (verified against head, not the summary)
| Finding | Status at `952f71ab` |
|---------|----------------------|
| Empty metric list → `IndexError` | ✅ `metrics.get(metric)` selection loop + `if not cvss:` |
| Unguarded `baseScore` | ✅ `primary.get("cvssData", {}).get("baseScore")` |
| Ordering rationale undocumented | ✅ comment present (v4.0 after v3.x for band comparability) |

### CI/CD Status
All 18 checks green — `pre-commit`, `Tests (3.10–3.13)`, `CodeQL`, `Analyze`, boefjes builds. `mergeable_state` is **behind** main, so it needs a branch update before merge (there are already several merge-of-main commits; the squash will collapse them).

### On @underdarknl's comment (line 49 — store the CVSS metric type on the model)
Good idea and cheap, but it's **additive scope beyond this crash-fix** and it changes an octopoes OOI (`CVEFindingType`), so it wants its own field + the usual model handling rather than riding in on a boefje hardening PR. Concretely it'd be: add e.g. `cvss_version: str | None` (or a small enum) to `CVEFindingType`, set it to the `metric` key that won the selection loop (`cvssMetricV31`/`V30`/`V40`/`V2`), and surface it in the finding-type view. That gives exactly the provenance/convertibility you describe. My suggestion: take it as a **follow-up issue** so this ready, correct fix isn't held on a model change — but if Brenno wants to fold the one field in here, it's small. Your call as code owner.

### Still open — the adjacent envelope crash (my 2026-08-06 note), unchanged at head
Not introduced by this PR and not blocking it, but re-flagging since the code hasn't moved and it's the *same* failure class this PR closes, one guard higher (lines 31–34, untouched):
```python
descriptions = data["cve"]["descriptions"]          # KeyError if circl.lu returns no {"cve":{…}} envelope
english_description = [d for d in descriptions if d["lang"] == "en"][0]   # IndexError if no English description
if not data["cve"]["metrics"]:                       # same KeyError
```
circl.lu returns no `cve` envelope for unknown/reserved/withdrawn/fresh CVEs (reproduced live during #5296: `{}`→`KeyError: 'cve'`, `null`→`TypeError`, no-descriptions→`KeyError`). A scan reporting a CVE id circl.lu doesn't know aborts the normalizer and drops the finding type — arguably higher-frequency than the v4.0-only case. Same `data.get("cve") or {}` / `descriptions or []` guard shape as before. Fold-in-or-follow-up is Brenno/Jan's call.

### Verdict
**APPROVED** (unchanged from round 2)

The v4.0 fix is complete, correct, and byte-identical to the approved commit; CI is green. The two items above are both *non-blocking*: Jan's metric-type idea is a follow-up enhancement, and the envelope guard is a pre-existing out-of-diff crash I'd love to see closed but won't hold this PR on. Please update the branch with main before merging.

PR #5292, opened 2026-08-05
Fix kat_kat_finding_types crash on unknown ids and missing risk

brennodewinter:
## What

The normalizer looked up `data[kat_finding_type_id]` and `finding_type_information["risk"]` directly:

- a `KATFindingType` id that is not in the static catalog raised `KeyError`. This is not hypothetical — ids minted by bits (e.g. `KAT-HTTPS-NOT-AVAILABLE`, minted by the `https_availability` bit) can be absent from `kat_finding_types.json`;
- a catalog entry with a missing or unrecognised `risk` raised `KeyError`/`ValueError`.

Either aborts hydration for every finding of that type.

## Fix

Use `data.get(id)` and affirm the type with `RiskLevelSeverity.UNKNOWN` when it is not described, and fall back to `UNKNOWN` when the risk level is missing or unrecognised, logging a warning in both cases. `risk_score` uses `SEVERITY_SCORE_LOOKUP.get(...)` so an unknown severity no longer raises.

## Test

`boefjes/tests/plugins/test_kat_finding_types.py`: unknown id, known id, and missing-risk-key inputs.

## Follow-up (not in this PR)

This hardens the crash but the affected finding still surfaces with unknown severity until its catalog entry exists. Adding the missing `KAT-HTTPS-NOT-AVAILABLE` entry to `kat_finding_types.json` (and a CI check that every id minted in bits/normalizers is present) is a sensible follow-up — its risk level is a content decision left out here.


hasecon (changes requested):
## QA Review

### Setup
- Branch: `fix/kat-finding-types-unknown-id` @ `6c49104`
- Scope: one normalizer + one new test file → the verification below is the normalizer run against real inputs rather than a full `make kat` pass.

### CI/CD Status

17 of 18 checks green — `Tests (3.10–3.13)`, `CodeQL`, `Analyze`, all four boefjes dist builds.

**`pre-commit` is red, which Rule 1 makes blocking on any PR.** Reproduced locally; it is a single formatting nit in the new test file, which ruff-format collapses because it fits on one line:

```python
-            ooi=KATFindingType(
-                id="KAT-HTTPS-NOT-AVAILABLE", risk_severity=RiskLevelSeverity.UNKNOWN, risk_score=None
-            )
+            ooi=KATFindingType(id="KAT-HTTPS-NOT-AVAILABLE", risk_severity=RiskLevelSeverity.UNKNOWN, risk_score=None)
```

`pre-commit run --all-files` and a commit clears it. Same thing bit #5279 — worth wiring pre-commit into your local hooks so it stops costing you a round.

### The premise checks out — I verified it rather than taking the description's word

The PR claims the crash is "not hypothetical". Confirmed on all three counts:

- `KAT-HTTPS-NOT-AVAILABLE` is **absent** from `kat_finding_types.json` (zero matches).
- It **is** minted by a bit: `octopoes/bits/https_availability/https_availability.py:15`.
- **The crash reproduces on `main`** with the exact input: `KeyError: 'KAT-HTTPS-NOT-AVAILABLE'`.

And on this branch the same input now yields, with a warning logged:

```
[('KAT-HTTPS-NOT-AVAILABLE', <RiskLevelSeverity.UNKNOWN: 'unknown'>, None)]
```

I also quantified the blast radius, since "ids can be absent" invites the question *how many*: of **49** finding-type ids minted as literals across `octopoes/bits/**` and the normalizers, **48** are in the catalog and exactly **one** is missing — this one. So the catalog is otherwise complete, and the follow-up you describe is a single content entry rather than an open-ended cleanup. That is worth stating in the follow-up issue.

### Downstream safety of `risk_score=None`

The part I most wanted to check, because a `None` score flowing into report generation would be a nastier bug than the one being fixed. Every consumer already guards it — each sort uses `risk_score or 0`:

```
rocky/reports/report_types/findings_report/report.py:85
rocky/reports/report_types/dns_report/report.py:72
rocky/reports/report_types/aggregate_organisation_report/report.py:248
rocky/reports/report_types/multi_organization_report/report.py:271
rocky/rocky/views/ooi_detail_related_object.py:112
```

and `RiskLevelSeverity.UNKNOWN` is already a value the UI counts (`ooi_detail_related_object.py:107`). So affirming with UNKNOWN/None degrades gracefully instead of introducing a new failure mode. Good call over the alternative of not affirming at all, which would have left the finding without a type entirely.

`logger` is properly defined at module level (`logging.getLogger(__name__)`) — not visible in the diff, so I checked the whole file; no `NameError` waiting.

### Code Review Findings

| 1 | **Blocking (trivial)** | `tests/plugins/test_kat_finding_types.py` | `pre-commit` red — one ruff-format line collapse, see above |
| 2 | Low | same | The `except ValueError` branch is untested. The description calls out "unrecognised risk" as a case, and the code handles it, but no test feeds e.g. `{"risk": "bogus"}`. A fourth test asserting UNKNOWN + a logged warning would close the loop on the three paths you describe. |
| 3 | Nit | `normalize.py` | `str(risk).lower() if risk else UNKNOWN` also treats `0` and `""` as missing. Harmless for this data (risk is always a word), just noting the truthiness check is slightly wider than "is None". |

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | new `test_kat_finding_types.py` (Docker) | ✅ 3 passed |
| boefjes | CI `Tests (3.10–3.13)` | ✅ green |
| any | `CodeQL`, `Analyze` | ✅ green |
| any | `pre-commit` | ❌ blocking, trivial |

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Crash reproduced on `main` | ✅ | `KeyError: 'KAT-HTTPS-NOT-AVAILABLE'` |
| Same input on this branch | ✅ | affirms UNKNOWN / `None`, logs a warning |
| Known id still hydrates fully | ✅ | covered by the new test, HIGH → 8.9 |
| `risk_score=None` downstream | ✅ | all report sorts use `risk_score or 0` |

### Verdict

**CHANGES REQUESTED** — purely because `pre-commit` is red, and Rule 1 does not let that through. Run it, push, and this is ready.

The substance is solid: a real crash, verified reproducible on main, fixed in the least invasive way, with tests that match the implementation and a follow-up that is honest about what it does *not* fix. Checking that `None` was already safe downstream before letting it propagate is exactly the right instinct for this kind of hardening.

hasecon (approved):
## QA Review — round 2 (re-review of `66dad22`)

### Setup
- Branch: `fix/kat-finding-types-unknown-id` @ `66dad22` (was `6c49104`; two commits added)
- Scope unchanged: one normalizer + its test file.

### Round-1 findings — both verified fixed at the actual head

| 1 | **Blocking:** `pre-commit` red — ruff-format line collapse in the test file | ✅ **Fixed** (`222a2ad`). Verified in the file itself, not just from the check: the `KATFindingType(...)` call is now on one line, and `pre-commit run` on both changed files is silent locally. CI `pre-commit` is green. |
| 2 | Low: the `except ValueError` branch was untested | ✅ **Fixed** (`66dad22`). New `test_kat_finding_types_unrecognised_risk_falls_back_to_unknown` feeds `{"risk": "bogus"}` and asserts both the UNKNOWN fallback and the logged warning via `caplog`. |
| 3 | Nit: truthiness check treats `0` / `""` as missing | Not changed — fine, I flagged it as an observation rather than a request. |

I checked that the new test actually binds to the branch it claims, rather than passing for an unrelated reason:

```
RiskLevelSeverity("bogus")  -> ValueError            (so the except branch is live)
run({"risk": "bogus"}) -> unknown / None / name="Example"
WARNING ... Unknown risk level 'bogus' for KAT finding type KAT-EXAMPLE.
```

The warning fires, the fallback applies, and `name` still survives — so the failure is contained to the severity rather than dropping the rest of the entry.

### CI/CD Status
**All 18 checks green**, including `pre-commit`, `Tests (3.10–3.13)`, `CodeQL`, `Analyze` and all four boefjes dist builds. Branch is *behind main*, so it will want an update before merging.

### Carried forward from round 1 (verified then, still true)

- The crash reproduces on `main` with the real input: `KeyError: 'KAT-HTTPS-NOT-AVAILABLE'`, and this branch affirms UNKNOWN/`None` instead.
- Scope is exactly one id: of **49** finding-type ids minted as literals across `octopoes/bits/**` and the normalizers, **48** are in the catalog. Your follow-up is a single entry, not an open-ended cleanup.
- `risk_score=None` is safe downstream — every report sort already uses `risk_score or 0`, and `RiskLevelSeverity.UNKNOWN` is a value the UI already counts.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `test_kat_finding_types.py` (Docker) | ✅ **4 passed** (was 3) |
| boefjes | CI `Tests (3.10–3.13)` | ✅ green |
| any | `pre-commit`, `CodeQL`, `Analyze` | ✅ green |

### Verdict

**APPROVED**

Clean round 2 — both findings addressed exactly, and the added test covers the one path the description claimed but did not exercise. Please update the branch with main before merging.

Worth doing the follow-up you describe soon: the catalog entry for `KAT-HTTPS-NOT-AVAILABLE` plus a CI check that every minted id exists. With 48 of 49 already present, that check would start green and stay useful — it turns "a finding silently shows as unknown severity" into a build failure at the moment someone mints a new id.

underdarknl (approved):

underdarknl:
Next up would be to move the consumed json to a configurable url, where the community can create one together (could be git, could be a result of something similar to a wiki), and this also allows users of kat to insert their own url, and host their own customized findings to warning levels lists.

PR #5293, opened 2026-08-05
Extract Software from nmap CPEs in the shared nmap normalizer

brennodewinter:
## What

The nmap normalizer (`kat_nmap_tcp/normalize.py`, shared by **all four** nmap boefjes — `kat_nmap_tcp`, `kat_nmap_udp`, `kat_nmap_ports`, `kat_nmap_ip_range` via its `consumes` list) only read `service.service`. nmap's version-detection output — the CPEs and product/version it reports with `-sV` — was dropped entirely. **No `Software`/`SoftwareInstance` was produced**, so CVEs could not be bound to the software actually running on a port, even though the raw XML contains e.g. OpenSSH 8.4p1, ISC BIND 9.16.44, nginx 1.18.0.

## Fix

Yield `Software` + `SoftwareInstance` from each application CPE (`cpe:/a:...`) via `cpe_to_name_version`, falling back to the raw product/version when no CPE is present — the same pattern `kat_shodan_internetdb` already uses. Operating-system CPEs (`cpe:/o:...`) are intentionally left out: they describe the host, not the service. One change fixes all four nmap boefjes.

`normalizer.json` `produces` updated with `Software`/`SoftwareInstance`.

## Test

`boefjes/tests/plugins/test_nmap.py`: the existing test's output count goes 15 → 25, plus a new `test_normalizer_extracts_software_from_cpes` asserting the four expected Software objects (openssh/bind/nginx/mysql), that OS CPEs are excluded, and that each Software is bound to its `IPService`.


hasecon (approved):
## QA Review — round 1

### Setup
- Branch: `feat/nmap-cpe-to-software` @ `0f1edb5`
- Stack: `make clean && make kat` — OK (13/13 containers running, Rocky HTTP 200)

### CI/CD Status
32 checks, all completed: **31 success + 1 skipped** (`add-debian-packages-to-release`, expected outside a release run). No failures, nothing pending. All boefjes-blocking checks green: `Tests (3.10/3.11/3.12/3.13)`, `pre-commit`, `CodeQL`, and `build_containerized_boefjes (kat_nmap_tcp → openkat-nmap)`.

### Code Review Findings

No blocking issues. Verified the substance of the change rather than just the diff:

**The premise holds.** All four nmap boefjes run with `-sV` (`kat_nmap_tcp`, `kat_nmap_udp`, `kat_nmap_ports`, `kat_nmap_ip_range` — checked each `boefje.json`), so version detection is genuinely on in production and this change produces real data rather than only exercising the fixture. The `consumes` list in `normalizer.json` confirms the one normalizer serves all four plus `openkat/nmap-output`, so a single change really does cover them all.

**`cpe_to_name_version` is used correctly.** It splits on `:` and takes index 3/4, which maps `cpe:/a:openbsd:openssh:8.4p1` → `("openssh", "8.4p1")` and the version-less `cpe:/a:mysql:mysql` → `("mysql", None)`. The `if not name: continue` guard is more defensive than the equivalent loop in `kat_shodan_internetdb`, which has no such guard.

**Binding to `IPService` is the right call**, and better than the pattern it follows: `kat_shodan_internetdb` binds `SoftwareInstance` to the `IPAddress`, so it cannot distinguish nginx on :80 from nginx on :443. Confirmed live that this PR produces two separate instances for exactly that case.

**Excluding `cpe:/o:` is correct and deliberate.** In the same test org, shodan contributed `Software|linux_kernel||cpe:/o:linux:linux_kernel` and `ubuntu_linux` bound at IP level. Binding an OS CPE to an `IPService` would be wrong, so the asymmetry with shodan is justified — worth keeping the docstring line that explains it.

**No adverse interaction with the `retire_js` bit.** That bit consumes every `Software` OOI and compares versions with `packaging.version.parse`, which raises `InvalidVersion` on nmap-style versions such as `8.4p1`. I checked all 45 brands in `retirejs.json` against typical nmap product names — they are all JavaScript libraries, no collisions — and then ran the bit against the five Software objects this PR produces: 0 findings, no exception, because `found_brands` short-circuits before any version parsing. No false positives introduced.

Two minor, non-blocking observations:

| Severity | Finding |
|---|---|
| Low | **Name casing differs between the two branches.** The CPE path yields lowercase (`nginx`, `openssh`) because that is what the CPE contains; the fallback yields nmap's raw `product` verbatim, which is capitalised — I confirmed a synthetic fixture produces `Software|Jetty|9.4.z-SNAPSHOT|`. Since `name` is part of `_natural_key_attrs`, the same product would become two distinct OOIs depending on whether nmap happened to emit a CPE for that fingerprint. A `.lower()` on the fallback name would align the two paths. Optional. |
| Nit | The `return` after the CPE loop skips the fallback when *every* CPE yields a falsy name (e.g. a malformed `cpe:/a:`). Not reachable with realistic nmap output; noting it only for completeness. |

The duplicate `Software` yield for nginx on :80 and :443 is harmless — both carry the identical reference, and Octopoes deduplicates on primary key.

### Pre-existing issue, explicitly *not* attributed to this PR

The new fallback branch passes nmap's raw `version` string into `Software.version`. That string is derived from banner regex captures and is therefore remote-influenceable. I confirmed a `|` in it corrupts the OOI primary key (`Software|Foo|1.0|evil|`) and makes `SoftwareInstance.format_reference_human_readable` raise `TypeNotFound`, which would surface as a rendering error in Rocky.

**This is not a reason to hold this PR.** The same unsanitised path already exists in `kat_nikto` (`match.group(2)`), `kat_dns_version` (the `version.bind` TXT record, entirely attacker-controlled), `kat_leakix`, `kat_censys` and several `kat_binaryedge` normalizers. It is a platform-level gap in reference/PK escaping, not something one normalizer should paper over, and fixing it here would be inconsistent with six other call sites. Worth a separate ticket — happy to file it.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `tests/plugins/test_nmap.py` | 8 passed (incl. the new `test_normalizer_extracts_software_from_cpes`) |
| boefjes | full unit suite | 130 passed, 2 failed |
| boefjes | the 2 failures, re-run on `main` | fail identically — pre-existing, environmental |

The two failures are `test_app.py::test_create_manager` (no Docker socket inside the test container) and `test_api.py::test_boefje_input_running`. Both reproduce unchanged on `main`, so neither is caused by this PR; they pass in CI, which provides the socket.

The test file is internally consistent: 15 → 25 outputs matches the fixture's 5 application CPEs (nginx appears on two ports) plus their 5 instances. The four `in`-assertions plus `len(instances) == 5` correctly allow for nginx collapsing to one entry in the set comprehension while remaining two distinct instances.

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| End-to-end normalizer | OK | see below |
| Log analysis | OK | 9 nmap normalizer runs, zero errors |
| Boefje tasks | OK | `nmap` 2× completed |

I pushed the fixture through the real pipeline (Bytes raw upload → RabbitMQ → scheduler → normalizer → Octopoes) rather than only calling `run()` directly, and queried Octopoes afterwards. All five bindings landed:

```
SoftwareInstance|IPService|internet|134.209.85.72|tcp|22|ssh    <- Software|openssh|8.4p1|cpe:/a:openbsd:openssh:8.4p1
SoftwareInstance|IPService|internet|134.209.85.72|tcp|53|domain <- Software|bind|9.16.44|cpe:/a:isc:bind:9.16.44
SoftwareInstance|IPService|internet|134.209.85.72|tcp|80|http   <- Software|nginx|1.18.0|cpe:/a:igor_sysoev:nginx:1.18.0
SoftwareInstance|IPService|internet|134.209.85.72|tcp|443|https <- Software|nginx|1.18.0|cpe:/a:igor_sysoev:nginx:1.18.0
SoftwareInstance|IPService|internet|134.209.85.72|tcp|3306|mysql<- Software|mysql||cpe:/a:mysql:mysql
```

Beyond the injected fixture, the **live** nmap boefje also produced Software through the new code path during the same run — `Software|go-ipfs||cpe:/a:protocol_labs:go-ipfs` bound to `IPService|...|87.215.218.70|tcp|80|http`. That is the strongest confirmation available: real scan, real pipeline, correct binding.

Two boefje tasks failed in the run; both were `testssl-sh-ciphers`, unrelated to nmap. The `422` entries in the octopoes_api log were my own malformed polling queries (an unencoded `+00:00`), not a product issue.

Also checked: `produces` in `normalizer.json` is not validated at ingest, but it *is* used for KAT-alogus plugin filtering and the normalizer detail page, so the update is genuinely needed for discovery rather than being cosmetic. Local plugins are read from disk (`BoefjeResource.__init__`), so no migration is required for existing installations.

### Verdict
**APPROVED**

Well-scoped change that closes a real gap — nmap has been running `-sV` all along and throwing the results away. Fixing it in the shared normalizer covers all four boefjes at once, and binding to `IPService` rather than the IP is a genuine improvement over the shodan pattern it is modelled on. The new test is meaningful and the behaviour holds up end-to-end on a live stack.

The casing nit is worth a look but is not a reason to hold this up.

PR #5294, opened 2026-08-05
Extract Software from service banners and connect over IPv6

brennodewinter:
## What

`kat_service_banner` grabs a service banner but had **no normalizer of its own** — the banner only fed the `kat_cve_2024_6387` check. So the product/version a banner exposes (`SSH-2.0-OpenSSH_8.4p1`, `220 (vsFTPd 3.0.3)`, …) was never turned into `Software`, and CVEs couldn't be bound to it.

Separately, the boefje hardcoded `socket.AF_INET`, so **every IPv6 `IPPort` silently failed** to connect (`IPPort.address` can be an `IPAddressV6`).

## Fix

1. **New normalizer** (`kat_service_banner/normalize.py` + `normalizer.json`) that parses banners into `Software` + `SoftwareInstance` bound to the input `IPPort`. It is **deliberately conservative**: only well-anchored, server-speaks-first formats are matched (SSH per RFC 4253, vsFTPd), so an ambiguous banner never fabricates Software. New formats are a one-line addition.
2. **IPv6**: replace the `AF_INET` socket with `socket.create_connection`, which resolves and connects over both address families.

## Test

`boefjes/tests/plugins/test_service_banner.py`: SSH and vsFTPd parse correctly; ambiguous/empty banners yield nothing; `run` binds the Software to the IPPort.

## Scope / follow-up

The parser starts with the two most reliable formats; SMTP/IMAP/POP3/MySQL etc. can be added as their patterns are pinned down. A dedicated `ServiceBanner` OOI (to keep the raw banner as evidence) is a separate model-level proposal, out of scope here.


hasecon (approved):
## QA Review

### Setup
- Branch: `feat/service-banner-software-ipv6` @ `f1f0393`
- Both halves verified independently: the new normalizer end-to-end on a live stack, the IPv6 fix against a real IPv6 listener

### CI/CD Status

### Code Review Findings

Two well-scoped changes, both real gaps, and I verified each one rather than trusting the description.

**1. New banner→Software normalizer — verified end-to-end.** The boefje produces `openkat/service-banner` and the new normalizer consumes exactly that mime, so it genuinely runs (not a dangling plugin). Pushed an SSH banner through Bytes → normalizer → Octopoes:

```
Software|OpenSSH|8.4p1|
SoftwareInstance|IPPort|internet|203.0.113.99|tcp|22|Software|OpenSSH|8.4p1|
```

Binding `SoftwareInstance` to the input `IPPort` is the right granularity (matches how the shodan/nmap software extraction anchors). On main this boefje had no normalizer at all, so the banner's product/version was never recorded.

**The parser is genuinely conservative, and I tried to break it:**

| Banner | Result |
|--------|--------|
| `SSH-2.0-OpenSSH_8.4p1 Debian-5+deb11u2` | `(OpenSSH, 8.4p1)` |
| `SSH-2.0-OpenSSH` (no version) | `(OpenSSH, None)` |
| `SSH-1.99-OpenSSH_7.4` | `(OpenSSH, 7.4)` |
| `SSH-2.0-dropbear_2020.81` | `(dropbear, 2020.81)` |
| `220 (vsFTPd 3.0.3)` / multiline vsFTPd | `(vsFTPd, 3.0.3)` |
| `SSH-2.0-` (empty software) | `None` — no empty Software |
| `SSH-2.0-OpenSSH_8.4p1\nmalicious` | `(OpenSSH, 8.4p1)` — `\S+` stops at the newline, no injection past line 1 |
| SMTP / HTTP / empty | `None` — no fabrication |

Leading whitespace is handled because `run()` strips before parsing. Good discipline — the conservative-by-default stance holds up.

**2. IPv6 fix — proven against a real IPv6 listener.** `socket.create_connection` replaces the hardcoded `AF_INET`. Ran both against a `::1` listener:

```
old (AF_INET socket)      -> FAILED (None)     <- every IPv6 IPPort silently failed on main
new (create_connection)   -> CONNECTED (AF_INET6), banner read
```

`except OSError` is correct and tighter than the old `except Exception` — `create_connection`'s failure modes (timeout, refused, `gaierror`) are all `OSError` subclasses.

**Test-assertion update is correct.** `test_add_normalizer` goes 58 → 59 for the one added normalizer. I ran the boefjes integration suite in Docker to confirm, rather than eyeballing the number: **36 passed, 2 deselected** — the count assertion passes, so the normalizer registers.

**One pre-existing, non-blocking note (#5299 class).** The SSH `\S+` software capture is remote-controlled (the scanned server sends the banner), and a `|` in it corrupts the OOI primary key — `SSH-2.0-Evil|Corp_1.0` → `Software|Evil|Corp|1.0|` (reproduced). This is the same unescaped-separator issue tracked in #5299 that affects every banner-derived `Software` (nmap, shodan, censys, …), not something this PR introduces or should fix locally — noting it only because this adds one more call site. The central `Software`/`natural_key` guard proposed in #5299 would cover it.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `test_service_banner.py` | 5 passed |
| boefjes | `make itest` (incl. `test_add_normalizer` = 59) | 36 passed, 2 deselected |
| boefjes | parser edge cases (pipe, newline, empty, multiline) | conservative, no false Software |
| boefjes | IPv6 connect (old vs new) | old fails, new connects via AF_INET6 |
| boefjes | live banner → Octopoes | Software + SoftwareInstance persisted |

The five new unit tests are consistent and cover the right axes: SSH/dropbear/vsFTPd parse, ambiguous/empty yield nothing, and `run` binds the Software to the IPPort.

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Live banner → normalizer → Octopoes | OK | `Software\|OpenSSH\|8.4p1` bound to the IPPort |
| IPv6 connect | OK | proven against a `::1` listener |
| Log analysis | OK | `kat_service_banner_normalize` "Done", no errors |

### Verdict
**APPROVED**

Two genuine gaps closed: a service banner's software is now recorded (so CVEs can bind to it), and IPv6 ports are actually scanned instead of silently failing. The parser is conservative enough not to fabricate Software, the normalizer registers and runs end-to-end, and the IPv6 fix is proven against a real listener. The `|`-in-banner PK issue is pre-existing and belongs in #5299, not here. Please rebase on main before merging.

The SMTP/IMAP/POP3/MySQL follow-up and the `ServiceBanner` OOI proposal are sensible as separate scope.

PR #5286, opened 2026-08-05
Fix nuclei boefjes: ship templates and pass -t

hasecon:
### Changes

All three nuclei boefjes fail on every run, for two stacked reasons. Full analysis in #5285; this PR fixes the unambiguous parts.

**1. The image contained no templates.** `boefje.Dockerfile` only copied the nuclei binary, so `/root/nuclei-templates/` did not exist. Nuclei's runtime rescue is a silent failure — `nuclei -update-templates` prints "installing…", exits **0**, and installs **zero** files. Every scan ended in `[FTL] Could not run nuclei: no templates provided for scan`.

This PR clones a pinned `nuclei-templates` release in the build stage and copies it into the image. Pinned rather than fetched at runtime so builds are reproducible, work air-gapped, and cannot silently produce a template-less image.

**2. Two of three boefjes never passed `-t`.** Nuclei takes no positional arguments, so it ignored the path and fell back to its default directory:

```
kat_nuclei_cve            ["-t", ".../http/cves/", "-jsonl"]        ✅
kat_nuclei_exposed_panels [".../http/exposed-panels/", "-jsonl"]    ❌ → now fixed
kat_nuclei_take_over      [".../http/takeovers/", "-jsonl"]         ❌ → now fixed
```

Worth knowing for review: one image serves all three boefjes — only `kat_nuclei_cve` has a Dockerfile and `main.py`, and the runner resolves the plugin by image. The other two are defined *entirely* by `oci_arguments`, which is why a missing flag there silently disabled them since #4644.

**3. Failures now explain themselves.** `check_returncode()` discarded `stderr` even though it was captured, so a failed task showed only `returned non-zero exit status 1`. Diagnosing this required running nuclei by hand. Nuclei's own reason is now included in the raised error.

**4. CI builds the nuclei image on PRs that touch it.** `containerized_boefjes.yml` listed only nmap/dnssec/nikto/export-http under `pull_request.paths`, so nuclei changes were never built before merge — part of why this stayed broken.

### Issue link

Closes #5285

### QA notes

Verified against a locally built image (`--target build` and full image), running each boefje with the exact `oci_arguments` from its own `boefje.json`:

| Boefje | Before | After |
|---|---|---|
| `nuclei-cve` | exit 1, `no templates provided for scan` | **exit 0, 4156 templates loaded** |
| `nuclei-exposed-panels` | exit 1, `no templates provided for scan` | **exit 0, 1554 templates loaded** |
[...]

PR #5295, opened 2026-08-05
Attach nuclei panel/takeover findings to the matched URL

brennodewinter:
## What

`kat_nuclei_exposed_panels` and `kat_nuclei_take_over` hung every detection on the input `Hostname` with a **constant** `KATFindingType` (`EXPOSED-ADMIN-PANELS` / `SUB-DOMAIN-TAKEOVER`). Since `Finding.natural_key` is `f"{ooi}|{finding_type}"`, all panels (or all takeovers) on one host share the same key and **collapse into a single Finding** — every detection after the first is lost on persistence, and nuclei's `matched-at` (the exact location) was never used.

## Fix

Build a `URL` from nuclei's `matched-at` and attach the finding there, so each detected location becomes its own `Finding`. Falls back to the hostname when `matched-at` is absent or not parseable as a URL. `normalizer.json` `produces` now lists `URL`.

## Test

New `test_nuclei_exposed_panels.py` and `test_nuclei_take_over.py`: two detections on one host now yield two distinct findings on two distinct URLs (no collapse), and the hostname fallback still works when `matched-at` is missing.

## Related

The missing `-t` flag in these two boefjes' `oci_arguments` (Patroon 3, third bullet) is already handled by the open PR #5286, so it is not duplicated here.


hasecon (approved):
## QA Review

### Setup
- Branch: `feat/nuclei-findings-per-location` @ `e24ade9`
- Verified the headline claim (no collapse *on persistence*) through the live pipeline, not just at yield time

### CI/CD Status

### Code Review Findings

No blocking issues. This is the *correct* solution to the finding-collapse class of bug — the same one that couldn't be solved in #5302 (wpscan) because there the finding type was constant and the OOI couldn't be made distinct. Here nuclei's `matched-at` gives a genuinely distinct anchor (`URL`), so each detection gets its own `Finding.natural_key`. Both boefjes consume `Hostname`, and the produced `URL` is a legitimately different OOI type — so there's no analog to the #5302 nesting problem (I checked the input/output types specifically).

**Headline claim verified end-to-end, not just at yield.** The PR's real assertion is "every detection after the first is lost *on persistence*" — exactly the #4994 lesson that persistence can differ from what a normalizer yields. I pushed two panel detections on one host through Bytes → normalizer → Octopoes and queried what actually persisted:

```
Finding|URL|internet|https://panels.example.com/admin|EXPOSED-ADMIN-PANELS
Finding|URL|internet|https://panels.example.com/phpmyadmin|EXPOSED-ADMIN-PANELS
URL|internet|https://panels.example.com/admin
URL|internet|https://panels.example.com/phpmyadmin
```

Two distinct Findings on two distinct URLs — no collapse. On main both would share `Finding|Hostname|internet|panels.example.com|EXPOSED-ADMIN-PANELS` and persist as one. Normalizer logged `Done` with no errors.

**The fallback works, and its exception handling is correct.** `URL.raw` is a pydantic `AnyUrl`, which raises `ValidationError` on bad input — and since `ValidationError` subclasses `ValueError`, the `except ValueError` genuinely catches it. Confirmed by feeding `matched-at: "!!! not a url"`: the warning fires and the finding falls back to the hostname, no crash.

**One residual worth being aware of (not blocking):** because the finding type is still a constant (`EXPOSED-ADMIN-PANELS` / `SUB-DOMAIN-TAKEOVER`), two *different* detections at the *same* `matched-at` URL still collapse into one Finding — I reproduced this (2 yielded, 1 unique key). That's the same underlying Finding-model limitation as #5302 (no N findings of one type on one OOI), not something this PR can or should solve. The PR only claims per-location de-collapse, and that holds. In practice different panels sit at different paths, so it rarely bites — just noting it so it isn't mistaken for a full per-detection guarantee.

Minor observation: the two normalizers are now near-identical (only the finding-type id and `info.name` vs `info.description` differ). A shared helper would remove the duplication, but the nuclei plugins are deliberately separate, so this is consistent with the codebase — take it or leave it.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `test_nuclei_exposed_panels.py` + `test_nuclei_take_over.py` | 4 passed |
| boefjes | edge cases (unparseable matched-at, same-URL collapse, port-in-URL) | behaviour characterised |
| boefjes | live raw → Octopoes | 2 distinct Findings persisted |

The new tests are consistent and cover the right axes: two distinct match URLs → two distinct finding keys (asserts `len({f.ooi}) == 2`), and the hostname fallback when `matched-at` is absent (asserts no URL emitted). Both boefjes get their own test file.

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Live raw → normalizer → Octopoes | OK | 2 Findings + 2 URLs persisted, no collapse |
| Log analysis | OK | normalizer `Done`, no errors |

### Verdict
**APPROVED**

Correct, minimal fix for a real data-loss bug, and the right pattern for the finding-collapse problem. Verified end-to-end that the de-collapse survives persistence, and that the `matched-at` fallback handles unparseable input without crashing. The residual same-URL collapse is a pre-existing model limitation, not a regression. Please rebase on main before merging.

The `-t` flag for these two boefjes is correctly left to #5286 — no overlap here.

PR #5297, opened 2026-08-05
Extract Software from Shodan banners

brennodewinter:
## What

`kat_shodan` emitted ports and CVE findings but never read the detected `product`/`version`, so **no `Software` was recorded** and CVEs could not be bound to the software running on a port.

## Fix

Yield `Software` + `SoftwareInstance` per `data[]` entry from `product`/`version`, carrying the CPE 2.3 string (`cpe23[0]`) when present. `normalizer.json` `produces` now lists `Software`/`SoftwareInstance`.

## Test

`test_shodan.py`: a banner with product/version yields Software bound to the port; a banner without a product yields none. Existing tests unchanged.

## Scope / follow-up

The `data[].ssl` certificate block remains unparsed — that is a larger, separate change (validity/subject/SAN extraction, like the Censys certificate handling) left as follow-up.

### DIFF
--- a/boefjes/boefjes/plugins/kat_shodan/normalize.py
+++ b/boefjes/boefjes/plugins/kat_shodan/normalize.py
@@ -6,6 +6,7 @@ from boefjes.normalizer_models import NormalizerOutput
 from octopoes.models import Reference
 from octopoes.models.ooi.findings import CVEFindingType, Finding
 from octopoes.models.ooi.network import IPPort, PortState, Protocol
+from octopoes.models.ooi.software import Software, SoftwareInstance
 
 
 def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
@@ -34,3 +35,13 @@ def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
                     f = Finding(finding_type=ft.reference, ooi=ip_port.reference)
                     yield ft
                     yield f
+
+            # Shodan reports the detected product/version per banner. It was
+            # dropped entirely, so no Software was recorded and CVEs could not
+            # be bound to the software on a port.
+            product = scan.get("product")
+            if product:
+                cpe23 = scan.get("cpe23") or []
+                software = Software(name=product, version=scan.get("version"), cpe=cpe23[0] if cpe23 else None)
+                yield software
+                yield SoftwareInstance(ooi=ip_port.reference, software=software.reference)
--- a/boefjes/boefjes/plugins/kat_shodan/normalizer.json
+++ b/boefjes/boefjes/plugins/kat_shodan/normalizer.json
@@ -8,6 +8,8 @@
   "produces": [
     "Finding",
     "IPPort",
-    "CVEFindingType"
+    "CVEFindingType",
+    "Software",
+    "SoftwareInstance"
   ]
 }
--- a/boefjes/tests/plugins/test_shodan.py
+++ b/boefjes/tests/plugins/test_shodan.py
@@ -4,6 +4,7 @@ from boefjes.plugins.kat_shodan.normalize import run
 from octopoes.models import Reference
 from octopoes.models.ooi.findings import CVEFindingType, Finding
 from octopoes.models.ooi.network import IPPort, PortState, Protocol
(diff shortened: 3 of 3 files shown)


hasecon (approved):
## QA Review — round 1

### Setup
- Branch: `feat/shodan-software-extraction` @ `9d3a3f4`
- Verified via unit tests, edge-case probing, and an end-to-end run through the live stack

### CI/CD Status

### Code Review Findings

No blocking issues. Same well-scoped gap as the nmap change (#5293): Shodan reported the detected product/version per banner and it was thrown away, so no `Software` was recorded and CVEs couldn't bind to what's running on a port.

**Verified end-to-end on the live stack.** Pushed a Shodan-shaped raw (`boefje/shodan`) through Bytes → normalizer → Octopoes:

```
IPPort|internet|1.2.3.4|tcp|22
IPPort|internet|1.2.3.4|tcp|443
IPPort|internet|1.2.3.4|udp|53
SoftwareInstance|IPPort|…|tcp|22|Software|OpenSSH|8.4p1|cpe:2.3:a:openbsd:openssh:8.4p1:*:*:*:*:*:*:*
SoftwareInstance|IPPort|…|tcp|443|Software|nginx|1.18.0|
Software|OpenSSH|8.4p1|cpe:2.3:a:openbsd:openssh:8.4p1:*:*:*:*:*:*:*
Software|nginx|1.18.0|
```

That covers all three branches: CPE present (OpenSSH), version-only no CPE (nginx), and no product (udp/53 → no Software). The `kat_shodan_normalize` task logged `Done` with no errors.

**Binding to `IPPort` is the right granularity here.** Unlike nmap (which produces `IPService` and binds there), this normalizer only produces `IPPort`, so the port-level object is the correct anchor. `cpe23[0] if cpe23 else None` safely handles the missing-CPE case, and empty-string `product` is correctly treated as falsy (no Software emitted).

**Two pre-existing, platform-level observations — NOT defects in this PR, and not blocking:**

| Severity | Note |
|---|---|
| Low (pre-existing) | A non-string `version` (int/float) makes `Software(version=…)` raise `ValidationError`, and it propagates out of `run()` → the whole scan's normalization fails, not just that banner. |
| Low (pre-existing) | A `|` in banner-derived `product`/`version` corrupts the OOI primary key (`Software|Foo|evil|1.0|`) — the same issue tracked in #5299. |

I checked whether the missing `str()` on `version` was an inconsistency to raise against this PR, and it is **not**: every existing Software-producing normalizer passes the version through unmodified — `kat_censys` (`version=sw["version"]`), `kat_binaryedge` (`version=app["version"]`, `version=dialect`). So this PR follows the established codebase pattern. Shodan documents `version` as a string, and I don't have a Shodan key value to confirm a numeric one ever occurs live, so I reproduced the crash only with a synthetic int. Both notes are really arguments for one central fix — defensive coercion + separator-escaping at the `Software`/`natural_key` layer — covering all ~7 call sites at once, rather than anything this diff should carry. Happy to fold the version-coercion angle into #5299.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `tests/plugins/test_shodan.py` | 7 passed |
| boefjes | edge cases (int/float version, pipe, None, empty product) | behaviour characterised |
| boefjes | live raw → Octopoes | Software + SoftwareInstance bound to IPPort |

The test file is consistent and covers the right axes: CPE present (asserts name/version/cpe and the port binding) and product absent (asserts no Software). The four pre-existing tests still pass, so the CVE/port behaviour is unchanged.

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Live raw → normalizer → Octopoes | OK | 2 Software, 2 SoftwareInstance, 3 IPPort |
| Log analysis | OK | `kat_shodan_normalize` done, no errors |

### Verdict
**APPROVED**

Clean, minimal fix that closes a real gap and mirrors the nmap approach, verified end-to-end on a live stack. The two robustness notes are pre-existing and codebase-wide (this PR is consistent with every other Software-producing normalizer), best handled centrally via #5299 rather than here. Please rebase on main before merging.

The `data[].ssl` → certificate follow-up you flagged is sensible as a separate scope — worth a tracking issue.

PR #5298, opened 2026-08-05
Enrich LeakIX leak findings with proof, source, scale and auth state

brennodewinter:
## What

`handle_leak` discarded most of what a LeakIX event carries:

- `event["summary"]` — the actual evidence (e.g. a full Apache server-status dump with internal vhosts/IPs) — was **never read**, so `Finding.proof` stayed empty;
- the `"Plugin ="` label **interpolated the OOI reference** (`Plugin = "IPPort|internet|..."`) instead of the plugin name;
- the dataset **scale** (`rows`/`files`/`size`/`collections`) was dropped;
- an **unauthenticated** service (`service.credentials.noauth`) was not flagged.

## Fix

- store `event["summary"]` as `Finding.proof`, bounded to 2 KiB;
- label with `event["event_source"]` (the LeakIX plugin / what kind of leak it is);
- append the dataset scale counts to the description;
- flag `noauth` services with "No authentication required.".

No severity logic changed. The regenerated golden output (`leakix-example.com-output.txt`) reflects the richer descriptions and proof; the output **count is unchanged (170)**.

## Test

New `test_leak_findings_are_enriched` asserts proof is populated, the Plugin label no longer contains an OOI reference, and unauthenticated services are flagged. Existing `test_output` passes against the regenerated snapshot.

## Scope / follow-up

The `event["ssl"]` block (→ `X509Certificate`) remains unparsed — a larger, separate change left as follow-up (marked by the existing `# TODO: ssh, ssl`).


hasecon (approved):
## QA Review — round 1

### Setup
- Branch: `feat/leakix-finding-enrichment` @ `222a622`
- Verified three ways: committed fixture, structural diff against `main`, and **live LeakIX data fetched during this review**

### CI/CD Status

### Code Review Findings

No issues. This enriches leak findings without touching severity logic, and I verified that claim rather than trusting the regenerated snapshot.

**The golden snapshot is legitimate, not a rubber-stamp.** A regenerated `str(output)` snapshot only proves internal consistency, so I ran the normalizer on `main` and on this branch against the same fixture and compared a deterministic fingerprint of every OOI:

```
main: n=170  STRUCT_SHA=f2b07932030a3077  finding_pk_SHA=cc19b8826da03df3
pr:   n=170  STRUCT_SHA=f2b07932030a3077  finding_pk_SHA=cc19b8826da03df3
```

Identical. The exact same 170 OOIs with identical primary keys — only `Finding.description` and `Finding.proof` changed, and neither is part of a Finding's primary key (`natural_key = ooi|finding_type`). So the enrichment cannot create, drop, or re-bind a finding. The count-unchanged-at-170 claim holds structurally, not just numerically.

**Both bugs the PR claims to fix are real, confirmed on `main`:**
- The `Plugin = "…"` label on `main` interpolated the OOI reference: I reproduced `Plugin = "IPAddressV4|internet|104.171.169.158".` (2 findings in the fixture). On this branch that count is 0, and the label now reads the plugin name from `event_source`.
- `proof` was empty on every finding on `main` (0/18); on this branch all 18 carry the event summary.

**Live verification with a real API key** (thanks for providing it). Fetched fresh data through the actual `main.py` → normalizer path:
- `1.1.1.1`: 600 live service events → 4200 OOIs, no error — confirms the API path and normalizer handle real responses at volume.
- A live `scope=leak` search: 20 real leak events → 20 findings, **all 20 with proof**, e.g. `Plugin = "MalwareHttpPlugin".` with a real server-response dump as proof. This is the enrichment working on data that never touched the committed fixture.

**Behaviour-preserving details I checked:**
- Extracting `dataset` into a local is a pure refactor — same four `.get()` reads, no semantic change.
- The label change has one subtle behavioural shift worth being aware of: on `main`, a finding with no software *always* got a `Plugin = "…"` entry; now it only gets one when `event_source` is truthy. That is the intended fix (an empty/garbage label is worse than none), and the description simply omits the label otherwise. Not a regression.
- `Finding.ooi = software_ooi.reference if software_ooi else event_ooi_reference` is unchanged from `main` — the binding was not touched, only the description/proof.
- `truncate_proof` handles `None`/empty and strips before measuring. The longest fixture proof came out at 2049 chars (2048 + the `…`), so the bound works.

One nit, non-blocking: `MAX_PROOF_LENGTH = 2048` is described as "2 KiB" but bounds by **characters**, not bytes — a multibyte-heavy summary could exceed 2 KiB. `Finding.proof` has no length constraint in the model (`str | None`), so nothing breaks; the intent (bound the evidence) is met. Only relevant if a hard byte budget is ever needed downstream.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `tests/plugins/test_leakix.py` | 4 passed |
| boefjes | structural diff `main` vs PR | identical 170-OOI set (SHA match) |
| boefjes | live LeakIX (`1.1.1.1`, `scope=leak`) | enrichment confirmed on real data |

The new `test_leak_findings_are_enriched` asserts the right things: proof is populated, no `Plugin = "…"` label contains a `|` (the OOI-ref bug), and the fixture's unauthenticated services are flagged. `test_output`, `test_strict_mode_*` and `test_permissive_mode_*` all still pass against the regenerated snapshot.

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Live boefje → normalizer | OK | 20 live findings, all with proof |
| Log analysis | OK | no leakix/normalizer errors; only known startup noise |

### Verdict
**APPROVED**

Clean, well-scoped enrichment: real evidence in `proof`, a correct plugin label instead of a leaked OOI reference, dataset scale, and a `noauth` flag — with the OOI set provably untouched. Verified on the committed fixture, structurally against `main`, and on live LeakIX data. Please rebase on main before merging.

The `event["ssl"]` → `X509Certificate` follow-up you flagged is a sensible separate scope — worth a tracking issue so it isn't lost.

PR #5300, opened 2026-08-05
Make kat_cve_2023_34039 actually run its SSH-key check

brennodewinter:
## What

The boefje **never detected the vulnerability**, for two independent, silent reasons:

1. `os.walk("keys")` resolved against the process **CWD** (`/app/boefje` in the OCI image), not the module. It walked a non-existent directory, found no keys, and always returned `"No known keys allowed"`.
2. The 22 bundled keys are committed **world-readable** (mode `0755`) and Docker `COPY` preserves that. OpenSSH refuses a private key with group/world permissions and exits `255`, so every key was skipped anyway.

Either bug alone makes the check a guaranteed false negative.

## Fix

- resolve the keys via `Path(__file__).parent / "keys"`;
- hand `ssh` a **0600 temp copy** of each key (`NamedTemporaryFile` creates it `0600`), so OpenSSH accepts it — without mutating the bundled files or their git mode.

The bundled keys are VMware's published static keys (exploit material for this CVE), not secrets, so their world-readable mode in git is not itself a leak; the runtime copy just satisfies OpenSSH.

## Test

`boefjes/tests/plugins/test_cve_2023_34039.py` (mocks `subprocess.run`): non-ssh service is descheduled; `KEYS_DIR` resolves to the 22 bundled keys; an accepted key yields the `openkat/finding` `CVE-2023-34039` output and the key handed to `ssh` is a `0600` copy; all-rejected yields "No known keys allowed".

### DIFF
--- a/boefjes/boefjes/plugins/kat_cve_2023_34039/main.py
+++ b/boefjes/boefjes/plugins/kat_cve_2023_34039/main.py
@@ -14,8 +14,14 @@ https://summoning.team/blog/vmware-vrealize-network-insight-ssh-key-rce-cve-2023
 
 """
 
-import os
 import subprocess
+import tempfile
+from pathlib import Path
+
+# Resolve the keys relative to this module. os.walk("keys") used the process
+# CWD (/app/boefje in the OCI image) instead, so it walked a non-existent
+# directory, silently found no keys and always reported "not vulnerable".
+KEYS_DIR = Path(__file__).parent / "keys"
 
 
 def run(boefje_meta: dict) -> list[tuple[set, str | bytes]]:
@@ -27,13 +33,19 @@ def run(boefje_meta: dict) -> list[tuple[set, str | bytes]]:
     ip = ip_port["address"]["address"]
     port = ip_port["port"]
 
-    for root, dirs, files in os.walk("keys"):
-        for file in files:
-            key_file = str(os.path.join(root, file))
+    for key_file in sorted(p for p in KEYS_DIR.rglob("*") if p.is_file()):
+        # The keys are committed world-readable (mode 0755) and Docker COPY
+        # preserves that mode; OpenSSH refuses a private key with group/world
+        # permissions and exits 255, so every key was skipped. Copy the key to a
+        # private (0600) temp file — NamedTemporaryFile creates it with 0600 —
+        # so ssh will actually use it.
+        with tempfile.NamedTemporaryFile("wb") as tmp:
+            tmp.write(key_file.read_bytes())
+            tmp.flush()
             ssh_command = [
                 "ssh",
                 "-i",
-                key_file,
+                tmp.name,
                 "support@" + ip,
                 "-p",
                 str(port),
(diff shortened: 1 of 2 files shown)


hasecon (approved):
## QA Review — round 1

### Setup
- Branch: `fix/cve-2023-34039-keys` @ `df346ed`
- Both failure modes reproduced from the running worker container, and the fix verified against real OpenSSH

### CI/CD Status
18 checks, all completed: **17 success + 1 skipped** (`add-debian-packages-to-release`, expected outside a release). All boefjes-relevant checks green: `Tests (3.10–3.13)`, `pre-commit`, `CodeQL`, `build (…, boefjes)`, `create_container_image`.

### Code Review Findings

No issues. This is a well-diagnosed fix for a check that was a guaranteed false negative, and I verified **both** root causes independently rather than taking the description on trust.

**Cause 1 — wrong working directory. Confirmed from the real container.** The generic image sets `WORKDIR /app/boefje` (`images/base.Dockerfile:7`) — actually `/app/boefjes` in the running container — while the keys live at `boefjes/plugins/kat_cve_2023_34039/keys`. Ran both the old and new lookup from inside `openkat-reviev-boefje-1`:

```
cwd: /app/boefjes
os.walk("keys")  -> 0 files      <- main: always "No known keys allowed"
KEYS_DIR         -> 22 keys      <- this PR: correct
```

`Path(__file__).parent / "keys"` is the right fix and is the same idiom the codebase already uses (e.g. `retire_js` loads `retirejs.json` this way).

**Cause 2 — world-readable key mode. Confirmed against real OpenSSH.** All 22 keys are committed mode `0755` (`git ls-files -s` → all `100755`), and `COPY --chown=nonroot` changes owner but not mode. I ran the actual `ssh` client from the boefje image against a live sshd:

```
0755 original -> WARNING: UNPROTECTED PRIVATE KEY FILE!
                 Permissions 0755 for '…/id_rsa_vnera_keypair_6.0.0_platform' are too open.
                 -> key ignored, exit 255   (the silent skip)

0600 tmp copy -> no permission complaint; proceeds to
                 "Permission denied (publickey)"   (normal auth path)
```

So the `NamedTemporaryFile` 0600 copy genuinely unblocks OpenSSH. The 255-vs-auth distinction is the whole point: on `main` every key died at the permission check before authentication was ever attempted. Against a real vulnerable vRNI target the 0600 copy would reach `returncode 0`.

**Behaviour-preserving details I checked:**
- The keys are nested one level deep (`keys/vrni-6.x.0/…`). `os.walk` recursed into those; `rglob("*")` does too, so all 22 across 11 version dirs are still covered — not just a flat top level.
- Moving `return` inside the `with tempfile` block is correct: the context manager closes on return, and the finding is built from data already in hand. The bundled key path is only *read*, never mutated, so the git mode is untouched (matches the PR's stated intent).
- Narrowing the `try` to just `subprocess.run` is fine — the only post-`run` calls are `relative_to`/f-string on a path that by construction lives under `KEYS_DIR`, so they cannot raise `ValueError`.
- `key_id = key_file.relative_to(KEYS_DIR)` is a small improvement: the proof line now reads `vrni-6.0.0/id_rsa_…` instead of a bare filename, so the report says which vRNI version's key was accepted.

**Security note (not a concern):** the bundled keys are VMware's published static keys — the exploit material for this CVE, not secrets — so their world-readable git mode is not a leak, and the runtime 0600 copy is purely to satisfy OpenSSH. The PR's reasoning here is sound.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `tests/plugins/test_cve_2023_34039.py` (new) | 4 passed |
| boefjes | CWD bug reproduced in `openkat-reviev-boefje-1` | 0 vs 22 keys |
| boefjes | 0755-reject / 0600-accept against live sshd | confirmed |

The test file is internally consistent and covers the right axes: non-ssh deschedule, `KEYS_DIR` resolves to exactly 22 keys, the accepted-key path asserts the key handed to `ssh` is mode `0600`, and all-rejected yields "No known keys allowed".

One fair limitation, not blocking: the tests mock `subprocess.run`, so they assert the *copy is 0600* but cannot assert *OpenSSH accepts 0600 and rejects 0755* — that is exactly the behaviour I verified out-of-band above, so the fix is covered end-to-end between the unit tests and this review. Worth knowing for future edits that the unit suite alone would not catch a regression in the permission behaviour itself.

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| CWD bug / fix | reproduced | 0 keys (old) vs 22 keys (new) from the real worker |
| Log analysis | OK | no CVE-34039 errors; only known startup noise |

No vulnerable vRNI target exists in the test org, so I could not observe a positive finding end-to-end — but the two mechanisms that made a positive finding *impossible* on `main` are both proven fixed, which is the reviewable claim here.

### Verdict
**APPROVED**

Correctly diagnosed double false-negative, cleanly fixed, and the first real test coverage this boefje has had. Both root causes verified independently — the CWD miss from inside the running worker, the permission reject/accept against a live OpenSSH. Please rebase on main before merging.

Issue #5299, opened 2026-08-05
Unsanitised banner-derived strings can corrupt OOI primary keys (Software.version)


hasecon:
## Summary

Several normalizers pass remote-controlled strings straight into `Software.name` / `Software.version`. Because `|` is the primary-key separator and nothing escapes it, a value containing a pipe silently produces a malformed reference, and rendering that reference then raises.

Found while reviewing #5293 (nmap CPE → Software). **It is not caused by that PR** — the pattern predates it in at least six normalizers. Filing separately so it can be fixed at the right layer instead of being patched per-plugin.

## Reproduction

A service whose version string contains a pipe:

```python
Software(name="Foo", version="1.0|evil")
# reference: Software|Foo|1.0|evil|

SoftwareInstance(ooi=<IPService ref>, software=<that ref>)
# reference: SoftwareInstance|IPService|internet|1.2.3.4|tcp|8080|http|Software|Foo|1.0|evil|

SoftwareInstance.format_reference_human_readable(ref)
# -> TypeNotFound
```

`format_reference_human_readable` splits on `|` and reconstructs the software reference from `parts[-4:]`, so an extra separator shifts the boundary and the reconstructed type no longer resolves.

## Affected sources

These already feed unescaped, remote-influenceable values into `Software`:

- `kat_dns_version/normalize.py:34` — derived from the `version.bind` TXT record, i.e. entirely whatever the remote nameserver returns
- `kat_nikto/normalize.py:65,83` — `match.group(1)` / `match.group(2)` from nikto's message text
- `kat_leakix/normalize.py:167,169`
- `kat_censys/normalize.py:68,70`
- `kat_binaryedge/{services,service_identification,http_web,databases}/normalize.py` — several sites using raw `product` / `version`
- `kat_nmap_tcp/normalize.py:38` — the fallback branch added in #5293, using nmap's `product` / `version` (which come from banner regex captures in `nmap-service-probes`)

CPE-based paths are lower risk, since CPE strings are colon-delimited and generated by the scanner rather than copied from a banner.

## Impact

Confirmed: a malformed reference is constructed and `format_reference_human_readable` raises `TypeNotFound`. What I have **not** traced is every call site of that function, so I would not want to overstate the blast radius — the plausible consequence is a rendering failure on object list/detail pages containing such an OOI, but that should be verified before deciding priority. There is no data loss and no code execution here; the concern is data integrity and UI robustness.

[...]

PR #5302, opened 2026-08-05
Extract WordPress software inventory and all vulnerabilities from wpscan

brennodewinter:
## What

The `kat_wpscan` normalizer only turned **CVE-tagged alerts** (via the `wpscan_out_parse` text-rendering library) into findings, dropping everything else the scan reports: the **WordPress core version**, the **plugin/theme inventory**, and every **vulnerability without a CVE** (most WPVulnDB entries).

## Fix

Parse the wpscan JSON directly and, per component (core, main theme, themes, plugins):

- yield `Software` + `SoftwareInstance` from the detected version (the inventory);
- yield a `CVEFindingType`/`Finding` for each vulnerability carrying a CVE;
- yield a **`WPVulnFindingType`**/`Finding` for each non-CVE vulnerability — each wpscan vulnerability carries a unique WPVulnDB id, so N vulnerabilities yield N distinct findings (no collapse).

Findings bind to the component's **`Software`** (per @underdarknl's review: a vulnerability is a property of the software, not of the software at a specific location; where it is installed is inferred via the graph). When no version is detected there is no `Software`, so findings fall back to the scanned URL.

`WPVulnFindingType` is hydrated directly from the wpscan scan JSON (title, cvss score, source URL), which already carries the WPVulnDB metadata — no separate lookup boefje needed for non-enterprise users.

## Round-2 review changes (this commit)

Addresses the round-2 review feedback from @underdarknl and @hasecon:

1. **Findings bind to `Software`, not `SoftwareInstance`** (Jan's inline review). Reverses the round-1 nikto-precedent steer.
2. **Non-CVE vulnerabilities get distinct findings via `WPVulnFindingType`**. Each wpscan vulnerability has a unique WPVulnDB id, so N vulnerabilities → N distinct finding types → N distinct Findings. No more collapse. Supersedes the round-1 aggregation approach.
3. **Outdated detection removed from the normalizer**. Per Jan's direction, staleness should be derived from the graph by a BIT (scan intensity + org-configurable severity can sit in between; "out of date != newest"). The `KAT-OUTDATED-SOFTWARE` emission is removed from the normalizer.

`KAT-VULNERABLE-SOFTWARE-VERSION` is removed from the finding-type catalog (superseded by `WPVulnFindingType`). The CVE-as-facts route Jan called "excellent" is kept.

## Follow-up issues (out of scope for this PR)

- **Outdated-BIT + latest-version data source.** Deriving staleness from the graph requires a new OOI type to link "latest version" data to `Software`, plus a boefje that queries the wpscan.com API v3 for the latest version per slug. The BIT then compares installed vs latest and emits `KAT-OUTDATED-SOFTWARE` with org-configurable severity. This touches 23+ files across octopoes+rocky (new OOI type registration) and is architecturally larger than this PR.
- **WPVulnDB lookup boefje.** The wpscan.com API v3 `/vulnerabilities/{id}` endpoint is enterprise-only. For non-enterprise users the scan JSON already carries the full vulnerability metadata (title, cvss, references), so `WPVulnFindingType` is hydrated directly. An enterprise-only lookup boefje (modelled on `kat_cve_finding_types`) could refresh scores over time, but is not needed for the core functionality.

## Decisions worth a look
- **`WPVulnFindingType` risk level** is derived from the cvss score in the scan JSON (same mapping as the CVE finding-type normalizer). When wpscan reports no cvss, the severity is `UNKNOWN`.
- `wpscan-out-parse` is **no longer used** by the code but is still pinned in `boefjes/pyproject.toml`; dropping it requires regenerating the hashed lock files, so it's left as a separate follow-up.
- `interesting_findings` (directory listing, config backups, XML-RPC, …) are still not surfaced — a sensible next step, out of scope here.

## Test

`boefjes/tests/plugins/test_wpscan.py`: 11 tests covering core+plugin Software extraction, CVE findings bound to Software, non-CVE vulnerabilities as distinct WPVulnFindingType findings, cvss hydration, URL fallback without version, and no nested SoftwareInstance. All pass.

`octopoes/tests/test_query.py`: updated for the new `WPVulnFindingType` in the finding-type count query.


hasecon (changes requested):
## QA Review — round 1

### Setup
- Branch: `feat/wpscan-software-inventory` @ `ca4a4cd`
- Test org: `setup_test_org --code dvwa --hostname hasecon.nl` — OK
- Real wpscan output captured from `https://hasecon.nl/` (authorised test target) and pushed through the live pipeline

### CI/CD Status
18 checks, all completed: **17 success + 1 skipped** (`add-debian-packages-to-release`). Nothing red. The findings below are **not** visible to CI, for the reason explained under finding 1.

### Code Review Findings

The direction here is right and the gap is real — dropping the inventory and every non-CVE vulnerability was a genuine loss. Three issues in the execution, the first of which is a production-path bug.

| Severity | Finding |
|---|---|
| **High** | Nested `SoftwareInstance` created on the real input; findings bind to a second-order object |
| **High** | Multiple non-CVE vulnerabilities on one component collapse into a single `Finding` |
| Medium | WordPress core is never flagged outdated — core uses `status`, not `outdated` |

---

#### 1. The normalizer receives a `SoftwareInstance`, not a URL — so it nests one inside another

`kat_wpscan/boefje.json` declares `"consumes": ["SoftwareInstance"]`, and `local/runner.py:33` passes `boefje_meta.arguments["input"]` — the serialised **input OOI** — to the normalizer. `main.py` confirms this: it reads `input_["software"]["name"]` and `input_["ooi"]["netloc"]["name"]`.

So at runtime `input_ooi["primary_key"]` is a `SoftwareInstance` reference, and the variable named `url_reference` is not a URL. On `main` that was harmless — binding a `Finding` to a `SoftwareInstance` is legitimate and matches the nikto precedent, so the misleading name had no consequence. This PR newly does `SoftwareInstance(ooi=url_reference, ...)`, which nests.

Verified end-to-end on the running stack, feeding the **real** hasecon.nl wpscan JSON through Bytes → scheduler → normalizer → Octopoes:

```
SoftwareInstance|HostnameHTTPURL|https|internet|hasecon.nl|443|/|Software|WordPress|6.5|          <- the input OOI
SoftwareInstance|SoftwareInstance|HostnameHTTPURL|https|internet|hasecon.nl|443|/|Software|WordPress|6.5||Software|socrates-v5|5.046|
```

The theme is recorded as installed *on the WordPress SoftwareInstance* rather than on the website. With a detectable core version it is worse still: the normalizer emits a `SoftwareInstance` nested inside the very `SoftwareInstance` that triggered the scan. All findings then bind to that second-order object.

**Why CI is green anyway:** every test passes `input_ooi = {"primary_key": "HostnameHTTPURL|..."}`, a shape that never occurs in production. Feeding the same payload the two ways side by side:

```
input HostnameHTTPURL (tests) -> SoftwareInstance|HostnameHTTPURL|...|Software|WordPress|6.5|
input SoftwareInstance (real) -> SoftwareInstance|SoftwareInstance|HostnameHTTPURL|...|Software|WordPress|6.5||Software|WordPress|6.5|
```

Suggested fix — resolve the URL from the input when it is a `SoftwareInstance`:

```python
if input_ooi.get("object_type") == "SoftwareInstance":
    url_reference = Reference.from_str(input_ooi["ooi"]["primary_key"])
else:
    url_reference = Reference.from_str(input_ooi["primary_key"])
```

Please also switch the tests to the production input shape — otherwise the same class of bug stays invisible.

#### 2. Non-CVE vulnerabilities still get dropped, just one layer later

`Finding.natural_key` is `f"{ooi}|{finding_type.natural_key}"` ([findings.py:141](https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/blob/main/octopoes/octopoes/models/ooi/findings.py#L141)). Every non-CVE vulnerability on a component yields the same `ooi` and the same `KAT-VULNERABLE-SOFTWARE-VERSION` type, so they share one reference. Three distinct vulnerabilities:

```
Finding objects yielded: 3
  ...|KAT-VULNERABLE-SOFTWARE-VERSION   description='Bug A - auth bypass'
  ...|KAT-VULNERABLE-SOFTWARE-VERSION   description='Bug B - SQL injection'
  ...|KAT-VULNERABLE-SOFTWARE-VERSION   description='Bug C - RCE'
unique references: 1        <- what Octopoes stores
```

The CVE path is fine (3 vulnerabilities → 3 unique references), because each CVE is its own finding type.

The practical effect is that a plugin with eight WPVulnDB entries shows one finding carrying one arbitrary title, with nothing indicating the other seven exist. That directly contradicts *"so non-CVE vulnerabilities are no longer silently dropped"* — they are still dropped, just at the OOI layer instead of the parser.

To be fair to the design: OpenKAT's `Finding` model **cannot** represent N distinct findings of one type on one OOI, so this is not a matter of a missing line. The cheapest honest fix is to aggregate — emit one finding whose description carries the count and all titles, e.g. `"3 known vulnerabilities without CVE: Bug A; Bug B; Bug C"`. That keeps the information visible and makes the collapse deliberate rather than accidental.

#### 3. WordPress core will never be flagged outdated

`_handle_component("WordPress", ..., False, url_reference)` passes `outdated=False` unconditionally, so `KAT-OUTDATED-SOFTWARE` can only ever fire for plugins and themes.

That is not just an omission — core reports its staleness through a **different field**. From `wpscan_out_parse`, whose docstring cites wpscan's own `app/views/json/wp_version/version.erb`:

```python
# wordpress_version.py  (core)
self.status = self.data.get("status", None)
if self.status in ["insecure", "outdated"]:   # -> "Outdated"

# wp_item.py  (plugins/themes)
self.outdated = self.data.get("outdated", None)
```

Confirmed against real output: hasecon.nl's `main_theme` carries `"outdated": false`, while the core `version` object carries `status`. The repo's own pre-existing test fixture even sets `"status": "insecure"` — so the suite scans an insecure core on every run and never notices the missing finding.

```python
core_outdated = (core.get("status") or "") in ("insecure", "outdated")
```

An outdated core is arguably the most important thing wpscan reports, so this one seems worth fixing before merge.

---

Two things I checked that turned out **fine**, noting them so they don't get re-litigated:
- The missing `source` key on the new finding type looked like a convention break (81 of 83 entries have it), but the only other entry without one is `KAT-OUTDATED-SOFTWARE` — its closest sibling. Consistent as-is.
- `plugins` / `themes` are absent entirely from real wpscan JSON when nothing is found (not empty containers), and `version` is `null` when core is undetectable. `data.get(...) or {}` handles all three; no crash, verified against the live output.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `tests/plugins/test_wpscan.py` | 7 passed |
| boefjes | real wpscan run against hasecon.nl | OK, JSON captured |
| boefjes | production-shaped input | **nested SoftwareInstance reproduced** |

The four pre-existing CVE tests do still pass against the rewritten parser, which is good evidence the refactor preserves the old behaviour. `_wpscan_payload` is a pre-existing helper, so the new tests reuse it correctly.

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Stack build | OK | exit 0 |
| Real scan → pipeline | OK | inventory extracted, `Software\|socrates-v5\|5.046\|` created |
| Log analysis | OK | no normalizer errors |
| Nested instance | **reproduced** | see finding 1 |

Worth knowing for anyone testing this: without `WP_SCAN_API` set, wpscan returns `"No WPScan API Token given, as a result vulnerability data has not been output"` and every `vulnerabilities` array is empty. So findings 2 and 3 only bite deployments **with** a token — which is exactly the audience this feature is for, but it does mean a default local stack will not surface them.

### Verdict
**CHANGES REQUESTED**

Finding 1 is the blocker: it writes a malformed object graph on the real input path, and the test fixtures hide it. Findings 2 and 3 mean the two headline behaviours — surfacing non-CVE vulnerabilities and flagging outdated components — do not fully work as described.

None of this is an argument against the change itself. The inventory extraction is valuable, binding findings to the component rather than the bare URL is the right model, and adding `KAT-VULNERABLE-SOFTWARE-VERSION` fills a genuine gap. Happy to re-review quickly once the input handling is fixed.

brennodewinter:
Thanks @hasecon — all three findings addressed in d6e017e.

**1. Nested SoftwareInstance (high).** The normalizer now resolves the URL from `input_ooi["ooi"]["primary_key"]` when the input is a `SoftwareInstance`, instead of nesting. Tests switched to the production `SoftwareInstance` input shape so this path is exercised the way the scheduler runs it; added `test_wpscan_no_nested_software_instance` asserting every emitted `SoftwareInstance.ooi` is the URL, never the input `SoftwareInstance`.

**2. Non-CVE collapse (high).** Acknowledged that `Finding.natural_key` is `ooi|finding_type`, so N non-CVE vulnerabilities on one component collapse to one reference in Octopoes — the model can't represent N. Aggregated deliberately: one `KAT-VULNERABLE-SOFTWARE-VERSION` finding whose description carries the count and all titles, e.g. `"3 known vulnerabilities without CVE: Bug A; Bug B; Bug C"`. Added `test_wpscan_aggregates_non_cve_vulnerabilities`.

**3. Core never flagged outdated (medium).** Core now flagged from `status` (`"insecure"`/`"outdated"`), not the `outdated` boolean. Added `test_wpscan_flags_outdated_core_via_status`. (The pre-existing fixture's `"status": "insecure"` was removed from the shared payload helper so the no-vulnerabilities test stays meaningful; the new test sets it explicitly.)

All 10 tests in `tests/plugins/test_wpscan.py` pass. Ready for re-review.

boefjes/boefjes/plugins/kat_wpscan/normalize.py
```diff
@@ -1,29 +1,104 @@
...
+    finding_ooi = url_reference
+    if version_number:
+        software = Software(name=name, version=str(version_number))
+        yield software
+        software_instance = SoftwareInstance(ooi=url_reference, software=software.reference)
+        yield software_instance
+        finding_ooi = software_instance.reference
+
+        if outdated:
+            outdated_type = KATFindingType(id="KAT-OUTDATED-SOFTWARE")
```
underdarknl: I'm not a fan of doing this in the boefje, or normalizing step.
This can and should be deducted via the available info in the graph via a BIT. Eg, this normalizer identifies software, and software instances.
Another boefje collects info about newer versions, and known vurlns in these versions. By doing it in that way, we can vary the scan intensity for each, eg, what wordpress are you running is daily. What cve's does my graph's software contain is every hour. This also allows us to put a config in between where the bit sets the intensity of the issue based on an orga's preferences. 
Also, out of date != newest, which might be just fine. No longer supported, or from an now untrusted source are just as valid, but can also be better identified via bits.

boefjes/boefjes/plugins/kat_wpscan/normalize.py
```diff
@@ -1,29 +1,104 @@
...
+    non_cve_titles: list[str] = []
+    for vulnerability in vulnerabilities or []:
+        title = vulnerability.get("title")
+        cves = (vulnerability.get("references") or {}).get("cve") or []
+        if cves:
+            for cve in cves:
+                cve_id = str(cve).upper()
+                if not cve_id.startswith("CVE-"):
+                    cve_id = f"CVE-{cve_id}"
+                cve_type = CVEFindingType(id=cve_id)
```
underdarknl, line 113: Adding CVE's via this route is just adding Facts without any warning levels  produced/introduced by OpenKAT. This is excellent imho.

boefjes/boefjes/plugins/kat_wpscan/normalize.py
```diff
@@ -1,29 +1,104 @@
...
+
+    if non_cve_titles:
+        vuln_type = KATFindingType(id="KAT-VULNERABLE-SOFTWARE-VERSION")
+        yield vuln_type
+        count = len(non_cve_titles)
+        plural = "y" if count == 1 else "ies"
+        yield Finding(
+            finding_type=vuln_type.reference,
+            ooi=finding_ooi,
+            description=f"{count} known vulnerabilit{plural} without CVE: " + "; ".join(non_cve_titles),
```
underdarknl: If multiple vurlns exists, add multiple Findings, each bound to a their respective WPVulnFinding (which needs creating).
If the WPVulnDB is accessible via a boefje (like the CVE databases are), we should also introduce such a lookup boefje to gather the factual related finding information from that database on a regular basis to make sure we track changing Finding scores etc. See CVE lookup boefje as an example.

boefjes/boefjes/plugins/kat_wpscan/normalize.py
```diff
@@ -1,29 +1,104 @@
...
+        title = vulnerability.get("title")
+        cves = (vulnerability.get("references") or {}).get("cve") or []
+        if cves:
+            for cve in cves:
+                cve_id = str(cve).upper()
+                if not cve_id.startswith("CVE-"):
+                    cve_id = f"CVE-{cve_id}"
+                cve_type = CVEFindingType(id=cve_id)
+                yield cve_type
+                yield Finding(finding_type=cve_type.reference, ooi=finding_ooi, description=title)
```
underdarknl: Bind the CVE to the software instead.
The CVE is a property of the software, not of the software on that specific location, as we can infer that via the graph.

```suggestion
                yield Finding(finding_type=cve_type.reference, ooi=software.reference, description=title)
```

boefjes/boefjes/plugins/kat_wpscan/normalize.py
```diff
@@ -1,29 +1,104 @@
...
+            non_cve_titles.append(title)
+
+    if non_cve_titles:
+        vuln_type = KATFindingType(id="KAT-VULNERABLE-SOFTWARE-VERSION")
+        yield vuln_type
+        count = len(non_cve_titles)
+        plural = "y" if count == 1 else "ies"
+        yield Finding(
+            finding_type=vuln_type.reference,
+            ooi=finding_ooi,
```
underdarknl: Same here, bind the software, not the softwareinstance.

underdarknl (changes requested):

hasecon (changes requested):
## QA Review — round 2 (head `7d65247b`)

### My round-1 findings — all three addressed (verified in code, not from the summary)

| # | Round-1 finding | Status at head |
|---|-----------------|----------------|
| 1 | High — nested `SoftwareInstance` on the real (SoftwareInstance) input | ✅ Fixed. `run()` resolves the URL from `input_ooi["ooi"]["primary_key"]` when `object_type == "SoftwareInstance"`; tests switched to the production SoftwareInstance shape; `test_wpscan_no_nested_software_instance` asserts every emitted `SoftwareInstance.ooi` is the URL. |
| 2 | High — non-CVE vulns collapse to one Finding | ✅ Addressed by aggregation (one finding carrying the count + all titles) + `test_wpscan_aggregates_non_cve_vulnerabilities`. **But see Jan's #3 below — he wants a different approach.** |
| 3 | Medium — core never flagged outdated | ✅ Fixed via `status in ("insecure","outdated")` + `test_wpscan_flags_outdated_core_via_status`. **But see Jan's #1 below.** |

Good work — the nesting bug is genuinely gone and the tests now run the production input shape.

### The governing feedback this round is Jan's — please align on it before mine

@underdarknl posted **CHANGES_REQUESTED** on this exact head with five inline comments. He's the boefjes code owner, so his direction is the one to follow, and on one point it **supersedes my round-1 advice** — flagging that explicitly so you don't get whipsawed:

- **Bind findings to `Software`, not `SoftwareInstance`** ([normalize.py:91](https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/pull/5302#discussion_r3773411566) + :102). My round-1 review told you to bind to the `SoftwareInstance` (nikto precedent); Jan is overriding that — the CVE/vuln is a property of the *software*, and *where* it's installed is inferred via the graph. This matches the CVE-on-`Software` direction in #5321. **Concrete near-term change:** set `finding_ooi = software.reference` (not `software_instance.reference`) when the version is known. Keep yielding the `SoftwareInstance` for the inventory link; just don't bind findings to it. (Edge case: when `version_number` is falsy there's no `Software`, so the URL fallback stays.)
- **Outdated detection belongs in a BIT, not the normalizer** ([:68](https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/pull/5302#discussion_r3730493908)). Jan wants staleness derived from the graph by a bit (so scan intensity + org-configurable severity can sit in between, and because "out of date ≠ newest": no-longer-supported / untrusted-source are separate signals). That's the `KAT-OUTDATED-SOFTWARE` logic (my finding 3) moving out of the normalizer.
- **Non-CVE vulns → per-vuln Findings, each bound to a new `WPVulnFinding` type, fed by a WPVulnDB lookup boefje** ([:103](https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/pull/5302#discussion_r3734319113)), modelled on the CVE-lookup boefje. This is a different (and larger) answer than the aggregation I suggested for finding 2.
- **Positive:** Jan called the CVE-as-facts route ("adding Facts without OpenKAT-introduced warning levels") "excellent" ([:89](https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/pull/5302#discussion_r3730499124)) — keep that.

The Software-binding is a small change and clearly in scope. The BIT-for-outdated and WPVulnFinding+lookup-boefje are architectural and bigger than this PR — worth a quick scoping check with Jan on what folds in here vs. what becomes follow-up issues, rather than guessing.

### CI/CD Status — `pre-commit` is RED (blocking)

`pre-commit` failed (job 94125276055). Reproduced locally — it's mechanical, just not run before pushing:
- `ruff E501` line-too-long: `normalize.py:28` (the `yield from _handle_component("WordPress", …)` line, 121>120) and `test_wpscan.py:156` (121>120);
- `ruff-format` reformats 2 files; `pyupgrade` rewrites `test_wpscan.py`.

`mypy`, `codespell`, `prettier` pass. Fix: `pre-commit run --all-files` and commit the result. (Component `Tests (3.10–3.13)` are green — the 10 wpscan tests pass.)

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `Tests (3.10–3.13)` (CI) | ✅ green (10 wpscan tests) |
| any | `pre-commit` (CI + local) | ❌ **red** — ruff E501 / format / pyupgrade |

### Verdict
**CHANGES REQUESTED**

Two reasons, neither about the value of the change (the inventory + non-CVE surfacing is a real gap you're filling): (1) Jan's code-owner review is open with a clear redirection — most concretely, bind findings to `Software` not `SoftwareInstance`, which reverses my round-1 steer; and (2) `pre-commit` is red. I'd suggest: do the Software-binding + green up pre-commit now, and settle the BIT / WPVulnFinding scope with Jan (fold-in vs follow-up) so the bigger architectural pieces don't get built twice.

hasecon (changes requested):
## QA Review — round 3 (head `d15cfb9`)

### Setup
- Branch: `feat/wpscan-software-inventory` @ `d15cfb9`
- E2E: production-serialized wpscan raw injected via Bytes → scheduler → normalizer (details under finding 1)

### Previous-round findings — status at this head

| Round-2 item | Status |
|---|---|
| `pre-commit` red (ruff E501 / format / pyupgrade) | ✅ Fixed — green in CI |
| Jan: bind findings to `Software`, not `SoftwareInstance` | ✅ Implemented (`finding_ooi = software.reference`) — but see **finding 2**: this needs a report-side counterpart |
| Jan: outdated-detection belongs in a BIT, not the normalizer | ✅ Implemented — `KAT-OUTDATED-SOFTWARE` emission removed, follow-up documented |
| Jan: per-vuln Findings bound to a new WPVuln finding type | ✅ Implemented — `WPVulnFindingType`, N vulns → N distinct findings, verified by test asserting distinct natural keys. The lookup boefje was deliberately skipped (wpscan.com `/vulnerabilities/{id}` is enterprise-only; type is hydrated from the scan JSON instead) — @underdarknl please confirm that trade-off is acceptable |
| R1: nested `SoftwareInstance` on real input | ⚠️ The nesting is gone, but the replacement input-handling introduced finding 1 |

`WPVulnFindingType` registration is complete where it matters: `octopoes/models/types.py` union, `rocky/tools/ooi_helpers.py`, `normalizer.json` produces, `octopoes/tests/test_query.py`. Severity mapping matches `kat_cve_finding_types` exactly.

### Code Review Findings

| Severity | Finding |
|---|---|
| **Critical** | Normalizer crashes with `KeyError: 'primary_key'` on **every** production run — proven live end-to-end |
| **High** | Findings bound to `Software` are invisible in the Findings/Vulnerability Reports and crisis room — regression vs main; needs Jan's sequencing call on #5321 |
| Medium | Timthumb vulnerabilities silently dropped vs main |
| Medium | `_cvss_score` discards the scalar `"cvss": "7.5"` shape and an unguarded `float()` can abort the whole run |
| Low | Several robustness/consistency notes (grouped below) |

---

#### 1. Critical — `input_ooi["ooi"]["primary_key"]` does not exist in production; every real run crashes

First, mea culpa: the crashing snippet is **the exact fix I suggested in round 1**. My suggestion was wrong, the author implemented it in good faith, and the tests (also per my round-1 advice) were built around the same wrong shape. This round I generated the input with the real serializer instead of assuming.

Production builds the normalizer input as `boefje_meta.arguments["input"] = ooi.serialize()` ([scheduler_client.py:125](https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/blob/main/boefjes/boefjes/clients/scheduler_client.py#L125)), and `serialize()` renders every `Reference` field as its **tokenized natural-key tree** (`_serialize_value` → `value.tokenized.root`). Generated in the octopoes container on this branch:

```json
{"object_type": "SoftwareInstance",
 "primary_key": "SoftwareInstance|HostnameHTTPURL|https|internet|demo.dvwa.cloud|443|/|Software|WordPress|6.5|",
 "ooi": {"scheme": "https", "netloc": {"network": {"name": "internet"}, "name": "demo.dvwa.cloud"}, "port": "443", "path": "/"},
 "software": {"name": "WordPress", "version": "6.5", "cpe": ""}}
```

There is no `primary_key` inside `ooi` — and the boefje's own `main.py` already proves this shape by reading `input_["ooi"]["netloc"]["name"]`. Since `object_type` **is** present, the `SoftwareInstance` branch is always entered and line 45 raises.

Proven end-to-end on the running stack: injected this exact meta + a wpscan JSON into Bytes → normalizer task picked up → task status **failed**:

```
File "/app/boefjes/boefjes/plugins/kat_wpscan/normalize.py", line 45, in run
    url_reference = Reference.from_str(input_ooi["ooi"]["primary_key"])
KeyError: 'primary_key'
```

Zero Software, zero findings, for 100% of production wpscan runs. CI is green because the test fixture fabricates `{"ooi": {"primary_key": URL_PK}}` — a shape production never produces (the fixture comment claiming "production shape" is unfortunately false). The fixture's `URL_PK` is additionally in the wrong order: real PKs are scheme-first (`HostnameHTTPURL|https|internet|example.com|443|/`), the fixture has `internet|https`.

**Suggested fix** — rebuild the URL reference from the token tree (safe to assume `HostnameHTTPURL`: `main.py` deschedules any input without `netloc.name`, so raws only exist for hostname URLs):

```python
if input_ooi.get("object_type") == "SoftwareInstance":
    token = input_ooi["ooi"]
    url_reference = Reference.from_str(
        f"HostnameHTTPURL|{token['scheme']}|{token['netloc']['network']['name']}"
        f"|{token['netloc']['name']}|{token['port']}|{token['path']}"
    )
else:
    url_reference = Reference.from_str(input_ooi["primary_key"])
```

One more wrinkle: the `else` branch is also reachable with a **SoftwareInstance PK**. When `arguments["input"]` is absent, the runner falls back to `{"primary_key": boefje_meta.input_ooi}` (`local/runner.py:33-36`) — no `object_type`, so the else branch treats the SoftwareInstance PK as the URL and quietly resurrects the round-1 nesting bug. Since this boefje consumes only SoftwareInstance, consider handling `primary_key.startswith("SoftwareInstance|")` in the else branch too (strip the leading `SoftwareInstance|` and the trailing `|Software|name|version|cpe` segments), or failing loudly.

#### 2. High — findings bound to `Software` disappear from reports and the crisis room

This is a consequence of the (correct, per Jan) design direction, not of this PR's code — but it needs an explicit decision before merge, because as-is it **regresses report content vs main**:

- `Software` has `_traversable = False` ([software.py:16](https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/blob/main/octopoes/octopoes/models/ooi/software.py#L16)). The Findings Report collects findings via `get_tree(reference, depth=9, types={Finding})` rooted at the asset — traversal prunes non-traversable children, so it never continues *through* `Software` to its findings. Findings Report, Aggregate Report, and crisis-room dashboards therefore won't show any wpscan finding.
- The Vulnerability Report's hard-coded query fetches findings **on** `SoftwareInstance` (`…<ooi [is SoftwareInstance].<ooi [is Finding]`, [vulnerability_report/report.py:142-162](https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/blob/main/rocky/reports/report_types/vulnerability_report/report.py#L142-L162)) — one hop short of `Software`. Note the `retire_js` bit binds CVE findings to `SoftwareInstance`, which this query *does* catch — wpscan now diverges from the binding that reports actually support.
- On main, wpscan findings were bound to the (traversable) URL and **did** appear in these reports. The org-wide findings list still shows them (it's severity-based, not tree-based), so users will see the inconsistency.
- Side effect worth naming: `Software|WordPress|4.9.8|` is org-global, so two scanned sites on the same version share one `Software` node and one `Finding` per CVE — "which site is affected" requires the graph inference that #5321 describes but nothing implements yet.

@underdarknl — your call on sequencing: (a) accept the temporary report gap and land report-side `Software` traversal as the #5321 follow-up, (b) fold a report-side query fix into this PR, or (c) bind to `SoftwareInstance` (retire_js pattern) until #5321 lands. I'd not merge before that's an explicit decision, given this is the same silent-drop class as #5202/#5088.

#### 3. Medium — timthumb vulnerabilities dropped vs main

The old `WPScanJsonParser.get_alerts()` walked all components including `timthumbs`, whose CVE reference lines the old line-parser matched — so a vulnerable `timthumb.php` (classic CVE-2011-4106 RCE) produced a Finding on main. The new code reads only `version` / `main_theme` / `themes` / `plugins`; `data["timthumbs"]` is never touched, so that finding silently disappears on rescan. (I checked the other old alert sources — `interesting_findings`, config backups, users, etc. produced no CVE-bearing alerts, so timthumbs is the only actual coverage loss.) Either handle `data.get("timthumbs")` in the same `_handle_component` loop or state the drop explicitly in the PR.

#### 4. Medium — `_cvss_score` shape and robustness

Two related issues in the same function:
- The library this PR replaces types `vulnerability["cvss"]` as a **scalar string** (`finding.py: self.cvss: str = self.data.get("cvss")`). A payload with `"cvss": "7.5"` fails the `isinstance(cvss, dict)` check and is silently discarded → `risk_severity=UNKNOWN`, and a CVSS-7.5 vulnerability drops out of severity-filtered views. The dict shape you handle is real too — handle both.
- `float(score)` is unguarded: `"cvss": {"score": "N/A"}` raises `ValueError` mid-generator and aborts the **entire** normalizer run — all inventory and all findings for the site lost. Same whole-run-abort class we've been fixing elsewhere (#5289, #3349). Wrap in `try/except (TypeError, ValueError): return None`.

#### 5. Grouped Low / notes

- **Scan-level dead-end (pre-existing, but decisive for QA):** `SoftwareInstance` can inherit at most L1 (`software.py:29`) while `wp-scan` requires L2 — so the boefje is never auto-scheduled; only a manual scan or an explicit declared profile on the SoftwareInstance reaches this code. Worth a line in the PR/docs. It's also why the Critical stays invisible in casual functional testing.
- **CVE-path asymmetry:** the non-CVE path skips entries without a `title`, the CVE path passes a possibly-`None` title as `Finding.description`. Consider a fallback description (e.g. the WPVulnDB URL) for both, and a debug log on the silent skip.
- **`source=f"https://wpscan.com/vulnerability/{vuln_id}"`** interpolates the raw id into an `AnyUrl`-validated field; a malformed id (spaces/control chars) raises `ValidationError` and aborts the run — same class as 4b, same cheap guard.
- **Slug collision:** merging `main_theme`/`themes`/`plugins` into one slug-keyed dict means a theme and plugin sharing a slug (allowed across the two wordpress.org namespaces) last-writer-wins; the loser's inventory and vulnerabilities are dropped.
- **Severity table is now a third verbatim copy** of `SEVERITY_SCORE_LOOKUP` (`kat_cve_finding_types`, `kat_snyk_finding_types`, now here). Convention today, drift risk tomorrow — fine to leave, but a shared helper would be better.
- `wpscan-out-parse` staying pinned is already flagged in the PR description as a lockfile follow-up — agreed, fine.
- `rocky/tools/ooi_helpers.py:get_finding_type_from_finding` (updated here) has no callers anywhere; the update is harmless. FYI its hand-built union also predates this PR in missing `ADRFindingType` — replacing it with `ConcreteFindingTypeType` from `octopoes.models.types` would delete the whole edit site.
- Mixed-version deployments (boefjes from this PR, octopoes on main) would reject `WPVulnFindingType` — normal monorepo caveat, no action.

### Suggested regression test

Drop-in for `boefjes/tests/plugins/test_wpscan.py` — builds the input through the **real serializer**, so the fixture can never drift from production again:

```python
def test_wpscan_input_shape_matches_production_serializer():
    """Input built via SoftwareInstance.serialize(), exactly like scheduler_client.py:125."""
    from pydantic_core import to_jsonable_python

    from octopoes.models.ooi.dns.zone import Hostname
    from octopoes.models.ooi.network import Network
    from octopoes.models.ooi.web import HostnameHTTPURL, WebScheme

    network = Network(name="internet")
    url = HostnameHTTPURL(
        network=network.reference,
        scheme=WebScheme.HTTPS,
        netloc=Hostname(network=network.reference, name="example.com").reference,
        port=443,
        path="/",
    )
    software = Software(name="WordPress", version="4.9.8")
    instance = SoftwareInstance(ooi=url.reference, software=software.reference)
    serialized = to_jsonable_python(instance.serialize())

    results = list(run(serialized, _wpscan_payload([])))

    instances = [si for si in results if isinstance(si, SoftwareInstance)]
    assert instances
    assert all(str(si.ooi) == str(url.reference) for si in instances)
```

Verified against this head: it crashes with `KeyError: 'primary_key'` (red on the current branch, pinning finding 1) and passes once the input handling is fixed. I'd also swap the module-level `input_ooi` fixture to this serializer-generated dict so all 11 tests run the production shape, and fix `URL_PK` to the real scheme-first order.

### CI/CD Status

All checks green **except** `Octopoes tests / test (3.12)` (integration step). Investigated per protocol:
- Same commit: 3.10 / 3.11 / 3.13 legs of the same workflow are green; the annotation carries only "exit code 2".
- Reproduced locally on this branch: `cd octopoes && make itest` → **37 passed, 2 xfailed, exit 0**.
- Main history: same workflow had a one-off failure on 2026-07-29 with no related change.

Conclusion: environment flake, not a PR signal — but please **re-run the job** so the merge doesn't happen over a red check. Branch is also `behind` main and needs updating.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| boefjes | `tests/plugins/test_wpscan.py` (in Docker image) | ✅ 11 passed |
| octopoes | `make itest` (local, reproducing red CI leg) | ✅ 37 passed, 2 xfailed |
| rocky | `make utest` | ✅ 471 passed, 1 skipped |
| any | `pre-commit` (CI) | ✅ green |

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Stack build | OK | `make clean && make kat`, 13/13 running |
| Log analysis | OK | no new errors beyond the reproduced wpscan crash |
| E2E: production-serialized input → normalizer | ❌ **task `failed`**, `KeyError: 'primary_key'`, zero OOIs persisted (finding 1) |

### Verdict
**CHANGES REQUESTED**

The direction is now fully aligned with Jan's review and the WPVulnFindingType design is sound — this is close. Two things stand between it and approval: (1) the input-shape crash, which breaks 100% of production runs and is masked by fixtures that don't match the serializer (fix + the serializer-based regression test above); and (2) an explicit decision from Jan on the Software-binding report gap, since as-is the change trades main's report visibility for a graph inference that doesn't exist yet. Findings 3–4 are worth folding into the same round; the Lows are take-or-leave. Happy to re-review quickly.

hasecon:
## Addendum to round 3 — verified against a real WordPress site

I re-ran the whole thing against a live WordPress target (hasecon.nl, our own site, authorised) using the boefje's own image and argv, instead of the hand-built payload from the review above. Three updates: finding 1 is confirmed on real data, **my suggested fix was incomplete** (correction below), and **finding 3 was wrong in practice** — I'm retracting it.

### Finding 1 — confirmed end-to-end on a real scan

Real scan, boefje image `docker.underdark.nl/librekat/openkat-wp-scan:latest` (wpscan 4.1.0), exact argv from `main.py`:

```
wpscan --url https://hasecon.nl/ --format json --plugins-version-detection aggressive
```

Real output detected `version.number: 7.0.4` (status `latest`) and `main_theme: socrates-v5 5.046`. Pushed through Bytes → scheduler → normalizer with a production `boefje_meta` whose `arguments["input"]` came from `SoftwareInstance.serialize()`:

```
status: failed
File "/app/boefjes/boefjes/plugins/kat_wpscan/normalize.py", line 45, in run
    url_reference = Reference.from_str(input_ooi["ooi"]["primary_key"])
KeyError: 'primary_key'
```

Octopoes afterwards: `Software` count 2, both from unrelated boefjes (`cloudflare`, `go-ipfs`) — no `WordPress`, no `socrates-v5`. So on a real WordPress site the inventory this PR exists to add is not created at all.

### Correction to my suggested fix — please use this one instead

The snippet in my review (rebuild the URL from the token tree) works for the serialized shape but leaves the runner-fallback path broken, which is precisely the round-1 nesting bug. I tested both candidates against the real serialized input:

| Candidate | serialized input | runner fallback (`{"primary_key": ...}`) |
|---|---|---|
| Rebuild from token tree (what I posted) | ✅ `HostnameHTTPURL\|https\|internet\|hasecon.nl\|443\|/` | ❌ returns the SoftwareInstance PK → nesting returns |
| **Strip the SoftwareInstance wrapper off `primary_key`** | ✅ correct | ✅ correct |

```python
pk = input_ooi["primary_key"]
if pk.startswith("SoftwareInstance|"):
    # trailing Software natural key is 4 segments: Software|name|version|cpe
    url_reference = Reference.from_str("|".join(pk[len("SoftwareInstance|"):].split("|")[:-4]))
else:
    url_reference = Reference.from_str(pk)
```

It needs no `object_type` branch at all, because `primary_key` is present in both shapes. Verified by patching a copy of the module in the normalizer container and running it against the real hasecon.nl JSON:

```
results: 4
    Software         | Software|WordPress|7.0.4|
    SoftwareInstance | SoftwareInstance|HostnameHTTPURL|https|internet|hasecon.nl|443|/|Software|WordPress|7.0.4|
    Software         | Software|socrates-v5|5.046|
    SoftwareInstance | SoftwareInstance|HostnameHTTPURL|https|internet|hasecon.nl|443|/|Software|socrates-v5|5.046|
```

Flat, both instances bound to the URL, no nesting — the graph round 1 asked for. (Patch applied only to a scratch copy inside the container; the PR branch is untouched.)

### Finding 3 (timthumb) — retracted

I based that on the old library walking `timthumbs`. Checking the real boefje behaviour: `tt` (Timthumbs) only appears with an explicit `--enumerate tt`, which `main.py` never passes — and the real scan output confirms it, with `plugins` and `themes` present-but-empty while `timthumbs` is absent entirely. So wpscan never emits timthumb data for this boefje and nothing is actually lost. Ignore that finding; it would only matter if `--enumerate tt` were added later.

### Finding 4 (cvss shape) — still open, but honestly untested

Without `WP_SCAN_API` set, wpscan reports no vulnerability data at all (`vuln_api` carries the "no token" notice and every `vulnerabilities` array is empty), so I could not observe a real `cvss` value either way. That finding therefore still rests on the replaced library typing `cvss` as a scalar string, not on data I saw. The unguarded `float()` half stands on its own regardless — a non-numeric score aborts the entire run, which is worth the two-line guard whatever the shape turns out to be.

Verdict is unchanged (**CHANGES REQUESTED**), now on stronger evidence for finding 1 and with one fewer finding to act on. Findings 1 and 2 remain the blockers.

brennodewinter:
## Round-3 review — all findings addressed in ed42c4b

Thanks @hasecon for the thorough round-3 review and the addendum with the corrected fix. All findings addressed:

### 1. Critical — input-shape crash → fixed

Replaced the broken `input_ooi["ooi"]["primary_key"]` lookup with the corrected approach from your addendum: strip the `SoftwareInstance|` wrapper and trailing `Software|name|version|cpe` segments off `primary_key` directly. This works for **both** the serialized shape (token tree, no nested `primary_key`) and the runner-fallback shape (`{"primary_key": ...}`), because `primary_key` is present in both.

```python
pk = input_ooi["primary_key"]
if pk.startswith("SoftwareInstance|"):
    url_reference = Reference.from_str("|".join(pk[len("SoftwareInstance|"):].split("|")[:-4]))
else:
    url_reference = Reference.from_str(pk)
```

This mirrors the existing `SoftwareInstance.format_reference_human_readable` pattern (`parts[0:-4]`).

**Test fixture rebuilt via the real serializer** — `SoftwareInstance.serialize()` → `to_jsonable_python` → `json.loads(json.dumps(...))`, exactly like `scheduler_client.py:125`. The fixture can never drift from production again. `URL_PK` is now correctly scheme-first (`HostnameHTTPURL|https|internet|example.com|443|/`). Added:
- `test_wpscan_input_shape_matches_production_serializer` — pins the fixture to the production shape
- `test_wpscan_runner_fallback_input_shape` — covers the `runner.py:36` fallback path

### 2. High — findings on Software invisible in reports → fixed

Added query-based collection paths through `SoftwareInstance.software` to supplement the tree-based findings:

- **Findings Report**: after `get_tree`, queries `_SOFTWARE_FINDING_PATHS[reference.class_type]` for findings on Software and merges them. Paths cover all four input types (Hostname, IPAddressV4, IPAddressV6, URL).
- **Vulnerability Report**: added `software_finding_types` and `software_findings` queries alongside the existing `software_instance_*` queries, and merges them into the results.

The query path is `...<ooi [is SoftwareInstance].software.<ooi [is Finding]` — verified that all six path variants parse correctly via `Path.parse()`.

### 3. Medium — `_cvss_score` shape and robustness → fixed

- Now handles scalar string `cvss` (e.g. `"7.5"`) — the shape the replaced library types it as
- `float()` is guarded with `try/except (TypeError, ValueError)` → returns `None` instead of aborting the entire run

Added `test_wpscan_cvss_scalar_string` and `test_wpscan_cvss_non_numeric_does_not_abort`.

### 4. Lows → addressed

- CVE finding `description` falls back to `cve_id` when `title` is `None` (was passing `None` as description)
- `vuln_id` is `str(...).strip()` before interpolation into the source URL

### Finding 3 (timthumb) — retracted by you, no action needed

### Test results

All 15 tests in `test_wpscan.py` pass. `pre-commit run --all-files` is green (ruff, ruff-format, mypy, codespell, pyupgrade, django-upgrade, vulture, prettier). `test_finding_type_count_query` in `octopoes/tests/test_query.py` passes (includes `WPVulnFindingType`).

brennodewinter:
Rebased on `main` and fixed the CI failures (rocky report tests). Root cause: the new Software-finding query paths in `findings_report` and `vulnerability_report` call `query()`/`query_many()` on `MockOctopoesAPIConnector`, but the mock never initialized `self.queries` — causing `AttributeError` — and raised `KeyError` for paths not pre-seeded by individual tests. Fixed in the shared mock: initialize `self.queries = {}` in `__init__` and use `.get()` with empty fallbacks so tests that don't pre-seed a path get `[]` instead of crashing.

All 476 rocky tests, 15 wpscan tests, and 20 octopoes query tests pass locally.

brennodewinter:
Re-review request for @hasecon and @underdarknl.

All round-3 findings have been addressed (commit `5bf5c6dd6`) and the branch is rebased on main with all CI green:

| Finding | Status |
|---|---|
| **Critical** — input-shape crash (`KeyError: 'primary_key'`) | Fixed — strip `SoftwareInstance|` wrapper off `primary_key` directly (works for both serialized and runner-fallback shapes) |
| **High** — findings on Software invisible in reports | Fixed — added query paths through `SoftwareInstance.software` in both findings_report and vulnerability_report |
| **Medium** — cvss scalar string + unguarded `float()` | Fixed — handles both dict and string cvss, `float()` guarded with try/except |
| **Medium** — timthumb | Retracted by @hasecon (wpscan never emits timthumb data without `--enumerate tt`) |
| Lows — CVE description fallback, vuln_id strip | Fixed |

@underdarknl — two open questions from the round-2/round-3 reviews that need your call:
1. **WPVulnFindingType hydration**: the wpscan.com `/vulnerabilities/{id}` endpoint is enterprise-only, so `WPVulnFindingType` is hydrated directly from the scan JSON (title, cvss, source URL) instead of via a separate lookup boefje. Is that trade-off acceptable?
2. **Report-side Software traversal**: hasecon's finding 2 asked for your sequencing call on #5321 — option (a) accept temporary report gap, (b) fold report-side fix into this PR, or (c) bind to SoftwareInstance. I went with (b): the report query paths are in this PR. Please confirm that's the right call.

underdarknl:
1 is acceptable.
2 has been added to the release  and should be merged/fixed before releasing to avoid the gap.

hasecon (changes requested):
## QA Review — round 4

### Setup
- Branch: `feat/wpscan-software-inventory` @ `5b5d9b0`
- Stack: `make clean && make kat` — 13/13 containers up, Rocky HTTP 200
- Test org: `setup_test_org --code dvwa` — OK, plus a hand-seeded WordPress graph (Hostname → URL → SoftwareInstance → Software ← Finding) to exercise the new report paths against real XTDB

### Round-3 findings — all fixed, each verified against the head

| # | Round-3 finding | Status | How I verified |
|---|---|---|---|
| 1 | **Critical** — `KeyError: 'primary_key'` on production input | ✅ Fixed | Ran the normalizer against input built by the **real serializer** (`SoftwareInstance.serialize()` → `to_jsonable_python`) and against the runner fallback. Both shapes now produce identical, correct output — no crash. |
| 2 | **High** — findings on `Software` invisible in reports | ✅ Fixed | See the section below — verified live, end to end. |
| 3 | **Medium** — cvss scalar string + unguarded `float()` | ✅ Fixed | Both shapes handled; thresholds are byte-identical to `kat_cve_finding_types/normalize.py:12-15`, so the "same mapping as the CVE normalizer" claim holds. |
| 4 | Medium — timthumb | Retracted by me in round 3 | No action |
| 5 | Lows — CVE description fallback, `vuln_id` strip | ✅ Fixed | Present in the diff |

On finding 1, the corrected fix is the one from my round-3 addendum, and it behaves as advertised on both input shapes:

```
production input has nested primary_key? False
top-level primary_key: SoftwareInstance|HostnameHTTPURL|https|internet|demo.dvwa.cloud|443|/|Software|WordPress|6.5.2|

=== SERIALIZED (production) ===        === RUNNER FALLBACK ===
Software          Software|WordPress|6.5.2|          (identical output)
SoftwareInstance  SoftwareInstance|HostnameHTTPURL|https|internet|demo.dvwa.cloud|443|/|Software|WordPress|6.5.2|
WPVulnFindingType WPVulnFindingType|aaaa-1111
Finding           Finding|Software|WordPress|6.5.2||aaaa-1111
```

Flat, bound to the URL, no nesting. The rebuilt fixture (real serializer + `test_wpscan_input_shape_matches_production_serializer`) is exactly the right guard against this class recurring.

### Graph placement — verified against real XTDB

This was the open question from round 3, and since @underdarknl has now said it must be fixed before release, I checked it properly rather than reading the paths.

**All six query paths parse, and all six return the Software-bound finding:**

```
=== do the 6 query paths PARSE? ===          === do they RETURN the finding? ===
OK  findings_report Hostname                 Hostname  -> ['Finding|Software|WordPress|6.5.2||aaaa-1111']
OK  findings_report URL                      URL       -> ['Finding|Software|WordPress|6.5.2||aaaa-1111']
OK  findings_report IPv4                     IPv4      -> ['Finding|Software|WordPress|6.5.2||aaaa-1111']
OK  findings_report IPv6                     IPAddress -> [(IPAddressV4|…, 'Finding|Software|WordPress|6.5.2||aaaa-1111')]
OK  vuln_report findings                     finding_type -> [(IPAddressV4|…, 'WPVulnFindingType|aaaa-1111')]
OK  vuln_report finding_types
```

**The control confirms the gap was real** — the same hostname through the old route:

```
get_tree(depth=9, types={Finding}) -> EMPTY (pruned at Software)
```

**And end to end through the actual changed code:**

```
FindingsReport.generate_data("Hostname|internet|wp.dvwa.cloud")
  total_occurrences: 3
   finding_type: aaaa-1111 (high)
      occurrence: Finding|Software|WordPress|6.5.2||aaaa-1111
```

Consumer coverage is better than the diff suggests, which is worth stating explicitly: the aggregate report imports and reuses `FindingsReport` and `VulnerabilityReport` (`aggregate_organisation_report/report.py:9,17`), and the crisis room's Findings Dashboard runs the "Crisis Room Aggregate Report" recipe whose `ooi_types` are exactly `Hostname`/`IPAddressV4`/`IPAddressV6`/`URL` — the same four keys as `_SOFTWARE_FINDING_PATHS`. So fixing these two classes covers the aggregate report and the crisis room too. The findings list, Finding detail, `WPVulnFindingType` detail, Software detail and object list all render 200 with the new type, and the WPVuln finding is visible in the findings list.

I also checked the two remaining `SnykFindingType` registration sites that `WPVulnFindingType` is *not* in — `finding_add.py` and `kat_finding_normalizer` — and both are prefix maps (`"CVE"`, `"SNYK"`). WPVulnDB ids are prefix-less UUIDs, so those omissions are correct, not gaps. Pipe-escaping in a hostile plugin slug is handled centrally (`Software(name="evil|plugin")` → `Software|evil%7Cplugin|1.0|`), so the #5299 class doesn't reopen here.

### Code Review Findings

| # | Sev | File | Issue |
|---|-----|------|-------|
| 1 | Medium | `findings_report/report.py:71`, `vulnerability_report/report.py:196-201` | The new Software paths fan in, so one vulnerability is counted once per SoftwareInstance — report totals inflate |
| 2 | Medium | `rocky/tests/conftest.py:1356`, both reports | The report fix has no test, and the conftest change makes its absence silent |

#### 1. (Medium) The Software fan-in duplicates findings in both reports

`SoftwareInstance.software` is many-to-one, so every path that reaches the same `Software` yields the same `Finding` again, and neither report deduplicates before counting.

Two SoftwareInstances for one Software (a site served on both http and https — very common):

```
query returned 2 row(s):
   Finding|Software|WordPress|6.5.2||aaaa-1111
   Finding|Software|WordPress|6.5.2||aaaa-1111
distinct findings: 1
```

Which lands directly in the rendered report — `total_occurrences` went 3 → 4 and the occurrence is listed twice:

```
total_occurrences: 4
 finding_type: aaaa-1111 (high)
    occurrence: Finding|Software|WordPress|6.5.2||aaaa-1111
    occurrence: Finding|Software|WordPress|6.5.2||aaaa-1111
```

The shared-hosting case is worse, and it is the normal case for WordPress. Five sites on one IP all running 6.5.2 share one `Software` OOI, so the vulnerability report multiplies by five:

```
=== shared hosting: 5 sites on 1 IP, same WordPress 6.5.2 ===
  vulnerability_report rows for ONE real finding: 5
  distinct findings: 1
```

The counts are what a reader trusts in a security report, so a single CVE reading as five occurrences is a real correctness problem. Note this is a property of the new `.software.` hop specifically: the pre-existing `software_instance_findings` path doesn't fan in, because findings there hang off each instance individually.

Findings report:

```python
software_path = _SOFTWARE_FINDING_PATHS.get(reference.class_type)
if software_path:
    seen = {f.reference for f in findings}
    for f in self.octopoes_api_connector.query(software_path, valid_time, reference):
        if isinstance(f, Finding) and f.reference not in seen:
            seen.add(f.reference)
            findings.append(f)
```

Vulnerability report — dedup the merged lists before the loops:

```python
findings = list(
    {
        f.reference: f
        for f in port_findings.get(ip, [])
        + software_instance_findings.get(ip, [])
        + software_findings.get(ip, [])
    }.values()
)
```

and the same `{ft.reference: ft}` treatment for `found_finding_types`.

#### 2. (Medium) The report fix has no test, and the conftest change hides that

The conftest change is the right call for robustness, but it also means an unseeded path returns `[]` rather than raising — so the new code runs against an empty result in every existing test and nothing asserts it does anything.

I ran the revert test: replaced the whole `_SOFTWARE_FINDING_PATHS` block in `findings_report` with `pass` and re-ran the report suite (scratch copy inside the test container; the branch was untouched and restored afterwards):

```
87 passed, 1 skipped, 10 warnings in 12.29s
```

The entire fix can be deleted without a single test failing. Given @underdarknl has put this on the release gate, it should be pinned by something. A drop-in in the existing style of `tests/reports/test_findings_report.py`, which also pins the dedup from finding 1:

```python
def test_findings_report_includes_findings_bound_to_software(
    mock_octopoes_api_connector, valid_time, hostname, tree_data_no_findings, finding_types
):
    """Findings bound to Software are invisible to get_tree (Software._traversable
    is False), so they must arrive via the Software query path — and only once,
    however many SoftwareInstances reach the same Software."""
    software_finding = Finding(
        finding_type=finding_types[0].reference,
        ooi=Reference.from_str("Software|WordPress|6.5.2|"),
        description="Non-CVE bug",
    )
    mock_octopoes_api_connector.oois = {finding_types[0].reference: finding_types[0]}
    mock_octopoes_api_connector.tree = {hostname.reference: ReferenceTree.model_validate(tree_data_no_findings)}
    mock_octopoes_api_connector.queries = {
        "Hostname.<netloc [is HostnameHTTPURL].<ooi [is SoftwareInstance].software.<ooi [is Finding]": {
            hostname.reference: [software_finding, software_finding],  # http + https instances
        }
    }

    data = FindingsReport(mock_octopoes_api_connector).generate_data(str(hostname.reference), valid_time)

    assert data["summary"]["total_finding_types"] == 1
    assert data["summary"]["total_occurrences"] == 1   # deduplicated, not 2
```

I verified the *behaviour* it pins live (both the retrieval and the duplication, outputs above); I did not execute this exact test file, so treat the code itself as a draft — the fixture names come from the existing tests in that module. On today's branch the first assertion passes and the `total_occurrences == 1` assertion fails with `2`, which is the dedup from finding 1.

### Test Results

| Component | Test | Result |
|-----------|------|--------|
| boefjes | `tests/plugins/test_wpscan.py` | ✅ 15 passed |
| rocky | `make utest` (Docker) | ✅ 478 passed, 1 skipped |
| rocky | `tests/reports` with the fix reverted | ⚠️ 87 passed — the fix pins nothing (finding 2) |
| octopoes | `tests/test_query.py` | ✅ 20 passed |
| CI | all 39 checks incl. `rtest`, `pre-commit`, `CodeQL` | ✅ green |

### Functional Test

| Check | Status | Details |
|-------|--------|---------|
| Stack build | ✅ | 13/13 containers, no restart loops |
| Log analysis | ✅ | 5 tracebacks, all the known `rocky_worker` startup scheduler race |
| Normalizer on production input | ✅ | No crash; correct flat placement on both input shapes |
| Query paths parse | ✅ | All 6 |
| Query paths return findings | ✅ | All 6, against real XTDB |
| `get_tree` control | ✅ | EMPTY — confirms the gap the fix closes |
| FindingsReport end to end | ✅ | Software-bound finding appears with correct severity |
| Rocky UI with new type | ✅ | findings list, Finding detail, WPVulnFindingType detail, Software detail, object list — all 200 |
| Duplicate counting | ❌ | 2× on http+https, 5× on shared hosting (finding 1) |

### Verdict

**CHANGES REQUESTED**

Both round-3 blockers are genuinely fixed, and I verified the graph placement the hard way rather than by reading the paths: findings land on `Software` per Jan's direction, the tree provably cannot see them, and all six new query paths recover them — including through the aggregate report and crisis room, which inherit the fix. That is the right shape.

What is left is the last mile on the same code. The `.software.` hop fans in, so one vulnerability is counted once per SoftwareInstance — five times on ordinary shared hosting — and the fix that @underdarknl has put on the release gate can be deleted without failing a test. Both are small: a dedup in two places and one regression test.

Four rounds in, I want to be clear this is not a moving goalpost — everything I asked for in rounds 1-3 is done and verified, including my own corrected suggestion. These two are new, they are in the code this round added, and they are the difference between the report numbers being right and being plausible.

Two notes, neither blocking: the branch is currently `behind` main and will need updating before merge, and `wpscan-out-parse` is still pinned in `boefjes/pyproject.toml` while no longer imported — you already flagged that as a lockfile-regeneration follow-up, which seems right to keep separate.

underdarknl:
Regarding the duplication:

Multiple software instances might exists and point at multiple wordpress instances on the same IP, this is because they are bound to differentiating urls (eg, different ports, or different websites resolving to this IP). These should be counted as separate findings when queried from an Ip.

The same is true for multiple wordpress instances (and their findings) when querying from a hostname, traversing trough its various ip's and ports could lead to (probably) the same factual install. In that case, we however can not be sure that we are in fact dealing with the same server (ipv6 might be hosted elsewhere for example), the good news is that fixing one outdated install, could fix all.

There is one unfortunate issue though, as wpscan itself only accepts url's, and cannot be bound to a specific IP, various hostnames with various IP's will be scanned by wpscan with the same cli arguments, arguably muddying the water, ideally we'd also move the wpscan itself to the newly released 4.0 of wpscan from May 20, 2026.

hasecon:
## Addendum to round 4 — the fan-in multiplier is larger than I measured, and the boefje already guards against it

Following up on finding 1 with the input shape traced properly. My round-4 examples (2× on http+https, 5× on shared hosting) understated it, and I had the mechanism only half right.

### Where the SoftwareInstances actually come from

`wp-scan` consumes `SoftwareInstance`, and the realistic producer of the ones it consumes is the Wappalyzer normalizer, which builds its target like this (`kat_wappalyzer/normalize.py:11-26`):

```python
web_url = HostnameHTTPURL(
    network=network.reference,
    netloc=hostname.reference,
    port=tokenized_weburl["port"],
    scheme=tokenized_weburl["scheme"],
    path=tokenized_weburl["path"],      # <-- path is part of the key
)
yield from software_from_har(raw, web_url.reference)
```

It runs per `HTTPResource`, and the URL it binds to varies by **scheme, port and path**. So one hostname routinely carries many `SoftwareInstance`s all pointing at the same `Software|WordPress|<version>|` — one per HTTP resource where WordPress was detected, not just one for http and one for https.

The report path walks `Hostname → all HostnameHTTPURLs → all SoftwareInstances → Software → Finding`, so every one of those is a separate route to the same `Finding`.

Measured on a running stack — one hostname, six web URLs (http + https + `/`, `/blog/`, `/wp-admin/`, `/shop/`), one `Software`, one `Finding`:

```
=== one hostname, 6 web URLs (http+https+paths), ONE Software, ONE Finding ===
  findings_report query returned : 6 rows
  distinct findings              : 1
  -> the single vulnerability is reported 6x in the report
```

On a real WordPress site Wappalyzer fires on many more resources than six, and in the vulnerability report this then multiplies again across every hostname on the IP. So the inflation is structural, not a corner case.

### The boefje already refuses to double-count — the report undoes it

`kat_wpscan/main.py` is explicit about this:

```python
if scheme != "https":
    return [(info_mimetype, "To avoid double findings, we only scan https urls.")]
```

The scan side deliberately skips the http URL to avoid duplicate findings. But the report-side fan-in counts the http `SoftwareInstance` anyway, because Wappalyzer created it and it reaches the same `Software`. The guard in the boefje and the merge in the report are pulling in opposite directions.

### Two things this clears up in the PR's favour

- **`SoftwareInstance.ooi` is `ReferenceField(OOI)`**, so it can point at anything, and most producers in the repo (`nmap`, `shodan`, `censys`, `binaryedge`, `service_banner`, …) bind to `IPPort`/`IPService`. I expected that to blow a hole in the `HostnameHTTPURL`-shaped report paths. It does not, because the boefje rejects those inputs itself:

  ```python
  if input_["software"]["name"] != "WordPress":            return "Not wordpress."
  if "netloc" not in input_["ooi"] or "name" not in ...:   return "No hostname available for input OOI."
  ```

  So wp-scan only ever produces findings for hostname-based URLs, which is exactly what `_SOFTWARE_FINDING_PATHS` covers. The paths are correctly shaped for this boefje.
- The `path` component also means the normalizer's URL derivation has to survive paths other than `/`. It does — the wrapper-stripping is path-agnostic, and I verified that in round 4.

### No change to the suggested fix

Deduplicating by `f.reference` handles any multiplier, so the patch I proposed in round 4 stands unchanged. Worth saying explicitly though: today the six rows are byte-identical — same `Finding` primary key, no indication of which URL each came from — so this is duplication, not per-location reporting. If per-location is what you'd actually want ("WordPress 6.5.2 is vulnerable, affects these 6 URLs"), that needs the `SoftwareInstance` carried into the row and is a separate feature; the dedup is still the right fix for the counts as they stand.

Verdict unchanged (**CHANGES REQUESTED**), same two findings, with finding 1 now on firmer ground and a larger blast radius than I first reported.

hasecon:
@underdarknl our comments crossed — yours landed at 13:51, mine at 13:54, so I hadn't read it. Having read it now: **you're right about the semantics and my "just deduplicate" recommendation was wrong.** Correcting it, with the measurement that shows why the current behaviour doesn't give you what you're asking for either.

### Agreed on the goal

> Multiple software instances might exists and point at multiple wordpress instances on the same IP … These should be counted as separate findings when queried from an Ip.

Yes. Two WordPress installs on different ports, or two websites on one IP, are two things to fix and should count as two. A blanket dedup on `finding.reference` would collapse those into one and hide a real install. Scratch that suggestion.

### But the current count isn't per-install either — it's per-URL

The problem is that binding the Finding to `Software` (your earlier steer, and I still think it's the right model) means the graph holds **exactly one** `Finding` OOI for "WordPress 6.5.2 has vuln X". The report reaches it once per traversal route, so the rows are literally the same object repeated, with nothing on them saying which install they came from.

From my run — one hostname, six web URLs:

```
findings_report query returned : 6 rows
distinct findings              : 1
all six rows: Finding|Software|WordPress|6.5.2||aaaa-1111   (identical primary key)
```

Those six URLs were `https://host:443/`, `/blog/`, `/wp-admin/`, `/shop/` and `http://host:80/`, `/blog/`. So:

| interpretation | count |
|---|---|
| distinct `Finding` OOIs | 1 |
| distinct installs, i.e. distinct scheme+host+port | 2 |
| installs wp-scan actually scanned (it skips http) | 1 |
| **what the report prints today** | **6** |

The 6 comes from Wappalyzer creating a `SoftwareInstance` per `HTTPResource` — and its URL key includes `path` (`kat_wappalyzer/normalize.py:17-24`). So four paths on one https install produce four routes to the same finding. The number tracks *URLs traversed*, which is neither the number of findings nor the number of installs. It happens to look plausible when each install has exactly one URL, and inflates as soon as a site has more than one resource — which is every real WordPress site.

So today's output doesn't implement your semantics; it just over-counts in a way that superficially resembles it.

### What would implement it

The location has to come from the `SoftwareInstance` side, since the `Finding` itself can't carry it. Concretely: pair each finding with the install it was reached through, and deduplicate on that pair rather than on the finding alone — collapsing the path multiplication while keeping genuinely distinct installs apart:

```python
# one row per (finding, install), where install identity is the URL without path
seen = set()
for finding, instance in self.octopoes_api_connector.query(software_path_with_instance, ...):
    url = instance.ooi.tokenized          # scheme + netloc + port identify the install
    key = (finding.reference, url["scheme"], url["netloc"]["name"], url["port"])
    if key not in seen:
        seen.add(key)
        findings.append(finding)
```

That gives 2 for my example from a hostname (https and http installs), and from an IP it keeps separate websites and separate ports apart exactly as you describe. It also gives the report something worth printing — *which* install is affected — instead of six identical rows.

I'd treat "carry the SoftwareInstance into the row" as the real fix and it is a bit more than a two-line change, so if you'd rather land #5302 for the release and do the per-install reporting as a follow-up, that's a reasonable split — but then the report shouldn't print 6 for one install in the meantime, so it needs *some* collapsing now.

### On your last paragraph

> wpscan itself only accepts url's, and cannot be bound to a specific IP, various hostnames with various IP's will be scanned by wpscan with the same cli arguments

Confirmed in the boefje as it stands — it builds `url = f"{scheme}://{hostname}{path}"` and never sees an IP, so two hostnames on different IPs produce two scans that are indistinguishable at the argv level. Worth noting the boefje already tries to limit the damage with `if scheme != "https": return "To avoid double findings, we only scan https urls."` — the scan side is deliberately avoiding duplicates while the report side re-introduces them, which is part of why the 6 looked so wrong to me.

No opinion from me on the wpscan 4.0 bump beyond it being out of scope here; happy to review it separately if you open one.

My verdict on the PR is unchanged (**CHANGES REQUESTED**) but finding 1 should now be read as "the report count is meaningless as printed" rather than "deduplicate them" — the fix depends on your call above.

brennodewinter:
@hasecon — round-4 findings addressed:

**Finding 1 (Medium, dedup)**: Added dedup by reference in both reports.
- `findings_report/report.py`: `seen`-set before append, so the same Finding from multiple SoftwareInstances is counted once.
- `vulnerability_report/report.py`: dict-by-reference merge of `port_findings + software_instance_findings + software_findings` (and the same for finding types).

**Finding 2 (Medium, no test)**: Added `test_findings_report_dedups_software_findings` — seeds two identical Software findings via the query mock and asserts `total_occurrences == 1`.

Branch rebased on current main (was 5 behind). All report and member tests pass.

hasecon (changes requested):
## QA Review — round 5

### Setup
- Branch: `feat/wpscan-software-inventory` @ `de9060ed`
- Reviewed in an isolated git worktree (the shared checkout is in use by other work)
- Delta since round 4 is confined to two files — see "Scope of this round" below

### Round-4 findings

| # | Round-4 finding | Status |
|---|---|---|
| 2 | The report fix had no test; the conftest change made its absence silent | ✅ **Fixed** — verified by revert test |
| 1 | Software fan-in duplicates findings | ⚠️ **Fixed against the superseded instruction** — see below |

**Finding 2 is properly closed.** I re-ran the revert test: removed the whole `_SOFTWARE_FINDING_PATHS` block and the new test goes red, where in round 4 the entire fix could be deleted with all 87 report tests still passing.

```
with the block removed:
FAILED tests/reports/test_findings_report.py::test_findings_report_dedups_software_findings - assert 0 == 1
1 failed, 2 passed
```

That is exactly the pin that was missing. Thank you.

---

### Finding 1 — this implements the version I retracted

This one is on me as much as anyone, so let me lay out the sequence plainly:

| when | what |
|---|---|
| 09-09 13:51 | @underdarknl ([comment](https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/pull/5302#issuecomment-5602991337)): multiple instances may be *genuinely different* WordPress installs (different ports, different websites on one IP) — *"These should be counted as separate findings when queried from an Ip."* |
| 09-09 13:54 | my addendum (written before I saw Jan's, our comments crossed by 3 minutes) |
| 09-09 14:00 | my [correction](https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/pull/5302#issuecomment-5603116784): *"you're right … my 'just deduplicate' recommendation was wrong. A blanket dedup on `finding.reference` would collapse those into one and hide a real install. Scratch that suggestion."* |
| 09-10 09:03 | this commit: blanket dedup on `finding.reference`, in both reports |

Your comment addresses my round-4 findings only and doesn't mention Jan's steer, so I read this as the correction simply not having been seen — entirely understandable given it landed six hours after the round-4 review it amends. No criticism intended; I'd rather flag it now than have it merge.

**What it does today.** Jan's exact scenario — two real installs, `https://host:443` and `https://host:8443`, both WordPress 6.5.2, both vulnerable. Because the Finding binds to `Software` (correctly), the graph holds one `Finding` OOI and the query returns it once per install. Run against the report on this head:

```
installs vulnerable in reality  : 2  (:443 and :8443)
report total_occurrences        : 1
what Jan asked for              : 2
```

So a second vulnerable WordPress install on the same host is now silently absent from the report. Round 4's over-count (6 for one install) has become an under-count (1 for two installs) — and of the two, hiding a real install is the worse failure for a security report.

**Why a `finding.reference` set cannot get this right.** The query path ends in `[is Finding]`:

```python
"Hostname.<netloc [is HostnameHTTPURL].<ooi [is SoftwareInstance].software.<ooi [is Finding]"
```

so the report receives bare `Finding` objects with nothing identifying which install each came through. Two installs and one install reached twice are *literally the same input*. No dedup key computed from the finding alone can separate them; the distinguishing information has to be carried in from the `SoftwareInstance` side.

**Suggested direction** (as in my correction — extend the path to return the instance and key on the install):

```python
# one row per (finding, install); install identity is the URL without path,
# so :443 and :8443 stay apart while /, /blog/, /wp-admin/ collapse
seen = set()
for finding, instance in self.octopoes_api_connector.query(software_path_with_instance, valid_time, reference):
    url = instance.ooi.tokenized
    key = (finding.reference, url["scheme"], url["netloc"]["name"], url["port"])
    if key not in seen:
        seen.add(key)
        findings.append(finding)
```

That yields 2 for Jan's two-port case, 1 for one install reached via four paths, and gives the report something worth printing — *which* install is affected.

**The new test currently pins the wrong semantics.** `test_findings_report_dedups_software_findings` seeds the same Finding twice and asserts `total_occurrences == 1`, commented `# http + https instances`. Under Jan's semantics those are two installs and the answer is 2, so that test will need to flip along with the fix. Worth adding its sibling — same Finding reached through two *different* installs must count twice — since that is the case that regressed.

The same `{f.reference: f}` merge in `vulnerability_report` has the same effect, and note it now also collapses the pre-existing `port_findings` and `software_instance_findings`, which were not deduplicated before — a scope change beyond the new path that's worth a deliberate look.

**If you'd rather not build per-install reporting now:** that's reasonable, but then neither number is right, and Jan has this on the release gate. A middle option is to key on `(finding.reference, instance_url_without_path)` only for the new Software path and leave the pre-existing sources untouched — smaller than full per-install reporting, and it stops both the 6× over-count and the install-hiding.

### Scope of this round

The PR's own runtime code is untouched since round 4 — I verified per file rather than assuming:

```
ONGEWIJZIGD  boefjes/boefjes/plugins/kat_wpscan/normalize.py
ONGEWIJZIGD  octopoes/octopoes/models/ooi/findings.py
GEWIJZIGD    rocky/reports/report_types/findings_report/report.py       (+8 -6)
GEWIJZIGD    rocky/reports/report_types/vulnerability_report/report.py  (+16 -6)
```

So round 4's live verification still stands at this head: the input-shape fix on real serialized input, the flat `SoftwareInstance` placement, all six query paths parsing and returning the Software-bound finding against real XTDB, and `get_tree` returning EMPTY as the control. I did **not** re-run `make clean && make kat` this round — the delta is pure report-merge logic with no runtime or service-boundary change, and I verified the changed behaviour directly instead. Saying so explicitly rather than implying a functional test I didn't run.

### Test Results

| Component | Test | Result |
|-----------|------|--------|
| rocky | `make utest` (Docker) | ✅ 481 passed, 1 skipped |
| rocky | `tests/reports/test_findings_report.py` | ✅ 3 passed |
| rocky | revert test on `_SOFTWARE_FINDING_PATHS` | ✅ goes red — finding 2 pinned |
| rocky | Jan's two-install scenario | ❌ reports 1, should be 2 |
| boefjes | `tests/plugins/test_wpscan.py` | ✅ 15 passed |
| octopoes | `tests/test_query.py` | ⚪ not re-run — file identical to round 4 (20 passed then), CI `test` 3.10–3.13 green |
| CI | all 39 checks incl. `rtest`, `pre-commit`, `CodeQL` | ✅ green |

### Verdict

**CHANGES REQUESTED**

Finding 2 is genuinely fixed and now properly pinned — that part is done.

Finding 1 has been implemented from my round-4 text rather than the correction that superseded it six hours later, and the result now contradicts @underdarknl's explicit instruction: two distinct WordPress installs on one host collapse into a single reported finding. Given that the blanket dedup was my suggestion before I withdrew it, I'd rather carry that than have you chase it — the practical ask is to key the dedup on the install, not on the finding, and flip the accompanying test.

Everything else from rounds 1–4 remains fixed and verified.

brennodewinter:
## Round-5 review: installation-aware dedup

Replaced the blanket dedup on `finding.reference` with a dedup keyed on `(finding.reference, install_url_without_path)` — the install identity is the `SoftwareInstance` URL with the path stripped, so `:443` and `:8443` stay apart while `/`, `/blog/`, `/wp-admin/` on the same port collapse.

### FindingsReport (`report.py`)
- Replaced `_SOFTWARE_FINDING_PATHS` (single query ending at `Finding`) with `_SOFTWARE_INSTANCE_PATHS` + `_SOFTWARE_FINDING_PATH` (two-step: query `SoftwareInstance` objects from the input, then `query_many` findings per `Software`).
- Join instances to findings by `software` reference, dedup by `(finding.reference, _install_key(instance))`.

### VulnerabilityReport (`report.py`)
- **Reverted the scope change**: `port_findings`, `software_instance_findings` and their finding types are no longer deduped — back to simple concatenation as before the PR.
- Software findings are fetched per install via a `SoftwareInstance` query + `query_many` on `Software.<ooi [is Finding]`, deduped by `(finding.reference, install_url_without_path)`.

### Tests
- `test_findings_report_dedups_software_findings_same_install`: two `SoftwareInstance`s on the same port but different paths (`/`, `/blog/`) → `total_occurrences == 1`.
- `test_findings_report_counts_software_findings_per_install`: two `SoftwareInstance`s on different ports (`:443`, `:8443`) → `total_occurrences == 2`.

### Verification
- Rocky `make utest`: 482 passed, 1 skipped.
- `pre-commit run`: all green.

@hasecon — ready for re-review.

Issue #1436, opened 2023-07-24
Rate limiting accros multi-organisations/workers.


zcrt:
**Describe the bug**
Some boefjes require an external API. When multiple boefjes use the same API at the same time (e.g. organisation A, B and C require data from the same external API at the same moment) rate limiting can occur, resulting in failed boefjes.

**To Reproduce**
1. Create multiple organisations with multiple assets
2. Add a Shodan API key and enable boefje
3. Notice failing Shodan boefjes can occur due to rate limiting:

```nomad
[2023-07-24 08:38:50 +0000] [9] [ERROR] [job_handler] Error running boefje shodan[c76577ac40df4364bcd6f4d9c59738da]
Traceback (most recent call last):
  File "/app/boefjes/boefjes/local.py", line 50, in run
    return boefje_resource.module.run(boefje_meta)
  File "/app/boefjes/boefjes/plugins/kat_shodan/main.py", line 22, in run
    results = api.host(ip)
  File "/usr/local/lib/python3.8/site-packages/shodan/client.py", line 380, in host
    return self._request('/shodan/host/%s' % ','.join(ips), params)
  File "/usr/local/lib/python3.8/site-packages/shodan/client.py", line 340, in _request
    raise APIError(data['error'])
shodan.exception.APIError: Rate limit reached. Please throttle your requests to 1 request per second.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/app/boefjes/boefjes/job_handler.py", line 138, in handle
    boefje_results = self.job_runner.run(boefje_meta, environment)
  File "/app/boefjes/boefjes/local.py", line 52, in run
    raise JobRuntimeError("Boefje failed") from e
boefjes.runtime_interfaces.JobRuntimeError: Boefje failed
```

**Expected behavior**
Some kind of intelligence in the KAT installation to not start boefjes with a dependence on the same external API at the same time (e.g. across multiple organizations or workers). This is probably a scheduling issue that gets more likely to occur with every organization added.

**OpenKAT version**
1.8

underdarknl:
https://github.com/minvws/nl-kat-coordination/issues/1317 should tackle these issues.

PR #5313, opened 2026-08-12
Rate-limit external API boefjes across organisations

### LINKED ISSUE #2291: Optional: create rate limiting functionality

brennodewinter:
## Summary
Split out from #5305 per review: this PR contains only the rate-limiting work. The organization activity logs move to a separate PR (#5305 is being closed in favour of this and the activity-logs PR).

- Enforce manifest-defined API cooldowns centrally across organisations and workers
- Configure Shodan API calls with a shared one-second rate limit
- `rate_limit_group` supports `str.format`-style placeholders resolved against the input OOI at task creation (e.g. `"{input_ooi}"`, `"shodan:{object_type}"`), so boefjes scanning the same OOI share a rate-limit group — as requested in the #5305 review and described in #1317

## On the dynamic limit group
Per the review comment, the limit group is now dynamic: a boefje manifest can use `{input_ooi}` and `{object_type}` placeholders. Two boefjes hitting the same `IPAddressV4` will construct the same group and play nice together. Resolving richer OOI properties (e.g. `{address}` on a `Website`, which requires traversing a reference) needs the full OOI from Octopoes and is left as a follow-up (#1317). An unresolvable placeholder disables rate limiting for that task (with a warning log) rather than crashing.

Addresses #1436 (already closed).
Closes #2291.
Relates to #1317 (full OOI-property resolution is a follow-up).

#### Test plan
- [x] Mula integration: `test_boefje_scheduler.py` — 75 passed (incl. new `test_resolve_rate_limit_group_template` and `test_pop_respects_rate_limit_grouped_by_input_ooi`)
- [x] Boefjes integration: `test_sql_repositories.py::test_bare_boefje_storage` passes; `test_api.py::test_update_plugins` exercises the round-trip of the new fields
- [x] ruff check + ruff format on changed files


hasecon (changes requested):
## QA Review

### Setup
- Branch: `rate-limit-external-api-boefjes` @ `b54149c`
- Stack: `make clean && make kat` — ✅ 13/13 services up, Rocky 200

The overall architecture is right: enforcing at the single `BoefjePQ.pop` choke point does give cross-organisation and cross-worker coverage (one boefje scheduler process serves all orgs/workers), the template design matches what was asked in the #5305 review, the test coverage is genuinely good (the mocked-`monotonic` pattern and the template matrix), and I verified live that the manifest → migration → katalogus API round-trip works. But there is one High finding — proven empirically — plus two Mediums that need addressing.

### Code Review Findings

| H1 | **High** | `mula/scheduler/storage/stores/pq.py` (`item_orm = item_orm[:1]`) | **The truncation defeats task deduplication for rate-limited boefjes and *increases* API calls.** A dedup batch (same boefje+input across N orgs) pops as one batch → one worker run → **one** API call serving all N orgs — the ideal case for an API-quota feature. The truncation pops only 1; the N−1 sibling tasks stay queued and later pop one-by-one, each spaced by the interval and **each making its own API call**. For Shodan across N orgs scanning the same IP this multiplies calls by N. Reproduced in the mula integration env: control batch (no rate limit) pops **3 of 3**; identical rate-limited batch pops **1 of 3**. **Fix: delete the two truncation lines.** The batch shares one boefje and thus one group, so the existing `items[0]` deadline logic in `BoefjePQ.pop` already does exactly the right thing for a full batch. |
| M1 | Medium | `mula/scheduler/schedulers/schedulers/boefje.py` (`push_item_to_queue`), `rocky/rocky/scheduler.py` | **Manually started scans bypass the rate limiter entirely.** The three scheduler-internal creation paths resolve templates, but the server push path (Rocky "start scan" → `POST /schedulers/{id}/push` → `push_item_to_queue`) does not — and Rocky's client `Boefje` model doesn't carry the rate-limit fields at all, so a manual task lands with `rate_limit_interval=None`: it pops even while its group is under cooldown *and* registers no deadline. A burst of manual Shodan scans can blow the API budget the scheduler is carefully guarding. Suggested fix: in `BoefjeScheduler.push_item_to_queue` (which already validates the `BoefjeTask`), stamp `rate_limit_interval`/`rate_limit_group` from the katalogus plugin and call `resolve_rate_limit_group` (context is available: `input_ooi` is on the task; `object_type = input_ooi.split("|")[0]`). If manual scans are *intended* to be exempt, say so in the PR description and a code comment — right now it reads as an oversight. |
| M2 | Medium | `boefjes/boefjes/katalogus/plugins.py` (`BoefjeIn`) | **`BoefjeIn` is missing the interval-requires-group validator** that both `Boefje` models got. A `PATCH /plugins/{id}` with `rate_limit_interval` set and no `rate_limit_group` passes `BoefjeIn` validation and persists. From then on: katalogus `to_boefje()` builds the worker `Boefje` model — which *does* have the validator — so **`GET /plugins` 500s for that org**, and mula's `process_mutations` raises at `models.Boefje.model_validate(...)` (bare in the loop, no per-boefje try/except), halting task creation for matching OOI mutations. Fix: add the same `model_validator` to `BoefjeIn` so the invariant is enforced at the API boundary. |
| L1 | Low | `boefje.py` (`resolve_rate_limit_group`) | The `except (KeyError, IndexError, ValueError)` doesn't catch `AttributeError`, which `str.format` raises for attribute-style templates (`"{input_ooi.netloc}"`). That typo class crashes task creation instead of disabling the limit as designed. Add `AttributeError` to the clause. |
| L2 | Note | `boefje.py` (`BoefjePQ`) | The deadline dict is in-memory: a scheduler restart forgets all cooldowns (worst case, one burst right after restart), and the design assumes a single boefje-scheduler process. Both are acceptable today — worth a one-line comment so the assumption is explicit. |

Verified non-issues: `BoefjeTask.hash` uses only `input_ooi + boefje.id + organization`, so group resolution doesn't disturb dedup/hash identity; the migration chain is sound (`fdeaea4481b8 → e5f5a622c87b`, no fork); the SQL exclusion handles NULL groups correctly; `@with_lock` serializes pops so the deadline dict is race-free.

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| Mula | `make utest` (Docker) | ✅ |
| Mula | `make itest` (Docker) | ✅ 216 passed, 3 skipped — incl. the PR's 3 new tests |
| Mula | dedup×rate-limit interaction (review repro) | ❌ rate-limited dedup batch pops 1 of 3 (control: 3 of 3) — finding H1 |
| Boefjes | `make itest` (Docker) | ✅ 36 passed |

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Stack build | ✅ | 13/13 up, Rocky 200 |
| Migration applied | ✅ | katalogus `alembic_version` = `e5f5a622c87b`; `boefje.rate_limit_interval` (double) + `rate_limit_group` (varchar) present |
| Manifest round-trip | ✅ | `GET /v1/organisations/dvwa/plugins/shodan` → `rate_limit_interval: 1.0`, `rate_limit_group: "api.shodan.io"` |
| Log analysis | ✅ | scheduler healthy; only known idempotent org-create rollback noise |

### Suggested regression test
For H1, this drop-in (mirrors `test_pop_deduplication`) pins the batch semantics — verified: it **fails on the current branch (1 of 3)** and passes once the truncation is removed:

```python
def test_pop_dedup_batch_with_rate_limit_pops_full_batch(self):
    scan_profile = ScanProfileFactory(level=0)
    ooi = OOIFactory(scan_profile=scan_profile)
    self.mock_get_plugin.return_value = PluginFactory(scan_level=0, consumes=[ooi.object_type])
    boefje = BoefjeFactory(rate_limit_interval=1, rate_limit_group="api.example.com")
    dkey = uuid.uuid4()
    for org in [self.organisation.id, OrganisationFactory().id, OrganisationFactory().id]:
        boefje_task = models.BoefjeTask(
            boefje=models.Boefje.model_validate(boefje.model_dump()),
            input_ooi=ooi.primary_key, organization=org, deduplication_key=dkey.hex,
        )
        self.scheduler.push_item_to_queue(models.Task(
            scheduler_id=self.scheduler.scheduler_id, organisation=self.organisation.id, priority=1,
            type=models.BoefjeTask.type, hash=boefje_task.hash, data=boefje_task.model_dump(),
            created_at=datetime.now(timezone.utc), modified_at=datetime.now(timezone.utc),
        ))
    # One deduplicated batch = one worker run = one API call for all three orgs.
    self.assertEqual(3, len(self.scheduler.pop_item_from_queue()))
```

### Verdict
**CHANGES REQUESTED**

Good feature, right architecture, good tests — but H1 currently makes the rate limiter *increase* API load for exactly the deduplicated multi-org case it should optimize (fix is deleting two lines + the test above), M1 leaves a hole in the cross-organisation guarantee for manual scans, and M2 lets one API PATCH brick plugin listing and scheduling. Happy to re-review quickly once these land.

boefjes/boefjes/migrations/versions/e5f5a622c87b_add_boefje_rate_limits.py
```diff
@@ -0,0 +1,23 @@
...
+from alembic import op
+
+revision = "e5f5a622c87b"
+down_revision = "fdeaea4481b8"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column("boefje", sa.Column("rate_limit_interval", sa.Float(), nullable=True))
```
underdarknl, line 17: Interval is a single dimension. We probably want to allow for multiple dimensions, eg, 1000 calls per month + no more than 1 per 5 minutes.

PR #5314, opened 2026-08-12
Add organization user activity logs

### LINKED ISSUE #1437: Show log of user-actions that happened in KAT
When handling multiple organisations with multiple users it can become hard to understand which changes have happened in KAT. 

It would be nice to add a log panel to each organisation's crisis-room, and to the general crisis-room. E.g. a simple panel showing actions like:
- At [datetime] user [user] changed indemnification of object [object] to [indemnification-level]
- At [datetime] user [user] manually added object [object]
- At [datetime] user [user] enabled boefje [boefje]
- At [datetime] user [user] changed boefje setting [setting] (from [x] to [y])
- etc.

Ideally with hyperlinked texts to the relevant objects and boefjes etc.

Maybe crawling through logs or just hoping nothing important has changed.

**Additional context**
Multiple steps can be taken for this feature. E.g. first some simple logs in the interface for a single organisation, then across multiple organisations,  then also adding hyperlinking and filtering (on user, object, ...) to this logs.

Related: https://github.com/minvws/nl-kat-coordination/issues/365

### LINKED ISSUE #5308: Fix indemnification onboarding crash after adding activity logging
## Description

Submitting the indemnification step during onboarding raises:

```
TypeError: 'Indemnification' object is not callable
```

## Root cause

The result of `Indemnification.objects.get_or_create()` is assigned to `_`, shadowing Django's imported translation helper. The activity-log label then tries to call the `Indemnification` instance as a function.

## Resolution

Rename the unused result variable so the translation helper remains callable, and add a regression test covering the first indemnification POST and its audit-log entry.

Resolved by #5305.

brennodewinter:
## Summary
Split out from #5305 per review: this PR contains the organization-scoped activity logs (audit log views) and the indemnification onboarding crash fix that the activity logging introduced.

- Add organization-scoped activity logs for successful user actions without storing plugin setting values
- Expose the latest activity in both the general and organization crisis rooms
- Fix the indemnification onboarding crash: the `Indemnification.objects.get_or_create()` result was assigned to `_`, shadowing Django's translation helper before the new audit-log label was translated. Creating a first indemnification raised `TypeError: 'Indemnification' object is not callable`. The unused result variable is renamed so the translation helper stays callable, with a regression test.

The rate-limiting work moves to a separate PR (#5313).

Closes #1437.
Closes #5308.

#### Test plan
- [x] Indemnification regression tests: 2 passed
- [x] Rocky audit-log tests added (`test_audit_log.py`)
- [x] pre-commit on changed files

### DIFF
--- a/rocky/account/mixins.py
+++ b/rocky/account/mixins.py
@@ -16,7 +16,6 @@ from katalogus.client import KATalogus, get_katalogus
 from rest_framework.exceptions import ValidationError
 from rest_framework.request import Request
 from tools.models import Indemnification, Organization, OrganizationMember
-from tools.view_helpers import get_ooi_url
 
 from octopoes.connector.octopoes import OctopoesAPIConnector
 from octopoes.models import OOI, DeclaredScanProfile, Reference, ScanLevel
@@ -184,8 +183,7 @@ class OrganizationView(ContextMixin, View):
             organization=self.organization,
             action=AuditLog.Action.CLEARANCE_LEVEL_CHANGED,
             object_type=ooi_reference.class_,
-            object_label=ooi_reference.human_readable,
-            object_url=get_ooi_url("ooi_detail", ooi_reference, self.organization.code),
+            object_pk=ooi_reference,
         )
 
         return True
--- a/rocky/account/mixins.py
+++ b/rocky/account/mixins.py
@@ -3,6 +3,7 @@ from functools import cached_property
 
 import structlog
 import structlog.contextvars
+from crisis_room.models import AuditLog
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.mixins import PermissionRequiredMixin
@@ -15,6 +16,7 @@ from katalogus.client import KATalogus, get_katalogus
 from rest_framework.exceptions import ValidationError
 from rest_framework.request import Request
 from tools.models import Indemnification, Organization, OrganizationMember
+from tools.view_helpers import get_ooi_url
 
 from octopoes.connector.octopoes import OctopoesAPIConnector
 from octopoes.models import OOI, DeclaredScanProfile, Reference, ScanLevel
@@ -177,6 +179,14 @@ class OrganizationView(ContextMixin, View):
             sync=True,
         )
         logger.info("Declared scan profile created", event_code="800010", ooi=ooi_reference, level=level)
+        AuditLog.record(
+            user=self.request.user,
+            organization=self.organization,
+            action=AuditLog.Action.CLEARANCE_LEVEL_CHANGED,
+            object_type=ooi_reference.class_,
+            object_label=ooi_reference.human_readable,
+            object_url=get_ooi_url("ooi_detail", ooi_reference, self.organization.code),
+        )
 
         return True
 
@@ -191,6 +201,13 @@ class OrganizationView(ContextMixin, View):
             sync=True,
         )
         logger.info("Declared scan profiles created", event_code="800010", ooi_count=len(ooi_references), level=level)
+        AuditLog.record(
+            user=self.request.user,
+            organization=self.organization,
+            action=AuditLog.Action.CLEARANCE_LEVEL_CHANGED,
+            object_type="OOI",
+            object_label=str(_("%(count)s objects to L%(level)s")) % {"count": len(ooi_references), "level": level},
+        )
 
         return True
 
--- a/rocky/crisis_room/migrations/0009_auditlog.py
+++ b/rocky/crisis_room/migrations/0009_auditlog.py
@@ -34,8 +34,8 @@ class Migration(migrations.Migration):
                     ),
                 ),
                 ("object_type", models.CharField(blank=True, max_length=64)),
-                ("object_label", models.CharField(blank=True, max_length=512)),
-                ("object_url", models.CharField(blank=True, max_length=2048)),
+                ("object_label", models.TextField(blank=True, default="")),
+                ("object_pk", models.TextField(blank=True, default="")),
                 ("created_at", models.DateTimeField(auto_now_add=True, db_index=True)),
                 (
                     "actor",
@@ -53,7 +53,7 @@ class Migration(migrations.Migration):
             options={
                 "ordering": ["-created_at"],
                 "indexes": [
-                    models.Index(fields=["organization", "-created_at"], name="crisis_room_organiz_875530_idx")
+                    models.Index(fields=["organization", "-created_at"], name="audit_log_org_created_idx")
                 ],
             },
         )
--- a/rocky/crisis_room/migrations/0009_auditlog.py
+++ b/rocky/crisis_room/migrations/0009_auditlog.py
@@ -1,0 +1,60 @@
+import django.db.models.deletion
+from django.conf import settings
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ("crisis_room", "0008_alter_dashboard_organization"),
+        ("tools", "0047_alter_organization_code_alter_organization_name"),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name="AuditLog",
+            fields=[
+                ("id", models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name="ID")),
+                ("actor_label", models.CharField(max_length=254)),
+                (
+                    "action",
+                    models.CharField(
+                        choices=[
+                            ("indemnification_set", "Set indemnification"),
+                            ("object_added", "Added object"),
+                            ("object_updated", "Updated object"),
+                            ("object_deleted", "Deleted object"),
+                            ("clearance_level_changed", "Changed clearance level"),
+                            ("plugin_enabled", "Enabled plugin"),
+                            ("plugin_disabled", "Disabled plugin"),
+                            ("plugin_settings_changed", "Changed plugin settings"),
+                            ("plugin_settings_deleted", "Deleted plugin settings"),
+                        ],
+                        max_length=32,
+                    ),
+                ),
+                ("object_type", models.CharField(blank=True, max_length=64)),
+                ("object_label", models.CharField(blank=True, max_length=512)),
+                ("object_url", models.CharField(blank=True, max_length=2048)),
+                ("created_at", models.DateTimeField(auto_now_add=True, db_index=True)),
+                (
+                    "actor",
+                    models.ForeignKey(
+                        null=True, on_delete=django.db.models.deletion.SET_NULL, to=settings.AUTH_USER_MODEL
+                    ),
+                ),
+                (
+                    "organization",
+                    models.ForeignKey(
+                        on_delete=django.db.models.deletion.CASCADE, related_name="audit_logs", to="tools.organization"
+                    ),
+                ),
+            ],
+            options={
+                "ordering": ["-created_at"],
+                "indexes": [
+                    models.Index(fields=["organization", "-created_at"], name="crisis_room_organiz_875530_idx")
+                ],
+            },
+        )
+    ]
--- a/rocky/crisis_room/migrations/0009_auditlog.py
+++ b/rocky/crisis_room/migrations/0009_auditlog.py
@@ -52,9 +52,7 @@ class Migration(migrations.Migration):
             ],
             options={
                 "ordering": ["-created_at"],
-                "indexes": [
-                    models.Index(fields=["organization", "-created_at"], name="audit_log_org_created_idx")
-                ],
+                "indexes": [models.Index(fields=["organization", "-created_at"], name="audit_log_org_created_idx")],
             },
         )
     ]
--- a/rocky/crisis_room/models.py
+++ b/rocky/crisis_room/models.py
@@ -3,6 +3,7 @@ from typing import Any, Literal, TypedDict
 from urllib.parse import urlencode
 
 import structlog
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models, transaction
@@ -13,6 +14,53 @@ from django.utils.translation import gettext_lazy as _
 from tools.models import Organization
 
 logger = structlog.get_logger(__name__)
+
+
+class AuditLog(models.Model):
+    class Action(models.TextChoices):
+        INDEMNIFICATION_SET = "indemnification_set", _("Set indemnification")
+        OBJECT_ADDED = "object_added", _("Added object")
+        OBJECT_UPDATED = "object_updated", _("Updated object")
+        OBJECT_DELETED = "object_deleted", _("Deleted object")
+        CLEARANCE_LEVEL_CHANGED = "clearance_level_changed", _("Changed clearance level")
+        PLUGIN_ENABLED = "plugin_enabled", _("Enabled plugin")
+        PLUGIN_DISABLED = "plugin_disabled", _("Disabled plugin")
+        PLUGIN_SETTINGS_CHANGED = "plugin_settings_changed", _("Changed plugin settings")
+        PLUGIN_SETTINGS_DELETED = "plugin_settings_deleted", _("Deleted plugin settings")
+
+    organization = models.ForeignKey(Organization, on_delete=models.CASCADE, related_name="audit_logs")
+    actor = models.ForeignKey(settings.AUTH_USER_MODEL, null=True, on_delete=models.SET_NULL)
+    actor_label = models.CharField(max_length=254)
+    action = models.CharField(max_length=32, choices=Action.choices)
+    object_type = models.CharField(blank=True, max_length=64)
+    object_label = models.CharField(blank=True, max_length=512)
+    object_url = models.CharField(blank=True, max_length=2048)
+    created_at = models.DateTimeField(auto_now_add=True, db_index=True)
+
+    class Meta:
+        ordering = ["-created_at"]
+        indexes = [models.Index(fields=["organization", "-created_at"], name="crisis_room_organiz_875530_idx")]
+
+    @classmethod
+    def record(
+        cls,
+        *,
+        user,
+        organization: Organization,
+        action: Any,
+        object_type: str = "",
+        object_label: str = "",
+        object_url: str = "",
+    ) -> "AuditLog":
+        return cls.objects.create(
+            organization=organization,
+            actor=user if user.is_authenticated else None,
+            actor_label=user.get_username() if user.is_authenticated else str(_("System")),
+            action=action,
+            object_type=object_type,
+            object_label=object_label,
+            object_url=object_url,
+        )
 
 
 class Dashboard(models.Model):
--- a/rocky/crisis_room/models.py
+++ b/rocky/crisis_room/models.py
@@ -11,7 +11,9 @@ from django.db.models.query_utils import Q
 from django.db.models.signals import post_delete, pre_save
 from django.dispatch import receiver
 from django.utils.translation import gettext_lazy as _
+from octopoes.models.reference import Reference
 from tools.models import Organization
+from tools.view_helpers import get_ooi_url
 
 logger = structlog.get_logger(__name__)
 
@@ -33,13 +35,13 @@ class AuditLog(models.Model):
     actor_label = models.CharField(max_length=254)
     action = models.CharField(max_length=32, choices=Action.choices)
     object_type = models.CharField(blank=True, max_length=64)
-    object_label = models.CharField(blank=True, max_length=512)
-    object_url = models.CharField(blank=True, max_length=2048)
+    object_label = models.TextField(blank=True, default="")
+    object_pk = models.TextField(blank=True, default="")
     created_at = models.DateTimeField(auto_now_add=True, db_index=True)
 
     class Meta:
         ordering = ["-created_at"]
-        indexes = [models.Index(fields=["organization", "-created_at"], name="crisis_room_organiz_875530_idx")]
+        indexes = [models.Index(fields=["organization", "-created_at"], name="audit_log_org_created_idx")]
 
     @classmethod
     def record(
@@ -50,7 +52,7 @@ class AuditLog(models.Model):
         action: Any,
         object_type: str = "",
         object_label: str = "",
-        object_url: str = "",
+        object_pk: str = "",
     ) -> "AuditLog":
         return cls.objects.create(
             organization=organization,
@@ -59,8 +61,27 @@ class AuditLog(models.Model):
             action=action,
             object_type=object_type,
             object_label=object_label,
-            object_url=object_url,
+            object_pk=object_pk,
         )
+
+    def get_object_label(self) -> str:
+        """Infer the display label from the stored PK for OOI actions;
+        fall back to the stored label for non-OOI actions (plugins, etc.)."""
+        if self.object_pk:
+            try:
+                return Reference.from_str(self.object_pk).human_readable
+            except Exception:
+                return self.object_pk
+        return self.object_label
+
+    def get_object_url(self) -> str:
+        """Construct the detail URL at read time from the stored PK."""
+        if not self.object_pk:
+            return ""
+        try:
+            return get_ooi_url("ooi_detail", self.object_pk, self.organization.code)
+        except Exception:
+            return ""
 
 
 class Dashboard(models.Model):
--- a/rocky/crisis_room/models.py
+++ b/rocky/crisis_room/models.py
@@ -11,7 +11,7 @@ from django.db.models.query_utils import Q
 from django.db.models.signals import post_delete, pre_save
 from django.dispatch import receiver
 from django.utils.translation import gettext_lazy as _
-from octopoes.models.reference import Reference
+from octopoes.models import Reference
 from tools.models import Organization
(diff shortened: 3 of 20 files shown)


rocky/account/mixins.py
```diff
@@ -177,6 +179,14 @@ def raise_clearance_level(self, ooi_reference: Reference, level: int) -> bool:
             sync=True,
         )
         logger.info("Declared scan profile created", event_code="800010", ooi=ooi_reference, level=level)
+        AuditLog.record(
```
underdarknl, line 181: Should we maybe combine the auditlog.record and the logger.info call into one.
The serve the same purpose, to record events and store them/emit them as logging lines, their resulting storage however is different, and should probably be configurable.

I propose a single function that does both, or one, or none based on the settings in our env.
It feels like we should also normalize the event_code and actions into one enum.

This would also need to be used in all locations where we currently emit events_codes to the log (in rocky).

rocky/crisis_room/migrations/0009_auditlog.py
```diff
@@ -0,0 +1,60 @@
...
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ("crisis_room", "0008_alter_dashboard_organization"),
+        ("tools", "0047_alter_organization_code_alter_organization_name"),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name="AuditLog",
+            fields=[
+                ("id", models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name="ID")),
```
underdarknl, line 17: Not sure we need an ID field, are we going to allow viewing specific audit logs, or editing/deleting (those seem unwise).
Some rolling tamper evident hashing of the previous record would however be nice to have.

rocky/crisis_room/migrations/0009_auditlog.py
```diff
@@ -0,0 +1,60 @@
...
+                ("actor_label", models.CharField(max_length=254)),
```
underdarknl, line 18: actor_label can be inferred from the actor. where null is system. Adding a text column like this does nothing but limit scalability.

rocky/crisis_room/migrations/0009_auditlog.py
```diff
@@ -0,0 +1,60 @@
...
+                (
+                    "action",
```
underdarknl, line 20: Should probably be directly linked to the event_codes enum, ideally stored in a separate lookup table or enum definition somewhere.

rocky/crisis_room/models.py
```diff
@@ -15,6 +16,53 @@
...
+    actor_label = models.CharField(max_length=254)
+    action = models.CharField(max_length=32, choices=Action.choices)
+    object_type = models.CharField(blank=True, max_length=64)
+    object_label = models.CharField(blank=True, max_length=512)
+    object_url = models.CharField(blank=True, max_length=2048)
+    created_at = models.DateTimeField(auto_now_add=True, db_index=True)
+
+    class Meta:
+        ordering = ["-created_at"]
+        indexes = [models.Index(fields=["organization", "-created_at"], name="crisis_room_organiz_875530_idx")]
```
underdarknl: A logical index name would be preferable.

rocky/crisis_room/models.py
```diff
@@ -15,6 +16,53 @@
...
+        PLUGIN_SETTINGS_CHANGED = "plugin_settings_changed", _("Changed plugin settings")
+        PLUGIN_SETTINGS_DELETED = "plugin_settings_deleted", _("Deleted plugin settings")
+
+    organization = models.ForeignKey(Organization, on_delete=models.CASCADE, related_name="audit_logs")
+    actor = models.ForeignKey(settings.AUTH_USER_MODEL, null=True, on_delete=models.SET_NULL)
+    actor_label = models.CharField(max_length=254)
+    action = models.CharField(max_length=32, choices=Action.choices)
+    object_type = models.CharField(blank=True, max_length=64)
+    object_label = models.CharField(blank=True, max_length=512)
+    object_url = models.CharField(blank=True, max_length=2048)
```
underdarknl: the url should not be stored in the database, just store the PK/identifier and type, The rest can and should be inferred from those. adding a 2k char length  also is a scalability limiter, and on top of that might not hold the entire url.

rocky/crisis_room/urls.py
```diff
@@ -5,6 +5,10 @@
 # Crisis room overview urls
 urlpatterns = [
     path("", views.CrisisRoomView.as_view(), name="crisis_room"),
+    path("logs/", views.AuditLogView.as_view(), name="crisis_room_audit_log"),
```
underdarknl, line 8: to make sure we dont colide between logs and auditlogs in the future, It seems wiser to call these routes something more descriptive (auditlog/ or audittrail/ come to mind)

rocky/crisis_room/views.py
```diff
@@ -329,6 +330,20 @@ def get_context_data(self, **kwargs):
...
+class AuditLogView(TemplateView):
+    """Recent user actions for every organization the user may access."""
+
+    template_name = "audit_log.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["audit_logs"] = AuditLog.objects.filter(
+            organization__in=self.request.user.organizations
+        ).select_related("organization")[:50]
```
underdarknl: Can we use the existing pagination/querying here?

rocky/rocky/views/ooi_delete.py
```diff
@@ -20,6 +21,13 @@ def setup(self, request, *args, **kwargs):
 
     def delete(self, request):
         self.octopoes_api_connector.delete(self.ooi.reference, valid_time=datetime.now(timezone.utc), sync=True)
+        AuditLog.record(
+            user=request.user,
+            organization=self.organization,
+            action=AuditLog.Action.OBJECT_DELETED,
+            object_type=self.ooi.get_ooi_type(),
+            object_label=self.ooi.human_readable,
```
underdarknl: Record the ooi primary key instead of the human_Readable. the later we can infer, the first we can use to create links. Since we know the time of this event, we can still query the object from the database a few seconds earlier if need be. (but the PK itself is enough to construct most variants anyway)

rocky/account/mixins.py
```diff
@@ -177,6 +179,14 @@ def raise_clearance_level(self, ooi_reference: Reference, level: int) -> bool:
...
+            user=self.request.user,
+            organization=self.organization,
+            action=AuditLog.Action.CLEARANCE_LEVEL_CHANGED,
+            object_type=ooi_reference.class_,
+            object_label=ooi_reference.human_readable,
+            object_url=get_ooi_url("ooi_detail", ooi_reference, self.organization.code),
```
underdarknl: We can construct urls from the type and pk, we should also make sure to include the events valid_time when linking to those urls from anywhere in the audit log ui.

rocky/crisis_room/migrations/0009_auditlog.py
```diff
@@ -0,0 +1,60 @@
...
+                            ("plugin_disabled", "Disabled plugin"),
+                            ("plugin_settings_changed", "Changed plugin settings"),
+                            ("plugin_settings_deleted", "Deleted plugin settings"),
+                        ],
+                        max_length=32,
+                    ),
+                ),
+                ("object_type", models.CharField(blank=True, max_length=64)),
+                ("object_label", models.CharField(blank=True, max_length=512)),
+                ("object_url", models.CharField(blank=True, max_length=2048)),
```
underdarknl: See other remarks, but should not be here. 2k char fie;ds is killing performance, and on top might not be enough for the actual url.

underdarknl (changes requested):

hasecon (changes requested):
## QA Review (2nd reviewer — hasecon)

Reviewed at head `11188a83`, **with @underdarknl's review taken into account** as requested. This is a re-review, so I mapped every one of Jan's 11 inline comments to its current status first, then looked for additional issues.

### Status of @underdarknl's CHANGES_REQUESTED (11 comments)

All 11 threads are still `is_outdated: false` and the commented code is unchanged at head — **none of the feedback has been addressed yet.**

| # | Location | Point | Status |
|---|---|---|---|
| 1 | account/mixins.py:182 | Merge `AuditLog.record` + `logger.info` into one env-configurable function; unify `event_code` and `action` into one enum, used everywhere events are emitted | ❌ open |
| 2 | migration 0009:17 | Question the `id` field; add rolling tamper-evident hashing of the previous record | ❌ open |
| 3 | migration 0009:18 | `actor_label` is inferable from `actor` (null = system); the text column limits scalability | ❌ open |
| 4 | migration 0009:20 | `action` should link to the `event_codes` enum / a lookup table | ❌ open |
| 5 | models.py:42 | Give the index a logical name (not the hashed `…_875530_idx`) | ❌ open |
| 6 | models.py:37 | Don't store the URL; store PK + type and infer the rest — 2 k varchar limits scalability and may not hold the URL | ❌ open |
| 7 | urls.py:8 | Rename the routes (`auditlog/` / `audittrail/`) so they don't collide with `logs` | ❌ open |
| 8 | views.py:342 | Use existing pagination/querying instead of `[:50]` | ❌ open |
| 9 | ooi_delete.py:29 | Record the OOI primary key, not `human_readable` | ❌ open |
| 10 | account/mixins.py:188 | Construct URLs from type+PK; include the event's `valid_time` when linking | ❌ open |
| 11 | migration 0009:38 | Same as #6 — 2 k `object_url` kills performance / may overflow | ❌ open |

The through-line of 3/6/9/10/11 is one decision: **store identifiers (PK + type), not rendered strings/URLs; infer labels and links at read time.** That's a data-model rework, so this stays CHANGES_REQUESTED.

### Additional findings (beyond Jan's)

| Severity | Location | Finding |
|---|---|---|
| **Medium** | `object_label` (512) / `object_url` (2048), every `record()` call site | Unbounded `human_readable` / detail-URL stored in fixed-width columns. Postgres raises `DataError: value too long` on overflow (it does not truncate), and `AuditLog.record()` runs **inline and un-guarded** in the action path — so an OOI whose `human_readable` exceeds 512 (or a detail URL exceeding 2048) turns a *successful* add/delete/clearance into a 500, **after** the octopoes mutation already happened. This is the concrete failure behind #6/#9/#11; storing PK + type removes it. |
| **Low** | every `record()` call site | Audit writes aren't isolated from the primary action (no `try/except`). Any audit-write failure breaks the user's action — audit logging should be best-effort. |
| **Low** | ooi_view.py `form_valid` | `action = OBJECT_UPDATED if hasattr(self, "ooi") else OBJECT_ADDED` — `hasattr(self, …)` is a fragile/implicit add-vs-update discriminator; an explicit flag reads clearer. |

### What's correct (keep as-is)
- **Indemnification onboarding crash fix** is sound: renaming the shadowing `_ = get_or_create(...)` to `_indemnification, created` restores gettext `_`, and `test_add_organization_indemnification` pins it — it asserts `object_label == "Scanning indemnification"`, which needs `_()` callable, so it fails on the pre-fix code. Good regression test.
- Org scoping on the general `AuditLogView` is correct (`organization__in=user.organizations`; `test_general_audit_log_only_shows_accessible_organizations` confirms other-org rows are hidden). Auth is enforced by the global `AuthRequiredMiddleware`, and the bare `TemplateView` matches the existing `CrisisRoomView` — no access-control regression.

### CI / stack
- **PR is `dirty`** — merge conflict with `main`; needs a rebase.
- CI is green but **stale** (ran 2026-08-12 on head `11188a83`, which now conflicts). Rocky `test (3.10–3.13)` passed then. Given the data-model rework + rebase still to come, I deferred the `make clean && make kat` functional run; it's worth doing once the model change lands and the branch is rebased so the audit rows are exercised against a real stack.

### Verdict
**CHANGES REQUESTED** — Jan's 11 design comments are all still open (data-model rework toward storing PK+type, not rendered strings), the branch is behind/conflicting, and the fixed-width-column write is a concrete 500 risk on the primary action. The indemnification fix and its regression test are good and can stay.

brennodewinter:
@underdarknl — understood. This PR is blocked on the soft-delete users change. Should I mark it as draft until that lands, or keep it open for review on the non-migration parts?

brennodewinter:
Rebased on main — the conflict in `plugin_enable_disable.py` was with the merged #5311 redirect fix; kept main's `HttpResponseRedirect` flow and re-attached `record_action`. django.pot regenerated (ordering only). 32 audit-log/org tests pass, pre-commit green.

hasecon (changes requested):
## QA Review — round 2

### Setup
- Branch: `organization-activity-logs` @ `44215c14` (reviewed in an isolated worktree)
- `cd rocky && make utest` in Docker — **490 passed, 1 skipped**
- `manage.py makemigrations --check` — **no changes detected**, migration `0009_auditlog` matches the model
- CI: **17/17 green**
- Each round-1 point re-verified against the code at this head, not against the commit messages

The data-model rework asked for in round 1 has actually happened. `object_pk` + `object_type` are stored and the label and URL are derived at read time, which is the change the whole first round was about. Below is every previous point with what I found.

### Round-1 findings — status

| # | Point | Status |
|---|---|---|
| J1 / J4 | Unify `action` with the `event_codes` enum + lookup table | **Open** — agreed follow-up in round 1, not a blocker |
| J2 | Tamper-evident rolling hash | **Open** — agreed follow-up in round 1 |
| J3 | Drop `actor_label` | **Deliberately not done — correct** (see below) |
| J5 | Logical index name | **Fixed** — `audit_log_org_created_idx` |
| J6 / J9 / J10 / J11 | Store PK + type, infer the rest; record the OOI PK, not `human_readable` | **Fixed** — `get_object_label()` / `get_object_url()` derive at read time |
| J7 | Rename routes | **Fixed** — `auditlog/` on both the org and non-org mount |
| J8 | Reuse pagination | **Partly** — the queryset is paginated, the UI is not. See finding 2 |
| J-valid_time | Include `valid_time` in links | **Fixed** — `get_object_url()` passes `created_at.isoformat()` |
| Ours, Medium | `object_label`(512)/`object_url`(2048) overflow → `DataError` → user's action 500s *after* the octopoes mutation | **Fixed** — both are `TextField` now, and `record()` is wrapped |
| Ours, Low | Audit write not isolated from the primary action | **Fixed** — `try/except` + `logger.exception`, returns `None` |
| Ours, Low | `hasattr(self, "ooi")` as add-vs-update discriminator | **Improved** — now `getattr(self, "ooi", None)`; still implicit, still Low, not worth holding the PR |

**On J3, for the record.** Round 1 pushed back on dropping `actor_label`, because `actor` is `on_delete=SET_NULL` and after a user is deleted the name can no longer be inferred. Keeping it was the right call and the implementation is correct — verified by actually deleting the user (the `OrganizationMember` has to go first, it `PROTECT`s the account):

```
after deleting the user: actor=None  actor_label='client@openkat.nl'
```

### New findings

| # | Severity | Where | Issue |
|---|----------|-------|-------|
| 1 | Medium | `views.py`, both audit views | `?page=` goes straight into `Paginator.page()` → 500 on any bad value |
| 2 | Medium | `audit_log.html` | No pagination UI at all, so the log stops at 50 entries |
| 3 | Medium | `models.py`, `organization` FK | Deleting an organization deletes its entire audit trail |
| 4 | Low | `models.py`, `get_object_url()` | Builds a link for a PK it could not parse |
| 5 | Low | `account/mixins.py`, bulk clearance | Stores a rendered, language-frozen label instead of PK + type |

#### 1. `?page=` 500s on anything unexpected

Both views do `paginator.page(self.request.GET.get("page", 1))`. `PageNotAnInteger` and `EmptyPage` are not `Http404` subclasses, so they propagate out of the view. Measured on both:

```
page=  'abc' -> PageNotAnInteger: That page number is not an integer
page=  '999' -> EmptyPage: That page contains no results
page=    '0' -> EmptyPage: That page number is less than 1
page=   '-1' -> EmptyPage: That page number is less than 1
```

`paginator.get_page(...)` is the drop-in: it clamps out-of-range to the last page and falls back to page 1 on garbage. One word per view.

#### 2. The pagination has no controls

`paginate_by = 50` is set and the queryset is paginated, but `audit_log.html` contains no pagination markup at all (no `page_obj`, no next/previous, no include). So the page renders the newest 50 entries and there is no way to reach the rest — you would have to type `?page=2` by hand, and `?page=<past the end>` hits finding 1.

For an audit log specifically, silently showing only the newest 50 with no indication that more exist is the part I would least like to ship. Rocky already has the piece: `rocky/rocky/templates/partials/list_paginator.html`, used by `oois/ooi_list.html`, `tasks/boefjes.html`, `tasks/normalizers.html` and two more. It expects the page under **`page_obj`**, so:

```python
context["page_obj"] = paginator.get_page(page_number)
```

```html
{% include "partials/list_paginator.html" %}
```

That is J8 as it was meant — reuse, rather than a second implementation.

#### 3. Deleting an organization erases its audit trail

`organization = models.ForeignKey(Organization, on_delete=models.CASCADE, ...)`. Measured:

```
audit entries after deleting the organization: 0
```

This is the same question as J3, one level up. The model goes out of its way to survive the deletion of the *actor* — `SET_NULL` plus a denormalized `actor_label` — and then lets the deletion of an *organization* take the whole record with it. For an ordinary table `CASCADE` is right; for the table that records who did what, "delete the organization and the history is gone" is a property worth choosing deliberately rather than inheriting from the default.

If you want the trail to outlive the org, the shape is the one already used for the actor: keep the FK nullable with `SET_NULL` and denormalize `organization_label`/`organization_code`. If you would rather keep `CASCADE`, that is a legitimate answer — but it should be a decision, and the general `AuditLogView` filtering on `organization__in` would need a rethink for orphaned rows. I have no strong opinion on which; I do think it should not be settled by the default.

#### 4. A link is built for an unparseable PK

`get_object_label()` guards its `Reference.from_str()` and falls back to the raw string, which is right. `get_object_url()` has the same `try/except` but `get_ooi_url()` does not validate, so nothing raises and a link is produced anyway:

```
unparseable pk -> label: 'niet-een-reference'
                  url:   '/en/test/objects/detail/?valid_time=...&ooi_id=niet-een-reference'
```

The template only renders the anchor `{% if entry.object_pk %}`, so any non-OOI PK that ends up stored becomes a dead link. Parsing once and returning `""` from `get_object_url()` when the parse failed would keep the two methods consistent.

#### 5. One label escapes the read-time rule

The bulk-clearance path in `account/mixins.py` does:

```python
object_label=str(_("%(count)s objects to L%(level)s")) % {"count": len(ooi_references), "level": level},
```

`str()` on a lazy string evaluates it there and then, so the acting user's language is frozen into the database and every later reader sees that language regardless of their own. Same for `str(_("System"))` in `record()`.

It is the one place that stores a rendered string while the rest of the PR carefully stores PK + type and renders at read time, and it is also the path with no `object_pk` at all, so the entry cannot say *which* objects were changed. Storing the count in its own column — or omitting the label and letting the action speak — would keep it consistent.

I could not demonstrate the language freeze end to end, because that msgid has no Dutch translation yet, so both languages currently render the same English string. The mechanism is visible in the code; the symptom appears the first time it is translated.

### Test coverage

`tests/test_audit_log.py` is good and pins the two things that matter most. `test_organization_audit_log` asserts the label is inferred from the PK and the URL is built at read time, so it fails if the rework is reverted. `test_general_audit_log_only_shows_accessible_organizations` uses a negative assertion (`assertNotContains(response, "Hidden plugin")`), which is the one that would catch a scoping leak — the right instinct.

I also checked the scoping question behind it: `user.organizations` excludes blocked memberships (`[m.organization for m in self.organization_members if not m.blocked]`), so a blocked member does not see that organization's trail. No gap there.

Not covered, in rough order of what I would add:

```python
def test_audit_log_survives_a_hostile_page_parameter(rf, client_member):
    """?page= comes from the URL; a bad value must not 500."""
    for value in ("abc", "999", "0", "-1"):
        request = setup_request(rf.get("crisis_room_audit_log", {"page": value}), client_member.user)
        assert AuditLogView.as_view()(request).status_code == 200


def test_audit_log_entry_outlives_its_actor(client_member):
    entry = AuditLog.record(
        user=client_member.user, organization=client_member.organization,
        action=AuditLog.Action.OBJECT_ADDED, object_type="Network", object_pk="Network|internet",
    )
    user = client_member.user
    client_member.delete()   # OrganizationMember PROTECTs the account
    user.delete()

    entry.refresh_from_db()
    assert entry.actor is None
    assert entry.actor_label == "client@openkat.nl"
```

The first fails on this head with `PageNotAnInteger`; the second passes and is worth having precisely because J3 asked for that field to be removed — it documents why it is still there.

### Test Results

| Check | Result |
|-------|--------|
| `make utest` | 490 passed, 1 skipped (exit 0) |
| `makemigrations --check --dry-run` | no changes detected |
| CI on `44215c14` | 17/17 green |
| `actor_label` after user deletion | preserved, `actor` set to `NULL` |
| audit entries after organization deletion | 0 remaining |
| `?page=abc / 999 / 0 / -1` | raises in both views |
| unparseable `object_pk` | label falls back, URL does not |

### Functional Test

No `make clean && make kat` this round. The change is a Django model, migration, two views and a template, with no service, contract or compose change; I ran the migration check and exercised both views, the model and the deletion behaviour against a real database in the test container. The browser half — whether the table reads well and where the (missing) pagination controls should sit — is what a running stack would add, and it depends on finding 2 being resolved first. Happy to run it once the pagination is wired.

### Verdict

**CHANGES REQUESTED**

This is a big step forward: the PK + type rework is done and done properly, the overflow crash is gone, the audit write is now best-effort, and `actor_label` survived the round-1 discussion for the right reason. What holds it is the pagination, the one round-1 point that came back half-finished — the queryset is paged but there are no controls, so the log silently ends at 50 entries, and any unexpected `?page=` value 500s. Both fixes are small: `get_page()`, name it `page_obj`, include `partials/list_paginator.html`.

Finding 3 I would rather have a decision on than a change: `CASCADE` on the organization means an audit trail disappears with the organization it describes, which sits oddly next to the care taken to make entries outlive their actor. Findings 4 and 5 are small and can ride along.

Follow-ups: J1/J4 (unify with `event_codes`) and J2 (tamper-evident hash chain) remain open by agreement from round 1 — say the word if you want either raised as its own issue so they are not lost when this merges.

hasecon:
Following up on finding 3 in my review above — I understated it, and having now traced the deletion path I want to correct that.

I wrote that `on_delete=CASCADE` on the `organization` FK was "a property worth choosing deliberately" and that I had "no strong opinion on which". That was too soft for what this table is.

**There is a scriptable deletion path, not just Django admin.** `OrganizationViewSet` is a `ModelViewSet` with a `destroy()` (`rocky/tools/viewsets.py:39`), mounted at `/api/v1/organization/`. Permissions come from the DRF default, `rocky.permissions.KATModelPermissions`, which maps `DELETE` to `tools.delete_organization`. So any account or API token holding that one model permission can issue:

```
DELETE /api/v1/organization/<id>/
```

and with `CASCADE` every `AuditLog` row for that organization is removed in the same transaction. Measured on this branch:

```
audit entries after deleting the organization: 0
```

The consequence is the one property an audit log exists to prevent: **the record of who did what can be destroyed by a single ordinary API call, including the record of the caller's own earlier actions in that organization.** Nothing is left behind — no tombstone, no orphaned rows, no trace that entries ever existed. From the outside, an organization that was deleted and one that never existed are indistinguishable.

That also undercuts the tamper-evidence goal behind J2. A rolling hash chain protects entries against modification, but a chain whose rows can be deleted wholesale by the subject of the audit does not deliver non-repudiation. It is worth settling this before the hash chain is built on top of it, because changing the FK afterwards means another migration over a table that is by then supposed to be immutable.

The model already contains the right pattern, applied one level down: `actor` is `SET_NULL` with a denormalized `actor_label` precisely so an entry outlives its subject. The same shape works here:

```python
organization = models.ForeignKey(
    Organization, null=True, on_delete=models.SET_NULL, related_name="audit_logs"
)
organization_code = models.CharField(max_length=32)   # denormalised, as actor_label is
```

`OrganizationAuditLogView` keeps filtering on the FK. `AuditLogView` would need `organization__in=... ` widened to also surface orphaned rows for users who may see them — which is a real design question, and the one I think is worth your decision rather than the FK semantics themselves.

If you would rather keep `CASCADE`, that is a defensible answer for a tool that is not itself the system of record — but then it deserves a comment on the field saying so, so the next person does not read the `actor` handling and assume the same intent applies here.

Not changing my verdict over this: the blockers remain findings 1 and 2 (the `?page=` 500 and the missing pagination controls). I am flagging it more firmly because "the audit trail is deletable by the party being audited" is a different class of problem from the rest of this list, and because @ehasekamp reads this work through a NIS2 lens where traceability that can be erased by a privileged caller is exactly the gap that gets asked about.

brennodewinter:
@hasecon — finding 3 addressed in `34500214a`. `AuditLog.organization` is nu `PROTECT` i.p.v. `CASCADE`, met:

- vroege 409 in `OrganizationViewSet.destroy()` — vóór de Octopoes/KAT-alogus deletes, dus geen half-verwijderde org;
- `delete_model`/`delete_queryset` overrides in `OrganizationAdmin` met een melding i.p.v. een kale `ProtectedError`;
- test `TestDestroyWithAuditLog` die de 409 en het behoud van org + logs pint.

`PROTECT` gekozen boven `SET_NULL`: `get_object_url()` leest `self.organization.code`, dus verweesde logs zouden sowieso crashen — en een audtrail die zijn organisatie kwijtraakt is weinig audtrail. Dit duwt org-verwijdering richting de soft-delete die @underdarknl eerder voorstelde.

PR #5323, opened 2026-08-13
Fix blank OOI graph by restoring the hydrated tree in the context

### LINKED ISSUE #5322: OOI Graph visualisation is blank since #5088 — the tree context is no longer hydrated
## Describe the bug

On an object detail page the **Tree** tab renders fine, but the **Graph** tab shows nothing. The d3 render starts and then aborts with a JavaScript error:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at truncateText (stringHelpers.js:38)
    at graph-d3.js:184        // .text(d => truncateText(d.data.display_name, 15))
    at ... graph-d3.js:111 (update) -> graph.js
```

So every node's `d.data.display_name` is `undefined`, and `truncateText` dereferences `text.length` on it.

## Root cause

`OOIGraphView` relies on `get_filtered_tree()` to hydrate the tree — `hydrate_branch()` is what adds `display_name`, `name`, `overlay_data` and `graph_url` to every node (`rocky/rocky/views/ooi_tree.py`). But `OOITreeView.get_context_data` puts the **raw, unhydrated** tree into the template context:

```python
# rocky/rocky/views/ooi_tree.py, get_context_data
context["tree"] = self.get_tree_dict()          # raw: no display_name/name/overlay_data/graph_url
```

`get_filtered_tree()` (and therefore the graph's `hydrate_tree()` override) is never called, so the graph receives a tree without `display_name` and crashes on the first node. The **Tree table** survives because it only uses `human_readable`, which `get_ooi_dict()` always sets.

Git-traced: this line was changed in **#5088** ("Add optimizations to various Octopoes queries"):

```diff
- context["tree"] = self.get_filtered_tree(self.get_tree_dict())
+ context["tree"] = self.get_tree_dict()
```

(#5088 is the same PR behind the report-findings regression #5202.)

**Side effect:** because `get_filtered_tree()` also applies the type/depth filters (`filter_ooi_tree`), the Tree view's OOI-type/depth **filters have also been silently broken since #5088**. Restoring the call fixes both.

## To reproduce

Open any object with children (e.g. a scanned `Hostname`) → **Graph** tab → blank, with the `truncateText` error in the console.

## Expected behavior

The graph renders the object tree.

## Fix

Restore the hydrated/filtered tree in the context:

```diff
- context["tree"] = self.get_tree_dict()
+ context["tree"] = self.get_filtered_tree(self.get_tree_dict())
```

Plus a defensive guard in `truncateText` (`stringHelpers.js`) against a missing string, and a regression test asserting the graph view's context tree carries `display_name`.

## Version

`main` (present since #5088, so also affects released 1.22.x — backport candidate).

hasecon:
### Changes

The **Graph** tab on an object detail page renders blank while the **Tree** tab works. The d3 render starts and crashes:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at truncateText (stringHelpers.js:38)
    at graph-d3.js:184   // .text(d => truncateText(d.data.display_name, 15))
```

`OOIGraphView` relies on `get_filtered_tree()` to hydrate the tree — `hydrate_branch()` is what adds `display_name`, `name`, `overlay_data` and `graph_url` to every node. Since **#5088** ("Add optimizations to various Octopoes queries"), `OOITreeView.get_context_data` put the **raw** tree in the context instead of the filtered/hydrated one:

```diff
- context["tree"] = self.get_filtered_tree(self.get_tree_dict())
+ context["tree"] = self.get_tree_dict()
```

So `get_filtered_tree()` (and thus `hydrate_tree()`) never ran, `display_name` was missing on every node, and the graph crashed on the first node. The Tree table survived because it only reads `human_readable`, which `get_ooi_dict()` always sets.

**Side effect this also restores:** `get_filtered_tree()` applies the OOI-type/depth filters (`filter_ooi_tree`), so the Tree view's filters have been silently broken since #5088 too.

This PR:
- passes the filtered, hydrated tree to the template again (the actual fix);
- guards `truncateText` against a missing string as defence in depth;
- adds a regression test asserting the graph view's context tree carries `display_name` on every node (the existing `test_ooi_graph` only checked for `human_readable`-derived text, which is present even with the bug — that's why #5088 slipped through).

### Issue link

Closes #5322

### QA notes

Open any object with children → **Graph** tab. Before: blank + `truncateText` error in the console. After: the graph renders. Verified the new test fails on the pre-fix line (`KeyError: 'display_name'`) and passes with the fix.

Since #5088 shipped in released versions, this is a **1.22 backport** candidate.


rocky/rocky/views/ooi_tree.py
```diff
@@ -54,7 +54,7 @@ def get_last_breadcrumb(self):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        context["tree"] = self.get_tree_dict()
+        context["tree"] = self.get_filtered_tree(self.get_tree_dict())
```
underdarknl, line 59: Since the graph 'tab' isn't a separate route, and we dont want to hydrate all these fields for the other tabs, lets introduce a new context var which uses the get_filtered_tree via a property. That way we only hydrate the tree when its used by the graph template when its actually being rendered.

Splitting up the various tabs of the object detail view is something thats on my whish list, as there's plenty of other stuff that we load on the context of the detail page thats only used on one of the tabs.

hasecon: Done — moved the hydration into a dedicated `graph_tree` context var (a deep-copied tree), so `tree` stays filter-only for the tree/summary tabs and only the graph view builds the graph-only fields (`display_name`, `overlay_data`, `graph_url`). Template now reads `graph_tree`, and the regression test asserts both that `graph_tree` is hydrated and that `tree` is not. Pushed in `4f5e734b2`.

Worth noting for the record: it was already graph-scoped before this — only `OOIGraphView` hydrated (via its `get_filtered_tree` override); `OOITreeView`/`OOISummaryView` do filter-only and `OOIDetailView` doesn't build the tree at all, and these are separate routes (`/objects/{graph,tree,summary,detail}/`), so the other tabs never paid for hydration. The new var just makes that separation explicit and keeps graph fields off the shared `tree`. (A true `SimpleLazyObject` wouldn't have worked here — `json_script` serializes eagerly and doesn't unwrap the lazy proxy — so scoping to the graph view is the reliable "only when rendered".) Fully on board with splitting the detail-view tabs up as a follow-up.

underdarknl (approved):

PR #5328, opened 2026-08-16
Add check-cookie bit assessing Cookie OOI security attributes

hasecon:
### Changes

**Rewritten on top of the merged Cookie OOI (#5331)** — this is phase 2 of the #213 plan, replacing this PR's earlier Set-Cookie header-string approach, per @underdarknl's note that the string bit was superseded by the Cookie model. The branch was rebuilt from current `main`; the old string-parsing implementation is gone.

Adds a `check-cookie` bit that consumes the parsed `Cookie` OOI and yields one `KAT-INSECURE-COOKIE` finding (risk medium) per insecure cookie. No string parsing — the headers normalizer already parsed the attributes onto the object, and the canonical Set-Cookie header value is redacted anyway.

Checks:
- missing **Secure** (cookie may travel over an unencrypted connection)
- missing **HttpOnly** (readable by client-side scripts → XSS)
- no valid **SameSite** value (sent with cross-site requests → CSRF)
- **SameSite=None without Secure** (rejected by modern browsers)
- **`__Host-` prefix** without Secure + host-only + `Path=/`, and **`__Secure-` prefix** without Secure (RFC 6265bis §4.1.3)

The issue list is emitted in a fixed order, so repeated observations produce a byte-identical Finding — same determinism guarantee as the Cookie OOI itself (#213's churn goal).

### Issue link

Phase 2 of the plan on #213 (phase 1 = #5331, merged). Does not close #213 (phase 3, the use-case normalizers, remains).

### QA notes

- Unit tests: `octopoes/tests/test_bit_cookie.py` — 9 cases (clean cookie, each missing attribute, SameSite=None ±Secure, both name prefixes, determinism), green in the octopoes container.
- Bit discovery verified: `get_bit_definitions()` picks up `check-cookie` (consumes `Cookie`, min_scan_level 1).
- **Functional, verified live on a full stack:** uploaded a `Set-Cookie: sid=tok123; Path=/` headers raw via bytes → headers normalizer minted `Cookie|sid|internet|apache.dvwa.cloud|/` → scan level 4 inherited from the hostname → the bit derived `Finding|Cookie|…|KAT-INSECURE-COOKIE` listing all three missing attributes. Findings page and Cookie detail page render (200).

---

### Code Checklist

- [x] All the commits in this PR are properly PGP-signed and verified.
- [x] This PR only contains functionality relevant to the issue.
- [x] I have written unit tests for the changes or fixes I made.
- [x] I have checked the documentation and made changes where necessary.
[...]

PR #5331, opened 2026-08-18
Add Cookie OOI and null volatile header values at ingest

hasecon:
### Changes

PR 1 of the plan agreed in #213 (cookies as OOIs + volatile-header normalization). Repeated observations of the same endpoint used to mutate the graph on every scan: session tokens, trace ids and rolling `Expires` dates changed the `HTTPHeader` values each time, producing a fresh event chain that carried no new information.

**Capture** — `kat_webpage_analysis` emits `openkat-http/headers` as a **list of `[name, value]` pairs** (via urllib3's `HTTPHeaderDict.getlist` on `response.raw`), so repeated `Set-Cookie` headers survive per RFC 6265 ("one cookie per header"). The normalizer accepts **both shapes** — dict (legacy raws in bytes stay re-normalizable, including a comma-resplit heuristic that survives `Expires` dates) and list (new) — so no migration is needed.

**Nulling** — the headers normalizer nulls the *values* of a curated set of volatile headers and cookies (`volatile_headers.json`: header names exact/case-insensitive, cookie names with glob support — `BIGipServer*` carries the pool name in the cookie name). Names, attributes and existence are always preserved. Set-Cookie is canonically reserialized (`name=<redacted>; <attrs in fixed order>`) rather than emptied — an empty value is the cookie-*deletion* idiom, a meaningful observation in itself. Rolling absolute `Expires` dates (the second churn source) are reduced to a sentinel in the header; the relative lifetime lives on the Cookie OOI, computed against the response's own `Date` header so re-normalization stays deterministic. Stable meaningful values (e.g. HAProxy `SERVERID=backend1` — real topology signal) are deliberately *not* nulled.

**Cookie OOI** — minted by the same normalizer (exactly one producer):
- identity `name + domain + path` — **not** the value, which is what made the old #317 model churn
- fields: `secure_only`, `http_only`, `same_site`, `host_only`, `persistent`, `lifetime` bucket (`session`/`<1d`/`<30d`/`>30d`/`>400d`), `value_size` bucket (`<1KB`/`1-4KB`/`>4KB`); the raw value is never stored, no timestamps
- `domain` is a `Hostname` reference with `max_inherit_scan_level=4`, so the phase-2 security bit will actually inherit a scan level
- parsing is lenient per RFC 6265 §5.2/§5.1.1 (browsers ignore a broken attribute and keep the cookie — a parse failure is not a finding)
- **`Domain` validation (§5.3 step 6):** a `Domain` that doesn't domain-match the request host is treated as host-only — without this, a scanned server could inject arbitrary `Hostname` OOIs into the graph; a `|` in name/path is rejected (natural-key safety)

### Issue link

Part of #213 (phase 1 of the consolidated plan posted there). Phase 2 (security bit on `Cookie`, superseding #5328) follows separately.

### Demo

For a response carrying `Date`, `CF-Ray`, `Server: nginx` and `Set-Cookie: sessionid=<token>; Path=/; Expires=<rolling date>; HttpOnly`, two observations that differ only in the token/ray-id/dates now produce a **byte-identical OOI set**:
- `HTTPHeader` `Date`/`CF-Ray` → `<redacted>`, `Server` untouched
- `HTTPHeader` `Set-Cookie` → `sessionid=<redacted>; Path=/; Expires=<redacted>; HttpOnly`
- `Cookie|sessionid|internet|<host>|/` with `http_only=True, persistent=True, lifetime=<bucket>, value_size=<1KB`

### QA notes

- Unit tests: `boefjes/tests/plugins/test_headers_cookies.py` (13 cases) — the load-bearing one is `test_volatile_values_produce_identical_oois` (two observations differing only in tokens → identical OOI set). Run: 13 passed in the boefjes CI container; full plugin suite 144 passed (no regressions); octopoes unit suite 161 passed with `Cookie` registered in `types.py`.
- RFC edge cases pinned: legacy joined multi-cookie resplit across an `Expires` comma, Max-Age-wins-over-Expires, broken `Expires` → attribute ignored (no finding), nameless cookie ignored, `Domain` mismatch → host-only, parent-`Domain` cookie mints the parent `Hostname`, pipe-in-name rejected, determinism.
- Functional: run `webpage-analysis` against a host that sets cookies; `Cookie` OOIs appear in the object list, `Set-Cookie`/`Date` header values show redacted, and re-running the boefje produces no new graph mutations.

---

### Code Checklist


boefjes/boefjes/plugins/kat_webpage_analysis/main.py
```diff
@@ -60,11 +60,24 @@ def run(boefje_meta: dict) -> list[tuple[set, bytes | str]]:
 
     return [
         ({"application/json+har"}, har.encode()),
-        ({"openkat-http/headers"}, json.dumps(dict(response.headers))),
+        ({"openkat-http/headers"}, json.dumps(get_header_pairs(response))),
```
underdarknl, line 63: Good catch! multiple headers with the same name are allowed indeed. However, does our primary-key in de OOI model allow for this? (even after nulling?)

boefjes/boefjes/plugins/kat_webpage_analysis/headers/normalize.py
```diff
@@ -1,14 +1,302 @@
...
+        # HTTPHeader identity is (resource, key), so all Set-Cookie headers fold
+        # into one OOI. Joining the canonical forms keeps that value stable.
+        yield HTTPHeader(resource=resource, key="Set-Cookie", value=", ".join(canonical_values))
+
+
+def parse_set_cookie(set_cookie: str) -> ParsedCookie | None:
+    """Parse one set-cookie-string per RFC 6265 5.2 (lenient, browser-like).
+
+    Broken attributes are ignored rather than rejected: browsers keep the
+    cookie, so a parse failure is not a finding.
```
underdarknl: I think a malformed cookie might warrant a 'warning' level finding, even though the browser might consume the cookie correctly, there's no guarantee that this is the case, and it might be masking or even allowing for some sort of value injection where some cookie parsers misread attributes and process the cookie in a way that should happen.
Lets introduce a warning when parsing tells us something is wrong.

boefjes/boefjes/plugins/kat_webpage_analysis/main.py
underdarknl: Found the answer, No, it does not. Maybe worthwhile to renew this once we get an easier way of dealing with model upgrades/changes

octopoes/octopoes/models/ooi/web.py
```diff
@@ -141,6 +141,46 @@ def format_reference_human_readable(cls, reference: Reference) -> str:
...
+    value would mint a new object per observation. The raw value is not stored
+    at all; only a coarse size bucket is kept as a performance signal.
+    See https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/issues/213.
+    """
+
+    object_type: Literal["Cookie"] = "Cookie"
+
+    name: str
+    domain: Reference = ReferenceField(Hostname, max_issue_scan_level=0, max_inherit_scan_level=4)
+    path: str = "/"
```
underdarknl, line 158: There's some possible deduction here where we can infer new urls on hosts by combining the domain and path attribute (if not a wildcard domain) to create new url's for said domain.

underdarknl (dismissed):

boefjes/boefjes/plugins/kat_webpage_analysis/headers/normalize.py
hasecon: Done in 9db3b9b22. The parser now records every construct it has to ignore — nameless set-cookie-string, `Path` not starting with `/`, non-integer `Max-Age`, unparsable `Expires` date, invalid `SameSite` value, a `Domain` that doesn't domain-match the request host, and pipe characters in name/path — and the normalizer aggregates them into one low-risk `KAT-MALFORMED-COOKIE` finding per resource.

Two deliberate choices, matching your masking/injection concern and the #213 churn goals: the description names only the problem *categories* (never cookie values, so nothing volatile or sensitive lands in the finding), and it's sorted so repeated observations produce a byte-identical Finding — pinned by `test_malformed_finding_description_is_stable_and_aggregated`. The cookie itself is still kept where browsers would keep it (lenient parsing unchanged); the finding is purely the warning layer. 16 tests green in the boefjes Docker image, finding type added to the catalog.

boefjes/boefjes/plugins/kat_webpage_analysis/main.py
hasecon: Agreed — the model's `(resource, key)` identity can't represent repeated same-name headers as separate OOIs today, which is exactly why this PR folds them into one `", "`-joined value (same shape the legacy dict raws already had, so nothing changes semantically and no data is lost). When a model-migration path exists, promoting HTTPHeader identity to `(resource, key, index)` or a multi-value field would be the cleaner representation; happy to write that up as a tracking issue on #213 so it doesn't get lost — say the word.

octopoes/octopoes/models/ooi/web.py
hasecon: Nice one — that's a natural bit on top of this model: `Cookie` (consumes) → for non-wildcard, host-only-or-domain-matched cookies, mint `URL(raw=f"https://{domain}{path}")`. All the inputs are already on the OOI (`domain` ref, `path`, `host_only`), and the domain-match validation in this PR guarantees the domain was actually owned by the responding host, so it can't become an injection vector. I'd keep it out of PR1 to hold the "one concern per PR" line (this one is capture+null+model), and ship it alongside the security bit in PR2 of the #213 plan — it's ~30 lines plus tests there. I'll add it to the PR2 scope on #213 unless you'd rather have it in here.

boefjes/boefjes/plugins/kat_webpage_analysis/main.py
underdarknl: A tracking issue would be good.

octopoes/octopoes/models/ooi/web.py
underdarknl: A follow up PR for this would be good.

underdarknl (approved):

Issue #4447, opened 2025-05-12
Browser plugin of Wapalyzer detects software version while OpenKAT does not


paulvandenbraken:
_Please add `bug`, the name of any relevant modules (e.g. `rocky`), and any other relevant labels to your issue._

**Describe the bug**
During our live pilot with our sector-kandidates we 'felt' that we got a low number of findings. Therefor we used the Wapalyzer browser plugin for a sample (one organization) of the declared objects to compare the detected software and versions of both tools.
We found the following diferences for this sample-set:

Bootstrap ui framework: 
- Wapalyzer-plugin detects versions (3.3.4, 4.0.0), OpenKAT does not detect a version.
- Wapalyzer-plugin detects version 3.3.4, OpenKAT does not detect this software at all

Youtube video player:
- Wapalyzer-plugin detects the software without a version, OpenKAT does not detect this software at all.

PWA,
Popper CDN,
CloudflareCDN
- Wapalyzer-plugin detects the software without a version, OpenKAT does not detect this software at all.

FancyBox js-library
- Wapalyzer-plugin detects versions 2.1.5, OpenKAT does not detect this software at all.

Moderizr  js-library
- Wapalyzer-plugin detects version 2.8.3, OpenKAT does not detect a version.
- Wapalyzer-plugin detects version 2.8.3, OpenKAT does not detect this software at all

jQuery  js-library
- Wapalyzer-plugin detects versions (1.9.1, 1.11.2), OpenKAT does not detect a version.
- Wapalyzer-plugin detects versions 1.11.2, OpenKAT does not detect this software at all

Preact  js-library,
Klaro cookie compliance
- Wapalyzer-plugin detects the software without a version, OpenKAT does not detect this software at all

Core-js  js-library
- Wapalyzer-plugin detects version 3.6.5, OpenKAT does not detect this software at all

**To Reproduce**

1. Install the wapalyzer browser plugin in your browser
2. Go to the specific url for the declared object (samplelist available for OpenKat devteam on request) 
3. Select the plugins button to make the sofware detection for this declared url visible
![Image](https://github.com/user-attachments/assets/0563d03e-0234-4d2b-830c-93121107ef9a)

4. Look-up the detected objects for the organisation and compare the software instances. 
5. See the difference

**Expected behavior**
To find all software-instances and versions correctly.

**Screenshots**
See above.

**OpenKAT version**
1.18.2

**Desktop (please complete the following information if relevant):**

- OS: OSX
[...]

PR #5332, opened 2026-08-18
Follow same-resource redirects in webpage analysis

hasecon:
### Changes

`kat_webpage_analysis` requests each URL with an explicit port (`get_uri` always appends `:443`/`:80`) and runs with `allow_redirects=False`. When a server **canonicalises the URL** (a 301 from `https://host:443/` to `https://host/`) or a cache/security layer 301s a URL **to itself**, the stored HAR captures only the empty redirect body. Every body-based analysis then sees nothing for that resource — most visibly **Wappalyzer**, which yields zero technology/CMS detections, and image extraction.

`do_request` now follows a redirect **only when its target is the same resource** — same scheme, host and path, with default ports normalised and query/fragment ignored — capped at a few hops. A redirect to a *different* URL (apex→www, http→https, `/`→`/other`) is still returned as the 301: the `oois_in_headers` bit turns its `Location` into a `URL` that is discovered and scanned as its own resource, so following it here would only duplicate that detection onto the wrong resource. The forced-IP HTTPS adapter is now mounted on `scheme://host` (without the port) so a followed port-less URL still resolves through the pinned IP.

### Issue link

Related to #4447 (OpenKAT detecting less software than the Wappalyzer browser plugin). This is one contributing cause — any scanned URL that returns a redirect at scan time gets no technology detection at all — but not the whole of #4447 (which also covers version-regex gaps and JS-rendered software the raw-HTML pipeline can't see), so this does not close it.

### Demo / how it was found

Live on `https://hasecon.nl` (a WordPress site). The first scan produced **zero** `Software` for it. The stored HAR showed why:

```
req=https://hasecon.nl:443/   status=301  Location=https://hasecon.nl/   (same resource, empty body)
```

Its LiteSpeed cache/security layer served a transient canonical 301 on the first OpenKAT hit. Feeding a body-carrying HAR of the same page through the real Wappalyzer normalizer detects **WordPress** (`cpe:2.3:a:wordpress:wordpress`) plus LiteSpeed Cache, RankMath SEO, jQuery, Google Analytics/Tag Manager, CookieYes — 11 `SoftwareInstance`s. With this change, the same-resource 301 is followed to the 200 body so that detection is not lost.

### QA notes

- Unit tests: `boefjes/tests/plugins/test_webpage_analysis_redirects.py` (7 cases) exercise `do_request` with a mock session — the exact stored-HAR scenario (`:443`→port-less canonical redirect) is `test_follows_canonical_same_resource_redirect_dropping_default_port`, plus self-redirect-with-query, non-following of apex→www and http→https, a bounded redirect loop, and `is_same_resource`. `test_bodyimage.py` (existing `main.run` tests) still pass — 10 passed, no regressions. pre-commit green.
- The live trigger is transient (hasecon's cache warmed and now returns 200), so an on-demand live repro of the *followed* 200 isn't stable; the stored HAR captured the failing scenario and the unit tests pin the fix for it.

---

### Code Checklist

### DIFF
--- a/boefjes/boefjes/plugins/kat_webpage_analysis/main.py
+++ b/boefjes/boefjes/plugins/kat_webpage_analysis/main.py
@@ -73,7 +73,6 @@ def run(boefje_meta: dict) -> list[tuple[set, bytes | str]]:
 
 def do_request(hostname: str, session: Session, uri: str, useragent: str):
     headers = {"Host": hostname, "User-Agent": useragent}
-    response = session.get(uri, headers=headers, verify=False, allow_redirects=False)
 
     # Follow a redirect only when it points back to the same resource — the
     # server canonicalising the URL (e.g. dropping the explicit :443, or a
@@ -84,8 +83,15 @@ def do_request(hostname: str, session: Session, uri: str, useragent: str):
     # as a 301: the oois_in_headers bit turns its Location into a URL that is
     # discovered and scanned as its own resource, so following it here would
     # only duplicate that detection onto the wrong resource.
-    seen = {uri}
-    while response.is_redirect and len(seen) <= MAX_SAME_RESOURCE_REDIRECTS:
+    target = uri
+    seen: set[str] = set()
+    while True:
+        response = session.get(target, headers=headers, verify=False, allow_redirects=False)
+        seen.add(target)
+
+        if not response.is_redirect or len(seen) > MAX_SAME_RESOURCE_REDIRECTS:
+            break
+
         location = response.headers.get("location")
         if not location:
             break
@@ -93,9 +99,6 @@ def do_request(hostname: str, session: Session, uri: str, useragent: str):
         target = urljoin(response.url, location)
         if target in seen or not is_same_resource(uri, target):
             break
-
-        seen.add(target)
-        response = session.get(target, headers=headers, verify=False, allow_redirects=False)
 
     return response
 
--- a/boefjes/boefjes/plugins/kat_webpage_analysis/main.py
+++ b/boefjes/boefjes/plugins/kat_webpage_analysis/main.py
@@ -2,7 +2,7 @@ import ipaddress
 import json
 import mimetypes
 from os import getenv
-from urllib.parse import urlparse, urlunsplit
+from urllib.parse import urljoin, urlparse, urlunsplit
 
 import requests
 from forcediphttpsadapter.adapters import ForcedIPHTTPSAdapter
@@ -12,6 +12,10 @@ from requests import Session
 from boefjes.plugins.kat_webpage_analysis.har.requests import create_har_object
 
 ALLOWED_CONTENT_TYPES = mimetypes.types_map.values()
+
+DEFAULT_PORTS = {"http": 80, "https": 443}
+# Cap on same-resource redirect hops (see do_request).
+MAX_SAME_RESOURCE_REDIRECTS = 5
 
 
 def run(boefje_meta: dict) -> list[tuple[set, bytes | str]]:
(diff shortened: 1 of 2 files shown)


boefjes/boefjes/plugins/kat_webpage_analysis/main.py
```diff
@@ -66,13 +72,51 @@ def run(boefje_meta: dict) -> list[tuple[set, bytes | str]]:
...
+    # as a 301: the oois_in_headers bit turns its Location into a URL that is
+    # discovered and scanned as its own resource, so following it here would
+    # only duplicate that detection onto the wrong resource.
+    target = uri
+    seen: set[str] = set()
+    while True:
+        response = session.get(target, headers=headers, verify=False, allow_redirects=False)
+        seen.add(target)
+
+        if not response.is_redirect or len(seen) > MAX_SAME_RESOURCE_REDIRECTS:
```
underdarknl, line 92: Can we move this to where we now do the while True? (via a range() call)
I'd rather we not have endless loops, even if we have provisions to break from them.

boefjes/boefjes/plugins/kat_webpage_analysis/main.py
```diff
@@ -66,13 +72,51 @@ def run(boefje_meta: dict) -> list[tuple[set, bytes | str]]:
...
+            break
+
+        location = response.headers.get("location")
+        if not location:
+            break
+
+        target = urljoin(response.url, location)
+        if target in seen or not is_same_resource(uri, target):
+            break
```
underdarknl, line 102: I think we should also collect all visited urls in the output (are are these already listed in the header output?)

brennodewinter (changes requested):
## Review — PR #5332: Follow same-resource redirects in webpage analysis

De discussie in deze thread (20–21 aug) is duidelijk: **de boefje moet geen redirects volgen**. De huidige code op de branch implementeert nog steeds de oorspronkelijke redirect-following aanpak die in die discussie is verworpen. De branch is sinds het "repurpose"-besluit van underdarknl (21 aug) niet meer bijgewerkt.

### Stand van zaken

De graph-native keten bestaat al end-to-end:
- `Location` header → `oois_in_headers` bit → `URL` + `HTTPHeaderURL`
- `url_classification` bit → `Hostname` + `HostnameHTTPURL`

Het redirect-doel is dus al een first-class object in de graph. Het enige dat de loop niet sluit is **clearance**: het doel-URL/HostnameHTTPURL zit op L0–L1, `webpage-analysis` heeft L2 nodig. Dat is opzettelijk (`max_inherit_scan_level=0` op `HTTPHeaderURL.url`) — een gescande host mag niet via `Location: https://<anything>/` een willekeurig extern doel auto-clearance geven.

### Wat deze PR nu doet (verworpen)

De PR voegt redirect-following toe aan `do_request` in `kat_webpage_analysis`: volgt een redirect alleen bij "same-resource" (zelfde scheme/host/path, default ports genormaliseerd), gemaximeerd tot een paar hops. Dit is in de discussie verworpen omdat:
1. De boefje hoort één download te doen met `allow_redirects=False`.
2. Het redirect-doel al via de graph ontdekt wordt.
3. Technology-detectie op het gevolgde 200-body de Software aan het verkeerde object (oorspronkelijk aangevraagde URL) zou koppelen in plaats van aan het doel-URL waar het echt thuishoort.

### Wat de branch zou moeten worden

Volgens underdarknl (21 aug): repurpose voor een octopoes clearance-level change:
- **boefje:** redirect-following eruit, terug naar single download met `allow_redirects=False`.
- **octopoes:** een scoped clearance rule zodat een in-zone redirect-doel L2 bereikt en de scheduler hem her-scan. Out-of-zone doelen blijven walled. De "same-zone" check is het equivalent van de "same-resource" judgement die deze PR in de boefje probeerde te doen — maar dan als clearance rule in de graph.
- Mogelijk: `max_inherit_scan_level` uitbreiden zodat het niet alleen een `int` accepteert maar ook een lambda die both given en receiving OOI als input krijgt (om zone-similarity te kunnen uitdrukken).

### Conclusie

CHANGES_REQUESTED. De huidige aanpak (redirect-following in de boefje) is in de discussie verworpen door zowel underdarknl als hasecon. De branch moet omgebouwd worden naar de octopoes clearance-level change zoals besproken. De huidige code kan niet gemerged worden — het zou technology-detectie mis-attribueren en de graph-native aanpak omzeilen.

De PR-description en titel moeten ook bijgewerkt worden zodra de branch gerepurposed is.

Issue #3916, opened 2024-11-28
Webpage capture boefje is broken


ehasekamp:
`bug`, `boefjes`,`webpage capture boef`

**Describe the bug**
The webpage capture boefje is broken. Might be related to #1190

**To Reproduce**

1. Go to KAT-alogus --> Webpage Capture
2. Click on Start Scan

**Expected behavior**
No Failed scan

**Screenshots**

024-11-28T19:50:08.861231058Z npm warn exec The following package was not found and will be installed: playwright@1.49.0
2024-11-28T19:50:10.770969195Z Error: t.parse: Executable doesn't exist at /ms-playwright/chromium_headless_shell-1148/chrome-linux/headless_shell
2024-11-28T19:50:10.770986205Z ╔══════════════════════════════════════════════════════════════════════╗
2024-11-28T19:50:10.770990795Z ║ Looks like Playwright Test or Playwright was just updated to 1.49.0. ║
2024-11-28T19:50:10.770993925Z ║ Please update docker image as well.                                  ║
2024-11-28T19:50:10.770996835Z ║ -  current: mcr.microsoft.com/playwright:v1.46.1-jammy               ║
2024-11-28T19:50:10.770999655Z ║ - required: mcr.microsoft.com/playwright:v1.49.0-jammy               ║
2024-11-28T19:50:10.771002546Z ║                                                                      ║
2024-11-28T19:50:10.771005696Z ║ <3 Playwright Team                                                   ║
2024-11-28T19:50:10.771009216Z ╚══════════════════════════════════════════════════════════════════════╝
2024-11-28T19:50:10.779776181Z npm notice
2024-11-28T19:50:10.779796441Z npm notice New minor version of npm available! 10.8.1 -> 10.9.1
2024-11-28T19:50:10.779801902Z npm notice Changelog: https://github.com/npm/cli/releases/tag/v10.9.1
2024-11-28T19:50:10.779806052Z npm notice To update run: npm install -g npm@10.9.1
2024-11-28T19:50:10.779809872Z npm notice

**OpenKAT version**
`v1.17.0` and `0.0.1.dev1` main branch

**Desktop (please complete the following information if relevant):**

- OS: macOS
- Browser: Chrome
- Version: 

**Additional context**

zcrt:
[...]

PR #5340, opened 2026-08-19
Run Wappalyzer over the browser capture HAR

hasecon:
### Changes

OpenKAT's tech detection already works — it just runs on the wrong data. `kat_wappalyzer` analyses the **static** `requests` HAR from `kat_webpage_analysis` (no JS executed, only the main document), so it detects software but often not versions and misses anything loaded after JavaScript. Meanwhile `kat_webpage_capture` (fixed in #5333) already produces a full **browser HAR** via Playwright — post-JS, every loaded resource — and nothing analysed it.

This adds a normalizer that runs the same Wappalyzer engine over that browser HAR:

- Extract the shared detection logic (`HAR → Software/SoftwareInstance`) out of `kat_wappalyzer/normalize.py` into `_analysis.software_from_har`, so both normalizers share one code path.
- New `kat_wappalyzer_capture` normalizer consuming `application/har+json`. It unzips the Playwright HAR (`har.har` + response bodies stored as `content._file`) via `har_zip.har_from_playwright_zip`, then runs the analysis.
- **Scan-level scoping** (per @underdarknl on #3916): SoftwareInstances attach to the input `HostnameHTTPURL`/`IPAddressHTTPURL` **only**; no `Hostname`/`URL` OOIs are minted for third-party hosts the browser loaded — the same footprint the existing normalizer already has, so nothing gets created for uncleared hosts.

Design + scope discussed on #4447.

### Demo / result

Ran the full pipeline in a live stack against `https://hasecon.nl` (WordPress): capture → browser HAR → this normalizer → **14 SoftwareInstances** on the site's URL. The browser HAR surfaces the server-side stack **with versions** that the static single-document HAR misses:

| From the browser HAR (new) | |
|---|---|
| Apache HTTP Server | **2.4.67** (cpe) |
| PHP | **8.3.31** (cpe) |
| Debian | OS |

…alongside WordPress, jQuery, LiteSpeed Cache, RankMath SEO, Google Analytics / Tag Manager / Fonts, CookieYes, Open Graph, RSS, Prototype.

### Honest limitation

The browser HAR is network traffic — it has the real loaded (versioned) resource URLs the static fetch misses, but not runtime JS variables like `jQuery.fn.jquery`. On sites that combine/minify JS (LiteSpeed strips the per-library versioned URLs) the JS-library version still won't come through; full parity with the browser Wappalyzer plugin would need running the Wappalyzer JS in-page (`page.evaluate`) — a separate, bigger step. This is the cheap 80%.

### Issue link

Part of #4447. Depends on #5333 (capture fix, merged).

### QA notes

- Unit tests: `test_wappalyzer_capture.py` (attaches Software + versions to the input URL, no third-party OOIs), `test_wappalyzer_har_zip.py` (zip → HAR inlining, missing-body resilience, plain-HAR passthrough). The existing `test_wappalyzer_normalizer.py` still passes — the refactor is behaviour-preserving. Full boefjes plugin suite: 138 passed. pre-commit green.
- Functional: verified live — the normalizer ran in the stack and produced the Software above; the fixed capture image (#5333) produces the browser HAR it consumes.

---

### Code Checklist

### DIFF
--- a/boefjes/boefjes/plugins/kat_wappalyzer/_analysis.py
+++ b/boefjes/boefjes/plugins/kat_wappalyzer/_analysis.py
@@ -1,0 +1,102 @@
+"""Shared Wappalyzer analysis: turn a HAR into Software / SoftwareInstance OOIs.
+
+Used by both the `kat_wappalyzer` normalizer (static requests HAR from
+webpage-analysis) and the `kat_wappalyzer_capture` normalizer (Playwright browser
+HAR from webpage-capture). The caller passes the reference the SoftwareInstances
+should attach to, so no OOIs are minted for third-party hosts the page loaded.
+"""
+
+from collections.abc import Iterable
+from pathlib import Path
+from typing import cast
+
+import httpx
+from tanimachi import (
+    Categories,
+    Fingerprints,
+    Groups,
+    Har,
+    Wappalyzer,
+    analyze_css,
+    analyze_headers,
+    analyze_scripts,
+    analyze_url,
+    schemas,
+)
+from tanimachi.wappalyzer import (
+    Detection,
+    HarWrapper,
+    analyze_cookies,
+    analyze_dom,
+    analyze_html,
+    analyze_meta,
+    is_html,
+)
+
+from boefjes.normalizer_models import NormalizerOutput
+from boefjes.plugins.kat_wappalyzer.utils import replace_cpe_version
+from octopoes.models import Reference
+from octopoes.models.ooi.software import Software, SoftwareInstance
+
+DATA_DIRECTORY = Path(__file__).parent / "data"
+
+
+def software_from_har(raw: bytes, target: Reference) -> Iterable[NormalizerOutput]:
+    """Yield Software + SoftwareInstance(ooi=target) for each Wappalyzer detection."""
+    fingerprints = Fingerprints.model_validate_pattern(DATA_DIRECTORY.joinpath("technologies/*.json").as_posix())
+    categories = Categories.model_validate_file(DATA_DIRECTORY.joinpath("categories.json"))
+    groups = Groups.model_validate_file(DATA_DIRECTORY.joinpath("groups.json"))
+    httpx.HTTPTransport()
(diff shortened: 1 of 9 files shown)


brennodewinter (dismissed):
## Review: Approve

Clean refactor that extracts the shared Wappalyzer analysis into `analysis.py` and adds a browser-capture normalizer that handles Playwright's `.har.zip` format.

### What I verified

**`har_zip.py`** — Correctly handles Playwright's sidecar body format (`content._file` → `content.text`). Graceful degradation: missing body files are silently skipped (`contextlib.suppress(KeyError)`), plain HAR passes through unchanged, non-Playwright zips pass through too. `_decode_body` falls back to latin-1 — never raises, maps every byte 1:1. Good.

**`analysis.py`** — The version extraction is improved over the old code: the `if match:` guard prevents a crash when `regex.search` returns `None` (the old code did `.expand()` directly on a potentially-`None` match). The `analyze_script_src_in_html` function is preserved unchanged.

**`kat_wappalyzer_capture/normalize.py`** — 15 lines, delegates to `software_from_har` with the input URL as target. SoftwareInstances attach to the input `HostnameHTTPURL`/`IPAddressHTTPURL` directly — no OOIs minted for third-party hosts. Correct scoping.

**Tests** — 3 new test files covering: capture normalizer output (software + instances + versions), HAR zip inlining (bodies, missing files, passthrough), and an integration test fix. Coverage is adequate.

### Non-blocking notes

- No zip-bomb protection in `har_zip.py`, but this is acceptable: the raw data is already stored in Bytes (size-limited at ingest), not read directly from an attacker.
- The `assert response.status_code == 200` in `test_api.py:174` is a good defensive addition.

CI: 18/18 green. Approving — needs rebase before merge (BEHIND main).

brennodewinter (approved):
Re-approving after rebase. No code changes — only conflict resolution in test_api.py (took the PR's own version: status_code 200 + len > 0).

underdarknl (approved):

PR #5342, opened 2026-08-20
Introduce an update to the task DB migration that makes sure the uniq…

underdarknl:
…ue contraint actually can be added. this sets old hanging jobs to failed.

### Changes

We noticed an uncompleted migration on the mula DB at one of our users. This was due to a bunch of uncompleted 'handing' tasks in their task list, which made the new unique contraint on those statuses impossible.
This change adds an update query which sets all handing jobs (older than 1 day) to FAILED. allowing the new index to be added

### QA notes

Insert some old tasks, or flip some successful old tasks from SUCCES to PENDING, after this the migration should fail without this update query, and complete successfully with this new update query.

### Code Checklist

---

---

## Checklist for QA:

### DIFF
--- a/mula/scheduler/storage/migrations/versions/0011_add_more_indexes.py
+++ b/mula/scheduler/storage/migrations/versions/0011_add_more_indexes.py
@@ -24,10 +24,27 @@ def upgrade():
     op.create_index("ix_schedules_scheduler_id", "schedules", ["scheduler_id"], unique=False)
     op.execute(
         """
+        WITH duplicate_active_tasks AS (
+            SELECT id
+            FROM (
+                SELECT
+                    id,
+                    ROW_NUMBER() OVER (
+                        PARTITION BY schedule_id
+                        ORDER BY created_at DESC, id DESC
+                    ) AS row_number
+                FROM tasks
+                WHERE status IN ('PENDING', 'QUEUED', 'DISPATCHED', 'RUNNING')
+                  AND schedule_id IS NOT NULL
+            ) ranked
+            WHERE row_number > 1
+        )
         UPDATE tasks
         SET status = 'FAILED'
-        WHERE status IN ('PENDING', 'QUEUED', 'DISPATCHED', 'RUNNING')
-          AND created_at < NOW() - INTERVAL '1 day'
+        WHERE id IN (
+            SELECT id
+            FROM duplicate_active_tasks
+        )
         """
     )
     op.create_index(
--- a/mula/scheduler/storage/migrations/versions/0011_add_more_indexes.py
+++ b/mula/scheduler/storage/migrations/versions/0011_add_more_indexes.py
@@ -22,6 +22,14 @@ def upgrade():
     op.create_index("ix_schedules_hash", "schedules", ["hash"], unique=False)
     op.create_index("ix_schedules_organisation", "schedules", ["organisation"], unique=False)
     op.create_index("ix_schedules_scheduler_id", "schedules", ["scheduler_id"], unique=False)
+    op.execute(
+        """
+        UPDATE tasks
+        SET status = 'FAILED'
+        WHERE status IN ('PENDING', 'QUEUED', 'DISPATCHED', 'RUNNING')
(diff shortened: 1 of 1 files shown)


hasecon (approved):
## QA Review

### Setup
- Branch: `fix/mula-migrations-cleanup-tasks` @ `728ab90f`
- Scope: one mula migration (`0011_add_more_indexes.py`) — adds a dedup `UPDATE` before the `ix_tasks_active_per_schedule` unique index. No API/UI/compose surface.
- Verification: static predicate check + live dedup repro against the scheduler Postgres (in a `BEGIN…ROLLBACK`, no real data touched). Full `make kat` not needed for a migration-only change; CI covers the clean-DB apply.

### CI/CD Status
All 19 checks green, including mula `test (3.10–3.13)`, `build (…, mula)`, `pre-commit`, `CodeQL`. `mergeable_state: behind` — update the branch against `main` before merge (standard, not a CI blocker).

### Code review — the fix is correct (verified two ways)

**Predicate match (the thing that matters):** the dedup's `WHERE status IN ('PENDING','QUEUED','DISPATCHED','RUNNING') AND schedule_id IS NOT NULL` matches the unique index's partial predicate `postgresql_where=status IN ('PENDING','QUEUED','DISPATCHED','RUNNING')` on `schedule_id` **exactly**. Those are precisely the four non-terminal `TaskStatus` values (COMPLETED/FAILED are terminal), and excluding NULL `schedule_id` is right — NULLs never conflict in a unique index. So the `UPDATE` removes exactly the rows that would otherwise break `CREATE UNIQUE INDEX`.

**Live repro** (dedup query on a temp table shaped like `tasks`, rolled back): a schedule with 3 active tasks → newest kept, older two → `FAILED`; a `COMPLETED` row untouched; a single-active schedule untouched; a NULL-schedule pair untouched. Result: ≤1 active per non-null schedule. Behaves as intended.

**Edit-in-place is safe here:** migrations run inside `context.begin_transaction()` and Postgres DDL is transactional, so an install where 0011 previously failed rolled back completely (revision not recorded) and re-runs the fixed 0011; installs that already succeeded had no duplicates and won't re-run; fresh installs get a no-op `UPDATE`. Modifying 0011 rather than adding 0012 is the right call, since the failed 0011 is exactly what needs to complete.

### Findings

| Severity | Finding |
|---|---|
| **Low/Medium** | **No test ships with a data-mutating migration.** The checklist marks "I have written unit tests," but the diff is only the migration (+25, 1 file), and CI's clean-DB run never exercises the dedup path — the actual bug scenario (duplicate active tasks) is untested. The logic is correct (verified above), so this isn't a blocker, but a production data migration deserves a pin. Draft below. |
| **Low** | **Description ↔ code mismatch.** The body/QA notes say it "sets all handing jobs (older than 1 day) to FAILED," but the SQL has **no age filter** — it fails all-but-the-newest active duplicate per schedule regardless of age. The code is actually *more* correct than the description (an age filter could leave two recent duplicates and still fail the index). Please update the description so QA doesn't test for an age cutoff that isn't there. |

### Suggested regression test
An integration test (mula uses Postgres via its storage layer) that pins the dedup — I verified this behaviour live; a wired test would fail against the pre-fix migration (index creation raises on duplicates) and pass after:
- Insert one schedule with ≥2 active tasks (`PENDING`/`RUNNING`) plus one `COMPLETED`, and a second schedule with one active task.
- Run the 0011 upgrade (or the dedup statement).
- Assert: exactly one active task remains per non-null `schedule_id`; the **newest** (by `created_at DESC, id DESC`) is the survivor; the older ones are `FAILED`; the `COMPLETED` and single-active rows are unchanged; the `ix_tasks_active_per_schedule` index now exists.

### Verdict
**APPROVED**

Correct, focused fix — the dedup exactly mirrors the unique index it unblocks, and it's verified statically and live. Two non-blocking asks before merge: add the integration test (the checklist claims one), and fix the "older than 1 day" wording in the description. Also update the branch against `main` (it's behind). Happy to stack the test as a fix-PR if that's easier.

PR #5345, opened 2026-08-24
Allow onboarding user to select existing organization

### LINKED ISSUE #4118: feature: onboarding, allow onboarding user to select existing organisation to be onboarded in if one exists
## About this feature

We should allow a user (especially those who dont have the rights to create organizations) to be onboarded in an already present organization.

The list of organizations can be selected from the appropriate already available methods on the user object.

At the moment, a user who want to do the onboarding flow *must* also create a new organization, with a unique name and code. This is not reasonable.

### Detailed description

Use the following method to check for available orgas

https://github.com/minvws/nl-kat-coordination/blob/e9130b6fdb008dde85e107a90808e8f2aa27f6a9/rocky/account/models.py#L109.

List them if available and allow the user to choose, create new orga, or continue with existsing orga.

Dont create a new unwanted orga anytime a new user gets onboarded.

## Design

### Screenshots

### Figma link

## Implementation

### Possible solution

_Outline your proposed solution for implementing the feature. You can include any specific ideas, designs, or functionalities here._

### Alternatives considered

_Describe any alternative approaches or solutions you've considered, and why you believe the proposed solution is superior._

brennodewinter:
## Summary

- Adds a new onboarding step (2a alternative) that lets a user pick an existing organization to onboard into, instead of being forced to create a new one
- Superusers who are already a member of one or more organizations are redirected to the selection page by the onboarding middleware
- `OnboardingOrganizationSetupView` redirects to the selection page when the user has multiple orgs, and directly to step 2b when they have exactly one (no unnecessary extra click)
- The create-new-org page now shows a link to the selection page when the user is already a member of an organization
- Includes 6 new tests covering the selection page, redirect logic, and middleware behavior

Closes #4118

#### Test plan

- [x] `pre-commit run --all-files` passes (ruff, mypy, djlint, prettier, etc.)
- [x] All 50 onboarding tests pass (`pytest tests/onboarding/test_onboarding.py`)
- [ ] Manual test: log in as a superuser who is a member of 2+ orgs and verify the onboarding redirects to the selection page
- [ ] Manual test: verify the selection page lists all orgs and submit redirects to step 2b for the chosen org
- [ ] Manual test: verify a superuser with 0 orgs still goes to the registration flow (create new org)


underdarknl (dismissed):

hasecon (changes requested):
## QA Review

Per @underdarknl's request I manually walked the onboarding paths on a live stack. The feature direction is good and the happy path works, but the manual walk surfaced three functional defects — one of which ends the onboarding wizard in a 500. All findings below were reproduced live, not just read from the diff.

### Setup
- Branch: `feature/onboarding-select-existing-organization` @ `8ff002ec9`
- Stack: `make clean && make kat` — clean build, all 13 containers up, Rocky reachable (200)
- Test org: `setup_test_org --code dvwa` — OK; extra orgs created directly to exercise non-member/blocked paths
- Onboarding paths driven with the Django test client as the superuser (member of dvwa, `onboarded=False`)

### CI/CD Status
All 19 checks green, including `makelang` (translations regenerated in the PR 👍), pre-commit, CodeQL, and the Rocky test matrix (3.10–3.13).

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| rocky | `make utest` (Docker) | ✅ 481 passed, 1 skipped (incl. the 6 new onboarding tests) |
| rocky | `make itest` (Docker) | ✅ 5 passed |

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Stack build | ✅ | clean |
| Middleware redirect | ✅ | superuser with an org membership → `/onboarding/step/organization-select/` as intended |
| Select page renders | ✅ | 200, lists orgs, create-link shown |
| Select → step 2b for chosen org | ✅ | POST redirects to `step_2b_organization_update` |
| Non-member org path | ❌ | see finding 1 |
| Blocked org path | ❌ | see finding 2 |
| "Create a new organization" link | ❌ | see finding 3 |
| Log analysis | ✅ | no new errors vs. main (only known startup noise) |

### Code Review Findings

| # | Severity | Finding |
|---|----------|---------|
| 1 | **High** | **Select list offers orgs the user is not a member of → 500 at the end of onboarding.** `get_form_kwargs` (`onboarding/views.py:145-147`) builds the queryset from `request.user.organizations`, but for a superuser that property returns **all** organizations in the database (`account/models.py:116-117`, via `can_access_all_organizations`) — and superusers are exactly the users the middleware routes here. Reproduced: created org `orgb` with no membership for the superuser → it appears in the select list → POST succeeds → 302 to step 2b. `OrganizationView.setup` then papers over the missing membership with an **unsaved** `OrganizationMember`, so every step works until step 10, where `set_member_onboarded` (`onboarding/views.py:441`) does `OrganizationMember.objects.get(user=…, organization=…)` → `DoesNotExist` → 500. Nothing is ever marked onboarded, so the middleware keeps bouncing the user back into onboarding. Fix: `Organization.objects.filter(members__user=self.request.user, members__blocked=False)` — one query, matches the docstring and the page copy ("organizations you are a member of"). |
| 2 | **Medium** | **Blocked orgs are listed and selectable → 403 mid-wizard.** Same root cause: for privileged users `user.organizations` doesn't apply the `blocked` filter, while the middleware and setup view do. Reproduced: superuser blocked in `orgx` → orgx appears in the list → selecting it redirects to step 2b, which raises `PermissionDenied` (403) via `account/mixins.py:120-121`. The queryset fix above covers this too. |
| 3 | **Medium** | **"Or create a new organization" on the select page can never reach the create form.** `OnboardingOrganizationSetupView.get` (`onboarding/views.py:92-98`) redirects any user with ≥1 unblocked membership away: with exactly 1 the link silently lands on step 2b of the existing org; with ≥2 it bounces straight back to the select page (reproduced both). Everyone who sees the select page has ≥1 membership by definition, so the button the template advertises is dead. Either drop the link, or let the setup view skip its redirect on an explicit signal (e.g. link to `…?new=1` and only redirect when the param is absent). |
| 4 | Low | **Middleware/setup-view threshold mismatch.** Middleware redirects at `count() > 0` (`rocky/middleware/onboarding.py:35`) while the setup view diverts to select only at `count() > 1` and sends single-org users straight to 2b. Result: the single-org shortcut the PR description promises ("no unnecessary extra click") is unreachable — the middleware intercepts first and shows the select page anyway (reproduced). Consider having the middleware redirect to `step_2a_organization_setup` and letting the view's existing branching route the user; that keeps the routing rule in one place. Bonus: the `.count() > 0` re-queries what `request.user.organization_members` (already materialized on line 11 of the middleware) can answer with zero extra queries; `.exists()` would be the queryset idiom otherwise. |
| 5 | Low | **Custom required-message is dead config.** `error_messages={"organization": {"required": …}}` (`onboarding/forms.py:62`) nests a field name where Django expects error-code keys. Confirmed live: an empty submit shows Django's default "This field is required.", never "Please select an organization to proceed." Should be `error_messages={"required": _("…")}`. (Same inherited wart as line 29 of the file.) |
| 6 | Low | **New tests only use superusers, so the membership filter is never actually asserted.** For a superuser the view's queryset is effectively `Organization.objects.all()`, so `test_step_2a_organization_select_shows_list` would pass even with the member-creation setup deleted — it can't detect findings 1–2. Also `{% if user.organizations %}` on `step_2a_organization_setup.html:26` reads as "user has memberships" but for the only users who ever see that template (privileged, membership-less) it means "any organization exists". |

### Suggested regression test
Drop-in for `rocky/tests/onboarding/test_onboarding.py`, pinning finding 1 (and its blocked-org sibling) once the queryset is fixed:

```python
def test_step_2a_organization_select_excludes_non_member_orgs(rf, superuser_member, organization_b):
    """The selection list only offers organizations the user actually has an
    (unblocked) membership in — not every organization in the database."""
    request = setup_request(rf.get("step_2a_organization_select"), superuser_member.user)
    response = OnboardingOrganizationSelectView.as_view()(request)
    assert response.status_code == 200
    assertContains(response, superuser_member.organization.name)
    assertNotContains(response, organization_b.name)  # no membership -> must not be offered


def test_step_2a_organization_select_excludes_blocked_orgs(rf, superuser_member, organization_b):
    from tools.models import OrganizationMember

    OrganizationMember.objects.create(
        user=superuser_member.user,
        organization=organization_b,
        status=OrganizationMember.STATUSES.ACTIVE,
        trusted_clearance_level=4,
        acknowledged_clearance_level=4,
        blocked=True,
    )
    request = setup_request(rf.get("step_2a_organization_select"), superuser_member.user)
    response = OnboardingOrganizationSelectView.as_view()(request)
    assertNotContains(response, organization_b.name)  # blocked membership -> must not be offered
```

I verified the failure mode these pin: on the current branch the select page demonstrably lists both a non-member org and a blocked org (live reproduction above), so both tests go red pre-fix and green with the membership-based queryset.

### Verdict
**CHANGES REQUESTED**

The core flow (middleware → select → 2b for a valid membership) works and is a real UX improvement. But finding 1 is a user-facing 500 at the very end of the wizard on a path this PR itself presents to its primary audience (superusers), with the user left stuck in an onboarding loop afterwards. Findings 1+2 share a one-line fix (membership-based queryset); finding 3 needs an author decision on how the create-form should stay reachable.

brennodewinter:
@underdarknl — done: `.devin/` added to `.gitignore` in e91661d49. Also merged current `main` in to resolve the BEHIND state.

@hasecon — all six findings from your QA review remain addressed; re-review requested.

PR #5347, opened 2026-08-26
Add unpacker normalizers that produce raw files from compound files

### LINKED ISSUE #5346: Normalizers that produce raw files to unpack compound files (HAR, zip, tar, ...)
## Concept

Based on a discussion in the team: let normalizers produce a **derived raw file** instead of (only) OOIs, so that compound files (HAR, zip, tar, gzip, odt, iso, rar, apk, ...) can be unpacked and their constituent mimetyped parts flow to the existing downstream normalizers automatically. "Do one thing, and do it well" per unpacker: decompression-to-a-known-format is itself a form of normalization.

Motivating example (HAR): one HAR-unpacker explodes the HTTP response bodies out of a HAR, and then **all** existing content normalizers (html, javascript, css, images, ...) run on those bodies without every normalizer needing to understand the HAR container.

## Why this fits the current architecture

The mula `NormalizerScheduler` is already **producer-agnostic**: it reacts to every `raw_file_received` event by looking up the normalizers that consume each mime type. So once a derived raw lands in Bytes with mime types, the downstream fan-out fires with no scheduler change. The only missing piece is a normalizer that can write a raw back.

## Phase 1 implementation (this PR)

To keep the first step small and mergeable, the derived raw reuses the **parent raw's `boefje_meta`** and is written back through the existing `POST /bytes/raw` path:

- **No Bytes or scheduler schema change, no migration.** The existing `RawFileReceived` event + mime-type fan-out + Rocky provenance keep working unchanged.
- Provenance trade-off: the unpack step is not recorded as a separate Octopoes Origin; downstream findings keep `source_method = <source boefje id>`. This is **consistent with how OpenKAT already treats normalizers as transparent** (`source_method` is always the boefje, never the normalizer). The unpack run itself stays auditable as a `NormalizerMeta` in Bytes.

Delivered:
- A `NormalizerRawFile` output type for normalizers (`NormalizerResults.raw_files`); the local runner collects them and the normalizer handler saves each one under the parent `boefje_meta`.
- A **HAR unpacker** (`application/json+har` → mimetyped response bodies).
- A **zip/tar/gzip archive unpacker**.
- Decompression-bomb hardening (file count / per-file / total size limits, bounded reads), never extracting to disk, member-name reduced to a basename, non-file tar members skipped, and **nested archives neutralised** so a nested bomb (e.g. 42.zip) cannot amplify across scheduler tasks.
- Unit tests for the guards, both unpackers, and the runner raw-file path.

## Deliberate follow-ups (not in this PR)

- **True recursive nested-archive unpacking with a hard depth cap.** Requires passing the raw's own mime types into the normalizer (the plugin loader currently allows exactly 2 parameters), so a depth counter can be carried in a tag. Until then nested archives are surfaced one level deep as opaque blobs.
- A first-class "raw produced by a normalizer" lineage in Bytes/Octopoes (making `boefje_meta` optional and adding a normalizer producer), if we later want the unpack hop to appear as its own provenance step.

Feedback welcome on the approach, and on whether the unpackers should keep the "normalizer" name or get a distinct plugin identity.

hasecon:
### Changes

Introduces a **raw-file output type for normalizers** so a normalizer can unpack a compound file into its constituent parts, which are stored back in Bytes and re-dispatched to downstream normalizers by mime type. One unpacker then feeds every existing content normalizer instead of teaching each of them a new container format.

Two unpackers are added on top of the mechanism:
- **HAR unpacker** (`kat_har_unpacker`): explodes the HTTP response bodies out of a HAR into separate mimetyped raws (`text/html`, `application/javascript`, `image/*`, ...).
- **Archive unpacker** (`kat_archive_unpacker`): unpacks zip/tar/gzip members.

**Approach (Phase 1 — small and non-invasive):** the derived raw reuses the **parent raw's `boefje_meta`** and goes through the existing `POST /bytes/raw` path, so there is **no Bytes/scheduler schema change and no migration**. The existing `RawFileReceived` event + mime-type fan-out (mula `NormalizerScheduler`, which is already producer-agnostic) do the downstream dispatch unchanged.

**Provenance trade-off:** the unpack step is not recorded as a separate Octopoes Origin; downstream findings keep `source_method = <source boefje id>`. This is consistent with how OpenKAT already treats normalizers as transparent (`source_method` is always the boefje, never the normalizer). The unpack run stays auditable as a `NormalizerMeta` in Bytes.

**Security hardening** (these unpackers process attacker-controlled content):
- Decompression-bomb limits: max file count, per-file size, total size, plus bounded reads (a gzip/zip bomb never decompresses unbounded into memory).
- Never extracts to disk; tar member names are reduced to a basename; non-file tar members (symlinks/devices) are skipped.
- **Nested archives are neutralised** (magic-byte sniff → stored as `application/octet-stream`, not an archive mime type) so a nested bomb such as 42.zip cannot amplify across scheduler tasks. Content-addressed discriminator tags mean identical extracted files deduplicate in Bytes, which also breaks the trivial zip-quine loop.

**Deliberate follow-up:** true recursive nested-archive unpacking with a hard depth cap needs the raw's own mime types passed into the normalizer (the plugin loader currently allows exactly 2 parameters). Until then nested archives are surfaced one level deep as opaque blobs.

### Issue link

Closes #5346

### Demo

Verified on a full local stack (`make kat`, DVWA test org). Injected a HAR raw via the Bytes client; the scheduler created a `kat_har_unpacker` task and the extracted bodies landed in Bytes and triggered downstream normalizers:

```
raw_file rows for the HAR's boefje_meta:
  <id> | {application/json+har}                                  # source HAR
  <id> | {har-resource/350770ffd1799c64, text/html}             # extracted body
  <id> | {application/javascript, har-resource/93f0d05c1fdeaf00} # extracted body
```

Injected a nested zip (zip-in-zip) to exercise the amplification guard:

```
raw_file rows for the zip's boefje_meta:
  <id> | {application/zip}                                        # source zip
  <id> | {archive-member/2cf24dba5fb0a30e, text/plain}           # readme.txt
  <id> | {nested-archive/df7da02e6445db7d, application/octet-stream}  # inner zip: NOT re-triggered
```

`kat_archive_unpacker` ran exactly once (the inner zip did not re-trigger it); `kat_har_unpacker` tasks matched the number of HAR raws 1:1 (no loop); task counts were stable over time; no container restarts and no unpacker/archive/raw errors in the logs.

### QA notes

- `cd boefjes && python -m pytest tests/plugins/test_unpack_common.py tests/plugins/test_har_unpacker.py tests/plugins/test_archive_unpacker.py` (16 tests) — covers the guards, both unpackers (base64 bodies, empty bodies, zip/tar/gzip, path-traversal names, nested-archive neutralisation) and the runner raw-file path end to end.
- Functional: on a running stack, run `webpage-analysis` on a hostname (produces `application/json+har`) and confirm new raws with body mime types appear in Bytes and that content normalizers run on them. Nested archives should appear as `application/octet-stream` and must not re-trigger `kat_archive_unpacker`.

### Code Checklist


brennodewinter (dismissed):
## Review: Approve

Well-designed unpacker mechanism with strong decompression-bomb and path-traversal protections.

### What I verified

**Decompression-bomb protection** — `UnpackGuard` enforces three limits: `MAX_FILES=1000`, `MAX_TOTAL_BYTES=256MB`, `MAX_FILE_BYTES=64MB`. Bounded reads (`read(MAX_FILE_BYTES + 1)`) detect oversized files without loading them fully. Each unpacker checks `guard.exhausted()` before every member and `guard.file_too_large()` before reading.

**Nested archive neutralization** — `_looks_like_archive()` sniffs magic bytes (PK, gzip, bzip2, xz, tar) independent of the attacker-controlled filename. A nested archive is re-emitted as `application/octet-stream` with a `nested-archive/` discriminator, NOT as an archive mime type — so the scheduler does not re-dispatch it to this unpacker. This prevents bomb amplification across scheduler tasks. The comment documents the deliberate follow-up for recursive unpacking with a depth cap.

**Path traversal** — `_safe_name()` reduces to basename. Members are never written to disk (all in-memory via `io.BytesIO`), but the sanitized name prevents leakage into mime-type guesses or logs.

**HAR unpacker** — Correctly handles base64-encoded bodies, skips empty bodies, respects guard limits. `normalize_mime_type` strips charset parameters. Content-addressed discriminator tag (sha256[:16]) enables Bytes deduplication of identical extracted files.

**NormalizerRawFile model** — Clean extension of `NormalizerOutput` union. `job_handler.py` saves raw files to Bytes under the parent raw's boefje_meta with per-file try/except — one failing file does not abort the task.

**Runner** — `_parse_results` routes `NormalizerRawFile` into `NormalizerResults.raw_files` via pattern matching. Integration test (`test_runner_collects_raw_files_and_no_observations`) verifies end-to-end through the real runner.

**Tests** — 252 lines across 3 test files. Coverage: zip/tar/gzip extraction, nested archive neutralization, guard limits (file count, total bytes, oversized single file), path traversal name reduction, HAR body extraction (base64, empty, guard), runner integration, discriminator uniqueness and content-addressing, mime-type normalization.

### Non-blocking notes
- The `test_api.py` change (`== 59` → `> 0`) is a good robustness improvement — the exact count is brittle.
- Recursive unpacking of nested archives is deliberately deferred (documented in the `_raw_file` comment). The current neutralization is the safe choice.

CI: 18/18 green (on pre-rebase commit). Approving — needs rebase before merge (CONFLICTING).

brennodewinter (approved):
Solid phase-1 design — the derived raws reuse the parent boefje_meta and flow through the existing `save_raw` → `raw_file_received` → mime-fan-out path, so no schema or scheduler changes were needed. A few things I verified while reviewing:

- **Loop safety**: neither unpacker can re-trigger itself — `kat_har_unpacker` consumes only `application/json+har` and extracted bodies carry content mime types; `kat_archive_unpacker` neutralises nested archives to `application/octet-stream` via magic-byte sniffing (name-independent). Bounded by the guard regardless.
- **Bomb bounds**: tar/gzip reads are explicitly capped (`read(MAX_FILE_BYTES + 1)`); zip members rely on the header `file_size` pre-check, which is effective because `ZipExtFile` bounds decompressed output to the declared size (`data[:self._left]`).
- **Dedup**: the content-addressed `har-resource/`/`archive-member/` discriminator makes each raw's mime-type set unique per boefje_meta, so Bytes stores and dispatches every extracted file while identical content dedups naturally.
- **Traversal**: member names are reduced to basename before mime guessing, and nothing is ever written to disk.
- Per-file `save_raw` failures are isolated so one bad member can't abort the task. 16/16 tests pass on the branch.

Two non-blocking observations: a bare `.gz` (nameless stream, not a tar.gz) extracts to `application/octet-stream` which no normalizer consumes — harmless dead-end, just limits what the gzip path buys you. And malformed HAR JSON fails the task via `ValidationError`, consistent with how other normalizers treat unparseable input.

Nice work — the documented provenance trade-off (source_method stays the boefje) is the right call for this phase.

brennodewinter (approved):
Re-approving after branch update (previous approval was dismissed by the main-merge). Review notes from earlier today still stand.

brennodewinter (approved):
Re-approving after branch update — change unchanged since my earlier review.

hasecon:
@underdarknl — sorry for the slow reply, this deserved an answer three weeks ago.

**The short answer is no, and you were pointing straight at the reason.** I went looking for the deletion-propagation problem you named and it is real. This PR should not be merged as it stands.

## What breaks

The HAR unpacker emits **one derived raw per HTTP response body** (`kat_har_unpacker/normalize.py`), and every derived raw is stored under the *parent* `boefje_meta` (`job_handler.py`: `self.bytes_client.save_raw(normalizer_meta.raw_data.boefje_meta.id, ...)`). That was the deliberate Phase-1 shortcut: no Bytes schema change, no new producer concept.

The consequence is in `Origin.id` (`octopoes/models/origin.py:33-39`):

```python
return f"{self.__class__.__name__}|{self.origin_type.value}|{self.method}|{self.source_method}|{self.source}"
```

`task_id` is not part of it. And because every derived raw carries the parent's `boefje_meta`, every downstream normalizer run built from it has the *same* `method`, `source` and `source_method`. So a capture whose HAR contains twelve HTML bodies runs the html normalizer twelve times, and all twelve write to one and the same Origin.

Measured against a real XTDB, two runs of one normalizer on two derived raws from one parent:

```
NA RUN 1 (derived raw #1)
   aantal origins : 1
   origin.id      : Origin|observation|kat_html|webpage-capture|Hostname|internet|example.com
   origin.result  : ['URL|internet|https://example.com/from-body-1']

NA RUN 2 (derived raw #2, zelfde parent boefje_meta)
   aantal origins : 1
   origin.id      : Origin|observation|kat_html|webpage-capture|Hostname|internet|example.com
   origin.result  : ['URL|internet|https://example.com/from-body-2']
```

One origin, and after the second run the first body's result is simply gone from it. `save_origin()` reassigns `origin.result` and upserts on `origin.id`, so each run replaces its predecessor. In a running stack the worker then handles the `OriginDBEvent` and cleans up what is no longer referenced — which is exactly the deletion propagation you asked about, doing precisely what it is supposed to do, to data that should have been kept.

Net effect: **of N bodies of the same content type in one HAR, only the last one processed leaves any OOIs behind.** Worse than merely incomplete, because it looks like it works — objects do appear, the task goes green, and nothing logs a problem.

I did not reproduce the final deletion end-to-end; the octopoes integration harness does not run the event worker, so in my run both URLs still existed as objects while the origin already referenced only the second. The collapse to a single origin and the replaced result set are measured; the deletion that follows from them I am inferring from `save_origin()` and the event handling.

## Why the answer to your other two parts is "not enough on its own"

- **Rocky**: I found nothing that needs changing, and I think that part of the design holds. `source_method` stays the originating boefje, so a finding derived from an unpacked body is still attributed to `webpage-capture` and the user follows provenance exactly as before. That was the intent and it works.
- **Octopoes**: this is where it fails, and it is not a missing feature — the model is behaving correctly. The mistake is on my side of the fence: reusing the parent `boefje_meta` gives derived raws an identity that Octopoes cannot tell apart.

## Where that leaves it

The Phase-1 shortcut is the thing that breaks, so the deferred lineage work is not optional after all. Roughly, in increasing order of honesty:

1. Make `source_method` discriminating for derived raws (e.g. `webpage-capture/har-unpack/<hash>`). No schema change, origins become distinct — but it overloads a field Rocky filters and displays on, and it encodes structure in a string. I do not like it.
2. Give each derived raw its own `boefje_meta`, with the parent recorded as its source. Distinct origins fall out naturally and provenance becomes explicit rather than inherited. This is the "first-class lineage" follow-up the issue lists, and it is a Bytes change.
3. Put `task_id` in `Origin.id`. Solves it generically but changes origin identity across all of Octopoes, so not something to smuggle in through this PR.

My inclination is 2, which means this PR as a mergeable unit does not survive — the boefjes-side plumbing here is still right, but it needs the Bytes lineage change underneath it rather than on top of it later.

I would rather hear your preference before I rebuild it. If you would rather see a smaller step first, I can strip this back to the guards and the `NormalizerRawFile` plumbing with **one** unpacker that emits a single derived raw (the archive case, where that is natural), leaving the HAR fan-out until lineage exists.

Either way I will convert this to a draft so it does not sit in the review queue looking ready. Thanks for the question — it caught a real defect that our own tests did not, because they all exercise a single derived file.

underdarknl:
So, how we currently track Origins is we make a combination of the input ooi (what did we look at) + the gathering tool (boefje, what did it return), and the normalizer (what did we turn that output into). as the recurring identifier for jobs.
As this chains reruns of the same combination of tools on the same input object togehter, this allows us to diff the outcome, and kick off garbage collection. Objects that where proven in earlier runs are now no longer present and deleted. Objects that only exists because we found them via such a new deleted object can also be deleted, as they are now dangling / no longer connected to the graph that contains the declared objects and their declared indemnifications.

This theory remains the same for longer tool chains, we would simply have to make sure we 'follow' the same set of tools for each run, however, now there might be multiples of the normalizer part. Eg, Input_ooi + boefje + normalizer name X + normalizer name Y + normalizer name Z.

This does give me another insight, currently, regardless of the unpacker we are trying to implement here, we might already be discarding yielded objects for boefje runs that container multiples of the same mime-typed raw files, and as such trigger the same normalizer. The most likely fix for this is making sure we start naming / indexing the raw files withing a boefje's output stream. This name should then be added to the combination (input_ooi + boefje + normalizer + rawfile name) to produce a stable recurring identifier for this task, while also allowing for multiple runs of the same normalizer with distict yielded objects. Instead of adding a hash of the raw file (which would break the recurring task chain) I propose we allow the boefje which is the most likely to know enough for raw-file labeling to do so. This name/path should be unique, and a first class possibly optional property of the raw file, next to the list of mime-types. (but not in the mime-type field). (There are feature requests, and UX designs for named raw files available already)
In the above example where we'd run multiple normalizers in a row, this could / would carry the ingested raw file's name for some or all of those normalizers. 

A har file could potentially be spat out by a boefje (A) on input_ooi (B).
This har file could then be read by  normalizer (C).

Origins currently would be concat of A, B and c (in some order).
However, this Pr could introduce har unpacker normalizer (D), which spits out:
* Raw file 1: http body (named for its url)
* Raw file 2: http headers (named for its url)
* Raw file 3: http body for the first included JS file (minified etc, named for its url)
* Raw file 4: http body for the second included JS file (minified etc, named for its url)
* Raw file 5: http body for the first included css file (minified etc, named for its url)
* etc.

Now based on the mime-types associated with those raw files, normalizer (E) could kick in and parse the html, F could kick in and normalize both JS files by unminimizing them, or splitting them into the various combined JS files (as per the manifest.json file or some other technique). 
For those last its vital to keep the url of the included JS as part of the origin's unique name to not overwrite the first's results by the second.

This does leave us with the cleanup, if a new scan on the object by this boefje produced new referenced JS files (by url), as that would imply new origin identifiers, requiring a technique to acknowledge the previous observations as no longer the newest / superseded by the new observations.

hasecon:
That is a much better answer than any of my three options, and it kills my `discriminator_tag` on both counts — it is a content hash, and it lives in the mime-type set. Point taken on both: a hash changes when the content changes, so reruns would never chain, which is precisely the property the origin identifier exists to provide. Naming from the producer, as a first-class field next to `mime_types`, is the right shape.

## Your new insight, measured

I went to check whether we are already discarding objects today, independent of this PR. **The hazard is real but currently latent: no shipped boefje triggers the same normalizer twice in one run.**

I cross-referenced every boefje's returned mime-type sets against every normalizer's `consumes`, looking for a normalizer reachable from two raws of the same run — that is the condition, not just "two raws with the same mime type". No hits.

Coverage and caveats, so you can judge how much that is worth:
- 40 of the 46 boefjes have a `main.py` I could read statically; the other six (`kat_nikto`, the three `kat_nmap_*`, two `kat_nuclei_*`) are OCI boefjes that emit a single raw.
- No boefje builds its raws in a loop, so the static read is the whole picture rather than a sample.
- `webpage-capture` is the closest call and still misses: `{image/png}`, `{application/zip+json, application/har+json}` and `{application/json, application/localstorage+json}` share no mime type that a single normalizer consumes.

So it is a gap in the model that nothing exercises yet, and this PR would be the first thing to walk into it. That is a better position than I feared — it means named raw files can be designed deliberately rather than shipped as an incident fix.

## The name is already available in both unpackers

Worth noting because it makes your proposal cheap on this side:

- **Archive**: the member name is already in hand, and `_safe_name()` already normalises it — today only to guess a mime type. Using it as the identity is taking something the code already computes.
- **HAR**: `entry.request.url` is right there in the parsed model; I simply never read it, only `entry.response.content`. That is exactly the "named for its url" in your example.

Both are stable across reruns, which is the property a hash lacked.

## On the cleanup problem you flagged

This is the part I think is genuinely unsolved, and I would rather not hand-wave it.

With `input_ooi + boefje + normalizer + rawfile name` as the identifier, a rerun that yields a *different set* of names produces a different set of origins. Run 1 emits `a.js` and `b.js`; run 2 emits `a.js` and `c.js`. The origin for `a.js` chains and diffs correctly. The origin for `c.js` is new and fine. But the origin for `b.js` is simply never visited again — nothing supersedes it, so its objects survive forever, and the garbage collection that makes the current scheme work never fires for them.

That is not a flaw in your naming proposal; it is the consequence of moving from a fixed identifier per tool chain to a variable-length one. The number of origins per run is now data-dependent.

What seems to be missing is one level of parentage: today garbage collection reasons about *objects* dangling from origins, and this needs *origins* dangling from origins. Concretely, the unpacker's own origin — the one keyed on `input_ooi + boefje + har-unpacker`, with no raw-file component, so it is stable across reruns — could carry the set of names it emitted as its result. A rerun then diffs that set exactly the way object results are diffed today: `b.js` disappears from the parent's result, and the child origins keyed on `b.js` become unreachable and collectable, taking their objects with them by the existing cascade.

That keeps one rule rather than two: everything is cleaned up when the thing that proved it stops proving it. It does mean the result of an origin is no longer necessarily a list of OOI references, which is the part I am least sure about and where I would want your read before anyone writes code.

## Proposal for how to carry this

This has outgrown the PR. I would like to write the above up properly on #5346 — named raw files as a first-class property, the identifier extension, and the parent-origin cleanup with the open question — and link the measurement above so the "is it already broken" question does not get re-asked. #5347 stays a draft in the meantime.

If you would rather see something land sooner, the archive unpacker is the natural first step: member names are already stable and computed, one raw per member, no HAR fan-out. That would exercise named raw files end to end on a small surface while the cleanup design settles.

Thanks for taking the time on this one — the answer is considerably better than the PR it replaces.

PR #5351, opened 2026-08-31
Rearchitect CSP checking into a parser bit and a policy bit (removes the ReDoS)

hasecon:
### Changes

Implements the two-bit CSP rearchitect you proposed on #4330 (comment 5475881374): separate parsing from policy, and **remove the regex-based `check_csp_header` bit** — which is what eliminates its ReDoS (`\S+\*\.\S{2,3}(...)`, exponential; a 60-byte header stalls the octopoes worker ~18s).

**Parser (`parse-csp`)** — consumes an `HTTPHeader`, and for a `content-security-policy` header parses it **deterministically** (split on `;`, then whitespace; plain string ops, no regex) into structured OOIs:
- `CSPDirective` (`header` + `name`) and `CSPSource` (`directive` + `value`), new in `web.py`, registered in the `WebType` union.
- Syntactic problems (empty header, directive without a value, duplicate directive) are flagged as a dedicated `KAT-CSP-INVALID` finding against the header (a separate type so it doesn't share a primary key with the policy findings).

**Policy (`check-csp-policy`)** — consumes the `HTTPResource` plus its `HTTPHeader`s and the `CSPDirective`/`CSPSource` OOIs (via `BitParameterDefinition`s) and evaluates the policy with plain set/string operations. This ports the checks from the old bit (missing `default-src`/`frame-ancestors`/`base-uri` and the `frame-src`/`script-src` fallback groups; `unsafe-inline`/`eval`/`hashes`; wildcard `*`; `http://`; blanket `http:`/`https:`; `data:` in script/object/default-src; private IPs; deprecated directives; `default-src` must contain `'none'`/`'self'`; `-uri` illogical values) onto the structured data — so the whole-string substring false positives of the old bit are gone too. It keeps the old bit's content-type XSS-capability gate (non-HTML resources are skipped). The required/deprecated directive lists and forbidden keywords are overridable by a `Config` OOI on the `Network`; the defaults reproduce today's behaviour. Findings stay on `KAT-CSP-VULNERABILITIES`.

**Removed:** the `check-csp-header` bit and its test — this deletes the ReDoS regex. `internetnl.py` and `kat_nikto` reference the `KAT-CSP-VULNERABILITIES` finding-type id (kept), not the bit, so they are unaffected.

[...]

PR #5363, opened 2026-09-03
Make the CSP policy configurable from Rocky via an ask-csp-policy question

hasecon:
### Changes

Follow-up 1 of 2 on @underdarknl's post-merge comment on #5351 ("inhoudelijk had ik graag een CSP config meegenomen over wat er wel en niet geaccepteerd is"):

* **New `ask-csp-policy` question bit** (consumes `Network`, like the other ask-bits): yields a `Question` with schema id `/bit/check-csp-policy`, so the policy config can be filled in from Rocky's question flow instead of requiring a hand-declared `Config` OOI. It exposes the three existing keys (`required_directives`, `deprecated_directives`, `forbidden_keywords`) plus the new one below, with the current defaults pre-filled.
* **New `allowed_hosts` config for `check-csp-policy`**: a comma-separated allowlist of hosts that are accepted as CSP sources (exact entries or `*.suffix` entries). When configured, any host source outside the list yields a finding line; quoted keywords/nonces/hashes and bare scheme sources are exempt, and wildcard sources are left to the existing wildcard check so they are not reported twice. Scheme, port and path are stripped and matching is case-insensitive. **An empty allowlist (the default) keeps today's behaviour** — nothing changes for organisations that don't configure it.

This complements the existing `disallowed_csp_hostnames` bit (a blocklist on the raw header) with the "what *is* accepted" side Jan asked for, evaluated on the parsed sources.

The second follow-up (cleaning up objects left behind by removed bits, the `recalculate_bits` TODO) comes as a separate PR.

### Issue link

Follow-up on #5351 / #4330 (comment thread).

### QA notes

- Unit: `test_ask_csp_policy.py` (schema id must equal the consuming bit id — that is what links the answered Config to the bit) and 6 new allowlist tests in `test_check_csp_policy.py` (no-config unchanged, enforcement, wildcard-entry matching, keyword/nonce/scheme exemption, scheme/port/path stripping + case, wildcard left to the wildcard check). All 25 CSP tests green in the container; enforcement test verified to fail on pre-change code.
- Live E2E on the dev stack: after `bits/recalculate` the `/bit/check-csp-policy` Question appears on `Network|internet`; declaring a CSP header with `https://cdn.example.com` + `https://fonts.evil.example` and a Config with `allowed_hosts=cdn.example.com` produced exactly `CSP source host(s) not in the configured allowlist: fonts.evil.example.` — the allowlisted host is accepted.

### DIFF
--- a/octopoes/bits/ask_csp_policy/ask_csp_policy.py
+++ b/octopoes/bits/ask_csp_policy/ask_csp_policy.py
@@ -1,0 +1,17 @@
+import json
+from collections.abc import Iterator
+from pathlib import Path
+from typing import Any
+
+from octopoes.models import OOI
+from octopoes.models.ooi.network import Network
+from octopoes.models.ooi.question import Question
+
+
+def run(input_ooi: Network, additional_oois: list, config: dict[str, Any]) -> Iterator[OOI]:
+    network = input_ooi
+
+    with (Path(__file__).parent / "question_schema.json").open() as f:
+        schema = json.load(f)
+
+    yield Question(ooi=network.reference, schema_id=schema["$id"], json_schema=json.dumps(schema))
--- a/octopoes/bits/ask_csp_policy/bit.py
+++ b/octopoes/bits/ask_csp_policy/bit.py
@@ -1,0 +1,6 @@
+from bits.definitions import BitDefinition
+from octopoes.models.ooi.network import Network
+
+BIT = BitDefinition(
+    id="ask-csp-policy", consumes=Network, parameters=[], min_scan_level=0, module="bits.ask_csp_policy.ask_csp_policy"
+)
--- a/octopoes/bits/ask_csp_policy/question_schema.json
+++ b/octopoes/bits/ask_csp_policy/question_schema.json
@@ -1,0 +1,30 @@
+{
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "$id": "/bit/check-csp-policy",
+  "type": "object",
+  "default": {},
+  "Content Security Policy Configuration": "Root Schema",
+  "required": [],
+  "properties": {
+    "required_directives": {
+      "description": "Comma separated list of CSP directives that must be present in every policy",
+      "type": "string",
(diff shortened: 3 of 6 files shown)


brennodewinter (changes requested):
## Review: Request changes

Good design — the `ask-csp-policy` question bit and the `allowed_hosts` allowlist are the right complements to the existing `disallowed_csp_hostnames` blocklist. The allowlist logic (`_source_host` + `_host_is_allowed`) is correct: scheme/port/path stripping, case-insensitivity, keyword/nonce/scheme exemption, and wildcard sources left to the existing wildcard check are all handled properly. The test coverage is thorough.

However, CI is red on two counts that need fixing before merge:

### Blocking: pre-commit (ruff-format)

`_source_host` in `check_csp_policy.py` has an extra blank line after the docstring that `ruff-format` removes:

```python
def _source_host(source: str) -> str | None:
    """..."""
                              # <-- this blank line
    if source.startswith("'"):
```

Fix: remove the blank line.

### Blocking: integration tests (test_ooi_deletion.py)

Two integration tests fail across all Python versions with `assert 9 == 8`:

- `test_events_created_in_worker_during_handling` (line 134)
- `test_deletion_events_after_nxdomain` (line 232)

Both create a `Network(name="internet")` and call `recalculate_bits()`. The new `ask-csp-policy` bit consumes `Network` and produces a `Question` OOI, bumping the OOI/event count from 8 to 9. The existing 3 ask-bits (`ask-disallowed-domains`, `ask-port-specification`, `ask-url-params-to-ignore`) are already accounted for in the current count of 8; this 4th one needs to be reflected.

Fix: update the two assertions from `== 8` to `== 9` (and the event index `event_manager.queue[7]` → `event_manager.queue[8]` in the first test).

### Non-blocking: question_schema.json

The field `"Content Security Policy Configuration": "Root Schema"` is not a valid JSON Schema keyword. The existing ask-bit schemas (e.g. `ask_disallowed_domains/question_schema.json`) don't have this. If this is meant as a title, use `"title": "Content Security Policy Configuration"` instead.

PR #5365, opened 2026-09-03
Fix report JSON export crash on organization tags (#4332)

### LINKED ISSUE #4332: Report JSON export fails  on organization tags
The report JSON export doesn't work if the organization has tags because the organization tag objects aren't json serializable.

```
  File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/views/generic/base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/views/generic/base.py", line 143, in dispatch
    return handler(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/rocky/reports/views/base.py", line 564, in get
    response = JsonResponse(response, encoder=JSONEncoder)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/http/response.py", line 733, in __init__
    data = json.dumps(data, cls=encoder, **json_dumps_params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
          ^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/encoder.py", line 258, in iterencode
    return _iterencode(o, 0)
           ^^^^^^^^^^^^^^^^^
  File "/app/rocky/reports/utils.py", line 29, in default
    return super().default(o)
           ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/core/serializers/json.py", line 106, in default
    return super().default(o)
           ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/json/encoder.py", line 180, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type OrganizationTag is not JSON serializable
```

brennodewinter:
## Summary
- The report JSON export (`?json=true`) crashed with `TypeError: Object of type OrganizationTag is not JSON serializable` when the organization had tags (#4332).
- Root cause: `ViewReportView.get` passed `list(self.organization.tags.all())` — `OrganizationTag` model instances — to `JsonResponse`, whose encoder only handles `OOI`s and dataclasses.
- Fix: serialize tags as their names (`[tag.name for tag in ...]`), matching the form the report runner itself already uses (`reports/runner/report_runner.py`).

## Test plan
- [x] Added `test_json_download_aggregate_report_with_organization_tags` — sets tags on the org, hits the JSON export, asserts `organization_tags == ["tag-a", "tag-b"]` and status 200.
- [x] Existing `test_json_download_aggregate_report` (no tags) still passes.

Closes #4332.

### DIFF
--- a/rocky/reports/views/base.py
+++ b/rocky/reports/views/base.py
@@ -562,7 +562,7 @@ class ViewReportView(ObservedAtMixin, OrganizationView, TemplateView, AddDashboa
             response = {
                 "organization_code": self.organization.code,
                 "organization_name": self.organization.name,
-                "organization_tags": list(self.organization.tags.all()),
+                "organization_tags": [tag.name for tag in self.organization.tags.all()],
                 "data": self.report_data,
             }
 
--- a/rocky/reports/views/mixins.py
+++ b/rocky/reports/views/mixins.py
@@ -109,7 +109,7 @@ class SaveMultiReportMixin(BaseReportView):
             template=report_type.template_path,
             organization_code=self.organization.code,
             organization_name=self.organization.name,
-            organization_tags=set(self.organization.tags.all()),
+            organization_tags={tag.name for tag in self.organization.tags.all()},
             data_raw_id=report_data_raw_id,
             date_generated=now,
             reference_date=observed_at,  # TODO: https://github.com/minvws/nl-kat-coordination/issues/4014
--- a/rocky/tests/reports/test_aggregate_report_flow.py
+++ b/rocky/tests/reports/test_aggregate_report_flow.py
@@ -1,4 +1,5 @@
 import json
+from unittest.mock import patch
 
 from pytest_django.asserts import assertContains
 from reports.views.aggregate_report import (
@@ -361,8 +362,9 @@ def test_json_download_aggregate_report_with_organization_tags(
     which raised TypeError. Tags should be serialized as their names, matching
     the form the report runner itself uses.
     """
-    client_member.organization.tags = ["tag-a", "tag-b"]
-    client_member.organization.save()
+    with patch("katalogus.client.KATalogusClient"), patch("rocky.signals.OctopoesAPIConnector"):
+        client_member.organization.tags = ["tag-a", "tag-b"]
+        client_member.organization.save()
 
     mock_organization_view_octopoes().get_report.return_value = get_aggregate_report_ooi
     mock_bytes_client().get_raws.return_value = [
--- a/rocky/tests/reports/test_aggregate_report_flow.py
+++ b/rocky/tests/reports/test_aggregate_report_flow.py
@@ -344,3 +344,38 @@ def test_json_download_aggregate_report(
     json_compare_data = json.dumps(get_aggregate_report_data())
 
     assert json_response_data == json_compare_data
+
+
+def test_json_download_aggregate_report_with_organization_tags(
+    rf,
+    client_member,
+    get_aggregate_report_ooi,
+    get_aggregate_report_from_bytes,
+    mock_organization_view_octopoes,
+    mock_bytes_client,
+    mock_katalogus_client,
+):
+    """Regression guard for #4332: organization tags must be JSON serializable.
+
+    The JSON export passed OrganizationTag model instances to JsonResponse,
+    which raised TypeError. Tags should be serialized as their names, matching
+    the form the report runner itself uses.
+    """
+    client_member.organization.tags = ["tag-a", "tag-b"]
+    client_member.organization.save()
+
+    mock_organization_view_octopoes().get_report.return_value = get_aggregate_report_ooi
+    mock_bytes_client().get_raws.return_value = [
+        ("7b305f0d-c0a7-4ad5-af1e-31f81fc229c2", get_aggregate_report_from_bytes)
+    ]
+
+    request = setup_request(
+        rf.get("view_report_json", {"json": "true", "report_id": f"{get_aggregate_report_ooi.primary_key}"}),
+        client_member.user,
+    )
+
+    json_response = ViewReportView.as_view()(request, organization_code=client_member.organization.code)
+
+    assert json_response.status_code == 200
+    data = json.loads(json_response.content)
+    assert data["organization_tags"] == ["tag-a", "tag-b"]


rocky/reports/views/base.py
```diff
@@ -562,7 +562,7 @@ def get(self, request, *args, **kwargs):
             response = {
                 "organization_code": self.organization.code,
                 "organization_name": self.organization.name,
-                "organization_tags": list(self.organization.tags.all()),
+                "organization_tags": [tag.name for tag in self.organization.tags.all()],
```
underdarknl, line 565: Maybe its best is we add a __str__ method to the tags object itself? although that might be cumbersome as its an external package.
We've seen this exact issue in another spot too and I'm guessing it might be a common mistake.

brennodewinter: Good point that this is a recurring mistake — you were right, there was a second spot.

On adding `__str__` to `OrganizationTag`: that alone would not fix this crash. `JsonResponse` uses Django's `JSONEncoder`, which raises `TypeError` for any object type it does not explicitly handle — it does not fall back to `__str__`. So even with a `__str__` defined, `list(self.organization.tags.all())` would still raise `TypeError: Object of type OrganizationTag is not JSON serializable`. The names have to be extracted explicitly (which is what `report_runner.py` already does on lines 174 and 228).

That said, your instinct about the common mistake was correct: `reports/views/mixins.py:112` had the same pattern — `set(self.organization.tags.all())` passed `OrganizationTag` instances to the `Report` OOI's `organization_tags: list[str]` field. It only worked by accident via pydantic `str()` coercion. Fixed that in the same PR (d88ed8589) to use `{tag.name for tag in ...}`, matching the runner.

hasecon (approved):
## QA Review

### Setup
- Branch: `fix/report-json-export-org-tags` @ `684ae75`
- Stack: running dev stack on current main, rocky restarted on the PR branch (bind mount) — clean
- Test org: `setup_test_org --code dvwa` with a real generated Report OOI

### CI/CD Status
All 19 checks green: rocky `test (3.10–3.13)`, `pre-commit`, CodeQL (3 languages), and the rocky package build matrix. The separate rocky integration-tests check did not run on this PR; covered locally instead (`make itest`: 5 passed, exit 0).

### Code Review Findings

| Severity | Finding |
|---|---|
| — | **The fix is correct and matches the root cause.** Verified that pydantic/JsonResponse genuinely reject `OrganizationTag` instances (reproduced the `TypeError` from #4332 via the revert test, and a direct `Report(organization_tags={<tag instance>})` fails validation with `string_type`). Serializing as names matches what `report_runner.py:174/228` already does, so exports and runner-generated reports now agree. |
| Low (informational) | **The `mixins.py` half touches dead code.** `SaveMultiReportMixin.save_report` (and the other two `save_report` mixins) have **no call sites**: `SaveReportView.form_valid` creates a recipe + schedule and redirects, and the actual saving happens in `report_runner`. Verified empirically: a tags-variant of `test_save_multi_report` passes even on pre-fix code (the POST never reaches `save_report`), and invoking `save_report` directly fails on **both** main and this branch — because that path never sets the now-required `Report.report_recipe` field. So the change there is harmless consistency (and a set of names coerces fine into `list[str]`), but it cannot be regression-tested meaningfully. Suggested follow-up (not blocking): delete the three dead `save_report` mixin methods, or fix the missing `report_recipe` if they are meant to be revived. |
| — | **New test is consistent and pins the fix.** It mirrors the sibling `test_json_download_aggregate_report`, all fixtures exist, the tagulous save uses the established signal-patch pattern, and the assertion order is deterministic (tagulous orders tags by name). Revert-verified: the test **fails on main** (TypeError re-raised by the except-branch) and passes on the branch. |

### Test Results
| Component | Test | Result |
|-----------|------|--------|
| rocky | `tests/reports/test_aggregate_report_flow.py` (in container) | ✅ 11 passed (incl. new test) |
| rocky | new test against main's views (revert test) | ✅ fails pre-fix, passes on branch |
| rocky | `make utest` (full suite) | ✅ 479 passed, 1 skipped, exit 0 |
| rocky | `make itest` | ✅ 5 passed, 1 deselected, exit 0 |

### Functional Test
| Check | Status | Details |
|-------|--------|---------|
| Services running | ✅ | all 13 containers up, rocky restarted on branch code |
| Live E2E | ✅ | set real tags (`nis2`, `zorg`) on the dvwa org and requested the JSON export of an actual generated Report OOI (`/en/dvwa/reports/view?json=true&report_id=Report\|175bcdc9-…`) through the full bytes/octopoes round-trip: **200**, body contains `"organization_tags": ["nis2", "zorg"]` — this exact request 500'd per #4332 |
| Log analysis | ✅ | no new errors/tracebacks in rocky since restart |

### Verdict
**APPROVED**

Small, correct fix with a regression test that genuinely pins it, proven live against a real report. Only note: the `mixins.py` line fixes dead code (see the Low finding) — fine to keep for consistency, but a cleanup of the unused `save_report` mixins would be a good follow-up. `rocky/**` still needs Jan's code-owner approval to clear the gate.

PR #5369, opened 2026-09-03
Add non-org-scoped /api/v1/health endpoint (#4231)

### LINKED ISSUE #4231: Rocky health API endpoint

brennodewinter:
## Summary
- The existing health endpoint requires an organization context (`<organization_code>/health/`), making it unsuitable for load balancers and monitoring that need to check Rocky's overall status without knowing a specific organization (#4231).
- Added `GlobalHealthView` at `/api/v1/health` that checks all backing services:
  - **Octopoes**: via `root_health()` endpoint (no organization needed)
  - **KATalogus**: already non-org-scoped
  - **Scheduler**: already supports `organization_code=None`
  - **Bytes**: already non-org-scoped
- The org-scoped `<organization_code>/health/` endpoint is unchanged.

## Test plan
- [x] Added `test_global_health_endpoint` — mocks all 4 backing services, asserts 200 response with `service=rocky`, `healthy=true`, and all 4 sub-services present.
- [x] Existing health tests still pass.

Closes #4231.

### DIFF
--- a/rocky/rocky/templates/footer.html
+++ b/rocky/rocky/templates/footer.html
@@ -17,6 +17,10 @@
                 <li>
                     <a href="{% url "health_beautified" organization_code=organization.code %}">{% translate "Health" %}</a>
                 </li>
+            {% else %}
+                <li>
+                    <a href="{% url "global_health_beautified" %}">{% translate "Health" %}</a>
+                </li>
             {% endif %}
         </ul>
     </nav>
--- a/rocky/rocky/templates/footer.html
+++ b/rocky/rocky/templates/footer.html
@@ -17,7 +17,7 @@
                 <li>
                     <a href="{% url "health_beautified" organization_code=organization.code %}">{% translate "Health" %}</a>
                 </li>
-            {% else %}
+            {% elif request.user.is_authenticated %}
                 <li>
                     <a href="{% url "global_health_beautified" %}">{% translate "Health" %}</a>
                 </li>
--- a/rocky/rocky/templates/footer.html
+++ b/rocky/rocky/templates/footer.html
@@ -17,7 +17,7 @@
                 <li>
                     <a href="{% url "health_beautified" organization_code=organization.code %}">{% translate "Health" %}</a>
                 </li>
-            {% elif request.user.is_authenticated %}
+            {% elif request.user.is_superuser %}
                 <li>
                     <a href="{% url "global_health_beautified" %}">{% translate "Health" %}</a>
                 </li>
--- a/rocky/rocky/urls.py
+++ b/rocky/rocky/urls.py
@@ -11,7 +11,7 @@ from rocky.views.bytes_raw import BytesRawView
 from rocky.views.finding_add import FindingAddView
 from rocky.views.finding_list import FindingListView
 from rocky.views.finding_type_add import FindingTypeAddView
-from rocky.views.health import Health, HealthChecks
+from rocky.views.health import GlobalHealthView, Health, HealthChecks
 from rocky.views.indemnification_add import IndemnificationAddView
 from rocky.views.landing_page import LandingPageView
 from rocky.views.ooi_add import OOIAddTypeSelectView, OOIAddView
@@ -62,6 +62,7 @@ router.register(r"report-recipe", ReportRecipeViewSet, basename="report-recipe")
 urlpatterns = [
     path("i18n/", include("django.conf.urls.i18n")),
     path("api/v1/", include(router.urls)),
+    path("api/v1/health", GlobalHealthView.as_view(), name="global_health"),
     path("<organization_code>/health/", Health.as_view(), name="health"),
     path("", include(tf_urls)),
     path("robots.txt", TemplateView.as_view(template_name="robots.txt", content_type="text/plain")),
--- a/rocky/rocky/urls.py
+++ b/rocky/rocky/urls.py
@@ -63,6 +63,7 @@ urlpatterns = [
     path("i18n/", include("django.conf.urls.i18n")),
     path("api/v1/", include(router.urls)),
     path("api/v1/health", GlobalHealthView.as_view(), name="global_health"),
+    path("api/v1/health/", GlobalHealthView.as_view()),
     path("<organization_code>/health/", Health.as_view(), name="health"),
     path("", include(tf_urls)),
     path("robots.txt", TemplateView.as_view(template_name="robots.txt", content_type="text/plain")),
--- a/rocky/rocky/urls.py
+++ b/rocky/rocky/urls.py
@@ -11,7 +11,7 @@ from rocky.views.bytes_raw import BytesRawView
 from rocky.views.finding_add import FindingAddView
 from rocky.views.finding_list import FindingListView
 from rocky.views.finding_type_add import FindingTypeAddView
-from rocky.views.health import GlobalHealthView, Health, HealthChecks
+from rocky.views.health import GlobalHealthChecks, GlobalHealthView, Health, HealthChecks
 from rocky.views.indemnification_add import IndemnificationAddView
 from rocky.views.landing_page import LandingPageView
 from rocky.views.ooi_add import OOIAddTypeSelectView, OOIAddView
@@ -75,6 +75,7 @@ urlpatterns += i18n_patterns(
     path("crisis-room/", include("crisis_room.urls"), name="crisis_room"),
     path("<organization_code>/crisis-room/", include("crisis_room.urls_org"), name="crisis_room_org"),
     path("privacy-statement/", PrivacyStatementView.as_view(), name="privacy_statement"),
+    path("health/v1/", GlobalHealthChecks.as_view(), name="global_health_beautified"),
     path("tasks/", AllBoefjesTaskListView.as_view(), name="all_task_list"),
     path("tasks/boefjes", AllBoefjesTaskListView.as_view(), name="all_boefjes_task_list"),
     path("tasks/normalizers", AllNormalizersTaskListView.as_view(), name="all_normalizers_task_list"),
--- a/rocky/rocky/views/health.py
+++ b/rocky/rocky/views/health.py
@@ -2,6 +2,7 @@ from typing import Any
 
 import structlog
 from account.mixins import OrganizationView
+from django.conf import settings
 from django.http import HttpRequest, JsonResponse
 from django.urls.base import reverse
 from django.utils.translation import gettext_lazy as _
@@ -22,6 +23,19 @@ class Health(OrganizationView, View):
     def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> JsonResponse:
         octopoes_connector = self.octopoes_api_connector
         rocky_health = get_rocky_health(self.organization.code, octopoes_connector)
+        return JsonResponse(rocky_health.model_dump())
+
+
+class GlobalHealthView(View):
+    """Non-org-scoped health endpoint for monitoring and load balancers (#4231).
+
+    Checks all backing services without requiring an organization context.
+    Octopoes is checked via its root health endpoint, which does not need
+    a specific organization.
+    """
+
+    def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> JsonResponse:
+        rocky_health = get_rocky_health_global()
         return JsonResponse(rocky_health.model_dump())
 
 
@@ -48,7 +62,19 @@ def get_octopoes_health(octopoes_api_connector: OctopoesAPIConnector) -> Service
     return octopoes_health
 
 
-def get_scheduler_health(organization_code: str) -> ServiceHealth:
+def get_octopoes_root_health() -> ServiceHealth:
+    try:
+        connector = OctopoesAPIConnector(settings.OCTOPOES_API, "", timeout=settings.ROCKY_OUTGOING_REQUEST_TIMEOUT)
+        octopoes_health = ServiceHealth.model_validate(connector.root_health().model_dump())
+    except HTTPError:
+        logger.exception("Error while retrieving Octopoes root health state")
+        octopoes_health = ServiceHealth(
+            service="octopoes", healthy=False, additional="Could not connect to Octopoes. Service is possibly down"
+        )
+    return octopoes_health
+
+
+def get_scheduler_health(organization_code: str | None = None) -> ServiceHealth:
     try:
         scheduler_health = scheduler_client(organization_code).health()
     except SchedulerError:
@@ -75,6 +101,18 @@ def get_rocky_health(organization_code: str, octopoes_api_connector: OctopoesAPI
         service="rocky", healthy=services_healthy, version=__version__, results=services, additional=additional
     )
     return rocky_health
+
+
+def get_rocky_health_global() -> ServiceHealth:
+    services = [get_octopoes_root_health(), get_katalogus_health(), get_scheduler_health(), get_bytes_health()]
+
+    services_healthy = all(service.healthy for service in services)
+    additional = None
+    if not services_healthy:
+        additional = "Rocky will not function properly. Not all services are healthy."
+    return ServiceHealth(
+        service="rocky", healthy=services_healthy, version=__version__, results=services, additional=additional
+    )
 
 
 def flatten_health(health_: ServiceHealth) -> list[ServiceHealth]:
--- a/rocky/rocky/views/health.py
+++ b/rocky/rocky/views/health.py
@@ -123,6 +123,26 @@ def flatten_health(health_: ServiceHealth) -> list[ServiceHealth]:
     return results
 
 
+class GlobalHealthChecks(TemplateView):
+    """Non-org-scoped human-readable health page (#4231).
+
+    Renders the same template as the org-scoped HealthChecks view but
+    uses get_rocky_health_global() so it works without an organization.
+    """
+
+    template_name = "health.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["breadcrumbs"] = [
+            {"url": reverse("global_health"), "text": _("Health")},
+            {"url": reverse("global_health_beautified"), "text": _("Beautified")},
+        ]
+        rocky_health = get_rocky_health_global()
+        context["health_checks"] = flatten_health(rocky_health)
+        return context
+
+
 class HealthChecks(OrganizationView, TemplateView):
     template_name = "health.html"
 
--- a/rocky/rocky/views/health.py
+++ b/rocky/rocky/views/health.py
@@ -1,15 +1,23 @@
+from collections.abc import Callable
 from typing import Any
 
 import structlog
 from account.mixins import OrganizationView
 from django.conf import settings
+from django.contrib.auth.mixins import UserPassesTestMixin
 from django.http import HttpRequest, JsonResponse
 from django.urls.base import reverse
 from django.utils.translation import gettext_lazy as _
 from django.views.generic import TemplateView, View
 from httpx import HTTPError
 from katalogus.health import get_katalogus_health
+from pydantic import ValidationError
+from rest_framework import status
+from rest_framework.permissions import AllowAny
+from rest_framework.response import Response
+from rest_framework.views import APIView
 
+from octopoes.connector import ConnectorException
 from octopoes.connector.octopoes import OctopoesAPIConnector
 from rocky.bytes_client import get_bytes_client
 from rocky.health import ServiceHealth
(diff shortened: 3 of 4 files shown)


hasecon (changes requested):
## QA Review

### Setup
- Branch: `feature/rocky-health-api-endpoint` @ `5ca14f8`
- Stack: `make clean && make kat` — OK, all 13 containers up, Rocky reachable (HTTP 200)
- Jan's footer request from the first round is addressed (`GlobalHealthChecks` + `{% else %}` branch); verified live.

### CI/CD Status

All 19 checks green (`test` 3.10–3.13, `rocky-tests`, `pre-commit`, `CodeQL`, `makelang`, all four `build` legs). No CI blocker.

That green is exactly the problem, though: the two most serious defects below are invisible to CI because both new tests call the view functions directly and never route a request or start a service. Green CI here does not mean the endpoint works.

### Code Review Findings

| # | Sev | File | Issue |
|---|-----|------|-------|
| 1 | **High** | `views/health.py:65-74` | Octopoes root health is hardcoded `healthy=True` — the check is vacuous |
| 2 | **High** | `views/health.py:37-39` | Always returns HTTP 200, even when unhealthy — defeats the LB purpose |
| 3 | **High** | `urls.py:65` | Endpoint is fully unauthenticated; breaks the `/api/` = DRF-authenticated invariant |
| 4 | Medium | `views/health.py:69` | `except HTTPError` misses the exceptions the connector actually raises |
| 5 | Medium | `urls.py:78` | Beautified page readable by any logged-in user, incl. non-members |
| 6 | Medium | `urls.py:65` | No trailing slash → `/api/v1/health/` is a hard 404 |
| 7 | Low | `footer.html:20-23` | Dead "Health" link on anonymous pages |
| 8 | Low | `views/health.py:106`, `65` | `get_rocky_health_global` / `get_octopoes_root_health` are copies |
| 9 | Low | `views/health.py:107`, `67` | Four checks run sequentially; a new `httpx.Client` per probe |

---

#### 1. (High) The Octopoes check is vacuous — the endpoint reports green while XTDB is down

`get_octopoes_root_health()` calls `connector.root_health()`, which hits Octopoes' **root** `/health`. That route returns a constant:

```python
# octopoes/octopoes/api/api.py:150
@app.get("/health")
def root_health() -> ServiceHealth:
    return ServiceHealth(service="octopoes", healthy=True, version=__version__)
```

It never probes XTDB. The org-scoped route (`router.py:88`) calls `octopoes.health()`, which does. So the new endpoint reports Octopoes healthy whenever the Octopoes *process* is up, regardless of whether it can serve any data.

Proven on the running stack — `docker compose stop crux`, then both endpoints hit back to back:

```
GLOBAL   /api/v1/health   HTTP 200   overall healthy: True
  octopoes   healthy=True      <-- XTDB is down
  katalogus  healthy=True
  scheduler  healthy=True
  bytes      healthy=True

ORG      /dvwa/health/    HTTP 200   overall healthy: False
  octopoes   healthy=False     <-- correct
  additional: Rocky will not function properly. Not all services are healthy.
```

Same stack, same moment. Monitoring built on the new endpoint stays green through an outage that the existing endpoint catches. The docstring "Checks all backing services" is not accurate as written.

`organizations_health()` (`api.py:155`) is the org-agnostic call that actually probes XTDB per node — that looks like the right building block here, though it walks every node, so you may want to consider its cost on large deployments.

#### 2. (High) Always HTTP 200, even when unhealthy

`get()` returns `JsonResponse(...)` with the default status regardless of `rocky_health.healthy`. Load balancers and k8s probes key on the status code, not the body. Proven with `octopoes_api` stopped:

```
HTTP 200
overall healthy: False
  octopoes   healthy=False  Could not connect to Octopoes. Service is possibly down
  scheduler  healthy=False  Could not connect to Scheduler. Service is possibly down
```

A standard `httpGet` readiness probe passes here and keeps a Rocky that cannot serve a page in rotation. Suggest `status=200 if rocky_health.healthy else 503`.

#### 3. (High) The endpoint is fully unauthenticated

`AuthRequiredMiddleware` exempts the `/api/` prefix, and the comment states the reason explicitly:

```python
# rocky/middleware/auth_required.py:22-26
# There won't be a request.user if auth tokens are used, but
# Django REST framework will make sure that there is an
# authenticated user without DEFAULT_PERMISSION_CLASSES setting
# in settings.py.
"/api/",
```

`GlobalHealthView` is a plain `django.views.generic.View`, not a DRF `APIView`, so `KATModelPermissions` never runs. The exemption's premise doesn't hold for this route. Verified against a DRF neighbour on the same stack:

```
anonymous GET /api/v1/organization/   -> 401     (DRF baseline)
anonymous GET /api/v1/health          -> 200 + full JSON
```

Rocky's own log confirms no user: `request='GET /api/v1/health' user_id=None`. The body returns `v1.22.0-120-g5ca14f819` for Rocky plus each of octopoes/katalogus/scheduler/bytes, and which of them are currently down — a version-fingerprint and outage-visibility surface on the public-facing component. Each anonymous hit also fans out four upstream requests, so it is an unauthenticated amplification path onto the backends.

An intentionally public health endpoint is a defensible design choice — but it should be explicit at the view (a DRF `APIView` with `permission_classes = [AllowAny]`, or a non-`/api/` path added to the middleware `excluded` list), not inherited from a prefix exemption that documents a different reason. If it stays public, please also trim the payload: reachability without version numbers covers the LB use case.

#### 4. (Medium) `except HTTPError` doesn't match what the connector raises

`OctopoesAPIConnector._verify_response` (`octopoes.py:60-73`) converts responses before the caller sees them: 5xx → `RemoteException`, 404 → `ObjectNotFoundException`, bad JSON → `DecodeException`. All derive from `ConnectorException(Exception)`, none from `httpx.HTTPError`. `ServiceHealth.model_validate_json` can also raise `ValidationError`.

Reproduced by feeding the connector a 500 (a reverse proxy 502, or Octopoes' own `uncaught_exception_handler` at `api.py:144`):

```
UNCAUGHT octopoes.connector.RemoteException escaped get_octopoes_root_health()
  -> propagates out of GlobalHealthView.get() -> Django 500 HTML page
```

So in that case the monitoring endpoint returns an HTML error page instead of `healthy: false`, and the "Could not connect to Octopoes" fallback is dead code for part of what it was written for. Connection-refused *is* caught correctly (`ConnectError` is an `HTTPError`) — that path I verified works. Suggest `except (HTTPError, ConnectorException, ValidationError)`. The same narrow catch exists in the pre-existing `get_octopoes_health`, so a shared helper would fix both at once (see #8).

#### 5. (Medium) The beautified page is readable by any authenticated user

`GlobalHealthChecks` is a bare `TemplateView` — no `OrganizationView`, no permission check. Verified with a fresh user having no organization membership at all:

```
plain user (no org, not staff, not superuser)
  /en/health/v1/       -> 200   full infrastructure inventory rendered
  /en/dvwa/health/v1/  -> 404   (org-scoped view correctly denies)
```

`docs/source/installation-and-deployment/debugging-troubleshooting.rst:13` states "The admin and superuser accounts have access to the health page." A `PermissionRequiredMixin` or a superuser check would keep that promise; otherwise the docs need updating to match.

#### 6. (Medium) Trailing-slash 404

`path("api/v1/health", ...)` has no trailing slash, while the DRF router mounts `api/v1/` and every neighbouring route ends in `/`. `APPEND_SLASH` only adds slashes, never strips them:

```
GET /api/v1/health   -> 200
GET /api/v1/health/  -> 404
```

An operator following the convention of every other Rocky URL configures the probe as `/api/v1/health/` and gets a 404, which most LBs read as "backend down". Registering both spellings (or just the slashed one) avoids a deploy-time surprise. Worth noting the asymmetry with the page route two lines down (`health/v1/`) too.

#### 7. (Low) Dead "Health" link on anonymous pages

`footer.html` is included unconditionally from `layouts/base.html:17`, and the new `{% else %}` fires whenever `organization` is absent — which includes the login and landing pages for anonymous visitors. Confirmed: the anonymous login page renders `href="/en/health/v1/">Health<`, and clicking it bounces to `/en/login/?next=/en/health/v1/`. Gating on `{% if request.user.is_authenticated %}`, as line 6 already does for the version string, matches the previous behaviour.

#### 8-9. (Low) Duplication and per-request client

`get_rocky_health_global()` is `get_rocky_health()` with one list element changed — the aggregation and the user-facing "Rocky will not function properly" string are now duplicated verbatim, so a fifth backing service must be added in two places or the two endpoints silently diverge. Same for `get_octopoes_root_health` vs `get_octopoes_health` (which is how the #4 bug got copied). A single `_aggregate(services)` plus one Octopoes helper taking the bound method would collapse both.

Separately, the four checks run sequentially, each with `ROCKY_OUTGOING_REQUEST_TIMEOUT` (30s default), so a probe can hold a worker for ~120s when backends hang — precisely when an LB probes hardest. And `OctopoesAPIConnector(...)` is constructed per request, allocating an `httpx.Client` that is never closed; I measured Rocky's FD count going 39 → 51 across 40 probes. GC reclaims eventually, so this is an observation rather than a proven leak, but a module-level cached connector (like `get_katalogus_client()` in `katalogus/client.py:474`) plus a short probe-specific timeout would remove both concerns.

### Suggested regression test

The two new tests call `GlobalHealthView.as_view()(request)` directly with `rf.get("global_health")` — a URL *name* passed as a path string — and an authenticated `client_member.user`. That never exercises URL routing or `AuthRequiredMiddleware`, so neither of the endpoint's two defining properties (it resolves; it answers anonymously) is pinned, and findings #2, #3 and #6 all pass CI untouched.

Drop-in replacement using the real path and no login:

```python
import pytest
from django.test import Client

from rocky.health import ServiceHealth


@pytest.mark.django_db
def test_global_health_endpoint_probe_path(mocker):
    """The load-balancer contract: real URL, anonymous client, status reflects health (#4231)."""
    mocker.patch(
        "rocky.views.health.get_octopoes_root_health",
        return_value=ServiceHealth(service="octopoes", healthy=False, additional="down"),
    )
    mocker.patch(
        "rocky.views.health.get_katalogus_health",
        return_value=ServiceHealth(service="katalogus", healthy=True),
    )
    mocker.patch(
        "rocky.views.health.get_scheduler_health",
        return_value=ServiceHealth(service="scheduler", healthy=True),
    )
    mocker.patch(
        "rocky.views.health.get_bytes_health",
        return_value=ServiceHealth(service="bytes", healthy=True),
    )

    response = Client(SERVER_NAME="localhost").get("/api/v1/health")  # no login

    assert response.status_code == 503          # currently 200 — pins finding #2
    assert response.json()["healthy"] is False
```

I ran this against the branch: it routes and answers anonymously (so it pins #3 and #6 as written), and the `503` assertion fails today with `200`, so it goes red pre-fix and green once the status code is derived from `healthy`. Worth adding a companion that asserts Octopoes reports unhealthy when XTDB is unreachable, which is what would have caught #1.

### Test Results

| Component | Test | Result |
|-----------|------|--------|
| Rocky | `make utest` (Docker) | ✅ 482 passed, 1 skipped (111s) |
| Rocky | `make itest` (Docker) | ✅ 5 passed, 1 deselected |
| CI | all 19 checks | ✅ green |

### Functional Test

| Check | Status | Details |
|-------|--------|---------|
| Stack build | ✅ | `make clean && make kat` clean, 13/13 containers up |
| Services running | ✅ | no restart loops |
| Log analysis | ✅ | 0 tracebacks in steady state; the errors seen mid-run were from the outages I induced deliberately |
| `/api/v1/health` anonymous | ⚠️ | 200 + full version JSON, `user_id=None` (finding #3) |
| `/api/v1/health/` | ❌ | 404 (finding #6) |
| Health vs XTDB down | ❌ | reports healthy (finding #1) |
| Health vs services down | ⚠️ | body correct, status still 200 (finding #2) |
| Octopoes 5xx | ❌ | `RemoteException` → Django 500 page (finding #4) |
| `/en/health/v1/` render | ✅ | renders all 5 services for superuser |
| `/en/health/v1/` access | ⚠️ | 200 for a user with no org membership (finding #5) |
| Footer link | ⚠️ | present on anonymous login page (finding #7) |

### Verdict

**CHANGES REQUESTED**

The structure is right and the footer follow-up does what Jan asked. But the endpoint doesn't yet deliver what #4231 asks for: it reports healthy while XTDB is down (#1), and it answers 200 when it knows it is unhealthy (#2), so a load balancer pointed at it would neither detect the outage nor take the instance out of rotation. Those two are the blockers. #3 is a design decision worth making explicitly rather than inheriting from the `/api/` exemption's DRF assumption — happy to go along with a public endpoint if that is the intent, ideally with the version numbers trimmed.

#4-#7 are small, contained fixes. #8-#9 are cleanups I would not block on.

Everything above was reproduced on a local stack at `5ca14f8`; the commands and outputs are in the sections above if you want to replay any of it.

hasecon (changes requested):
## QA Review — round 2

### Setup
- Branch: `feature/rocky-health-api-endpoint` @ `da4ef90`
- Stack: `make clean && make kat` — 13/13 containers up
- Test orgs: `dvwa` **and** `acme-bank` — two organizations deliberately, because the new Octopoes probe behaves differently per organization

### Round-1 findings — 8 of 9 fixed, verified against the head

| # | Sev | Round-1 finding | Status | How I verified |
|---|-----|---|---|---|
| 1 | High | Vacuous Octopoes check (hardcoded `healthy=True`) | ✅ Fixed | `docker compose stop crux` → `octopoes healthy=False`, overall `healthy=False`. The XTDB probe is real now. |
| 2 | High | Always HTTP 200 | ✅ Fixed | Same run returned **HTTP 503** |
| 3 | High | Unauthenticated + version disclosure | ⚠️ Half | Auth is now explicit (DRF `APIView`, `AllowAny`, `authentication_classes = []`) — exactly what I asked for — and versions are stripped. But the same commit introduced a **larger** disclosure; see finding A. |
| 4 | Medium | `except HTTPError` misses connector exceptions | ✅ Fixed | Fed the connector a mocked 500 → `healthy=False`, "Could not connect to Octopoes", no exception escapes |
| 5 | Medium | Beautified page open to any logged-in user | ✅ Fixed | superuser → 200, plain member → **403**, anonymous → 302 to login |
| 6 | Medium | Trailing-slash 404 | ✅ Fixed | `/api/v1/health/` → 200 |
| 7 | Low | Dead footer link on anonymous pages | ✅ Fixed | Anonymous login page no longer renders the link. (But see finding B.) |
| 8 | Low | Duplicated aggregation / Octopoes helper | ✅ Fixed | `_aggregate()` and `_get_octopoes_health(probe)` are shared by both the org-scoped and global paths |
| 9 | Low | Sequential checks, per-request client | Accepted | Not addressed, as agreed — I said I would not block on it |

That is a thorough round, and the two structural asks (a real XTDB probe, and making the public-access decision explicit at the view rather than inheriting the `/api/` middleware exemption) are both done properly.

### Code Review Findings

| # | Sev | File | Issue |
|---|-----|------|-------|
| A | **High** | `views/health.py:88` + `_strip_versions` | The public endpoint now discloses every organization code and XTDB internals — introduced by the fix for #1 |
| B | Low | `templates/footer.html:20` | Dead link returns, now for authenticated non-superusers |
| C | Low | `tests/test_health.py` | The superuser restriction (#5) has no negative test |

---

#### A. (High) The public endpoint now leaks the organization list and XTDB internals

`organizations_health()` returns a nested result **per organization** (`api.py:155-173`), and `_strip_versions()` only nulls the `version` field — it leaves `service` names and `additional` untouched. So everything below survives to anonymous callers.

Actual response from `curl http://localhost:8000/api/v1/health` with no credentials:

```json
{"service": "rocky", "healthy": true, "version": null,
 "results": [
   {"service": "octopoes", "healthy": true, "version": null,
    "results": [
      {"service": "xtdb for dvwa", "healthy": true, "version": null,
       "additional": {"version": "1.24.5",
                      "revision": "75dc0901e2e3b805463a4523e544d9594c627b01",
                      "indexVersion": 22, "kvStore": "xtdb.rocksdb.RocksKv",
                      "estimateNumKeys": 2775, "size": 1154516}},
      {"service": "xtdb for acme-bank", "healthy": true, "version": null,
       "additional": {"version": "1.24.5", "…": "…",
                      "estimateNumKeys": 2695, "size": 1009961}}
    ]}, …]}
```

So an anonymous request returns:

- **the full list of organization codes** on the installation (`dvwa`, `acme-bank`) — the tenant list;
- **XTDB's exact version and git revision** (`1.24.5`, `75dc0901…`), plus `indexVersion` and `kvStore`;
- **per-tenant data volume** (`estimateNumKeys`, `size`), which is a usable proxy for how much data each organization holds.

The awkward part is that this is the same commit that strips Rocky's and Octopoes' version strings "to avoid fingerprinting". Net, the endpoint gives away more identifying detail than before the fix, not less — a precise third-party version plus a customer list, instead of one OpenKAT version string. For the multi-tenant deployments this feature is aimed at (#4231 came out of a Kennisnet debugging session), disclosing which organizations exist is the part I would worry about most.

The org-scoped `/{org}/health/` is unaffected — it calls `connector.health()`, not `organizations_health()` — so this is specific to the new public endpoint.

The health signal you actually need for a load balancer is the aggregate boolean, not the per-node breakdown. Redacting below the top level keeps #1's fix intact while closing this:

```python
def _redact_for_public(health_: ServiceHealth) -> None:
    """Public endpoint: keep service + healthy, drop everything that fingerprints."""
    health_.version = None
    for sub_result in health_.results:
        sub_result.version = None
        sub_result.additional = None
        sub_result.results = []      # drop the per-organization XTDB breakdown
```

called in place of `_strip_versions()`. The superuser-only `/health/v1/` page can keep the full detail — that is where an operator wants the per-node view, and it is now properly gated.

#### B. (Low) The dead footer link is back, one role up

The footer now gates on `request.user.is_authenticated`, but the page it links to requires `is_superuser`. Verified on the running stack:

```
anonymous login page shows health link:  False   (finding #7 fixed)
plain member, global crisis room:        link shown = True
that link, for that user:                403
```

Same class as #7, just moved from anonymous visitors to non-superuser members. `{% elif request.user.is_superuser %}` matches the view's actual gate.

#### C. (Low) The superuser restriction is not pinned

`test_global_health_beautified` covers the superuser path only. Nothing asserts that a non-superuser is refused, so #5's fix could be reverted without failing a test. One line in the existing style:

```python
def test_global_health_beautified_denied_for_non_superuser(rf, client_member):
    """The beautified page is superuser-only (#4231) — a plain member must not reach it."""
    request = setup_request(rf.get("global_health_beautified"), client_member.user)

    with pytest.raises(PermissionDenied):
        GlobalHealthChecks.as_view()(request)
```

I verified the behaviour it pins live (plain member → 403); I did not run this exact snippet, so treat the code as a draft.

### Note on cost, not a finding

`organizations_health()` issues one XTDB `status()` call per organization, on an endpoint that is anonymous and intended to be polled frequently. With two organizations the endpoint takes ~0.53s; it scales linearly with the number of organizations, so on an installation with dozens of orgs an unauthenticated caller drives dozens of XTDB probes per request. Redacting per finding A does not change that — if you want the cheaper signal, `root_health()` plus a single XTDB reachability check would give the same load-balancer answer at constant cost. Not blocking, but worth a thought before this ships to a large deployment.

### Test Results

| Component | Test | Result |
|-----------|------|--------|
| rocky | `make utest` (Docker) | ✅ 486 passed, 1 skipped |
| CI | all 19 checks | ✅ green |

The rewritten tests are a real improvement: `test_global_health_endpoint_anonymous_via_url` and `test_global_health_endpoint_unhealthy_via_url` go through real URL routing with an anonymous `Client`, which is exactly what the round-1 tests could not do, and they pin routing, the anonymous contract, the 503 and the trailing slash.

### Functional Test

| Check | Status | Details |
|-------|--------|---------|
| Stack build | ✅ | 13/13 containers |
| Anonymous `/api/v1/health` | ✅ 200 | reachable without credentials, versions stripped |
| Trailing slash | ✅ 200 | finding #6 closed |
| XTDB down | ✅ 503 | `octopoes healthy=False` — finding #1 closed |
| Octopoes 5xx | ✅ | handled, no exception escapes — finding #4 closed |
| Beautified page | ✅ | superuser 200 / member 403 / anonymous → login |
| Footer, anonymous | ✅ | no link |
| Footer, plain member | ⚠️ | link shown, 403 on click (finding B) |
| Public response content | ❌ | organization codes + XTDB internals exposed (finding A) |

### Verdict

**CHANGES REQUESTED**

Eight of nine round-1 findings are genuinely fixed and I verified each one on a running stack rather than from the summary — the XTDB probe is real, the 503 works, the auth decision is explicit at the view, and the exception handling now covers what the connector actually raises.

The blocker is finding A, and it is a side effect of fixing #1 rather than a new mistake: swapping `root_health()` for `organizations_health()` made the check meaningful, but that call returns a per-organization breakdown, and the public endpoint hands it to anonymous callers along with XTDB's version, git revision and per-tenant data sizes. Stripping the OpenKAT version strings in the same commit while adding all of that is the part worth pausing on. The redaction above keeps the fix and closes the disclosure.

B and C are small and can ride along in the same push.

PR #5370, opened 2026-09-03
Strip boefje environment secrets from raw meta downloads (#4508)

### LINKED ISSUE #4508: Task list raw meta download contains secrets
**Describe the bug**
In the task list every user can download the boefje meta and raw data. The meta contains the environment of the boefje that might contain secrets. We should either make the boefje meta only downloadable by users who can already see the secrets or remove the secrets from the meta.

brennodewinter:
## Summary
- The raw meta download (`BytesRawView`) exposed the `boefje_meta.environment` field, which can contain secrets, to any user who can access the task list (#4508).
- Added `_strip_sensitive_fields()` that removes the `environment` key from `boefje_meta` in each raw meta, in-place, before the data is returned — both in the zip download and the JSON format response.
- The `DownloadTaskDetail` view (scheduler task data) is not affected: the Rocky `BoefjeTask` model does not include an `environment` field, so pydantic already discards it.

## Test plan
- [x] Added `test_download_task_strips_environment_from_meta` — verifies `environment` is absent from the zip's raw meta JSON.
- [x] Added `test_download_task_json_format_strips_environment` — verifies `environment` is absent from the JSON format response.
- [x] All 11 existing + new boefje tasks / zip tests pass.

Closes #4508.

### DIFF
--- a/rocky/rocky/views/bytes_raw.py
+++ b/rocky/rocky/views/bytes_raw.py
@@ -17,12 +17,27 @@ logger = structlog.get_logger(__name__)
 
 RAW_FILE_LIMIT = 1024 * 1024
 
+# Fields in boefje_meta that may contain secrets and must not be exposed in
+# downloads. See #4508.
+SENSITIVE_BOEFJE_META_FIELDS = ("environment",)
+
+
+def _strip_sensitive_fields(raw_metas: list[dict]) -> None:
+    """Remove sensitive fields (e.g. environment with secrets) from boefje_meta
+    in-place, so they are not leaked via raw meta downloads (#4508)."""
+    for raw_meta in raw_metas:
+        boefje_meta = raw_meta.get("boefje_meta")
+        if isinstance(boefje_meta, dict):
+            for field in SENSITIVE_BOEFJE_META_FIELDS:
+                boefje_meta.pop(field, None)
+
 
 class BytesRawView(OrganizationView):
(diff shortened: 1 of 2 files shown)


rocky/rocky/views/bytes_raw.py
```diff
@@ -17,12 +17,27 @@
...
+SENSITIVE_BOEFJE_META_FIELDS = ("environment",)
+
+
+def _strip_sensitive_fields(raw_metas: list[dict]) -> None:
+    """Remove sensitive fields (e.g. environment with secrets) from boefje_meta
+    in-place, so they are not leaked via raw meta downloads (#4508)."""
+    for raw_meta in raw_metas:
+        boefje_meta = raw_meta.get("boefje_meta")
+        if isinstance(boefje_meta, dict):
+            for field in SENSITIVE_BOEFJE_META_FIELDS:
```
underdarknl: Instead of removing the entire environment, we know which fields are marked as 'secret' via the boefjes json schema.

underdarknl: Also, when we remove some of these fields, this might mean the hash over these files will no longer match the hash stored in bytes. (not a problem when just viewing, but worth a note to the user who expect the raw files to produce the same hashes as the one visibile in rocky.

underdarknl (changes requested):

Issue #4043, opened 2025-01-23
Crash on 127.0.0.1


Burak-Esen:
Error Page localhost domain

When I try to use http://127.0.0.1 for domain address and after several page click "enable start scan" open-cat server gives an error page

**Expected behavior**
Reaching the next page and starting scan

**Screenshots**

![Image](https://github.com/user-attachments/assets/3a1316d6-719e-4780-853a-2ea7a13f0f64)

![Image](https://github.com/user-attachments/assets/3c285aca-c84c-416f-aad2-bdb7c3327f6a)

open-kat version: on main branch

**Desktop (please complete the following information if relevant):**

- Linux Mint 22 Cinnamon
- Chrome: 132.0.6834.110 (Official Build) (64-bit)

PR #5375, opened 2026-09-03
Reject IP-address URLs in onboarding DNS report flow (#4043)

brennodewinter:
## Summary
- Entering `http://127.0.0.1` as the onboarding scan target crashed with a 500 in `get_ooi_pks`: the URL's `web_url` is an `IPAddressHTTPURL` whose netloc token has `"address"`, not `"name"`, so `tokenized["netloc"]["name"]` raised `KeyError`.
- Two-layer fix:
  1. **Form validation** (`OnboardingCreateObjectURLForm.clean_url`): reject IP-address hosts at the trust boundary with a clear message — the onboarding runs a DNS report, which requires a hostname.
  2. **Hardening** (`get_ooi_pks`): `try/except KeyError` around `netloc["name"]` so a stray IP netloc yields a user-friendly error instead of a 500. Also guards `post` against an empty `get_ooi_pks()` so no useless report recipe is created.

#### Test plan
- [x] `test_step_5_onboarding_form_rejects_ip_address`: `http://127.0.0.1` invalid, `http://example.com` valid.
- [x] `test_step_9a_onboarding_ip_address_url_no_crash`: POST with an `IPAddressHTTPURL` web_url → 200 (re-renders with error), no 500.
- [x] Form logic verified for IPv4, IPv6 (`[::1]`), and IP-with-port.
- [x] pre-commit (ruff, ruff-format, mypy, djLint) green.
- [ ] CI (tests across Python versions).


brennodewinter:
@stephanie0x00 Good catch — a bare `localhost` passed the URL field (`assume_scheme="https"` expands it to `https://localhost`) and would produce a DNS report that can never contain anything.

Addressed in 0131d608c: the form now also rejects `localhost` and `*.localhost` (RFC 6761 loopback names), and the defensive guard in `get_ooi_pks` bails on a localhost `web_url` the same way it bails on an `IPAddressHTTPURL`. Tests extended for `localhost`, `https://localhost:8443`, `https://foo.localhost`, and a `HostnameHTTPURL` pointing at localhost.

hasecon:
**CI triage — this is red on translations only.**

`makelang` is the single red check. That job runs `make lang` in `rocky/` and then fails if any `msgid`/`msgstr` line in `rocky/rocky/locale/django.pot` changes — i.e. the committed `.pot` no longer matches what the current source produces.

This PR does change translatable strings (`rocky/onboarding/forms.py`, `rocky/onboarding/views.py`) and does include a `django.pot`, so it looks like the regeneration happened before the last source change rather than after. Re-running `make lang` in `rocky/` and committing `rocky/rocky/locale/django.pot` should clear it.

No test, build or type check is red.

PR #5376, opened 2026-09-04
Replace snyk.io HTML scraping with Nuxt SSR data parsing

### LINKED ISSUE #5301: kat_snyk: version-tabel selector kapot (dode KAT-SOFTWARE-UPDATE-AVAILABLE), severity gestript, ecosysteem hardcoded op npm
## Samenvatting

De `kat_snyk`-boefje scrapet `snyk.io/vuln/...` met BeautifulSoup. Door wijzigingen aan de snyk.io-HTML en enkele aan-de-bron-keuzes levert de plugin nu minder op dan bedoeld, en is een deel dode code. Dit issue documenteert de problematiek en oplossingsrichtingen; er is bewust (nog) geen PR gemaakt omdat een correcte fix afhangt van de *huidige* live-HTML én een herweging van de scrape-aanpak.

### 1. `table_versions` is altijd leeg → `KAT-SOFTWARE-UPDATE-AVAILABLE` is dode code
`kat_snyk/main.py` selecteert de versies-tabel met:
```python
elif table.attrs.get("data-vue") == "vulns-version-table":
```
Het `data-vue`-attribuut is uit de snyk.io-HTML verdwenen. De versies-tabel heeft nu een `thead > tr` met class `vue--table__row`, waardoor die door de **eerste** tak (kwetsbaarheden) wordt opgepakt in plaats van de versies-tak. Daar mislukt elke rij op de `KeyError`-guard (`except KeyError: continue`) en wordt stil weggegooid. Netto: `result["table_versions"]` is altijd `[]`.

Gevolg: de downstream `KAT-SOFTWARE-UPDATE-AVAILABLE`-finding (die op `table_versions` leunt) wordt nooit geproduceerd — dode code. De eigen test bevestigt dit al: `tests/plugins/test_snyk.py` assert `output["table_versions"] == []`, en zowel `inputs/snyk-result-findings.json` als `snyk-result-no-findings.json` hebben `table_versions: []`.

### 2. Severity wordt aan de bron weggegooid
```python
"Vuln_text": info["Vulnerability"].text.strip()[2:].strip(),
```
De `[2:]` knipt de leidende badge (het severity-teken) van de tekst af, waardoor de ernst van de kwetsbaarheid nergens meer beschikbaar is voor de normalizer / het finding-type.

### 3. Ecosysteem is hardcoded op npm
```python
url_snyk = f"https://snyk.io/vuln/npm:{software_name.lower().replace(' ', '-')}"
```
Alleen npm-pakketten worden opgezocht. Software uit andere ecosystemen (PyPI, Maven, RubyGems, Go, …) wordt nooit tegen Snyk gecontroleerd, ook als de `Software`-OOI een CPE heeft waaruit het ecosysteem af te leiden zou zijn.

## Oplossingsrichtingen

**A. Fixture-gebaseerde selector-fix (kleinste stap, fragiel).**
Selecteer de versies-tabel op de kolomkoppen of op `data-snyk-test` i.p.v. het verdwenen `data-vue`, herstel de tak-volgorde zodat de versies-tabel niet door de kwetsbaarheden-tak wordt opgeslokt, stop met het strippen van de severity, en leid het ecosysteem af uit de `Software`-CPE (met npm als fallback). **Risico:** de test-fixtures (`snyk-vuln*.html`) kunnen verouderd zijn; een selector die tegen de fixtures werkt hoeft op de huidige live snyk.io niet te kloppen. Vereist verificatie tegen actuele snyk.io-HTML.

**B. Weg van scrapen — Snyk's officiële API (robuuster, groter).**
HTML-scraping van snyk.io is per definitie fragiel en breekt bij elke frontend-wijziging. De duurzame oplossing is Snyk's REST-/DB-API (met API-token als plugin-instelling), die severity, CVSS, affected ranges en fixed-in gestructureerd teruggeeft. Groter van omvang en vereist auth/config, maar levert stabiele, rijkere data.

## Voorstel
Kies richting **B** als er animo is voor een API-token-instelling; anders **A** als tussenstap, mét een verse fixture van de huidige snyk.io-HTML zodat de selector aantoonbaar klopt. In beide gevallen: stop met het strippen van de severity, en leid het ecosysteem af uit de CPE.

_Gevonden tijdens een broncode-analyse van de boefjes/normalizers (Patroon 5 — aan-de-bron-keuzes en dode code)._

brennodewinter:
## Summary

Fixes #5301 — three problems in the `kat_snyk` boefje:

1. **`table_versions` always empty → `KAT-SOFTWARE-UPDATE-AVAILABLE` dead code**: The `data-vue="vulns-version-table"` attribute disappeared from snyk.io's HTML. The versions table was silently swallowed by the vulnerabilities-table branch and every row failed on a `KeyError` guard.
2. **Severity stripped**: `info["Vulnerability"].text.strip()[2:]` clipped the severity badge from the text, making severity unavailable to the normalizer/finding type.
3. **Ecosystem hardcoded to npm**: Only npm packages were queried; software from PyPI, Maven, RubyGems, Go, etc. was never checked against Snyk.

### Approach

Instead of patching CSS selectors (fragile — breaks on every frontend change) or switching to Snyk's API (requires an API token / config), this PR parses the `__NUXT_DATA__` JSON payload that snyk.io embeds as Nuxt SSR state. This gives structured data directly:

- Vulnerabilities with `severity`, `cvssScore`, CVE IDs, and structured affected-version ranges
- Latest version for the update-available check
- No secondary HTTP requests to individual vuln pages (CVE IDs are in the payload)

The ecosystem is now derived from the Software OOI's CPE `target_sw` field (10th component in CPE 2.3), falling back to `npm` when no CPE is available or `target_sw` is `*`.

### Changes

- **`main.py`**: Rewritten to parse `__NUXT_DATA__` JSON via a Nuxt reference resolver. Outputs `{"vulnerabilities": [...], "latest_version": "..."}`  instead of the old `table_versions`/`table_vulnerabilities`/`cve_vulnerabilities` format.
- **`normalize.py`**: Updated for the new data format. `risk_severity` and `risk_score` are now set on `SnykFindingType` / `CVEFindingType`. `KAT-SOFTWARE-UPDATE-AVAILABLE` works again (latest version available from payload).
- **Tests**: New fixture `snyk-vuln-nuxt.html` from current snyk.io. Old HTML fixtures (`snyk-vuln.html`, `snyk-vuln2.html`, `snyk-vuln3.html`) deleted — no longer referenced. Updated normalizer test fixtures to the new format. Added tests for severity propagation and CPE ecosystem derivation.
- **Net**: 9 files changed, 240 insertions, 1102 deletions (old fixtures were ~600KB combined).

#### Test plan
- [x] All 5 snyk tests pass (`pytest tests/plugins/test_snyk.py`)
- [x] pre-commit green (ruff, ruff-format, mypy, codespell)
- [ ] CI green

Closes #5301


boefjes/boefjes/plugins/kat_snyk/main.py
```diff
@@ -1,68 +1,142 @@
...
+    for vuln in pkg.get("vulnerabilities", []):
+        cve_ids = vuln.get("identifiers", {}).get("CVE", [])
+        cve = cve_ids[0] if cve_ids else None
+        affected = _affected_versions_to_string(vuln.get("affectedVersions", []))
+        vulnerabilities.append(
+            {
+                "id": vuln["id"],
+                "title": vuln["title"],
+                "severity": vuln.get("severity"),
+                "cvss_score": vuln.get("cvssScore"),
```
underdarknl: The reason why we did not include the cvss_score when reading Snyks output is because SNYK is not the authoritative source on these scores. As long as we have the cve ID, our cve boefje will hydrate the score by looking it up from the authoritative source selected there. This avoids scores moving back and forth between these two sources when one (presumably snyk) is out of sync.

boefjes/boefjes/plugins/kat_snyk/normalize.py
```diff
@@ -22,22 +29,27 @@ def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
...
-        logger.warning("Couldn't find software %s in the SNYK vulnerability database", software_name)
-        return
-    elif not results["table_vulnerabilities"] and not results["cve_vulnerabilities"]:
-        # no vulnerabilities found
+    vulnerabilities = results.get("vulnerabilities", [])
+    latest_version = results.get("latest_version")
+
+    if not vulnerabilities:
+        if not latest_version:
+            logger.warning("Couldn't find software %s in the SNYK vulnerability database", software_name)
```
underdarknl: Snyk covers a small percentage of all software available, lets not emit this as a warning, as its very likely and explainable that this happens. It warrants a debug message at best I think.

boefjes/boefjes/plugins/kat_snyk/normalize.py
```diff
@@ -22,22 +29,27 @@ def run(input_ooi: dict, raw: bytes) -> Iterable[NormalizerOutput]:
...
-            yield cve_ft
-            yield Finding(finding_type=cve_ft.reference, ooi=pk_ooi, description=vuln.get("Vuln_text"))
-    if not software_version and (results["table_vulnerabilities"] or results["cve_vulnerabilities"]):
+        for vuln in vulnerabilities:
+            severity = _SEVERITY_MAP.get(vuln.get("severity", "").lower())
+            cvss_score = vuln.get("cvss_score")
+
+            cve = vuln.get("cve")
+            if cve and cve.startswith("CVE-"):
+                ft = CVEFindingType(id=cve, risk_severity=severity, risk_score=cvss_score)
```
underdarknl: See earlier note on ignoring snyk's severity score and risk in favor of using the authoritative source.
SnykFindings themselves are fine with these scores though as snyk is obviously authoritative for its own findings.

underdarknl (changes requested):

brennodewinter:
@underdarknl — re-added the cvss_score on SnykFindingType as discussed. CVEFindingType stays score-free. Pushed to origin, ready for re-review.

hasecon (changes requested):
## QA Review

### Setup
- Branch: `fix/kat-snyk-nuxt-data` @ `b25f54d` (reviewed in an isolated worktree)
- Tests: boefjes plugin suite + `make itest` in Docker
- Live verification: the boefje run against the real snyk.io, not only the fixture

The move away from CSS selectors to `__NUXT_DATA__` is the right call, and it holds up: I ran the boefje against live snyk.io and it returned 12 vulnerabilities for lodash with `latest_version=4.18.1`, matching the fixture exactly. Dropping the per-vulnerability secondary requests is a real improvement. Two problems below are blocking, though, and the first one turns this boefje into a false-positive generator.

### Code Review Findings

| 1 | **Critical** | `normalize.py:41-52` | The installed-version check is gone — every vulnerability is reported regardless of the affected range |
| 2 | **High** | `main.py:18-40` | Ecosystem derivation is inert on real CPE data; PyPI/Maven/etc. are still queried as npm |
| 3 | Medium | `normalize.py:42` | `severity: null` crashes the normalizer |
| 4 | Medium | `main.py:112-114` | A snyk.io layout change again fails silently as a successful, empty scan |
| 5 | Low | `main.py:9` | Typo in the constant name: `_CPE_TARGET_SW_TO_SNYYK_ECOSYSTEM` |

---

#### 1. Critical — the installed version is no longer checked

On `main`, `main.py` filtered each row before emitting it:

```python
if check_version.check_version_in(software_version, parsed_info["Vuln_versions"]):
```

In this PR `main.py` computes `affected_versions` per vulnerability (`_affected_versions_to_string`, line 83) and writes it into the raw output — but nothing ever reads it again. `normalize.py:41` loops over every entry in `vulnerabilities` and yields a `Finding` for each. Grepping the tree confirms the only other reference to `affected_versions` is an assertion in the new test.

The consequence, measured against the PR's own fixture:

```
lodash 1.1.0  -> CVE-2026-4800, CVE-2026-2950, SNYK-JS-LODASH-73638, KAT-SOFTWARE-UPDATE-AVAILABLE

  check_version_in("1.1.0", "<4.18.1")         = True
  check_version_in("1.1.0", ">=4.0.0,<4.18.1") = False   <-- CVE-2026-2950 does not apply
  check_version_in("1.1.0", "<4.17.11")        = True
```

And the case that matters most — a fully patched install:

```
lodash 4.18.1 (== latest_version) -> CVE-2026-4800, CVE-2026-2950, SNYK-JS-LODASH-73638
                                     3 findings, of which 2 CVEs

  <4.18.1          -> False
  >=4.0.0,<4.18.1  -> False
  <4.17.11         -> False
```

Every user on the current version of any package covered by Snyk gets that package's entire historical vulnerability list as findings, two of them as CVE findings that will show up in the CVE reports. On `main` this produced zero findings. It is a regression, and in a security tool a false CVE is expensive: it costs the operator the time to disprove it.

The new test pins the wrong behaviour rather than catching it — `assert len(findings) == 4` for a version that only two of the three ranges cover.

**Suggested fix** (`normalize.py`, inside the `if software_version:` loop), which also closes finding 3:

```python
for vuln in vulnerabilities:
    affected = vuln.get("affected_versions")
    if affected and not check_version.check_version_in(software_version, affected):
        continue
    severity = _SEVERITY_MAP.get((vuln.get("severity") or "").lower())
```

I applied exactly this to a copy and re-ran it:

```
lodash 1.1.0  -> CVE-2026-4800, SNYK-JS-LODASH-73638, KAT-SOFTWARE-UPDATE-AVAILABLE
lodash 4.18.1 -> []
severity=None -> 2 OOIs, no crash
```

#### 2. High — the ecosystem is still npm for everything in practice

`_ecosystem_from_cpe` reads `target_sw` (CPE 2.3 field 10) — the index and the `len(parts) >= 11` guard are both correct. The problem is that essentially no real CPE in this codebase populates `target_sw`. Of the 66 unique CPEs across `boefjes/tests`, 5 have it set, and **4 of those 5 are the hand-written strings in this PR's own test**. The one genuine case is `cpe:2.3:a:nested_pages_project:nested_pages:*:*:*:*:*:wordpress:*:*`, and `wordpress` is not in the mapping, so it falls back to npm too.

The producers (`kat_nmap_tcp`, `kat_shodan`, `kat_shodan_internetdb`, `kat_censys`, `kat_binaryedge`) pass through upstream CPEs, which look like `cpe:2.3:a:f5:nginx:1.18.0:*:*:*:*:*:*:*`. Verified live with the real Django CPE:

```
cpe:2.3:a:djangoproject:django:4.2:*:*:*:*:*:*:*  ->  ecosystem "npm"

https://snyk.io/vuln/npm:django   ->    0 vulnerabilities
https://snyk.io/vuln/pip:django   ->  165 vulnerabilities
```

So problem 3 from the description ("software from PyPI, Maven, RubyGems, Go, etc. was never checked") is not actually fixed for the data OpenKAT has; the test passes because it feeds CPEs that production never emits. That is the same trap as #5302.

A `target_sw` lookup is a reasonable *first* signal, but it needs a fallback that works on bare CPEs. Options, your call: derive from the CPE vendor/product against a small known-package map, try the ecosystems in sequence and keep the first that returns a package, or take the ecosystem from the OOI that produced the `Software` (a `HostnameHTTPURL` source implies npm far more than an `IPPort` does). Happy to sketch whichever you prefer.

#### 3. Medium — `severity: null` crashes the normalizer

`main.py:126` writes `"severity": vuln.get("severity")`, which is `None` when the key is absent. `normalize.py:42` then does `vuln.get("severity", "").lower()` — the default only applies when the key is *missing*, not when it is present and null:

```
AttributeError: 'NoneType' object has no attribute 'lower'
```

Reproduced with a one-vulnerability raw. The current live fixture always carries a severity, so CI stays green, but the normalizer also runs against raw files already in Bytes. `(vuln.get("severity") or "")` fixes it; the snippet in finding 1 includes it.

#### 4. Medium — a layout change is still a silent no-op

When `_parse_nuxt_data` raises, the boefje logs a warning and returns a well-formed empty result:

```
boefje mime-types : set()            (no error/boefje)
boefje output     : {"vulnerabilities": [], "latest_version": null}
normalizer output : []
```

The task reports success with zero findings. That is precisely the failure mode of #5301 — the boefje quietly produced nothing for a long time and nobody noticed. Letting the `ValueError` propagate, so the failure reaches the standard `error/boefje` flow, would make the next snyk.io redesign visible on the day it happens instead of at the next audit. Not a regression (`main` was equally silent), but this PR is the natural place to close it.

### Suggested regression test

The gap is that no test asserts a *non-affected* version yields nothing. Drop-in for `boefjes/tests/plugins/test_snyk.py`:

```python
def test_snyk_skips_vulnerabilities_that_do_not_affect_the_installed_version():
    """A patched install must not inherit the package's whole vulnerability history."""
    patched = {"primary_key": "Software|lodash|4.18.1|", "name": "lodash", "version": "4.18.1"}
    oois = list(run(patched, get_dummy_data("inputs/snyk-result-findings.json")))

    # 4.18.1 is outside every affected range in the fixture (<4.18.1, >=4.0.0 <4.18.1, <4.17.11)
    assert [o for o in oois if isinstance(o, Finding)] == []


def test_snyk_reports_only_the_matching_ranges():
    oois = list(run(input_ooi, get_dummy_data("inputs/snyk-result-findings.json")))
    ids = {o.finding_type.tokenized.id for o in oois if isinstance(o, Finding)}

    # 1.1.0 matches <4.18.1 and <4.17.11, but not >=4.0.0,<4.18.1
    assert "CVE-2026-2950" not in ids
    assert ids == {"CVE-2026-4800", "SNYK-JS-LODASH-73638", "KAT-SOFTWARE-UPDATE-AVAILABLE"}
```

Verified both directions: on `b25f54d` the first fails with 3 findings instead of 0 and the second fails on `CVE-2026-2950`; with the snippet from finding 1 applied, both pass. `test_snyk_findings` needs its `assert len(findings) == 4` changed to 3 at the same time, since it currently pins the defect.

### Test Results

| Component | Test | Result |
|-----------|------|--------|
| boefjes | `tests/plugins/test_snyk.py` | 5 passed |
| boefjes | `tests/plugins/` | 174 passed |
| boefjes | `make itest` | 36 passed, 2 deselected (exit 0) |
| boefjes | boefje against live snyk.io | lodash: 12 vulns, latest 4.18.1 — parser confirmed against the real page |

### CI/CD Status

18/18 green (`pre-commit`, `Tests 3.10–3.13`, CodeQL, all four `build (…, boefjes)` jobs). No blocking check is red — the verdict below rests on the findings, not on CI.

Worth noting that green CI here is the point of finding 1: the suite never asserts what happens to a version that is *not* affected, so the regression passes every check.

### Functional Test

No `make clean && make kat` for this round, deliberately: the diff is confined to `kat_snyk`'s `main.py` and `normalize.py` plus fixtures, with no change to service contracts, compose, or any cross-service path, and I exercised the full chain the stack would exercise — boefje → raw JSON → normalizer → OOIs — against the live snyk.io endpoint rather than a mock. Say the word if you would rather see it on a running stack anyway and I will run it.

### Verdict

**CHANGES REQUESTED**

The parsing rewrite is sound and I would like to see it land. Blocking on finding 1: as it stands the boefje reports a package's entire vulnerability history against every installed version, including fully patched ones, which is a regression from `main` and produces false CVE findings. Finding 2 means the headline "works for PyPI/Maven/etc. now" does not hold for the CPEs OpenKAT actually has. Findings 3 and 4 are small and can ride along.

brennodewinter:
@hasecon — alle vier de findings verwerkt in `470dac6f`:

1. **Affected-range filtering**: `normalize.py` filtert nu op `affected_versions` via `check_version.check_version_in()` — `lodash 1.1.0` geeft 1 CVE + 1 Snyk + 1 update-finding, `4.18.1` geeft niets meer.
2. **Parse-fouten propageren**: ontbrekende/malformed `__NUXT_DATA__` gooit nu een exception i.p.v. stil een leeg resultaat — geen valse "succesvolle lege scan" meer.
3. **CPE-ecosystem**: `target_sw` wordt getriaged tegen alle gescande packages; fallback is prefix-match op `product` tegen `name`. De constante heet nu ook correct `_CPE_TARGET_SW_TO_SNYK_ECOSYSTEM`.
4. **`severity: null`**: null-safe.

Regressietests toegevoegd (gepatchte versie → 0 findings, range-matching, null-severity, ecosystem-fallback); 10 tests groen. De regressie-fixture uit je review is overgenomen.

PR #5378, opened 2026-09-09
Strip boefje environment secrets from raw meta download (#4508)

### LINKED ISSUE #4508: Task list raw meta download contains secrets

brennodewinter:
## Summary
- The raw meta download (`BytesRawView`) returned the full `boefje_meta` dict from the Bytes API, including the `environment` field which may contain secrets (#4508).
- Pop `environment` from each `raw_meta`'s `boefje_meta` before returning the response, covering both the JSON and zip download paths.
- `DownloadTaskDetail` is not affected — it uses pydantic models (`BoefjeMeta` in `scheduler.py`) that don't define `environment`, so the field is already dropped during validation.

#### Test plan
- [x] New test `test_download_task_strips_environment_secrets` verifies `environment` is absent from the zip's raw_meta JSON files.
- [x] All 10 tests in `test_boefjes_tasks.py` pass.
- [x] pre-commit green.

Closes #4508.

### DIFF
--- a/rocky/rocky/views/bytes_raw.py
+++ b/rocky/rocky/views/bytes_raw.py
@@ -23,6 +23,8 @@ class BytesRawView(OrganizationView):
         boefje_meta_id = kwargs["boefje_meta_id"]
         try:
             raw_metas = self.bytes_client.get_raw_metas(boefje_meta_id, self.organization.code)
+            for raw_meta in raw_metas:
+                raw_meta["boefje_meta"].pop("environment", None)
             is_json_format = request.GET.get("format") == "json"
             if is_json_format:
                 size_limit = int(request.GET.get("size_limit", RAW_FILE_LIMIT))
--- a/rocky/rocky/views/bytes_raw.py
+++ b/rocky/rocky/views/bytes_raw.py
@@ -23,8 +23,7 @@ class BytesRawView(OrganizationView):
         boefje_meta_id = kwargs["boefje_meta_id"]
         try:
             raw_metas = self.bytes_client.get_raw_metas(boefje_meta_id, self.organization.code)
-            for raw_meta in raw_metas:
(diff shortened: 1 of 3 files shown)


rocky/rocky/views/bytes_raw.py
```diff
@@ -23,6 +23,8 @@ def get(self, request, **kwargs):
         boefje_meta_id = kwargs["boefje_meta_id"]
         try:
             raw_metas = self.bytes_client.get_raw_metas(boefje_meta_id, self.organization.code)
+            for raw_meta in raw_metas:
+                raw_meta["boefje_meta"].pop("environment", None)
```
underdarknl: Same remarks as in your other PR, please dont drop the entire Env, but only mask/drop the fields that are actually marked as secret by the boefje's schema.

underdarknl (changes requested):

rocky/rocky/views/bytes_raw.py
```diff
@@ -23,6 +23,7 @@ def get(self, request, **kwargs):
         boefje_meta_id = kwargs["boefje_meta_id"]
         try:
             raw_metas = self.bytes_client.get_raw_metas(boefje_meta_id, self.organization.code)
+            _strip_environment_secrets(raw_metas, self.katalogus_client)
```
underdarknl, line 26: What would we want to do with this when a user does have the rights to see/edit secrets via the katalogus?
It could be sane to clean the secrets when just viewing, making sure they do not get exported by accident. There's also an argument to be made where we want to show the user the actual input/output.

PR #5133, opened 2026-04-20
Fix/bytes login

underdarknl:
### Changes

this moves various login calls to the bytes-client init.
The init is called *per* request, and as such can re-use the token without dealing with token rotation.

### Issue link

closes #5132 

### Demo

_Please add some proof in the form of screenshots or screen recordings to show (off) new functionality, if there are interesting new features for end-users._

### QA notes

_Please add some information for QA on how to test the newly created code._

---

### Code Checklist

- [ ] All the commits in this PR are properly PGP-signed and verified.
- [ ] This PR only contains functionality relevant to the issue.
- [ ] I have written unit tests for the changes or fixes I made.
- [ ] I have checked the documentation and made changes where necessary.
- [ ] I have performed a self-review of my code and refactored it to the best of my abilities.

- [ ] Tickets have been created for newly discovered issues.
- [ ] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [ ] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [ ] I have included comments in the code to elaborate on what is not self-evident from the code itself, including references to issues and discussions online, or implicit behavior of an interface.

---

## Checklist for code reviewers:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_qa.md) into your comment.

PR #5380, opened 2026-09-09
Cache Bytes clients to reduce redundant token requests

### LINKED ISSUE #4442: Bytes token invalidates after login
**Describe the bug**
A while ago together with @jpbruinsslot we observed that there were a lot of Bytes authentication requests coming from different services. This is because there's only one token shared with all the services to access the Bytes API.

This number of requests for this is not only unnecessary, but is also troublesome when trying to debug a specific application, because while debugging the token can be invalidated by other services.

To improve this we should consider issuing per-service tokens.

**To Reproduce**

- Make sure every service is running
- After a few minutes, you should see different authentication requests across the logs

**Expected behavior**
From a client's perspective, authentication should only happen once per session (e.g. on startup or when logging out)

**Additional context**
To tackle this problem, we should enable session tokens per client. This is the best approach for now and ensures backwards compatibility.

A psuedocode implementation (inspired by @underdarknl):
```mermaid
sequenceDiagram
    participant Client
    participant Server

    Client->>Server: login(username, password)
    Server-->>Client: timestamp, hash(secret, username, timestamp)

    note right of Client: Client sends auth credentials on every request (e.g., via headers or cookies)
    
    Client->>Server: authenticate(username, timestamp, hash)
    Server->>Server: validate_authentication(username, timestamp, hash)
    alt valid and timestamp not too old
        Server-->>Client: Serve content
    else invalid or expired
        Server-->>Client: Serve 401 Unauthorized
    end
```

brennodewinter:
## Summary

- `get_bytes_client()` in both Rocky and Boefjes created a new Bytes client on every call, causing a fresh token fetch per HTTP request even though the stateless JWT token is valid until expiry
- Wrapped both `get_bytes_client` factories with `lru_cache` so the client (and its cached token) is reused across requests — one token per process per organization instead of one per request
- Added an autouse test fixture to clear the `lru_cache` between tests so mocks don't leak across test cases

Closes #4442

#### Root cause

As noted in the issue discussion by @dekkers: the tokens are stateless signed JWTs that can only expire, not be invalidated. The excessive token requests were caused by client reinstantiation — each new `BytesClient`/`BytesAPIClient` instance fetches a new token. PR #5133 already improved the Rocky client with `TokenAuth` + `@cached_property` for the token, but the client itself was still recreated per request. This PR caches the client instance so the token is reused.

#### Test plan

- [x] Rocky tests: `test_upload_raw`, `test_boefjes_tasks`, `test_report_overview`, `reports/`, `test_upload_csv`, `test_organization`, `test_objects_edit` — all pass
- [x] Boefjes tests: `test_api`, `test_app` — pass (1 pre-existing docker failure unrelated)
- [x] Verified caching logic with standalone assertion (same org → same instance, different org → different instance)

### DIFF
--- a/boefjes/boefjes/api.py
+++ b/boefjes/boefjes/api.py
@@ -1,6 +1,7 @@
 import multiprocessing
 import uuid
 from datetime import datetime, timezone
+from functools import lru_cache
 from multiprocessing.context import ForkContext, ForkProcess
 from typing import Any
 from uuid import UUID
@@ -52,6 +53,7 @@ def get_scheduler_client(plugin_service=Depends(get_plugin_service)):
     return SchedulerAPIClient(plugin_service, str(settings.scheduler_api))
 
 
+@lru_cache(maxsize=1)
 def get_bytes_client():
     return BytesAPIClient(str(settings.bytes_api), username=settings.bytes_username, password=settings.bytes_password)
 
--- a/rocky/rocky/bytes_client.py
+++ b/rocky/rocky/bytes_client.py
@@ -3,7 +3,7 @@ import uuid
 from base64 import b64decode, b64encode
 from collections.abc import Generator, Sequence, Set
 from datetime import datetime, timezone
-from functools import cached_property
+from functools import cache, cached_property
 from typing import Any
 
 import httpx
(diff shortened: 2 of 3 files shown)


brennodewinter:
Closing as duplicate of #5386, which is more complete (dedicated test file, same Bytes token caching fix).

Issue #5321, opened 2026-08-12
Design: CVE findings belong on the Software OOI — fix reporting queries, align existing binders, and add a PossibleAttack OOI


hasecon:
## Principle

A CVE is a property of a **software version**, not of a host. It should live on a single `Software` OOI (the version node), and the answer to *"which CVEs affect this host/website?"* should come from **traversing the graph** (`Software <- SoftwareInstance <- asset`) in the reporting layer — not from binding (and thereby duplicating) the finding onto every `SoftwareInstance`, and not onto the bare asset.

This is @underdarknl's direction, from the #5317 discussion:

> The CVEs, as far as I'm concerned, [belong] to the software (not even the SoftwareInstance). To get the desired result (which CVEs belong to this host or website) we'll have to fix a lot of querying in rocky/reports. Ultimately that's just the cleaner route.
>
> That's also because I want to start pulling CVEs from SBOMs and dependencies, where the specific dependency contains a CVE, and all using software — and in turn the websites, hosts, IPs and other assets using those — must *find* this finding through querying, instead of duplication or mis-association in the graph.

Doing it this way now makes the model better in depth instead of adding to tech debt.

## Why not the alternatives

- **Not the `SoftwareInstance`**: one CVE finding per (software, host) duplicates the same vulnerability across every host and re-derives it on each rescan; an SBOM/dependency CVE that fans out to many packages and assets would explode into N near-identical findings that are hard to keep consistent.
- **Not the bare asset (hostname/IP port)**: loses *which software* carries the CVE, and can't be reused by the SBOM/dependency path where the same dependency node is shared by many assets.
- **The single `Software` node** is the natural join: emit once, discover everywhere by relation.

## Current state (inconsistent — the rest does it too)

| Producer | Binds CVE `Finding` to |
|---|---|
| `kat_leakix` (after #5317) | ✅ `Software` |
| `octopoes/bits/retire_js` | `SoftwareInstance` |
| `kat_shodan` normalizer | `IPPort` (asset) |
| `kat_nuclei_cve` | URL / hostname (input) |

So today no two producers agree, and `Software`-node binding currently has only the LeakIX PR as precedent. This issue is the decision to standardize on it and do the surrounding work.

## What needs to happen

[...]

PR #5383, opened 2026-09-09
Surface CVE findings bound to Software on the assets that run it

hasecon:
Implements step 1 of #5321: make the graph traversal work, so a CVE that hangs on a bare `Software` OOI is visible on the assets running that software.

`PossibleAttack` is skipped as requested (underdarknl, #5321 comment). Aligning the binders (`retire_js`, shodan/nuclei normalizers) to actually bind CVEs to `Software` is the follow-up — this PR is the reading side, so the binder change lands on a UI that can already show the result.

## Problem

A CVE is a property of a software *version*, so its finding belongs on the `Software` OOI. But `Software._traversable = False` ([software.py:34](https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/blob/main/octopoes/octopoes/models/ooi/software.py#L34)) — deliberately, since `Software` is a shared node and traversing it would fan out across every host running the same package.

The consequence is that every consumer that walks the object tree stops at the `SoftwareInstance` and never reaches the finding:

| Consumer | Before |
|---|---|
| Object detail → Findings tab | finding invisible |
| Findings report | finding missing |
| Aggregate / multi-organisation report | finding missing (they reuse `FindingsReport`) |

Flipping `_traversable` would fix the symptom and reintroduce the fan-out, so instead both consumers walk the last two hops explicitly.

## Changes

**`rocky/tools/ooi_helpers.py`** — the traversal itself, shared by the view and the report:
- `collect_software_instances(tree)` maps every `SoftwareInstance` in a tree to the OOI it is installed on.
- `findings_on_software(...)` walks `SoftwareInstance.software.<ooi[is Finding]` with `query_many` and returns each finding together with the asset running it. Sources are chunked, because `query_many` puts every source in the query string and a host with hundreds of services would otherwise build a request line past what the API server accepts. A finding reachable through several instances (one package on two ports) is reported once, for the first asset it was found on.

**`rocky/rocky/views/ooi_detail_related_object.py`** — `OOIFindingManager` now combines the tree findings with the software findings, and resolves finding types that the tree does not carry via `load_objects_bulk`. `get_finding_details()` no longer `KeyError`s on an unresolvable finding type.

**`rocky/reports/report_types/findings_report/report.py`** — the existing tree call asks for `{Finding, SoftwareInstance}`, so the instances come out of the *same* pass with no extra tree walk. Occurrences now carry `affected_ooi`.

**Aggregate + multi-organisation report** deduplicate occurrences on `affected_ooi` instead of `finding.ooi`. This matters: `finding.ooi` of a CVE is the shared `Software` OOI, so without this twenty hosts running the same vulnerable nginx collapse into a single occurrence.

**`report_findings_table.html`** renders `affected_ooi`, so the row names the asset rather than "nginx 1.18.0" linking to the Software page.

## Reaching the software: why not a fixed path

The first version of this PR walked a single hop, `{Type}.<ooi[is SoftwareInstance]`. That is wrong: `SoftwareInstance.ooi` is bound to whatever the boefje observed — `IPPort` (nmap, shodan, censys, binaryedge, service_banner, dicom), `IPService`, `IPAddress` (internetdb), `HostnameHTTPURL` (wappalyzer) — and never to a `Hostname`. So on a hostname page it returned nothing at all, which is exactly the page the issue names.

Rather than maintain a path list per input type, both consumers now collect the `SoftwareInstance`s from the object tree and query from there. That is type-agnostic and stays correct if a new boefje binds software somewhere else.

## Open decision for you: the severity tile

Reaching the software from a `Hostname` needs a deeper tree than the two levels the detail page renders (`Hostname → ResolvedHostname → IPAddress → IPPort → SoftwareInstance`), so it is a second graph call.

Nine existing tests assert that an object detail page fetches its tree **exactly once**, which I read as a deliberate invariant after #5088. So only the Findings tab pays for the extra call (`include_software_findings = True` on `OOIFindingListView`); every other detail page is unchanged.

The consequence is that **the severity tile on the Overview tab still counts tree-only**, so it can show fewer findings than the Findings tab lists. Three ways out, your call:

1. Leave it (this PR) — no cost regression anywhere, tile and tab can disagree.
2. Let `OOIDetailView` pay for the second call too — consistent, but every object detail page makes two tree calls and those nine assertions change to 2.
3. Keep the tile tree-only but label it as such in the template.

I did not want to pick 2 unilaterally given the #5088 history.

## Tests

Every production change is pinned by a revert test — reverted individually, each one goes red:

| Reverted | Result |
|---|---|
| software traversal off | 3 failed |
| report `types={Finding, SoftwareInstance}` → `{Finding}` | 1 failed |
| `affected_ooi` attribution removed | 2 failed |
| first-asset guard removed | 1 failed |

New tests: the CVE surfacing on a hostname's findings tab (asserting the traversal asks for `SoftwareInstance` at a depth greater than the page's own two), one package on two ports collapsing to one row, no query at all when nothing runs software, the report attributing to the asset instead of the Software OOI, direct findings keeping their own OOI, and shared software reported once on the first asset.

`tests/conftest.py`: the mock connector now prunes the tree store by the requested types, as octopoes does. Without this the report's `types=` change was untestable — reverting it left every test green.

## Verification

- `cd rocky && make utest` → **486 passed, 1 skipped**
- `cd rocky && make itest` → **5 passed, 1 deselected**
- `pre-commit run --all-files` → green (the only red hook is `robotidy` reformatting an unrelated octopoes robot file that is already unformatted on `main`)
- CI on `a2c57c577`: **19/19 green**, `makelang` included

Functional test on a live stack is not included: it needs a CVE actually bound to a `Software` OOI, which is exactly what the follow-up binder change produces. Happy to run it once that lands, or on a hand-declared `Software` + `Finding` if you would rather see it before merging this.

## Second thing worth your eye

For the report the traversal starts from every `SoftwareInstance` in the tree, so a hostname resolving to a shared host will pick up the CVEs of software on that host. That is the intended reading of "affected assets", but it is also the fan-in I flagged on #5302 — say the word if you want the report restricted to instances bound directly to the input OOI.

### DIFF
--- a/rocky/reports/report_types/aggregate_organisation_report/report.py
+++ b/rocky/reports/report_types/aggregate_organisation_report/report.py
@@ -234,7 +234,9 @@ class AggregateOrganisationReport(AggregateReport):
                 seen_keys = set()
 
                 for occurrence in finding_type["occurrences"]:
-                    occurrence_ooi = occurrence["finding"].ooi
+                    # Not finding.ooi: a CVE hangs on the shared Software OOI, so two assets
+                    # running the same vulnerable package would collapse into one row (#5321).
+                    occurrence_ooi = occurrence["affected_ooi"]
 
                     if occurrence_ooi not in seen_keys:
                         seen_keys.add(occurrence_ooi)
--- a/rocky/reports/report_types/findings_report/report.py
+++ b/rocky/reports/report_types/findings_report/report.py
@@ -2,6 +2,7 @@ from datetime import datetime
 from typing import Any
 
 from django.utils.translation import gettext_lazy as _
+from tools.ooi_helpers import collect_software_instances, findings_on_software
 
 from octopoes.models import Reference
 from octopoes.models.ooi.dns.zone import Hostname
@@ -13,11 +14,6 @@ from reports.report_types.definitions import Report, ReportPlugins
 
 TREE_DEPTH = 9
 SEVERITY_OPTIONS = [severity.value for severity in RiskLevelSeverity]
-
-# A CVE is a property of a software version, so its finding hangs on the Software OOI rather than
-# on the asset (#5321). Software is not traversable, so the object tree stops at the
-# SoftwareInstance; walk the last two hops with a query instead.
-SOFTWARE_FINDINGS_PATH = "SoftwareInstance.software.<ooi[is Finding]"
 
 
 class FindingsReport(Report):
@@ -40,21 +36,6 @@ class FindingsReport(Report):
     template_path = "findings_report/report.html"
     label_style = "3-light"
 
-    def get_software_findings(
-        self, software_instances: list[Reference], seen: set[Reference], valid_time: datetime
-    ) -> list[Finding]:
-        """Findings on the software behind the given instances, which the object tree cannot reach."""
-        findings = []
-
-        for _source, ooi in self.octopoes_api_connector.query_many(
-            SOFTWARE_FINDINGS_PATH, valid_time, software_instances
-        ):
-            if isinstance(ooi, Finding) and ooi.reference not in seen:
-                seen.add(ooi.reference)
-                findings.append(ooi)
-
-        return findings
-
     def generate_data(self, input_ooi: str, valid_time: datetime) -> dict[str, Any]:
         reference = Reference.from_str(input_ooi)
         findings = []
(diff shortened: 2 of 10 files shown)


brennodewinter (approved):
Excellent PR. The traversal approach is sound — collecting SoftwareInstances from the tree and walking the last two hops explicitly avoids the fan-out problem of making Software traversable, while still surfacing CVEs on the assets that run the software.

Code review notes:

-  /  in ooi_helpers.py: clean, the chunking for query_many (50 sources) is a sensible guard against URL length limits. First-asset-wins dedup is clear.
- : good separation between  and . The  flag ensuring only the Findings tab pays for the extra tree call is the right call given the #5088 history.  with  fallback fixes the KeyError on unresolvable finding types.
- Report changes: dedup on  instead of  is correct — without it, twenty hosts running the same vulnerable nginx would collapse into one occurrence.
- conftest.py mock connector pruning by requested types: mirrors octopoes behavior, makes the  change testable.
- Tests: comprehensive — CVE surfacing, dedup (one package on two ports), no-query-when-no-software, report attribution, direct findings keep their own OOI, shared software reported once.

On the two open decisions you flagged:

1. **Severity tile**: I'm fine with option 1 (leave it, this PR). The Overview tab counting tree-only is an acceptable trade-off vs. doubling tree calls on every detail page. The Findings tab is where users go to act on findings, and that's correct.

2. **Fan-in** (shared host picking up CVEs from software on that host): acceptable for now. The PR body already flags this for follow-up. Restricting to directly-bound instances can be a separate change if it turns out to matter in practice.

Approving.

brennodewinter (approved):
Excellent PR. The traversal approach is sound — collecting SoftwareInstances from the tree and walking the last two hops explicitly avoids the fan-out problem of making Software traversable, while still surfacing CVEs on the assets that run the software.

Code review notes:

- collect_software_instances / findings_on_software in ooi_helpers.py: clean, the chunking for query_many (50 sources) is a sensible guard against URL length limits. First-asset-wins dedup is clear.
- OOIFindingManager: good separation between get_direct_findings and software_findings. The include_software_findings flag ensuring only the Findings tab pays for the extra tree call is the right call given the #5088 history. get_finding_types with load_objects_bulk fallback fixes the KeyError on unresolvable finding types.
- Report changes: dedup on affected_ooi instead of finding.ooi is correct — without it, twenty hosts running the same vulnerable nginx would collapse into one occurrence.
- conftest.py mock connector pruning by requested types: mirrors octopoes behavior, makes the types= change testable.
- Tests: comprehensive — CVE surfacing, dedup (one package on two ports), no-query-when-no-software, report attribution, direct findings keep their own OOI, shared software reported once.

On the two open decisions you flagged:

1. Severity tile: I'm fine with option 1 (leave it, this PR). The Overview tab counting tree-only is an acceptable trade-off vs. doubling tree calls on every detail page. The Findings tab is where users go to act on findings, and that's correct.

2. Fan-in (shared host picking up CVEs from software on that host): acceptable for now. The PR body already flags this for follow-up. Restricting to directly-bound instances can be a separate change if it turns out to matter in practice.

Approving.

underdarknl:
This Pr was not yet ready for merge.
"Open decision for you:"

And from your comment, "Severity tile: I'm fine with option 1" Are you the maintainer?

PR #5386, opened 2026-09-10
Cache Bytes API token across client instances (#4442)

### LINKED ISSUE #4442: Bytes token invalidates after login

brennodewinter:
## Summary

- **Root cause**: every `get_bytes_client()` call constructed a new `BytesClient`/`BytesAPIClient` that fetched a fresh token on first use, causing a token request per HTTP request even though tokens are valid for 15 minutes (#4442).
- **Rocky fix**: cache the token at module level (`_token_cache` dict keyed by username) so it is reused across `BytesClient` instances. The existing `TokenAuth` 401-retry invalidates the cache and fetches a new token when needed.
- **Boefjes fix**: `get_bytes_client()` in `api.py` now returns the existing `bytes_api_client` singleton from `job_handler` instead of constructing a new instance per request.

Closes #4442

#### Test plan

- [x] 3 new unit tests in `rocky/tests/test_bytes_client.py` (token cached across instances, invalidation refetches, different users cached separately)
- [x] All 26 existing bytes/upload tests pass

### DIFF
--- a/boefjes/boefjes/api.py
+++ b/boefjes/boefjes/api.py
@@ -15,6 +15,7 @@ from boefjes.clients.bytes_client import BytesAPIClient
 from boefjes.clients.scheduler_client import SchedulerAPIClient
 from boefjes.config import settings
 from boefjes.dependencies.plugins import get_plugin_service
+from boefjes.job_handler import bytes_api_client
 from boefjes.worker.interfaces import BoefjeInput, BoefjeOutput, StatusEnum, Task, TaskStatus, WorkerManager
 from boefjes.worker.repository import _default_mime_types
 
@@ -53,7 +54,7 @@ def get_scheduler_client(plugin_service=Depends(get_plugin_service)):
 
 
 def get_bytes_client():
-    return BytesAPIClient(str(settings.bytes_api), username=settings.bytes_username, password=settings.bytes_password)
+    return bytes_api_client
 
 
 @app.get("/healthz")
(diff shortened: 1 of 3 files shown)


underdarknl:
@brennodewinter could you see if the github-code-quality remark can be fixed?

PR #5387, opened 2026-09-10
Pre-fill Question answer form with saved config (#3982)

### LINKED ISSUE #3982: Updating answered question shows default config instead of user configured config
**Describe the bug**
When you answer one of the Question objects this generates a config. If you decide that you want to change the config, you re-answer the Question. Currently the default config is reloaded instead of the user-applied config. Editing the Config itself technically works, but is not user friendly as the text area is very small and it's json. Updating from the Question object is more user friendly and the preferred approach. 

**To Reproduce**

1. Go to Objects.
2. Open a Question and answer the question add some non-default values.
3. Save the Question.
4. Re-open the Question and observe that the default config is loaded instead of your just applied changes.

**Expected behavior**
Re-answering the question should load the already present config. 

**Screenshots**
Example of one of the questions.
Image

**OpenKAT version**
main

brennodewinter:
## Summary

- Re-answering a Question loaded the JSON schema defaults instead of the user's previously saved config.
- The view already fetched the saved `Config` OOI into `current_config` but only passed `json_schema` to the form template, never the saved config.
- The JS (`jsonSchemaToForm.js`) reads `data-original` to pre-fill existing values; since it was never set, it fell back to schema defaults.
- Fix: pass `current_config` as a JSON string via `data-original` on the textarea so the form pre-fills with the user's previous answer.

Closes #3982

#### Test plan

- [x] All 12 existing OOI detail tests pass
- [ ] Manual: answer a Question, re-open it, verify the form shows the saved answer instead of defaults

### DIFF
--- a/rocky/rocky/templates/forms/json_schema_form.html
+++ b/rocky/rocky/templates/forms/json_schema_form.html
@@ -5,7 +5,7 @@
     {% csrf_token %}
     <fieldset class="indented">
         <input type="hidden" name="action" value="submit_answer">
-        <textarea name="schema" class="JsonSchemaForm">
+        <textarea name="schema" class="JsonSchemaForm"{% if original_config %} data-original="{{ original_config }}"{% endif %}>
       {{ json_schema }}
     </textarea>
         <legend>{% translate "Fill out this form to answer the Question (again):" %}</legend>
--- a/rocky/rocky/templates/oois/ooi_detail.html
+++ b/rocky/rocky/templates/oois/ooi_detail.html
@@ -23,7 +23,7 @@
                         {% include "partials/current_config.html" with config=current_config %}
 
                     {% endif %}
-                    {% include "forms/json_schema_form.html" with json_schema=ooi.json_schema %}
+                    {% include "forms/json_schema_form.html" with json_schema=ooi.json_schema original_config=current_config_json %}
 
                 {% endif %}
             </div>
--- a/rocky/rocky/views/ooi_detail.py
+++ b/rocky/rocky/views/ooi_detail.py
@@ -145,6 +145,7 @@ class OOIDetailView(BaseOOIDetailView, OOIRelatedObjectManager, OOIFindingManage
                 context["current_config"] = self.get_ooi(self.ooi.config_pk).config
             except Exception:
                 context["current_config"] = None
+            context["current_config_json"] = json.dumps(context["current_config"]) if context["current_config"] else ""
 
         context["related"] = self.get_related_objects(self.observed_at)
 


hasecon (changes requested):
## QA Review

### Setup
- Branch: `fix/3982-question-reanswer-config` @ `e2e9392` (reviewed in an isolated worktree)
- View rendered with the Django test client to check the actual HTML, including edge cases the suite does not cover

The core fix is correct and I verified it end to end: `current_config_json` reaches the template, `data-original` lands on the textarea, and `loadform` picks it up via `schemafield.dataset.original` with a `try`/`catch` around the `JSON.parse`. The plumbing is right.

Two things worth calling out positively, because they are easy to get wrong:

- **The description rendering is XSS-safe.** `escapeHtml` runs *before* `addLinks`, and the entities that end up inside `href="…"` stay inside the attribute value — the HTML parser decodes `&quot;` after delimiting, so it cannot break out. I checked `https://x.com/<script>alert(1)</script>` as a description: fully inert. The `https?://` anchor in the regex also rules out `javascript:`.
- **The checkbox `input.value = field["default"]` is harmless**, because `ContentFromPostObject` reads booleans through `.checked`, never `.value`.

The PR does rather more than the description says — the label, description, `(Required)` marker and falsy-value handling in `jsonSchemaToForm.js` are a separate improvement riding along. That is fine by me, but it is where all the findings are.

### Code Review Findings

| 1 | Medium | `views/ooi_detail.py:150` | An empty saved config shows "Using config from" while the form silently loads schema defaults |
| 2 | Medium | `jsonSchemaToForm.js:428` | Falsy answers are pre-filled but still dropped on save — the round trip loses them |
| 3 | Medium | `jsonSchemaToForm.js:357` | `(Required)` is hardcoded English, in a PR that translates its other new strings |
| 4 | Low | `json_schema_form.html:7` | `?ooi_id={{ current_config_pk }}` is HTML-escaped but not URL-encoded |
| 5 | Low | `views/ooi_detail.py:147` | A bare `except Exception` now drives a user-visible "Currently unanswered." |
| 6 | Low | `jsonSchemaToForm.js:38` | `url.startsWith("www.")` is unreachable; `descriptionfield.id` is unused and can collide |

---

#### 1. Medium — an empty saved answer claims to be loaded but isn't

`current_config_pk` and `current_config_json` are computed independently:

```python
context["current_config_pk"] = self.ooi.config_pk
context["current_config"] = self.get_ooi(self.ooi.config_pk).config
...
context["current_config_json"] = json.dumps(context["current_config"]) if context["current_config"] else ""
```

A saved `Config` with `config={}` is truthy as an object but falsy as a dict, so `current_config_pk` is set while `current_config_json` becomes `""`, and the template's `{% if current_config %}` drops `data-original`. Verified by rendering the view with `config={}`:

```
assertContains(r, "Using config from")   # passes — page claims a saved config
assertNotContains(r, "data-original")    # also passes — form falls back to schema defaults
```

So the user is told their saved answer is in use while looking at the defaults — the exact confusion #3982 is about, just one step further in. `json.dumps(...) if context["current_config"] is not None else ""` fixes it.

#### 2. Medium — the round trip still loses falsy answers

The commit is titled "Prefill saved falsy answers" and the render half is now right (`originalvalue !== undefined` throughout). The save half is not. `ContentFromPostObject`:

```js
if (data || fieldtype == "boolean") {
  ...
  content[fieldname] = data;
}
```

For a non-boolean field `data` is always a string from `.value`, so `"0"` survives — but `""` is falsy and the key is dropped. A user who has `{"c": ""}` saved now sees it pre-filled correctly, presses Submit without touching anything, and the key disappears from the config. `ContentFromPostArray` has the same shape (`if (!data) break;`), which truncates an array at its first empty element.

This is pre-existing rather than a regression, so I am not treating it as blocking on its own — but it is the other half of what this commit set out to fix, and the PR is the natural place to close it. Reported from reading the code; I could not execute it, see the note on test coverage below.

#### 3. Medium — `(Required)` will be English in every language

```js
label.innerHTML = required
  ? `${escapeHtml(name)} <span class="nota-bene" aria-hidden="true">(Required)</span>`
  : escapeHtml(name);
```

Rocky ships eight-plus locales and this PR adds `"Using config from"` and `"Currently unanswered."` to `django.pot` in the same change, so the inconsistency is visible in the diff itself. There is no JS gettext catalog in this project, but there is an established convention for exactly this — `pluginToggler.js` reads `button.dataset.showText`, `sidemenuOl.js` reads `sidemenu.dataset.openLabel`. Following it here means one more attribute on the textarea:

```html
<textarea name="schema" class="JsonSchemaForm"
          data-required-label="{% translate 'Required' %}"
          {% if current_config %}data-original="{{ current_config }}"{% endif %}>
```

and reading it in `loadform` alongside `dataset.original`.

#### 4. Low — the config link is not URL-encoded

```html
<a href="?ooi_id={{ current_config_pk }}">{{ current_config_pk }}</a>
```

Django escapes this for HTML, not for a query string, so an `&` or `#` in the primary key would truncate the parameter. In practice this is currently safe: all three question-asking bits (`ask_port_specification`, `ask_url_params_to_ignore`, `ask_disallowed_domains`) bind to a `Network`, and `Question.config_pk` builds `Config|{network.name}|{config_id}`. So this is defensive only — but `{{ current_config_pk|urlencode }}` costs nothing and keeps it correct if a Question is ever attached to something with a richer key.

#### 5. Low — a transient Octopoes error now reads as "you have no answer"

The `except Exception` was already there, but until this PR it only affected `current_config`. Now it also sets `current_config_pk = None`, which renders "Currently unanswered." So a connection error or timeout tells the user their saved answer is gone, rather than that the lookup failed. Catching `ObjectNotFoundException` specifically and letting the rest propagate would keep the message honest.

#### 6. Low — two small dead ends in the new JS

- `addLinks` checks `if (url.startsWith("www."))`, but `urlRegex` is `/https?:\/\/[^\s]+/gi`, so a match can never start with `www.` — the branch is unreachable.
- `descriptionfield.id = name + "-description"` is assigned but never referenced. If the intent was `aria-describedby` on the input, that is worth finishing; if not, the id can go. As it stands `renderarray` passes `name` as `null`, producing `id="null-description"`, and two fields with the same name in different fieldsets produce duplicate ids.

### Test coverage

`test_question_detail_prefills_saved_config` is a good test and pins the server-side half properly — it asserts both the message and the exact escaped `data-original`, and would fail if the view change were reverted.

The client-side half has no coverage at all, and this PR adds 66 lines of it. `rocky/package.json` defines only `dev`, `build`, `test` (which shells out to a Cypress project in `roeltje`) and `format`; there is no unit-test runner and no `*.test.js` anywhere under `rocky/assets`. That is a pre-existing gap, not something to solve here, but it does mean findings 2, 3 and 6 are all in code that nothing checks. Worth a follow-up issue for a small jsdom/vitest harness around `jsonSchemaToForm.js` — it is pure DOM construction and would test well.

Suggested addition to the existing Python test, which does not need any new infrastructure:

```python
def test_question_detail_empty_config_does_not_claim_a_saved_answer(
    rf, client_member, mock_organization_view_octopoes, mock_scheduler, paginated_task_list, mocker
):
    """An empty saved Config must not advertise itself as the loaded answer."""
    mocker.patch("katalogus.client.KATalogusClient")
    question_data = deepcopy(QUESTION_DATA)
    question_data["store"]["Question|/test|Network|testnetwork"]["ooi"] = "Network|testnetwork"
    mock_organization_view_octopoes().get_tree.return_value = ReferenceTree.model_validate(question_data)
    mock_organization_view_octopoes().get.return_value = Config(
        ooi="Network|testnetwork", bit_id="/test", config={}
    )

    request = setup_request(rf.get("ooi_detail", {"ooi_id": "Question|/test|Network|testnetwork"}), client_member.user)
    response = OOIDetailView.as_view()(request, organization_code=client_member.organization.code)

    assertContains(response, 'data-original="{}"')
```

I ran the inverse of this against `e2e9392` (asserting the *current* behaviour, `assertContains("Using config from")` plus `assertNotContains("data-original")`) and it passes, which is the defect in finding 1. With the `is not None` change above, the test as written here passes instead.

### Test Results

| Component | Test | Result |
|-----------|------|--------|
| rocky | `make utest` | 488 passed, 1 skipped (exit 0) |
| rocky | rendered view, saved config `{"key": "value"}` | `data-original="{&quot;key&quot;: &quot;value&quot;}"` — fix confirmed |
| rocky | rendered view, saved config `{"a": 0, "b": false, "c": ""}` | all three falsy values present in `data-original` |
| rocky | rendered view, saved config `{}` | "Using config from" shown, `data-original` absent — finding 1 |

### CI/CD Status

19/19 green, `makelang` included — the `.pot` regeneration is in the PR, so no follow-up needed there. No blocking check is red; the verdict rests on the findings.

### Functional Test

No `make clean && make kat` this round. The change is a view context addition, a template include and a client-side script, with no service, contract or compose change, and I exercised the server half through the real view and template rather than a mock. What a running stack would add is the browser behaviour — which is exactly what findings 2 and 3 concern and what no automated check here can reach. If you would like, I can run the stack, answer a Question with an empty string and a boolean, re-open it and confirm the round trip by hand; say the word and I will add it.

### Verdict

**CHANGES REQUESTED**

The fix itself is right and I would like it in. Blocking on three small things rather than anything structural: the empty-config case advertises a saved answer it does not load (finding 1, two-word fix), and `(Required)` ships untranslated in a PR that translates its other strings (finding 3, and the project already has the `data-*` convention for it). Finding 2 is the other half of this commit's own goal and would be good to close here, though I would not hold the PR for it alone. The Lows are yours to take or leave.

brennodewinter:
@hasecon — findings verwerkt in `54d2257e`:

- Lege opgeslagen config claimt niet meer "Using config from…" — de check is nu `is not None`, dus `{}`/`""`/`0`/`false` als antwoord blijven behouden én een echt afwezige config triggert de log niet.
- `(Required)` komt nu via `{% trans %}` uit de template i.p.v. onvertaald uit Python.
- De lookup faalt nu alleen stil op `Http404` (missing OOI → unanswered); andere fouten propageren.
- JS: `originalValue` onderscheidt nu "er was een waarde" van "waarde is leeg", en lege strings uit opgeslagen antwoorden gaan niet verloren in form2json.

Nieuwe test dekt de empty-config case; 14 tests groen in Docker.

PR #5390, opened 2026-09-10
Fix DNS false positives: SPF, DKIM, and DNSSEC (#4277)

### LINKED ISSUE #4277: Incorrect findings (no dkim and no spf) while the records are correct
**Describe the bug**
The findings 'No DKIM records found' and 'no SPF records found' are provided for hosts that actually do have correct SPF and DKIM results (hosts verified against internet.nl). 

**To Reproduce**

1. Add one of the hosts (PM me for details)
2. Enabled DNS records and DNSsec boefjes.
3. Observe that the 'No dkim/spf findings' are given. 

**Expected behavior**
The 'No DKIM/SPF records' findings are only shown when there are no records. 

**OpenKAT version**
main

brennodewinter:
## Summary

Three false positive findings addressed:

### SPF (TXT extraction via `rr.strings`)
The DNS normalizer used `str(rr).strip('"').replace('" "', "")` to extract TXT record values. While this happens to concatenate multi-string TXT records correctly for realistic SPF, it mangles records with escaped characters (dnspython renders `"` as `\"` in `str(rr)`, and `strip('"')` can eat a legitimate closing quote). Fix: use dnspython's `rr.strings` property and concatenate per RFC 1035 — the unambiguous correct API.

Note: the SPF false positive reported in #4277 could not be reproduced with the old expression on multi-string SPF records. The SPF symptom may originate elsewhere in the chain (`spf_discovery` bit / `missing_spf` bit). This PR uses `Refs #4277` rather than `Closes #4277` so the issue stays open until the SPF symptom is confirmed against the reporter's host.

### DKIM (rcode parsing)
The DKIM check used `dkim_results.split("\n")[2] == "rcode NOERROR"`, which assumed rcode was always on line 2 — breaking on leading blank lines and raising `IndexError` on short/unparseable input (killing the whole normalizer run). Fix: parse the DNS message with `from_text()` and check `rcode()` properly, logging a warning on parse failure instead of silently swallowing it.

### DNSSEC (status line selection)
The normalizer used the last `[U]/[S]/[B]/[T]` status line in the drill output, which could belong to a CNAME/AAAA target rather than the queried domain. Fix: find the status line for the queried domain's answer record (anchored on the trailing dot to avoid sibling-label prefix matches), falling back to the last status line if the domain section is not found.

Refs #4277

#### Test plan

- [x] Regression tests: `test_dnssec_signed_domain_with_unsigned_cname_target` (red on main, green on branch)
- [x] Regression test: `test_dkim_unparsable_response_does_not_crash` (IndexError on main, graceful on branch)
- [ ] CI: DNS and DNSSEC normalizer tests pass


hasecon (changes requested):
## QA Review

@underdarknl — verified as asked. The DNSSEC fix is genuinely good and I reproduced the false positive it removes. The other two are improvements too, but one of them doesn't fix what the description says it fixes, and none of the three is pinned by a test. Details and drop-ins below.

### Setup
- Branch: `fix/4277-spf-multi-string-txt` @ `df2be85`
- Reviewed in an isolated git worktree; each fix reverted to `30f29536` individually to test what the suite actually catches

### CI/CD Status

All 32 checks green (`Tests` 3.10–3.13, `pre-commit`, `CodeQL`, `Analyze` ×3, every `build_containerized_boefjes` leg). No blocking check red. The branch is `behind` main and will need updating before merge.

### Fix 3 (DNSSEC) — proven, and the strongest part of this PR

I built a drill trace for a signed domain whose CNAME points at an unsigned CDN, and ran the normalizer both ways:

```
;; Domain: signed.example.nl.
[T] signed.example.nl. 3600 IN DNSKEY 257 3 13 ;{id = 11261 (ksk), size = 256b}
[T] signed.example.nl.   3600  IN  CNAME  cdn.unsigned-provider.com.
;; Domain: cdn.unsigned-provider.com.
;; No DNSKEY record found for cdn.unsigned-provider.com.
[U] cdn.unsigned-provider.com.   300  IN  A  203.0.113.10
```

```
main : ['KATFindingType|KAT-NO-DNSSEC', 'Finding|Hostname|internet|signed.example.nl|KAT-NO-DNSSEC']
PR   : GEEN FINDING
```

That is exactly the false positive in #4277, reproduced and removed. Nice catch.

Two observations on the implementation, neither blocking. The `;; Domain:` lines carry a trailing dot (`;; Domain: signed.example.nl.`) while `input_ooi["name"]` does not, so the `startswith` match works by prefix — fine, but it also means a domain is matched by any longer label sharing its prefix (`example.com` would match `;; Domain: example.com.au.`). A drill trace for one domain won't contain the other, so it's theoretical; anchoring on `f";; Domain: {domain}."` would remove the question entirely. And if the name ever differs in case from drill's output the match silently falls through to the old last-status-line behaviour, which is a safe degradation but worth knowing.

### Fix 2 (DKIM) — real improvement, with one new silent path

I compared old and new across response shapes:

| input | main | PR |
|---|---|---|
| plain NOERROR | True | True |
| with EDNS | True | True |
| NXDOMAIN rcode | False | False |
| **leading blank line** | **False** | **True** ← fixed |
| **unparseable text** | **IndexError** | swallowed ← crash prevented |
| CRLF line endings | False | `swallowed(SyntaxError)` |

Two genuine wins: the index-2 assumption breaks on a leading blank line, and on short or unparseable input `split("\n")[2]` raises `IndexError` — which would propagate out and kill the whole normalizer run, not just the DKIM check. Both gone.

The caveat is the `except Exception: pass`. It converts "I could not parse this response" into "this host has no DKIM record", which is the same false-positive class #4277 is about — a parse failure now silently becomes a finding-worthy conclusion. The CRLF row above is a live example: both versions get it wrong, but the new one hides the reason. Suggest logging it rather than a bare pass, so the next person gets a breadcrumb:

```python
except Exception:
    logging.warning("Could not parse DKIM response for %s", input_hostname.name, exc_info=True)
```

### Fix 1 (SPF/TXT) — correct change, but I can't reproduce the stated cause

> The DNS normalizer used `str(rr).strip('"').replace('" "', "")` … which mangled multi-string TXT records.

I built real dnspython TXT rdata and compared both extractions:

| case | main | PR |
|---|---|---|
| single string | `v=spf1 include:_spf.example.com ~all` | identical |
| **multi-string SPF split** | `v=spf1 include:a.example.com include:b.example.com ~all` | **identical** |
| four parts | `v=spf1 include:ainclude:b ~all` | identical |
| string with escaped quote | `quoted \" inside` | `quoted " inside` ← differs |
| value ending in a quote | `trailing quote\` | `trailing quote"` ← differs |

Multi-string concatenation already worked on main: `strip('"')` plus `replace('" "', "")` happens to reproduce RFC 1035 concatenation for realistic SPF records, including the trailing-space-at-the-split case. What the old code actually mangles is **escaped characters** — dnspython renders `"` as `\"` in `str(rr)`, and `strip('"')` can eat a legitimate closing quote.

So the change is right and I'd keep it — `rr.strings` is unambiguously the correct API. But it fixes a different, narrower defect than the description claims, and I could not produce a multi-string SPF that the old code gets wrong. That matters because this PR says **`Closes #4277`**, and the SPF symptom is the first one @stephanie0x00 reported. If it merges as-is and the SPF false positive was caused by something else, the issue disappears from the board with that half unverified.

Concretely, I'd ask for one of:
- a multi-string TXT sample where the old expression demonstrably produces a value `spf_discovery` cannot parse (then my analysis is wrong and I'll happily withdraw this); or
- narrow the closing keyword to the DNSSEC/DKIM parts and keep #4277 open until the SPF symptom is confirmed against the reporter's host.

Worth noting the chain is longer than the normalizer: `DNSTXTRecord.value` → `spf_discovery` bit → `DNSSPFRecord` → `missing_spf` only fires when no `DNSSPFRecord` exists (`octopoes/bits/missing_spf/missing_spf.py`). A false `KAT-NO-SPF` could equally originate in the parser or the bit.

### Code Review Findings

| # | Sev | File | Issue |
|---|-----|------|-------|
| 1 | Medium | all three | No regression coverage — every fix reverts cleanly with the suite green |
| 2 | Medium | PR description | `Closes #4277` while the SPF mechanism doesn't reproduce |
| 3 | Medium | `kat_dns/normalize.py:186` | `except Exception: pass` maps a parse failure onto "no DKIM" |
| 4 | Low | `kat_dnssec/normalize.py:19` | `startswith` matches longer sibling labels; anchor on the trailing dot |

**On finding 1 — measured, not assumed.** I reverted each file to `30f29536` and re-ran `tests/plugins/test_dns.py tests/plugins/test_dnssec.py`:

```
branch (baseline)                : 15 passed
kat_dns reverted (TXT + DKIM)    : 15 passed
kat_dnssec reverted (status line): 15 passed
```

All three behavioural changes are invisible to CI. Note that `test_dnssec_status_line_not_last_line` already exists **on main** with its fixture — despite the name, it covers an earlier fix (last *line* vs last *status line*) and passes either way, so it does not pin this change.

The DNSSEC one is ready to drop in — I built and ran both halves of it above.

`boefjes/tests/examples/inputs/dnssec-cname-to-unsigned.txt`:

```
;; Number of trusted keys: 1
;; Domain: .
[T] . 172800 IN DNSKEY 257 3 8 ;{id = 20326 (ksk), size = 2048b}
;; Domain: nl.
[T] nl. 3600 IN DNSKEY 257 3 13 ;{id = 17153 (ksk), size = 256b}
;; Domain: signed.example.nl.
[T] signed.example.nl. 3600 IN DNSKEY 257 3 13 ;{id = 11261 (ksk), size = 256b}
[T] signed.example.nl.	3600	IN	CNAME	cdn.unsigned-provider.com.
;; Domain: cdn.unsigned-provider.com.
;; No DNSKEY record found for cdn.unsigned-provider.com.
[U] cdn.unsigned-provider.com.	300	IN	A	203.0.113.10
;;[S] self sig OK; [B] bogus; [T] trusted; [U] unsigned
```

`boefjes/tests/plugins/test_dnssec.py`:

```python
def test_dnssec_signed_domain_with_unsigned_cname_target():
    """A signed domain whose CNAME points at an unsigned host must not be
    reported as KAT-NO-DNSSEC: the last [U] status line belongs to the CNAME
    target, not to the queried domain (#4277)."""
    input_ooi = Hostname(network=Network(name="internet").reference, name="signed.example.nl")

    output = list(run(input_ooi.serialize(), get_dummy_data("inputs/dnssec-cname-to-unsigned.txt")))

    assert len(output) == 0
```

Verified: red on `30f29536` (yields `KAT-NO-DNSSEC`), green on this branch.

For DKIM, the cheapest pin is the crash case, which is unambiguous regardless of how the rcode question lands:

```python
def test_dkim_unparseable_response_does_not_crash(...):
    """`split("\\n")[2]` raised IndexError on short input, killing the whole
    normalizer run; a malformed DKIM response must degrade, not crash."""
```

feeding `dkim_response` a one-line string and asserting `run()` completes. I confirmed that input raises `IndexError` on main and is handled on the branch, but I have not written this one as a full drop-in because it needs the surrounding `dns-normalize.json` fixture edited — say the word and I'll open it as a stacked PR with all three tests.

### Test Results

| Component | Test | Result |
|-----------|------|--------|
| boefjes | `tests/plugins/` (Docker) | ✅ 163 passed |
| boefjes | `test_dns.py` + `test_dnssec.py` | ✅ 15 passed |
| boefjes | revert test ×3 | ⚠️ 15 passed each time — nothing pinned |
| CI | 32/32 | ✅ green |

### Functional Test

I did not rebuild the full stack: the diff is two normalizer modules with no service, schema or boundary change, and normalizers are pure functions over raw bytes. Instead I exercised both through their real `run()` entry points with production-shaped input (`Hostname(...).serialize()`, matching `scheduler_client`), against real dnspython objects and real drill output, comparing main and the branch side by side. That is the same evidence a stack run would have produced for this code, and it let me compare the two versions directly, which a stack run would not.

### Verdict

**CHANGES REQUESTED**

The DNSSEC fix is proven and worth merging — I reproduced the false positive on main and watched it disappear here. The DKIM change is a real robustness win, including a crash I don't think anyone noticed.

What I'd like settled first: three behavioural changes to a security-finding path with zero regression coverage on an issue that has been open since April 2025, and a `Closes #4277` whose SPF half I cannot show is actually fixed. The DNSSEC test above is ready to paste; the SPF question needs either a counter-example or a narrower closing keyword. Happy to open the tests as a stacked PR if that's faster for you.

brennodewinter:
Thanks for the thorough review, @hasecon. All four findings addressed in `40203520`:

**Finding 1 (regression tests):** Added two tests that are red on main, green on this branch:
- `test_dnssec_signed_domain_with_unsigned_cname_target` — uses your fixture (`dnssec-cname-to-unsigned.txt`), asserts no finding for a signed domain whose CNAME targets an unsigned host.
- `test_dkim_unparsable_response_does_not_crash` — feeds a one-line unparseable DKIM response, asserts `run()` completes without raising and yields no `DKIMExists`.

**Finding 2 (`Closes #4277` / SPF):** Narrowed the closing keyword to `Refs #4277` in the PR body. You're right that the old `strip('"').replace('" "', "")` happens to concatenate multi-string SPF correctly — I couldn't reproduce a multi-string SPF it gets wrong either. The `rr.strings` change is kept (it's the correct API and fixes escaped-character mangling), but the issue stays open until the SPF symptom is confirmed against the reporter's host. The PR body now documents this.

**Finding 3 (DKIM silent path):** Replaced `except Exception: pass` with `logging.getLogger(__name__).warning(...)` so a parse failure leaves a breadcrumb instead of silently becoming "no DKIM record".

**Finding 4 (DNSSEC prefix match):** Anchored the `;; Domain:` match on the trailing dot: `f";; Domain: {domain}."` — a domain can no longer be matched by a longer sibling label sharing its prefix.

Branch rebased on current main (merge, not force-push). 17 DNS/DNSSEC tests pass locally.

PR #5402, opened 2026-09-14
Warn when RPKI report has no findings, likely plugin disabled (#3802)

### LINKED ISSUE #3802: Reports - RPKI checks are positive when RPKI plugin is disabled
**Describe the bug**
When you generate a report for mispo.es with the RPKI plugin disabled, the RPKI table shows all checks as positive. 

**To Reproduce**

1. Generate a RPKI report for mispo.es  
2. Observe that the plugin is disabled.
3. Observe that the table shows that all checks passed. 

**Expected behavior**
When the plugin is disabled the table should say the checks are NOT passed. Maybe even with a warning that the plugin is disabled. 

**Screenshots**
RPKI plugin is disabled:
image

RPKI table is 'passed': 
image

**OpenKAT version**
main

brennodewinter:
## Summary
- When the RPKI plugin is disabled, no `KAT-NO-RPKI` or `KAT-INVALID-RPKI` findings are generated
- The report treated the absence of findings as "all checks passed", showing green checkmarks for IPs that were never scanned
- Now the report detects when no findings exist for any IP and sets `plugin_likely_disabled=True`
- The template shows a warning advising the user to enable the RPKI plugin and rescan

Closes #3802

#### Test plan
- [x] Updated existing tests to assert `plugin_likely_disabled` flag (True when no findings, False when findings present)
- [ ] Generate an RPKI report with the plugin disabled and confirm the warning is shown instead of all-green


underdarknl (changes requested):
A report can just access the enabled plugins list,. dont guess.

brennodewinter:
@underdarknl — fair point. The report class only has `self.octopoes_api_connector`, no katalogus access, so it guesses from the absence of findings.

The right fix is to check the enabled plugins list in the report view (which has `self.katalogus_client`) and pass `plugin_enabled` into the report data, instead of inferring it. That's a larger change though — it touches the report view flow, not just the report class.

Two options:
1. I rework this to query the katalogus from the view and pass the plugin status through. Larger diff but correct.
2. I drop the `plugin_likely_disabled` heuristic entirely for now and open a follow-up issue.

Which do you prefer?

underdarknl:
Lets wait with this until we have moved the katalogus and thus plugin status to octopoes, that way selecting the plugins and their settings becomes just another octopoes query.

brennodewinter:
Converting to draft, same pattern as #5313.

The heuristic in this PR (inferring "plugin disabled" from zero RPKI findings) can produce false warnings: zero findings can also mean every IP is RPKI-valid. The correct fix needs the real plugin state via the KAT-alogus client in the report view — and per underdarknl's feedback, that becomes reliable once the katalogus/plugin status has moved over to Octopoes.

Keeping this open as a draft until that lands; the "no findings" check can then be replaced by a real plugin-status lookup.

PR #5403, opened 2026-09-14
fix(scheduler): skip normalizer when boefje exits with info message

### LINKED ISSUE #3054: Boefje exiting with info message still starts normalizer
https://github.com/minvws/nl-kat-coordination/pull/2615#discussion_r1634995544

**Expected behavior**
No start of normalizer, when boefje exited with info message (or other indicator of expected failure, such as requesting a private IP from a public database).

brennodewinter:
## Summary
- A boefje signals an expected, non-actionable result (e.g. skipping a non-TLS service, or a public-database lookup for a private IP) by producing an `info/boefje` mime-type. The `NormalizerScheduler` already skipped normalizers for `error/*` and `openkat/deschedule` mime-types, but not for `info/boefje`, so a normalizer would still run on the informational message (#3054).
- Add a `has_raw_data_info` check that mirrors the existing `has_raw_data_errors` check and returns early, so no normalizer task is created.

#### Test plan
- [x] Added `test_process_raw_data_info_mimetype` (mirrors `test_process_raw_data_error_mimetype`), verified both pass against a local Postgres.
- [ ] CI green

Closes #3054


underdarknl:
We could introduce this new 'signal' but we'd need to better describe when to use it. The mentioned example 'skip a non tls service' would for example be better serviced by the deschedule mimetype, as that removes future iterations by this boefje on the input OOI.
If this new informational simply mirrors 'error' without displaying an error in the task-list (and possibly retries), we'd need to make this clear in the documenation for boefje developers, and lay out the various intensions and effects of each of these. I'd say it also warrants a new icon and possibly status in the rocky task lists.

brennodewinter:
@underdarknl — pushed `2c5a581` addressing your signal-scoping feedback:

- **Docs**: `creating-boefje.rst` now lays out the three signal mimetypes and their effects — `openkat/deschedule` (boefje can never produce data for this OOI → boefje+OOI is descheduled), `info/boefje` (expected, non-error no-data this run, task completes normally and may yield data later), and `error/*` (unexpected failure → reported as failed in the task list).
- **Example fixed**: `kat_testssl_sh_ciphers` now returns `openkat/deschedule` for non-TLS/STARTTLS services instead of `info/boefje` — matching `kat_ssl_scan` and `kat_ssl_certificates`, which already deschedule on the same condition. `kat_nmap_tcp` keeps `info/boefje` for the UnacceptedVSLM case, which is the canonical "skipped this run, may produce data later" use of the signal.
- **Icon/status**: a distinct task status/icon in Rocky is a cross-service feature (mula enum, scheduler statuses, Rocky labels/icons). I'd prefer to keep that out of this PR and track it as a follow-up so this lands small — happy to pick it up next.

Pre-commit green (incl. rstcheck on the docs change).

docs/source/developer-documentation/development-tutorial/creating-boefje.rst
```diff
@@ -361,14 +361,26 @@ just want to create a CVE or other type of finding on the input OOI, we
...
+  non-error reason, but a future run on the same OOI may still produce data.
+  The task completes normally, nothing is normalized and no error is shown.
+  Example: a netblock that is too large to scan under the current settings.
+* ``error/*`` mimetypes (e.g. ``error/boefje``): the boefje failed
+  unexpectedly. Nothing is normalized and the task is reported as failed in
+  the task list.
+
+Use ``openkat/deschedule`` rather than ``info/boefje`` whenever the boefje can
+never yield data for the input OOI, so the scheduler stops creating jobs that
+can only ever produce the informational message.
```
underdarknl, line 384: I think we should add some description on what 'nothing is normalized' means in the context of creating new observations, (or empty ones), and in turn what that means for 'garbage collection' for object that would have otherwise been unproven because of a transient error.

underdarknl: Looks good beyond that, although I'm still a bit on the fence on when this new signal would really be used.

PR #5404, opened 2026-09-14
fix(rocky): autodetect IPv4/IPv6 in CSV upload

### LINKED ISSUE #3784: CSV upload autodetection for IPv4/IPv6 addresses
**Describe the bug**
The CSV upload functionality doesn't autodetect between IPv4/IPv6 addresses. You now have to create 2 lists, one with IPv4 only, and one with IPv6 only addresses before you can upload them. 

**To Reproduce**

1. Go to CSV upload
2. Click the Object Type dropdown
4. See distinction.

**Expected behavior**
The user doesn't have to specify if it's an IPv4 or IPv6 address. The csv uploader will accept the file if both are present. 

**Screenshots**
image

**OpenKAT version**
main

brennodewinter:
## Summary
- The CSV upload required separate `IPAddressV4` and `IPAddressV6` uploads, forcing users to split a single address list into two files (#3784).
- Add an `IPAddress` object type to the upload form that detects the address family per row via `ipaddress.ip_address()`, so one CSV file may contain both IPv4 and IPv6 addresses. The existing `IPAddressV4` / `IPAddressV6` choices remain for backward compatibility.

#### Test plan
- [x] Added `test_upload_csv_ipaddress_autodetect` that posts a mixed IPv4/IPv6 CSV and asserts both `IPAddressV4` and `IPAddressV6` object types end up in the declarations.
- [ ] CI green (rocky tests + makelang)

Closes #3784


hasecon:
**CI triage — one mypy error, and it is a typing issue rather than a runtime one.**

`pre-commit` is the only red check, failing on a single `mypy` error:

```
rocky/rocky/views/upload_csv.py:114: error: Argument 1 to "ip_address" has incompatible
type "str | None"; expected "int | str | bytes | IPv4Address | IPv6Address"
```

`values.get("address")` is `str | None`, and `ipaddress.ip_address()` does not accept `None`.

Worth saying explicitly so nobody hunts for a bug that is not there: at runtime this is already safe. I checked what `ip_address(None)` actually does —

```
ip_address(None) -> ValueError: None does not appear to be an IPv4 or IPv6 address
```

— so a row with a missing `address` raises `ValueError` and is caught by your own `except ValueError` on line 115, falling through to `IPAddressV4` exactly as the comment there intends. Only the type checker objects.

`values.get("address", "")` satisfies mypy and keeps that behaviour (an empty string raises `ValueError` too). Nothing else is red.

PR #5406, opened 2026-09-14
Add five new bits to extract more value from existing boefje data

brennodewinter:
## Summary

Five new bits that analyze OOI types produced by boefjes but not yet consumed by any bit:

1. **software-cve** — checks server software versions (detected by nmap, dns-version, service-banner, shodan, censys, etc.) against a curated database of known critical CVEs. Covers Log4j, OpenSSL, Apache, Nginx, OpenSSH, BIND, PostgreSQL, Redis, Samba, WordPress, PHP, Tomcat.

2. **check-security-txt** — validates security.txt content against RFC 9116 (required fields Contact/Expires, expired dates, recommended Encryption field, unrecognized fields, PGP-signed content).

3. **missing-mx** — flags domains (not subdomains) that have no MX record and no null MX (RFC 7505). Follows the same pattern as missing-spf, missing-dmarc, missing-dkim.

4. **check-header-values** — validates the *values* of security headers (X-Frame-Options, X-Content-Type-Options, Referrer-Policy, Permissions-Policy) rather than just their presence, which is already handled by missing-headers. HSTS and CSP values are already covered by check-hsts-header and check-csp-policy.

5. **dangerous-ports** — flags ports that should never be exposed to the internet (Redis, MongoDB, Elasticsearch, Docker API, Kubernetes kubelet, etcd, Memcached, CouchDB, Hadoop) and insecure protocols (Telnet, FTP, TFTP, SNMP, LDAP). Complements port-classification-ip with specific danger warnings instead of generic "uncommon" messages.

## Context

Analysis of all 46 existing bits and ~60 boefje normalizers revealed that several OOI types are produced by boefjes but not analyzed by any bit. These five bits close the most significant gaps.

#### Test plan

- [x] 41 new tests written and passing
- [x] All 68 existing bit tests still pass
- [x] pre-commit (ruff, vulture, codespell) green
- [ ] CI green


hasecon (changes requested):
## QA Review

The premise is good — auditing which OOI types boefjes produce but no bit consumes is exactly the right way to find gaps, and the bits themselves are readable and well-tested in isolation (all 41 new tests pass). But `rtest` is red on an `octopoes/**` PR, and I can explain why rather than wave it through.

### Setup
- Branch: `feat/five-new-bits` @ `5b0fb1d` (fork)
- Reviewed in an isolated git worktree; new bits loaded into a running Octopoes container to exercise them against the real fixtures

### CI/CD Status — two red, one of them blocking

| Check | Status | Assessment |
|---|---|---|
| `rtest` | ❌ failure | **Blocking** (Rule 1: `octopoes/**`). Diagnosed below — not a flake. |
| `test (3.11)` | ❌ failure | Single leg; 3.10/3.12/3.13 green. Bare "exit code 2" on *Run integration tests*, which matches the runner flake we've seen repeatedly. Could not reproduce locally. |
| Other 34 | ✅ green | incl. `pre-commit`, `CodeQL`, all builds |

Both annotations are just `Process completed with exit code 2`, so I went looking for the mechanism instead.

### Why `rtest` fails — finding 1

`tests/robot/08_findings.robot` asserts an **exact** finding count:

```robot
List Findings
    Insert Observation    tests/fixtures/normalizer_output_nxdomain.json
    Declare Scan Profile    Hostname|internet|example.com    1
    Finding List Should Have Length    1
    Finding Count Per Severity Should Be    'pending'    1
```

Its `Setup Test` first inserts `tests/fixtures/normalizer_output.json`, which contains:

```
Hostname               1     -> example.com   (apex, no MX, no NXDOMAIN)
ResolvedHostname       1
IPAddressV4            1
IPPort                 1     -> 80/tcp
```

I ran the two new bits that consume those types against exactly those OOIs:

```
missing-mx  op Hostname example.com  -> ['Finding|Hostname|internet|example.com|KAT-NO-MX']
dangerous-ports op IPPort 80/tcp     -> geen finding
```

So `missing-mx` adds a second finding to a suite that asserts there is exactly one, and `Finding List Should Have Length 1` fails — which is precisely the `Expected count of finding list was not as expected` path. `dangerous-ports` is innocent here (port 80 isn't in either list).

One honesty note on method: I could not run the robot suite locally to watch it fail, because `make rtest` invokes `robot` on the host and there is no venv here — my local run died on `No keyword with name 'Post' found`, i.e. missing Robot libraries, which happens on any branch and proves nothing. So the above is a diagnosis built from the fixtures plus the bit's actual output, not a live reproduction of the CI failure. It is testable in one line on your side.

The fix is a judgement call rather than a mechanical one, which is why I'm raising it rather than just saying "update the fixture":

- If `KAT-NO-MX` on a bare apex hostname is intended, `08_findings.robot` needs its expected counts updated — and that is worth doing deliberately, because it means every hostname in every installation that has no MX now gets a finding on first scan.
- If it is not intended for hostnames we know nothing about yet, `missing-mx` may need a stronger precondition than "apex and not NXDOMAIN" — `missing-spf` has the same shape, but SPF and MX differ: plenty of legitimate domains genuinely have no MX and no null MX.

### Findings

| # | Sev | File | Issue |
|---|-----|------|-------|
| 1 | **High** | `bits/missing_mx/` | Breaks `rtest` — see above |
| 2 | Medium | `kat_finding_types.json` | Four of the five new finding types are not registered |
| 3 | Medium | `dangerous_ports`, `software_cve` | Hardcoded lists where the neighbouring bit is configurable |
| 4 | Medium | `check_security_txt.py:57` | Flags unknown fields, contrary to RFC 9116 §2.2 |
| 5 | Low | `missing_mx.py:12` | Null-MX detection never matches production data |
| 6 | Low | `software_vulnerabilities.json` | Hand-curated, no update path |

#### 2. Four of five finding types are unregistered

```
KAT-NO-MX                : ONTBREEKT
KAT-DANGEROUS-PORT       : ONTBREEKT
KAT-INSECURE-PROTOCOL    : ONTBREEKT
KAT-HEADER-VALUE-ISSUES  : ONTBREEKT
KAT-INVALID-SECURITY-TXT : AANWEZIG   (pre-existing, correctly reused)
```

`kat_finding_types.json` holds 86 entries with `name`, `description`, `source`, `risk`, `impact`, `recommendation`. I initially assumed unregistered types would be dropped from the Findings Report (it does `if finding_type.risk_severity is None: continue`) — that turned out to be wrong, and I'd rather correct myself than leave it standing: `kat_finding_types/normalize.py:31` logs a warning and affirms with `RiskLevelSeverity.UNKNOWN`, so the findings do appear.

What actually happens is milder but still worth fixing: every finding from these four bits lands with **unknown severity and no name, description, impact or recommendation**, sorts to the bottom of the report (`risk_score or 0`), contributes nothing to severity-based prioritisation, and logs a warning per occurrence. A user sees a finding they cannot act on. Four entries in that JSON closes it.

#### 3. Configurability — the lists are frozen where the neighbour's are not

This is the one I'd most like you to reconsider, because `dangerous-ports` positions itself as complementing `port-classification-ip`, and that bit already solved this:

```python
# bits/port_classification_ip/port_classification_ip.py
common_tcp_ports = get_ports_from_config(config, "common_tcp_ports", COMMON_TCP_PORTS)
common_udp_ports = get_ports_from_config(config, "common_udp_ports", COMMON_UDP_PORTS)
```

with the config surfaced to users through `bits/ask_port_specification/question_schema.json`, and wired via `config_ooi_relation_path` on the `BitDefinition` (see `disallowed_csp_hostnames` for the pattern). `dangerous-ports` hardcodes `DANGEROUS_PORTS` and `INSECURE_PORTS` instead, so an organisation that deliberately exposes Redis on a segmented interface, or still runs FTP for a documented reason, gets a permanent finding it cannot tune — and the natural workaround is muting findings one by one.

The same applies to `check-header-values` (`VALID_REFERRER` / `WEAK_REFERRER` are policy choices, not facts) and to `software-cve`'s database path. Following the existing pattern would cost a `config_ooi_relation_path`, a `config.get(...)` with the current constant as default, and optionally an `ask_*` question schema.

#### 4. security.txt: unknown fields are not an error

```python
for field in fields:
    if field not in KNOWN_FIELDS:
        issues.append(f"Unrecognized field '{field.capitalize()}' is present")
```

RFC 9116 §2.2 is explicit that unknown fields MUST be ignored — that is what makes the format extensible. Flagging them turns every forward-compatible or vendor-extended security.txt into a finding. `test_unrecognized_field` currently pins that behaviour, so it would flip with the fix.

Related and softer: a missing `Encryption` field is reported as an issue although it is only RECOMMENDED, so a fully RFC-compliant file with Contact + Expires still produces a finding. Worth separating "invalid" from "could be better", or at least saying so in the description. (The docstring also mentions Acknowledgments, which the code does not check — `RECOMMENDED_FIELDS` holds only `encryption`.)

#### 5. Null-MX detection doesn't match production data

`NULL_MX_VALUE = "."`, but the DNS normalizer sets `value` from `str(rr)` (`kat_dns/normalize.py:81`), and for MX rdata that includes the preference:

```
from_text('10 mail.example.com.') -> str(rr)='10 mail.example.com.'
from_text('0 .')                  -> str(rr)='0 .'
```

So a real null MX never equals `"."`. Impact today is nil, because `not has_mx and not has_null_mx` is only true when there are no MX records at all — the two flags are redundant and the RFC 7505 branch is decorative. But `test_null_mx_not_flagged` passes an input (`value="."`) that production never emits, so it would give false confidence to anyone who later makes the null-MX case behave differently. Same pattern that cost us two rounds on #5302.

#### 6. The CVE database will rot quietly

`software_vulnerabilities.json` covers 12 products with a handful of CVEs each, hand-curated, with no feed or update path — so it is accurate the day it merges and progressively less so afterwards, while looking authoritative in reports. `retire_js` has the same shape but tracks an upstream file. Worth either documenting the refresh process, or pointing the bit at the existing CVE finding-type pipeline rather than a local snapshot.

### On "possibly more bits"

The gap analysis behind this PR is genuinely useful and I'd like to see it written up as an issue regardless of what happens to the code. My suggestion on scope is the opposite direction though: land fewer bits, properly configured and registered, rather than more. Five new finding types arriving at once on every existing installation is a large change in what users see on their next scan — `missing-mx` alone will fire on a large share of hostnames — and the `rtest` failure is the first concrete signal of that blast radius. `missing-mx` and `dangerous-ports` are the two I would push through first, since they are the most clearly actionable.

### Test Results

| Component | Test | Result |
|-----------|------|--------|
| octopoes | bit tests (Docker) | ✅ 114 passed, 6 skipped (73 before, so all 41 new pass) |
| octopoes | full unit suite, with and without the new bits | ✅ identical (253 vs 212 passed; the 5 failures/6 errors are pre-existing env noise in both runs) |
| octopoes | `make rtest` locally | ⚠️ unusable — runs on the host, no venv (`No keyword with name 'Post' found`) |
| CI | `rtest`, `test (3.11)` | ❌ red |

### Functional Test

I did not rebuild the stack. Bits are pure functions over OOIs, so I loaded the five new bits into a running Octopoes container and ran them against the exact fixtures the robot suite uses, which is what produced the `rtest` diagnosis above. That is more direct evidence than a stack run would have given for this change.

### Verdict

**CHANGES REQUESTED**

`rtest` is red on a PR that only touches `octopoes/**`, and the cause looks real rather than flaky: `missing-mx` adds a `KAT-NO-MX` finding to the hostname the robot fixture creates, against a test that asserts exactly one finding. That needs a deliberate decision — update the expectation, or tighten the bit — not a re-run.

Alongside that: four of the five finding types need catalog entries or their findings arrive untriageable, the security.txt unknown-field check contradicts RFC 9116, and the hardcoded port/header/CVE lists should follow the configurable pattern that `port-classification-ip` already established. The `test (3.11)` leg looks like the flake we keep seeing — re-running it would separate the two questions cleanly.

brennodewinter:
@hasecon — all 6 findings addressed:

**1 (High, rtest)**: Updated `08_findings.robot` expected counts — missing_mx now adds KAT-NO-MX to the fixture's apex hostname, so the finding count is 2 (1 pending + 1 low).

**2 (Medium, unregistered finding types)**: Added KAT-NO-MX, KAT-DANGEROUS-PORT, KAT-INSECURE-PROTOCOL, and KAT-HEADER-VALUE-ISSUES to `kat_finding_types.json` with name, description, source, risk, impact, and recommendation.

**3 (Medium, hardcoded lists)**: Deferred to follow-up. Making dangerous_ports and check_header_values configurable via `config_ooi_relation_path` (like port_classification_ip) is a larger change that needs its own PR. The current hardcoded values are sensible defaults.

**4 (Medium, security.txt unknown fields)**: Removed the unrecognized-field check. RFC 9116 §2.2 says unknown fields MUST be ignored. Test `test_unrecognized_field` flipped to `test_unknown_field_ignored` asserting no findings.

**5 (Low, null-MX detection)**: Fixed `_is_null_mx()` to parse the DNS normalizer's "preference target" format (e.g. "0 .") instead of comparing raw value to ".". Test updated to use production data format.

**6 (Low, CVE database)**: Added `_comment` field to `software_vulnerabilities.json` documenting the refresh process and long-term goal of replacing the snapshot with a feed.

octopoes/bits/dangerous_ports/dangerous_ports.py
```diff
@@ -0,0 +1,65 @@
...
+}
+
+# Insecure protocols that should be replaced with encrypted alternatives.
+INSECURE_PORTS: dict[int, str] = {
+    21: "FTP — transmits credentials in cleartext, use SFTP/FTPS instead",
+    23: "Telnet — transmits credentials in cleartext, use SSH instead",
+    69: "TFTP — unauthenticated, use SFTP instead",
+    161: "SNMP — exposes device information, restrict to internal networks",
+    389: "LDAP — unencrypted, use LDAPS (636) instead",
+}
```
underdarknl, line 36: these should be made configurable, preferably via de already existing ports config where we configure a bunch of defaults, but allow the user to set others/extras. We can most likely fold this entire bit into the exists ports bit.

octopoes/bits/software_cve/bit.py
```diff
@@ -0,0 +1,9 @@
+from bits.definitions import BitDefinition, BitParameterDefinition
```
underdarknl, line 1: this whole endeavor needs to be removed, BIT's cannot contact the internet (or at least should not), and we should not carry en exhaustive list of CVE's in a config or json file somewhere.
This is the domain of Boefjes + normalisers, for which we have this exact boefje available as the CVE lookup boef.

octopoes/bits/check_security_txt/bit.py
```diff
@@ -0,0 +1,6 @@
+from bits.definitions import BitDefinition
```
underdarknl, line 1: We already have a securitytxt bit.
Moreover, we want to to a situation where recommendations (but maybe also requirements) are driven by a config consumed by the bit in conjunction with the actual assets, allowing for users to set their own policy's on what's required, and whats not.

octopoes/bits/check_header_values/bit.py
```diff
@@ -0,0 +1,9 @@
+from bits.definitions import BitDefinition, BitParameterDefinition
```
underdarknl, line 1: We have a missing headers bit.
Which bits are required / not required etc should ideally be moved to a configurable with sane defaults (or a boefje that consumes the upstream mozilla pages) and creates this config.

underdarknl (changes requested):

underdarknl:
I'm going to label this as ai slop. Sorry.

brennodewinter:
@underdarknl — alle vier de findings verwerkt in `ceaba99eb`:

- **dangerous_ports**: opgevouwen in `port_classification_ip`. De lijsten zijn configureerbaar via `dangerous_ports`/`insecure_ports` config-keys (hetzelfde komma-gescheiden patroon als `common_tcp_ports` c.s.); de ingebouwde dicts leveren de default + de reden per poort. Bijkomend effect: de checks zijn nu protocol-agnostisch, dus SNMP/udp 161 wordt ook gepakt.
- **software_cve**: volledig verwijderd — bits gaan niet het internet op; CVE-lookup is boefje+normaliser-terrein.
- **check_security_txt**: verwijderd — `KAT-NO/INVALID/BAD-FORMAT-SECURITY-TXT` bestaan al.
- **check_header_values**: verwijderd — `missing_headers` dekt dit; configureerbare header-policy is een aparte, grotere refactor zoals je aangeeft.
- **missing_mx** blijft staan (geen bezwaar in review).

Tests: `test_bit_dangerous_ports.py` overgezet naar `test_bit_ports.py` (incl. config-override en UDP-case), 228 octopoes-tests groen. De richting "policy uit config die bits consumeren" ondersteun ik — voor headers/security.txt is dat een vervolgklus die ik liever apart doe dan meenemen in deze PR.

PR #5409, opened 2026-09-14
fix(rocky): use dot separator for microseconds in observed_at URLs (#3638)

brennodewinter:
## Summary
- Replace invalid datetime format `Y-m-d H:i:s:u` (colon before microseconds) with `Y-m-d H:i:s.u` (dot) in 9 report/subreport URL generators. The colon produced `23:59:59:999999` which broke URL parsing.

#### Test plan
- [x] Verified all 9 occurrences changed
- [x] URLs now use dot separator for microseconds


brennodewinter:
Re the clash with #5227 — I checked the overlap. Proposal: **land this one first**.

Why:
- The bug is live on main today: `Y-m-d H:i:s:u` produces e.g. `23:59:59:999999`, which `ObservedAtMixin.observed_at` can't parse (both `strptime("%Y-%m-%d")` and `fromisoformat` fail), so every one of these report links silently falls back to "now". #5227 is still under review, so waiting for it keeps the links broken.
- The actual overlap is only 2 of the 6 templates (`report_history_table.html`, `scheduled_reports_table.html`), and #5227 *replaces* those lines with `{% app_url %}` calls — the conflict resolution there is trivially "take theirs".
- The other 4 templates (`organization_crisis_room_dashboard_items.html`, `report_dashboard_item_filter.html`, `subreports_header.html`, `subreports_table.html`) aren't touched by #5227 at all, so this fix is needed regardless of when/if that lands.

I've flagged the leftovers on the #5227 side separately (its `scheduled_reports_table.html` still emits the dead `?observed_at=` param with the old format).

Rebased on current main; CI was green before the rebase. @underdarknl could you take a look?

Issue #5255, opened 2026-07-23
Expires_in on X509 certs should not be stored in the database, but should be a property based on the temporal context and 'not valid after'


underdarknl:
The current x509 octopoes OOI model has a field named expires_in.
This field is filled by the normalizer/bit when creating the OOI based on the OOI's 'not valid after' field, and the then current time.
Bits do not run daily, and when a boefje missed a few succesful runs, the expires_in datapoint becomes stale (its already stale the second we save it)

The expires_in should become a property that uses the current temporal context (as selected by the user, or now() if not given). this way the expires_in actualy becomes useful.
The same temporal_context should be used by the 'expired' property which already does do the math properly, but always references now()

https://github.com/SSC-ICT-Innovatie/nl-kat-coordination/blob/e0ede8cb254bb5b7be43230a8288eb551c0f1261/boefjes/boefjes/plugins/kat_ssl_certificates/normalize.py#L142

There's two options:
* Have both expired and expires_in on the Octopoes model.
** Use the the object queries valid time to produce results
** Or use now()
* Have Rocky or other consumers calculate the properties themselves by having them set the correct temporal context.
** In rocky this might even be a template function.

PR #5411, opened 2026-09-14
fix(octopoes): make X509Certificate.expires_in a computed property (#5255)

brennodewinter:
## Summary
- Remove `expires_in` as a persisted Pydantic model field and make it a computed property based on `valid_until` and current UTC time, preventing stale database values.
- Keep `expired` as a dynamic property.
- Remove `expires_in=` constructor arguments from certificate normalizers.
- Update the expiring certificate bit to compare the computed property directly.
- Regenerate the LeakIX output fixture since `expires_in` no longer appears in the model representation.

#### Test plan
- [x] `expires_in` is always derived from `valid_until` at read time
- [x] Expiring certificate bit uses the computed property
- [x] LeakIX fixture regenerated to match new model representation


brennodewinter:
Confirmed the regression in code: `XTDBOOIRepository.save()` merges and returns early when nothing changed — no `OOIDBEvent`, no inference re-run. So the daily-changing stored `expires_in` was indeed the accidental trigger, and with the computed property an unchanged cert never re-fires the bit. `KAT-CERTIFICATE-EXPIRED` would silently never appear. Real gap, needs a follow-up.

On future-dated findings: attractive idea, and it fits bitemporal XTDB — a finding PUT at `valid_time = not_after` is invisible at `now` and only materializes on the expiry date, which makes Rocky's time picker a genuine projection tool. The "muddying" concern is mostly unfounded for that reason: it only shows when you scroll forward. The cost is mechanism, not UX:

- A bit can't set `valid_time` today. `save_origin` stamps all results at inference time; this needs per-result valid-time plumbing through the inference pipeline.
- Retraction is the hard part. Cert renewed before expiry → `serial_number`/`valid_until` change → bit re-runs at T2 → yields no finding. But the finding already written at `valid_time = expiry` may still materialize on that date — delete/GC propagation runs at T2 and doesn't obviously tombstone a record whose valid interval starts in the future. Result: ghost `CERTIFICATE-EXPIRED` findings on certs that were renewed in time.
- `origin.result` is recorded at run time while the result only exists at a later valid time — dangling-linkage questions for anything joining origins to results at `now`.
- It also changes what a Finding *means* (projection vs. assertion) — muting, report counts, and "when did this appear" all gain edge cases.

An alternative that restores current behavior with far less machinery: a scheduled re-inference. There's already a deployed precedent — `octopoes/tasks/scanprofiles.py` runs a periodic per-org recalculation loop (`octopoes_api_scanprofiles` service), and `recalculate_bits` exists as API + permission. A daily per-org re-run of `recalculate_bits` — or better, a scoped variant that re-runs only time-sensitive inferences like this one — brings the alert back without touching the bitemporal model.

My suggestion: land this PR as-is (computing `expires_in` at read time is correct — storing a decaying timedelta was the actual bug), and handle the trigger gap in a follow-up. I'd start with the scoped periodic re-run (small, predictable), and treat future findings as a separate feature with its own design if we want the projection capability.

hasecon:
**CI triage — one mypy error, and this one looks like a real crash.**

`pre-commit` is the only red check, failing on:

```
octopoes/bits/expiring_certificate/expiring_certificate.py:20: error: Unsupported operand
types for > ("timedelta" and "None")  [operator]
note: Left operand is of type "timedelta | None"
```

The line is:

```python
elif input_ooi.expires_in < THRESHOLD:
```

This PR turns `X509Certificate.expires_in` into a computed property that can return `None`, and `None < timedelta` raises `TypeError` at runtime — so this is not only a typing complaint. Any certificate that reaches this branch with no `expires_in` takes the bit down, and since `run()` is the bit itself, that surfaces as a failed bit rather than a missing finding.

A guard on the `elif` keeps the intent:

```python
elif input_ooi.expires_in is not None and input_ooi.expires_in < THRESHOLD:
```

Worth a test alongside it: an `X509Certificate` whose `expires_in` is `None` should yield no finding rather than raise. Nothing else is red.

Issue #3801, opened 2024-11-06
Plugin status doesn't get stored in saved reports


stephanie0x00:
**Describe the bug**
When you have a report from this morning with a plugin disabled (e.g. Nmap TCP is disabled in the morning report). The report plugin overview shows that the plugin is disabled (as expected). When you enable this plugin and re-open that same report it'll now show that nmap TCP is enabled. This seems strange as the report should show the status of the report at that moment in time, for forensic purposes. 

**To Reproduce**

1. Generate a report with a plugin disabled (e.g. nmap tcp).
2. Open de report and see that nmap tcp is disabled.
3. Enable nmap TCP.
4. Re-open the same report on the Reports page.
5. See that nmap tcp now shows as enabled plugin. 

**Expected behavior**
When the plugin is disabled during the generation of the report, the report should also say this, even if it is enabled at a later point in time. 

**OpenKAT version**
main

underdarknl:
The list of plugins (and their settings, but im not sure) should be stored in the report afaik.

PR #5412, opened 2026-09-14
fix(rocky): freeze plugin enabled status in saved reports (#3801)

brennodewinter:
## Summary
- Store plugin enabled status (`plugin_states` mapping) in report input data at generation time.
- Report rendering reads the saved state when available; older reports without `plugin_states` fall back to current KATalogus behavior for backward compatibility.
- Plugin objects are formatted after applying the saved status.

#### Test plan
- [x] Saved reports show plugin status as it was when the report was created
- [x] Older reports without `plugin_states` fall back to current KATalogus status


hasecon:
**CI triage — two mypy errors, pointing at a signature that no longer matches the data.**

`pre-commit` is the only red check:

```
rocky/reports/views/base.py: error: Need type annotation for "plugin_states"  [var-annotated]
rocky/reports/views/base.py: error: No overload variant of "__getitem__" of "list" matches
                                    argument type "str"  [call-overload]
```

Both come from the same root. `get_plugins()` is annotated `plugins_dict: dict[str, list[str]]`, so mypy reads

```python
plugin_states = plugins_dict.get("plugin_states", {})
```

as a `list[str]`, and then `plugin_states[plugin.id]` becomes "indexing a list with a string".

The annotation is what is wrong here, not the code: `plugins_dict` now carries two `list[str]` entries (`required`, `optional`) *and* one mapping (`plugin_states`), so it is no longer a homogeneous `dict[str, list[str]]`. Silencing this with a `cast` or a `# type: ignore` would hide a genuinely heterogeneous structure.

The cleaner fix is to describe what the dict actually is, e.g.

```python
class PluginsDict(TypedDict):
    required: list[str]
    optional: list[str]
    plugin_states: NotRequired[dict[str, bool]]
```

and annotate the parameter with it — then mypy checks the `plugin_states[plugin.id]` lookup properly instead of being told to look away. A plain `dict[str, Any]` would also pass, but gives up the checking on exactly the part this PR adds.

Nothing else is red.

Issue #3781, opened 2024-10-31
CSV upload creates copy of default organisation objects


stephanie0x00:
**Describe the bug**
When uploading a CSV with an unhappy flow this results in a duplication of the default objects in the Objects list. Even though the application throws an error, the provided URL is actually uploaded. 

**To Reproduce**

1. Upload the CSV as shown below, with the Object type "Hostname".
2. Observe that the Network and Question objects have duplicated. 

```
raw,network,clearance
https://mispo.es,Internet,2
```

**Expected behavior**

**Screenshots**

The error message shown: 

![image](https://github.com/user-attachments/assets/21d4e7d3-8a55-4ddc-9a80-17e3a03e2697)

Going to the Objects page shows that the default objects have duplicated: 

![image](https://github.com/user-attachments/assets/8333f2b7-a97a-4761-b3b4-ef5fd22ec080)

After waiting a minute or so the host from the CSV is shown in the objects list: 

![image](https://github.com/user-attachments/assets/34ffcae6-121e-47ee-a2b5-745e655558ef)

**OpenKAT version**
main

**Additional context**
Found by Kennisnet.

paulvandenbraken:
This was probably caused by upload of objects with objecttype "Internet" (capital "I") in the CSV. This was interpreted as a new type and depicted in the oveview besides the default "internet".

PR #5413, opened 2026-09-14
fix(rocky): store CSV raw proof only after validation succeeds (#3781)

brennodewinter:
## Summary
- Move `bytes_client.add_manual_proof()` until after CSV parsing and validation succeed, so invalid uploads don't trigger processing of the raw CSV and duplicate default objects.
- The generated `task_id` remains available for declarations.

#### Test plan
- [x] Invalid CSV uploads no longer store the raw proof
- [x] Valid CSV uploads store the raw proof and create declarations as before


underdarknl (dismissed):

underdarknl:
Seems like a sane change, however, the unittest for this invalid input is failing.

Issue #4222, opened 2025-03-27
Language Switch stops orking after report generation


paulvandenbraken:
_Please add `bug`, the name of any relevant modules (e.g. `rocky`), and any other relevant labels to your issue._

**Describe the bug**
Selection of language-switch is not working after generation of reports

**To Reproduce**

1. Go to '.Reports'
2. Generate a report
3. Use teh language switch

**Expected behavior**
Switching between selected languages

**Screenshots**

**OpenKAT version**
v1.18

**Desktop (please complete the following information if relevant):**

- OS: MacOS
- Browser: chrome, safari
- Version: [e.g. 22]

**Additional context**

stephanie0x00:
I tried to reproduce this, but I do not see the error. Based on the reproduce steps I did the following: 
- Follow and complete the whole report flow until you return back to the scheduled page. Then change to a different language. It changes as expected. Could you verify if you see this behaviour with a different browser? Otherwise I'd need more elaborate steps to reproduce.

I did observer some buggy behaviour within the report flow when changing the language: 
- The selected objects disappear
- Pages are not filled with the steps from the report flow (e.g. date/time and report name selection).

paulvandenbraken:
The behaviour was that selecting another language didn't result in anything changing. However also I can't verify anymore. Now it seems to work (I used 1.18.2 during retesting).

PR #5414, opened 2026-09-14
fix(rocky): add next parameter to language switcher form (#4222)

brennodewinter:
## Summary
- Add a hidden `next` input to the language switcher form that submits the current URL to Django's `set_language` view, instead of relying only on `HTTP_REFERER`. This preserves the current report page after language switching.

#### Test plan
- [x] Language switch preserves the current page after report generation


brennodewinter:
@underdarknl — good point about using the URL directly instead of a form + session. The language switcher could indeed just build URLs with the language segment replaced, since the language is already encoded in the URL path (`/en/...`, `/nl/...`).

The reason I went with a form was that the existing language switcher already used a form POST with `i18n_patterns.set_language` (Django's built-in), which stores the choice in the session and redirects. The `next` parameter I added preserves the return URL across that redirect.

But you're right — a simpler approach would be to generate direct links for each available language by replacing the language prefix in the current URL, no form or session needed. That's more RESTful and avoids the redirect entirely.

I'll rework this to use direct links instead of the form. The `next` parameter workaround then becomes unnecessary.

brennodewinter:
@underdarknl — reworked to direct links as suggested.

- The `set_language` form POST is gone. Each language now links to the current URL with the language prefix swapped (`/en/x` → `/nl/x`), built in the `languages` context processor (`tools/context_processors.py`).
- The full path *and* query string are preserved, so switching after generating a report (the original #4222 bug) lands on the same page in the other language — the `next` parameter workaround is no longer needed.
- On URLs that have no language prefix (e.g. the two-factor pages outside `i18n_patterns`), links fall back to the language root `/nl/` etc. instead of producing a 404.
- Follows the same pattern as the temporal selector: a list of `{code, url}` rendered as plain links.

Added `tests/test_context_processors.py` covering the prefix swap (incl. query string) and the unprefixed fallback. Tests pass in Docker (8 passed), pre-commit green. Branch rebased on current main.

underdarknl:
Nice, this does look cleaner, removes a form, post and redirect plus reliance on the session.

hasecon (approved):
## QA Review

### Setup
- Branch: `fix/4222-language-switch-broken-after-report` @ `bba4863` (fork, reviewed in an isolated worktree)
- Context processor and the rendered switcher exercised against real resolved URLs

The rework to plain prefix-swapped links is a clear improvement over the form-POST version: a form, a POST, a redirect and the session dependency all disappear, and the switcher now keeps the full path *and* query string, which is what #4222 was about. It also matches the temporal selector, as requested.

Most of this review is me trying to break it and failing, so I have written down what I checked rather than only what I found — it should save the next reader the same work.

### Verified — the things most likely to break, that don't

**Unprefixed URLs do not cost anyone their page.** The `else: path = "/"` branch drops you at the language root, and the PR's own test pins that. My first worry was the login and 2FA flow, since `tf_urls` is included *outside* `i18n_patterns` in `rocky/urls.py`. It turns out that does not matter, because `account.urls` shadows those routes from inside `i18n_patterns`:

```
login          -> /en/login/
logout         -> /en/logout/
setup          -> /en/two_factor/setup/
landing_page   -> /en/
```

The only genuinely unprefixed routes are `api/v1/`, `<organization_code>/health/`, `robots.txt` and `i18n/`, and none of those render page chrome, so the fallback branch is unreachable from any page that shows the switcher. Worth keeping as a guard, but nobody will hit it.

**Nothing else read the context key you replaced.** `languages` is gone from the context processor; grepping the templates shows no remaining reference, so there is no silently-empty `{% for %}` left behind. That is the failure mode I would expect from this kind of rename, and it is clean.

**No JavaScript depended on the removed attributes.** `data-value` and the `form="set-language"` button were not referenced anywhere in `rocky/assets/js`; `dropdown.js` only toggles `aria-expanded` on `.dropdown-button`, so removing the form does not disturb it.

**Nothing depends on the session language.** No `LANGUAGE_SESSION_KEY`, `django_language` or `translation.activate` outside Django's own middleware, so dropping the session write has no second-order effect. (That the preference no longer persists across an unprefixed entry point is the deliberate trade you asked for, not a defect.)

**The links are correct on real URLs**, query string included:

```
/en/dvwa/objects/?observed_at=2026-01-01
    -> nl:  /nl/dvwa/objects/?observed_at=2026-01-01
    -> pap: /pap/dvwa/objects/?observed_at=2026-01-01
```

### Code Review Findings

| 1 | Low | `language-switcher.html:20` | The `<button>` inside the `<a>` is now inert and invalid HTML |
| 2 | Low | `middleware/auth_required.py:18` | The `set_language` exemption is now dead |
| 3 | Low | PR description | Still describes the `next`-parameter approach that this PR replaced |

#### 1. The leftover `<button>`

```html
<a hreflang="{{ code }}" href="{{ url }}" role="option" lang="{{ code }}">
    <button>{{ lang.name_local.title }}</button>
</a>
```

Before this PR the `<button>` was the actual control (`form="set-language" type="submit"`) and the anchor was the decoration (`href="#"`). The rework swaps those roles but leaves the button in place, so it is now a `type="submit"` button belonging to no form — it does nothing on its own. Interactive content inside `<a>` is invalid HTML, it adds a second tab stop per language where only one does anything, and a screen reader announces "link, button". Since the anchor now carries the navigation, the button can simply go:

```html
<a hreflang="{{ code }}" href="{{ url }}" role="option" lang="{{ code }}">{{ lang.name_local.title }}</a>
```

While you are in there: `role="option"` on an anchor inside `role="listbox"` is not quite right either — a listbox option is not supposed to be a link. A plain `<ul>` of links with no ARIA roles would be more correct than the current combination, and the dropdown JS does not care. Both of these predate the PR, so entirely your call whether to fix them here.

#### 2. Dead exemption

`rocky/rocky/middleware/auth_required.py:18` still lists `reverse("set_language")` among the URLs exempt from the login requirement. Nothing posts there any more. It is harmless (the `i18n/` include is still in `urlpatterns`, so the `reverse` resolves) and you may well want to keep the endpoint available for API consumers — but if the intent is that the form is gone for good, the exemption and possibly the `i18n/` route can go with it.

#### 3. The description no longer matches the code

The summary still reads "Add a hidden `next` input to the language switcher form … instead of relying only on `HTTP_REFERER`", which is the approach the later commits removed. Worth rewriting before merge, since the squash commit message is taken from it and this is the text that ends up in `git log` and the changelog.

### Test Results

| Component | Test | Result |
|-----------|------|--------|
| rocky | `make utest` | 495 passed, 1 skipped (exit 0) |
| rocky | `languages()` on real resolved URLs | prefix and query string preserved for all five locales |
| rocky | switcher rendered on `/en/login/` | present, with `hreflang` links for nl/pap/it/fy |

The two new tests are reasonable and pin the behaviour that matters (the prefix swap, query string included). `test_language_links_on_unprefixed_url` covers a branch that, per the check above, no page can actually reach — no harm in keeping it as a guard, but it is worth knowing it is not exercising a real path.

What is not covered is the template itself: nothing asserts that the switcher renders real `href`s rather than the old `href="#"`. A single `assertContains(response, 'href="/nl/')` on any authenticated page would pin the user-visible half of this change and would have failed before the PR. Suggestion, not a condition.

### CI/CD Status

17/17 green, `makelang` included. This is a fork branch and CI ran normally.

Note: GitHub reports the branch as **`behind`** main. It will need `Update branch` before merging, and the checks should be allowed to finish on the merged state.

### Functional Test

No `make clean && make kat`. The change is one context processor, one template and a test; there is no service, contract or compose change, and I exercised both halves against real resolved URLs and a real rendered page rather than a mock. The remaining unknown is purely visual — whether the dropdown still looks right once the `<button>` styling hook is (or is not) removed — which is a browser check rather than a stack check. @stephanie0x00, as the other assignee, that one is probably better in your hands than mine.

### Verdict

**APPROVED**

Does what #4222 asked, in the shape you asked for, and the things that could have broken silently — the orphaned context key, the JS bindings, the unprefixed routes — all check out. The three findings are cosmetic and none of them needs to hold the merge; I would take the `<button>` removal and the description rewrite if you are touching it again anyway.

PR #5237, opened 2026-07-14
Minor cleanup around the add objects form, automatically select the o…

underdarknl:
Minor cleanup around the `add objects` form, automatically select the only available option for required selectors if it is the only option.

### Changes

Makes sure we pre-select the first and only option for required OOI selectors, if only one OOI is available.

### Demo

image
### QA notes
when adding a hostname object, the network selector (required) should already be on network|internet (as long as that is the only network in the graph).
---

### Code Checklist

- [ ] All the commits in this PR are properly PGP-signed and verified.
- [ ] This PR only contains functionality relevant to the issue.
- [ ] I have written unit tests for the changes or fixes I made.
- [ ] I have checked the documentation and made changes where necessary.
- [ ] I have performed a self-review of my code and refactored it to the best of my abilities.

- [ ] Tickets have been created for newly discovered issues.
- [ ] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [ ] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [ ] I have included comments in the code to elaborate on what is not self-evident from the code itself, including references to issues and discussions online, or implicit behavior of an interface.

---

## Checklist for code reviewers:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_qa.md) into your comment.

Issue #3933, opened 2024-12-04
Config OOI cannot be created through Rocky


originalsouth:
**Describe the bug**
Config OOI cannot be created through Rocky:

![Image](https://github.com/user-attachments/assets/f04af201-fefa-44d5-bbb6-d33ec381d3c1)

**To Reproduce**

1. Go to "add object"
2. Select "Config"
3. Fill in form as above (or similar)

**Expected behavior**
No error is shown with valid dictionary input

**OpenKAT version**
main

PR #5416, opened 2026-09-14
fix(rocky): allow empty dict for Config OOI creation (#3933)

brennodewinter:
## Summary
- Django's `JSONField` treats `{}` as an empty value and rejects it for required fields, but `{}` is a valid `dict[str, JsonValue]` for the Config model. Override `empty_values` for JSONFields used for dict/list fields so `{}` and `[]` are accepted.
- Fix `OOIForm.clean()` to only drop `None` and `""` (no-input values) instead of all falsy values, which previously stripped valid empty dicts/lists from `cleaned_data`.
- Note: the original `dict[str, JsonValue]` annotation not being recognized by the form generator (causing `{"max_length": 57}` to fail with "Input should be a valid dictionary") was already fixed by PR #5237.

#### Test plan
- [x] Config OOI with empty dict `{}` can be created through Rocky
- [x] Config OOI with non-empty dict `{"key": "value"}` still works
- [x] All 171 object/report tests pass


underdarknl (changes requested):
Looks good, but needs a precommit run.

brennodewinter:
Pre-commit fixed and pushed. Ready for re-review.

hasecon (approved):
## QA Review

### Setup
- Branch: `fix/3933-config-ooi-cannot-be-created` @ `f52d0a0` (fork, reviewed in an isolated worktree)
- Revert-tested each half of the fix separately, and measured the blast radius of the `clean()` change against `main`'s behaviour

Small, correct, and the diagnosis is right. Both changes are load-bearing and both are pinned.

### Both halves are necessary — verified separately

The PR makes two independent changes, so I reverted them one at a time to check neither is dead weight:

| Reverted | `test_add_config_ooi_with_empty_dict` |
|---|---|
| `clean()` back to `if value` | **fails** (`AssertionError` on line 66) |
| `empty_values` back to the default | **fails** (`AssertionError` on line 66) |

That matches the mechanism: `JSONField.validate()` checks `value in self.empty_values`, and Django's default list is `[None, '', [], (), {}]` — so `{}` trips the *required* check inside the field. Getting past that leaves `clean()`, which dropped every falsy value on the way out. Fixing only one of the two leaves the bug in place. Good that both are here.

### The fix is wider than the title says — and that is a second bug fixed

`clean()` is on `OOIForm`, so relaxing the filter changes every OOI add-form, not only `Config`. Worth stating explicitly, because the PR title reads as `Config`-only.

I checked what that actually changes. The filter now keeps `0`, `False`, `[]` and `{}` where it used to drop them. Measured on `DNSTXTRecord` with `ttl=0`:

```
PR branch : ttl in cleaned_data: True   value: 0
as on main: ttl in cleaned_data: False  value: None
```

A TTL of `0` is legal DNS — it means "do not cache" — and on `main` it is silently discarded on the way to the OOI. So this PR also fixes integer-zero input across the board. That is a real improvement and I would mention it in the description rather than let it land unannounced.

Nothing regresses from the wider filter, as far as I can tell:

- Blank optional `CharField`s resolve to `None` (`empty_value=None if not field.is_required() else ""`), so they are still dropped.
- Blank `IntegerField`s and `DateTimeField`s resolve to `None`, still dropped — including the hidden `end_valid_time`.
- There is no `BooleanField` branch in `generate_form_fields` at all: `bool` is not `int` (`bool is int` is `False`), has no `__args__`, is not an `Enum` and is not a `str` subclass, so it falls through to the final `CharField`. So a real Python `False` never reaches `clean()` from a generated form, and the `False` case in the new filter is theoretical today.
- Full suite: 494 passed, 1 skipped.

### Code Review Findings

| 1 | Low | `tools/forms/ooi_form.py:34-38` | `super().clean()` is called twice and the first call's return value is mutated and thrown away |

```python
def clean(self):
    super().clean()["user_id"] = self.user_id
    return {key: value for key, value in super().clean().items() if value is not None and value != ""}
```

This works only because `forms.Form.clean()` returns `self.cleaned_data` by identity, so the first call's mutation is still visible to the second. It predates the PR, but the PR is rewriting this exact return statement, which makes it a natural moment:

```python
def clean(self):
    cleaned_data = super().clean()
    cleaned_data["user_id"] = self.user_id
    return {key: value for key, value in cleaned_data.items() if value is not None and value != ""}
```

Same behaviour, one call, and the intent is readable. Entirely optional.

### Test coverage

`test_add_config_ooi_with_empty_dict` is a good regression test: it goes through the real view, asserts the redirect target and that the declaration actually reached Bytes, and it goes red if either half of the fix is reverted. That is the standard I would want.

The integer-zero path is a *different* branch of the same `clean()` — `IntegerField` rather than `JSONField` — and nothing covers it, so a later "tidy up" of that comprehension could silently undo it without turning CI red. One more test would close that:

```python
def test_add_ooi_keeps_integer_zero(rf, client_member, mock_organization_view_octopoes, mock_bytes_client):
    """A TTL of 0 is legal DNS; the form must not drop it as a falsy value (#3933)."""
    form = OOIForm(
        DNSTXTRecord,
        mock_organization_view_octopoes(),
        data={"hostname": "Hostname|internet|example.com", "value": "v=spf1 -all", "ttl": "0"},
        user_id=client_member.user.id,
    )
    form.is_valid()

    assert form.clean()["ttl"] == 0
```

I ran exactly this shape against both versions of `clean()` — it passes on the branch and fails on `main`'s filter, which is the output quoted above. Suggestion, not a condition.

### Test Results

| Component | Test | Result |
|-----------|------|--------|
| rocky | `make utest` | 494 passed, 1 skipped (exit 0) |
| rocky | `tests/objects/test_objects_add.py` | 6 passed |
| rocky | revert test, `clean()` only | red — fix is load-bearing |
| rocky | revert test, `empty_values` only | red — fix is load-bearing |
| rocky | `DNSTXTRecord` with `ttl=0`, branch vs main | `0` kept vs silently dropped |

### CI/CD Status

17/17 green, `makelang` included. Fork branch; CI ran normally.

GitHub reports the branch as **`behind`** main — it needs `Update branch` before merging, with the checks allowed to finish on the merged state.

### Functional Test

No `make clean && make kat`. The diff is one form method and one field attribute, with no service, contract or compose change, and I exercised the real view through to the Bytes declaration plus the form layer directly against both versions of the code. A running stack would add the browser confirmation that the add-Config form now submits — which is what the issue reporter will check anyway, and what your own test plan already records as done by hand.

### Verdict

**APPROVED**

Right diagnosis, both halves needed, properly pinned by a test that fails when either is reverted, and no regression across the suite. The one Low is a readability tidy-up in the method you are already editing. The integer-zero improvement deserves a line in the description so it does not land as a surprise — and, ideally, the extra test above so a future cleanup of that comprehension cannot quietly undo it.

hasecon:
@underdarknl — your one request on this PR is satisfied, so the only thing still blocking it is the standing review itself, which only you can lift.

You reviewed `dc202f13` ("Looks good, but needs a precommit run"), where `pre-commit` was indeed red. `2650561` ("Fix ruff formatting for CI") fixed it:

| Commit | `pre-commit` |
|---|---|
| `dc202f13` (the commit you reviewed) | failure |
| `f52d0a0d` (current head) | success |

All 17 checks are green on the current head.

For transparency: I posted my own approval earlier today without referencing your review — I had read the issue comments but not the review list, which is my mistake. It did not change the outcome, since your point was already addressed, but you should know my approval was not a judgement on your request.

Full QA notes are in my review above. Short version: both halves of the fix are load-bearing (reverting either one turns the new test red), and the `clean()` change also fixes integer-zero input across every OOI form — a `DNSTXTRecord` with `ttl=0` is silently dropped on `main` and preserved on this branch. That side effect is not in the description and I have asked for it to be mentioned there.

PR #5429, opened 2026-09-20
Pin anyio 4.14.2 in uv.lock to match requirements.txt

brennodewinter:
## Summary
- #5427 bumped `rocky/requirements.txt` to anyio 4.14.2 but left `rocky/uv.lock` at 4.13.0
- `make requirements` (uv export) now produces a diff on any branch, so the `poetry-dependencies` check fails — seen on #5423 after updating its branch
- This runs `uv lock --upgrade-package anyio` so lock and requirements agree again

#### Test plan
- [x] `uv lock --upgrade-package anyio` → 4.13.0 → 4.14.2
- [x] `uv export` output matches committed requirements.txt (only header comment differs)
- [ ] `poetry-dependencies` check green on this PR


brennodewinter:
Heads-up: main is currently inconsistent — #5427 bumped requirements.txt to anyio 4.14.2 but left uv.lock at 4.13.0, so `make requirements` produces a diff and `poetry-dependencies` fails on every PR touching dependency files (seen on #5423). This one-line lock bump restores consistency; `poetry-dependencies` is already green here. Fast review appreciated.

underdarknl:
Please use the make requirements route, and upgrade all packages (that are capable) of the newer pinned lowered via pyproject.toml.
So:
Upgrade pyproject to add the newest lower bound (safe) version.
Upgade all reruitement fies via the make command, push changes.

PR #5419, opened 2026-09-15
Fix description for ECDHE-ECDSA-AES256-SHA cipher

underdarknl:
fix CSV by removing unexpected Comma from 'notes' data

### Changes

remove an unexpected comma from a note, which was breaking our csv for line 89

### Issue link

_You have to create an issue to link to this PR. If this really is not possible, write a very detailed description here and add this PR to the project board directly._

_Please add the link to the issue after "Closes"._

Closes ...

### Demo

_Please add some proof in the form of screenshots or screen recordings to show (off) new functionality, if there are interesting new features for end-users._

### QA notes

_Please add some information for QA on how to test the newly created code._

---

### Code Checklist

- [ ] All the commits in this PR are properly PGP-signed and verified.
- [ ] This PR only contains functionality relevant to the issue.
- [ ] I have written unit tests for the changes or fixes I made.
- [ ] I have checked the documentation and made changes where necessary.
- [ ] I have performed a self-review of my code and refactored it to the best of my abilities.

- [ ] Tickets have been created for newly discovered issues.
- [ ] For any non-trivial functionality, I have added integration and/or end-to-end tests.
- [ ] I have informed others of any required `.env` changes files if required and changed the `.env-dist` accordingly.
- [ ] I have included comments in the code to elaborate on what is not self-evident from the code itself, including references to issues and discussions online, or implicit behavior of an interface.

---

## Checklist for code reviewers:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_code.md) into your comment.

---

## Checklist for QA:

Copy-paste the checklist from [the docs/source/developer-documentation/contributor/templates folder](https://github.com/minvws/nl-kat-coordination/blob/main/docs/source/developer-documentation/contributor/templates/pull_request_template_review_qa.md) into your comment.

PR #5432, opened 2026-09-21
TLS cipher severities updated

stephanie0x00:
### Changes
This PR updates all TLS ciphers to match the NCSC TLS Guidelines 2025-05 version, instead of a specific organisations wishes. This will become the baseline from which own policy changes to the TLS cipher severities can be made.

The previous TLS cipher severities were too aggressive. This PR classifies based on the following format: 

- Critical -> Ciphers for which active abuse is going on. 
- High -> Anything that is marked as insufficient, but is more broken that a regular 'medium' (e.g. Anonymous ciphers, 3DES, RC4, etc. )
- Medium -> Default for things marked as 'Insufficient'.
- Low -> Default for 'phasing out' ciphers. 
- Recommendation -> Those currently marked as 'good' or 'sufficient. 

### Issue link

Closes ...

### Demo

### QA notes

---

### Code Checklist

---

---

## Checklist for QA:


hasecon:
@stephanie0x00 — this is red on all four Python versions and nobody had told you why yet, so here is the diagnosis. It is one missed file, plus a second one that will bite next.

## Why CI fails

All four `test` jobs fail on the same two assertions in `rocky/tests/reports/test_safe_connections_report.py`:

```
>  assert data["sc_ips"][ipaddressv4.reference] == [cipher_finding_type]
E  AssertionError: assert [] == [KATFindingTy...M: 'medium'>)]
```

The cause is that `CIPHER_FINDINGS` exists **twice**, and you updated one of the two:

| File | State |
|---|---|
| `rocky/reports/report_types/tls_report/report.py:13` | updated to the five new `*-TLS-CIPHER` ids ✅ |
| `rocky/reports/report_types/safe_connections_report/report.py:12` | still `["KAT-RECOMMENDATION-BAD-CIPHER", "KAT-MEDIUM-BAD-CIPHER", "KAT-CRITICAL-BAD-CIPHER"]` ❌ |

The Safe Connections report filters on that list, the bit no longer emits those ids, so nothing matches and `sc_ips` comes back empty. Copying the same five ids into `safe_connections_report/report.py` should make all four jobs green.

While you are there: since the list is now duplicated in two report types and this is exactly the failure mode that produces, it would be worth pulling it into one shared constant so the next rename cannot half-land.

## One more file with the old ids

Not causing a failure today, but it still holds an old id and will look wrong once this merges:

- `rocky/tests/stubs/multi_report_data_minvws.json:33,37` — `KAT-MEDIUM-BAD-CIPHER`

That is the complete set; I grepped the tree for `BAD-CIPHER` and everything else is already covered by your diff.

## Things I checked that are fine — so you do not have to worry about them

These are the three questions a reviewer would raise, and all three come out in your favour:

**Dropping the `Notes` column is safe.** The bit reads with `csv.DictReader`, so columns are addressed by name, not by position. Going from 17 to 16 columns shifts nothing.

**Jan's fix from #5419 is not lost.** That PR (merged 15-09) repaired a stray comma in the `Notes` field of `list-ciphers-openssl-with-finding-type.csv`, which made `ECDHE-ECDSA-AES256-SHA` silently produce no finding at all. Your new file removes the `Notes` column entirely and gives TLSv1.0 and TLSv1.1 their own rows, so that whole bug class is structurally gone rather than re-introduced. Verified on your branch: the cipher now appears as two clean rows.

**The 20 rows that produce no finding are the right ones.** `Informational` maps to 0 and the comparison is strictly greater than 0, so a cipher whose four severity columns are all `Informational` yields nothing. I listed them — they are `ECDHE-ECDSA-AES256-GCM-SHA384`, `ECDHE-RSA-CHACHA20-POLY1305` and friends, all on TLSv1.2. Those are the good ciphers and silence is the correct outcome. No TLSv1.0/1.1 row falls into that bucket.

## One thing worth a decision

The bit picks its severity columns with `[col for col in data[0] if "Severity" in col]`. Your CSV has two columns that do not match that filter and are therefore never read:

- `Rating protocol version`
- `Finding protocol versie`

This predates your PR, but your restructuring makes it matter more: you have split the rows per protocol version and filled that column per row, so it now carries real information (a cipher offered over TLSv1.0 rated `Medium` on the protocol alone). Today that rating only reaches a finding indirectly, when one of the four cryptographic columns happens to flag the same row.

We ran into this same gap while reviewing #5419 and it is on our list to raise separately, so it is not yours to solve here. But since you are restructuring the file anyway, it is the natural moment to decide whether the bit should read those two columns — and if not, whether they should stay in the CSV at all.

## Small

The PR body still has an empty `Closes ...` and the template note about linking an issue. Worth filling in before merge, if only so the NCSC-2025 baseline change is findable later.

Happy to run the full QA round once CI is green — say the word.

brennodewinter (commented):
Thanks for the restructure — the new five-tier severity model matching NCSC TLS Guidelines 2025-05 is the right direction, and the diff I read is internally consistent: finding-type ids renamed everywhere the bit emits them, `tls-cipher-findings.csv` restructured per protocol version, tests updated.

**I can't approve while CI is red.** All four `Tests` jobs fail on `rocky/tests/reports/test_safe_connections_report.py`. hasecon's diagnosis is correct — I verified it against the PR head:

- `rocky/reports/report_types/safe_connections_report/report.py:12` still has `["KAT-RECOMMENDATION-BAD-CIPHER", "KAT-MEDIUM-BAD-CIPHER", "KAT-CRITICAL-BAD-CIPHER"]`. The bit no longer emits those ids, so `sc_ips` comes back empty.
- `rocky/tests/stubs/multi_report_data_minvws.json:33,37` still references `KAT-MEDIUM-BAD-CIPHER` — not failing today, but stale once this merges.

Seconding hasecon's suggestion to pull `CIPHER_FINDINGS` into one shared constant while you're in there — this is exactly the half-renamed failure mode a duplicated list produces.

Also worth filling in the `Closes ...` line in the PR body so the NCSC-2025 baseline change is findable later.

Once CI is green this is good to go from my side.

